PR #6, opened 2015-04-19
Wait for port on deploy

jakolehm:
This PR adds "wait_for_port" option to service deploy. If port is defined, service deployer will wait that port is open before proceeding to next container.


jakolehm:
ping @nevalla

nevalla:
Looks good!

PR #7, opened 2015-04-26
Replace sidekiq with sucker_punch

jakolehm:
This PR removes sidekiq with sucker_punch. Using sucker_punch we can get rid of separate sidekiq process that sucks a lot of memory.


jakolehm:
ping @nevalla

nevalla:
Looks good. One minor thing: there are still some debug output in clean up job.

PR #9, opened 2015-05-04
allow user to register Kontena account without invitation to Kontena server

### LINKED ISSUE #8: User not allowed to register Kontena account before user is invited to server
Currently user has to be invited to a Kontena server before he/she can register a Kontena account. This has to be changed so that user can also register a Kontena account without invitation, but error is returned when user is not allowed to use current Kontena server and further invitation is needed.

Supported workflow:
1. Admin user invites user to Kontena server
2. User connects to Kontena server
2. If user has not Kontena account, user can register account for himself/herself
3. Use can log in to Kontena server with Kontena account

What also is needed to support:
1. User connects to Kontena server
2. User registers Kontena account
3. Admin user invites user to Kontena server
4. User can log in to Kontena server with Kontena account

nevalla:
Currently user is not able to register Kontena account if he/she has not invitation to a Kontena server. This PR fixes this issue.
Fixes #8

### DIFF
--- a/app/mutations/users/register.rb
+++ b/app/mutations/users/register.rb
@@ -8,25 +8,29 @@ module Users
     end
 
     def execute
+
+      kontena_user = register_user(email, password)
+      if kontena_user.nil?
+        add_error(:external_id, :invalid, 'Kontena account registration failed')
+        return
+      end
       if User.count == 0
         user = create_admin_user(email)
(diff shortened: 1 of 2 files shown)


nevalla:
I Close this PR now since code structure has been changed. I will make a new PR.

PR #14, opened 2015-05-22
Orphan node cleanup

jakolehm:
This PR implements recurring job that removes orphaned nodes from grid.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #15, opened 2015-05-22
Cli vpn commands

jakolehm:
This PR adds initial cli commands for managing an OpenVPN service. OpenVPN can be used to access internal virtual network and it's services that are not exposed to internet.


nevalla:
Looks very cool!

Kontena::Cli::Grids::Vpn#create method seems to be quite a long. At some point you might want to refactor it a little bit.

PR #17, opened 2015-05-24
Show container id in service logs

jakolehm:
This PR adds container id to logs. Each container id is shown with different color.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #25, opened 2015-05-30
Tx off from already running containers

jakolehm:
This PR fixes bug where weave-helper process does not modify (`ethtool tx off`) already running containers.


jakolehm:
@nevalla ping

nevalla:
:+1:

PR #34, opened 2015-06-01
Improve cli container exec behaviour

### LINKED ISSUE #32: Kontena container exec does not return errors correctly
`kontena container exec` does not return proper exit code from container. It also outputs stderr to stdout which is not good behaviour.

jakolehm:
This PR redirects container stderr to console stderr and exits with container exit code.

Fixes #32


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #35, opened 2015-06-05
Add service option to kontena deploy command

### LINKED ISSUE #23: Add optional parameter(s) for kontena deploy to target only selected services
Sometimes it is needed to deploy just single or few selected services from kontena.yml file. Add support to do this. For example: `kontena deploy <servicename1> <servicename2> <servicenameN>`

nevalla:
This PR adds service option `kontena deploy` command.

If service option is provided only given services will be created/updated and deployed.

Fixes #23


jakolehm:
Looks good!

PR #39, opened 2015-06-10
Add prefix wildcard support to kontena.yml files

### LINKED ISSUE #21: Add prefix wildcard support to kontena.yml files
When using Kontena Deploy feature, all services are prefixed by default with a parent directory name containing the kontena.yml file. Also, user has possibility to prefix the services with a custom prefix. This might affect the kontena.yml file contents since for example when linking services, it is required to name the linked services (including the prefix). Therefore, it is required to have some kind of wildcard support in kontena.yml file to use currently selected prefix instead of hard coded values.

nevalla:
This PR add prefix wildcard support to kontena.yml files. `%{prefix}` tag will be replaced automatically with given prefix or name of the current working directory.

Fixes #21

### DIFF
--- a/cli/lib/kontena/cli/stacks/stacks.rb
+++ b/cli/lib/kontena/cli/stacks/stacks.rb
@@ -18,8 +18,9 @@ module Kontena::Cli::Stacks
       require_token
 
       filename = options.file || './kontena.yml'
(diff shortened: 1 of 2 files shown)


jakolehm:
LGTM!

PR #40, opened 2015-06-12
Refactor connect, login and register flows

### LINKED ISSUE #38: refactor connect, login and register flows
In order to streamline the usage of Kontena, following enhancements should be made:

1. `kontena register` should be changed so that it is possible to register at any time. The command without any arguments should register user to Kontena's default and public **auth provider** service at address https://auth.kontena.io. User may provide address to some other auth provider service as an argument for this command.
2. Master Node should have configurable parameter for defining the **auth provider** address. Any login attempts to Master Node should be authenticated via this auth provider. By default, this address should be Kontena's default and public auth provider service at address https://auth.kontena.io.
3. `kontena login` should be changed so that it is requiring an argument describing the Master Node address. Show Kontena ASCII art during login :)
4. Get rid of `kontena connect` as there is no more need for it.

nevalla:
This PR updates connect, login and register flows:
* `kontena register` does request directly to auth.kontena.io service (or to custom auth provider)
* `kontena connect` command is removed.
* `kontena login` takes url of master node as an argument

Fixes #38


jakolehm:
Looks good!

nevalla:
@miskun I have updated the documentation if you want to take a look before merge:
https://github.com/kontena/kontena/commit/0f04d48ce9abd4dea47be8fbb7f0c6b3552ab1e5

PR #41, opened 2015-06-15
Etcd as dns backend

jakolehm:
This PR changes dns backend from master node to etcd cluster. This change allows agent nodes to resolve dns even if master node is not available.


nevalla:
Looks good!

PR #42, opened 2015-06-15
Hide internal services

### LINKED ISSUE #20: Hide internal system containers
At the moment, some internal system containers (like VPN) are shown by default. We need to have ability to hide any such system containers as they might clutter the list of services and containers.

jakolehm:
This PR hides internal services from grid service listings. 

Fixes #20 .


nevalla:
LGTM!

PR #43, opened 2015-06-16
Node label based affinity filtering

jakolehm:
This PR implements new scheduler filtering based on node labels. We already collect & save labels from nodes so let's use them.


jakolehm:
@nevalla ping

miskun:
looks good to me

nevalla:
Looks good!

PR #50, opened 2015-06-23
Enhance kontena grid commands

nevalla:
This PR updates `kontena grid create` command so that created grid is in use automatically after grid creation.

There was also a bug when tried to remove existing grid. It is now fixed.

Also some minor tweaking is done related output messages.


jakolehm:
Looks good!

miskun:
looks good to me too

PR #52, opened 2015-07-07
Grid discovery

jakolehm:
This PR implements basic discovery for agents (nodes). It includes automatic weave + etcd peer discovery. Agent needs to know only master url and grid token to join.


nevalla:
Looks good!

PR #61, opened 2015-07-15
Fix FloatDomainError when showing service stats

jakolehm:


jakolehm:
@miskun @nevalla ping

nevalla:
Looks good to me!

PR #62, opened 2015-07-15
Use Alpine Linux 3.2 instead of edge

jakolehm:


jakolehm:
@miskun @nevalla ping

nevalla:
:+1:

PR #65, opened 2015-07-15
Update roda to 2.5

nevalla:
This PR updates [Roda](http://roda.jeremyevans.net/) framework to latest version (2.5). 

The only [backwards compatibility issue](http://roda.jeremyevans.net/rdoc/files/doc/release_notes/2_3_0_txt.html#label-Backwards+Compatibility) was that `ext` option was removed from render plugin. I fixed this by using 'json.jbuilder' as `engine`. That way Roda can resolve template file correctly.


jakolehm:
:+1: 

Can we include this in 0.7?

miskun:
yes, why not

PR #68, opened 2015-07-20
DigitalOcean node provisioner

jakolehm:
This PR provides cli helper commands to provision droplets to DigitalOcean that are automatically provisioned and joined to Kontena grid as nodes.

```sh
$ kontena node provision digitalocean --token=<do_token> --ssh-key=<path_to_ssh_public_key>

$ kontena node destroy digitalocean --token=<do_token> --name=<node_name>
```

### DIFF
--- a/cli/lib/kontena/cli/nodes/commands.rb
+++ b/cli/lib/kontena/cli/nodes/commands.rb
@@ -1,6 +1,7 @@
 module Kontena::Cli::Nodes; end;
 
 require_relative 'nodes'
+require_relative 'digital_ocean'
 
 command 'node list' do |c|
   c.syntax = 'kontena node list'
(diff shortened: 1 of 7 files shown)


miskun:
we should still consider those verbs; i'd prefer:
* provision --> **add** (or add, create, deploy)
* destroy --> **remove** (or delete, terminate)

"destroy" is just seems too scary :) "add" is easy to remember.

jakolehm:
We already have `node remove`, it removes node from the grid but does not destroy the server. That's why I thought that scary `destroy` would be better because it will eat your server (and data)?

miskun:
Ok, I see. What does node remove actually do? I think it should be considered as "kick"; so basically revoking access token for a node. But this might require heave redesign of node "join" process and workflows...

jakolehm:
We have only grid tokens atm so we cannot revoke node access token. So current `node remove` command is handy when you shutdown/terminate node via some other tool and want it to disappear from `node list` immediately.

PR #70, opened 2015-07-20
Remove fist of fury

jakolehm:
This PR removes fist_of_fury gem and replaces it's functionality with much simpler approach. One dependency less is better.


miskun:
looking good!

nevalla:
Lgtm

PR #71, opened 2015-07-20
Expire container logs after 3 months

jakolehm:
We don't want to keep container logs forever. Let's expire them after 3 months.


miskun:
looking good!

nevalla:
:+1:

PR #72, opened 2015-07-20
Expire container stats after 1 month

jakolehm:
We don't want to keep container stats forever. Let's expire them after 1 month.


miskun:
looking good!

nevalla:
:+1:

PR #75, opened 2015-07-22
Fix weave default iprange

jakolehm:
Default iprange conflicts with DigitalOcean, this should fix it (fix taken from weave master)


miskun:
looking good!

PR #78, opened 2015-07-22
Fix ContainersApi

nevalla:
There are couple of errors on Containers API:
- load_grid_container fails when validating grid
- /v1/containers/:grid_name/:service_name/:name/top fails because of invalid host_id field
- Container#to_path is not correct

this PR fixes these issues.


jakolehm:
LGTM!

PR #80, opened 2015-07-24
Sort logs by _id

jakolehm:
We should sort logs by _id (instead of $natural) because it allows mongo to use indexes.


jakolehm:
@miskun @nevalla ping

nevalla:
Looks good!

PR #86, opened 2015-07-25
Change grid default initial size to 1

jakolehm:
This changes grid default initial size from 3 to 1. I would like to change this because everyone is starting to test with only one node and it confuses if grid is fully functioning only after third node joins.


jakolehm:
@nevalla @miskun ping

nevalla:
Looks good!

PR #89, opened 2015-07-26
Fix kontena deploy

### LINKED ISSUE #88: kontena deploy fails with "not found"
I made a kontena.yml that looks similar to the following:

    # jenkins configuration
    jenkins:
        image: jenkinsci/jenkins:latest
        stateful: true
        instances: 1
        user: jenkins
        ports:
            - "9181:8080"
            - "9150:50000"
        affinity:
            - container!=gitblit

    # gitblit configuration
    gitblit:
        image: elderresearch/docker-gitblit:latest
        stateful: true
        instances: 1
        user: gitblit
        ports:
            - "9180:80"
            - "9118:9418"
            - "9122:29418"
        affinity:
            - container!=jenkins

However, when I run `kontena deploy` with this kontena.yml file, it creates the services then fails with `error: {"error":"Not found"}.`. For some reason it also creates the services as `demo/vagrant-jenkins` and `demo/vagrant-gitblit`. If I call it to deploy manually with `kontena service deploy vagrant-jenkins` and `kontena service deploy vagrant-gitblit`, it works.

Why does it append the `demo/` portion if we're already on the `demo` grid? (Surely that's redundant and confusing, because my first instinct was to do `kontena service deploy demo/vagrant-jenkins`, which doesn't work.) Secondly, it looks that `kontena deploy` is calling the services wrong.

nevalla:
There is regression on 0.7 release that breaks `kontena deploy` command. This PR will fix this issue.

Fixes #88

### DIFF
--- a/cli/lib/kontena/cli/stacks/stacks.rb
+++ b/cli/lib/kontena/cli/stacks/stacks.rb
@@ -44,7 +44,7 @@ module Kontena::Cli::Stacks
           data[:strategy] = service['deploy']['strategy'] if service['deploy']['strategy']
           data[:wait_for_port] = service['deploy']['wait_for_port'] if service['deploy']['wait_for_port']
         end
-        deploy_service(token, service['id'], data)
+        deploy_service(token, service['id'].split('/').last, data)
       end
     end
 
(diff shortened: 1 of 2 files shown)


jakolehm:
LGTM

PR #91, opened 2015-07-26
Fix cli service id handling

### LINKED ISSUE #90: Hide grid prefix from service names
As mentioned in #88 it's confusing to show full id/path for services because user has already switched to grid. We should at least remove grid prefix from `service list` and autocomplete for now.

jakolehm:
This PR improves service id handling and fixes #90 .


jakolehm:
@nevalla ping

nevalla:
LGTM

PR #93, opened 2015-07-26
Fix kontena bintray repo path in Vagrantfile

robxu9:
Provisioning agents fail otherwise.


nevalla:
Thanks! I thought I already fixed this, but apparently I never commited it.

PR #102, opened 2015-08-05
Fix kontena container exec

nevalla:
This PR fixes `kontena container exec` command by adding service name to request url.


jakolehm:
LGTM!

PR #103, opened 2015-08-08
Non-blocking subscribe

jakolehm:


jakolehm:
@nevalla please run tests in a loop and report if you bump into errors.

PR #104, opened 2015-08-09
Remove --bridge=weave usage

jakolehm:
This PR removes usage of Docker option `--bridge=weave` with more officially supported way, using weaveexec container (and weave shell script) to attach container to overlay network.

Why: http://blog.weave.works/2015/07/16/bridge-over-troubled-weavers/


jakolehm:
Current implementation is based on unreleased master version of weave (aka upcoming 1.1).

jakolehm:
@miskun @nevalla ping

jakolehm:
Moved launching weave/etcd/cadvisor to agent. Now it's easy to start agent with plain-old-docker commands (complex packaging is not needed). That said, we should come up with some way to modify hosts dns settings (resolv.conf).

```
docker run --name=kontena-agent \
      -e KONTENA_URI=$KONTENA_URI \
      -e KONTENA_TOKEN=$KONTENA_TOKEN \
      -e KONTENA_PEER_INTERFACE=$KONTENA_PEER_INTERFACE \
      -v=/var/run/docker.sock:/var/run/docker.sock \
      kontena/agent:$KONTENA_VERSION
```

jakolehm:
@nevalla @miskun ok to merge?

agent/lib/kontena/weave_adapter.rb
```diff
@@ -0,0 +1,130 @@
+require_relative 'helpers/node_helper'
+
+module Kontena
+  class WeaveAdapter
+    include Helpers::NodeHelper
+
+    WEAVE_VERSION = 'git-bd19b8625908'
```
nevalla: If this will be changed regurarly, should we define it somewhere else?

nevalla:
Just a minor comment above, otherwise looking so good!

agent/lib/kontena/weave_adapter.rb
jakolehm: Do you have any suggestions where we could define it?

jakolehm: We will change it to '1.1.0' when weave releases it.. after that it will change only when needed. One option is to override version through environment variable.

nevalla: I would have suggested env variable as well, but I don't know is it worth of it if we don't update it so often.

PR #106, opened 2015-08-10
Add etcd RPC api

jakolehm:
This PR implements "passthrough" api to etcd. Purpose for this api is to give master access to grid etcd k/v store.


miskun:
Looking good!

PR #107, opened 2015-08-12
Check that initial nodes are setup before allowing deploys

jakolehm:
Check that grid has all the initial nodes registered before allowing service deploys.


jakolehm:
@nevalla ping

nevalla:
LGTM!

jakolehm:
Rebased

PR #108, opened 2015-08-12
Don't allow to destroy initial node without force

jakolehm:


nevalla:
Looks good! This should be taken into account also on #112 or is it already done?

jakolehm:
Let's move this to 0.9.

PR #111, opened 2015-08-17
Do not re-deploy containers that are not changed since last deploy

### LINKED ISSUE #22: Avoid updating of service that has not changed with Kontena deploy
At the moment, when Kontena deploy is used, all services are deployed even if there was no change in the service configuration. This is unnecessary and might cause downtime for example for database services. Ideally, only the services that have changed will be updated with `kontena deploy` command.

nevalla:
At the moment services are deployed even if there was no change in the service configuration. This PR changes this behavior by checking is Docker image or service configuration changed before deploy is made.

Fixes #22

### DIFF
--- a/server/app/models/container.rb
+++ b/server/app/models/container.rb
@@ -9,7 +9,7 @@ class Container
   field :driver, type: String
   field :exec_driver, type: String
   field :image, type: String
-  field :image_id, type: String
+  field :image_version, type: String
   field :env, type: Array, default: []
   field :network_settings, type: Hash, default: {}
   field :state, type: Hash, default: {}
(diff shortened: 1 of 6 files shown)


jakolehm:
It seems that relation from Image -> Container is still missing, we should add that. Otherwise looks really good!

nevalla:
We don't have that relation yet and container's `image_id` field is for docker container image_id. But good point that if we want to add relation between image and container that `image_id` field should be renamed to something else, maybe `image_version`?

jakolehm:
`image_version` sounds good to me.

miskun:
should we target this for 0.8.0 ? if so, let's add the milestone

nevalla:
I think we need to test this feature carefully before we can merge this, so I will postpone this to 0.9 release.

jakolehm:
I agree.

jakolehm:
Rebased and tested, seems to work. Should we merge? @nevalla @miskun

nevalla:
+1 for merge

PR #112, opened 2015-08-19
Refactor from commander to clamp

jakolehm:
This PR changes cli to use clamp instead of commander. This will improve cli user experience a lot because subcommands are nested (each command will show available subcommands etc..). Option parser is also much more user friendly in clamp.


nevalla:
Looks good!

PR #125, opened 2015-08-28
Switch to static etcd bootstrapping

### LINKED ISSUE #118: Use static etcd bootstrapping
Discovery seems to cause some race conditions when nodes are booted in parallel. We know node overlay ip's before hand so static bootstrapping should be a better option. And it does not call to internet.

jakolehm:
Fixes #118


nevalla:
Looks good!

PR #126, opened 2015-08-28
Prevent deleting linked service

nevalla:
Currently when deleting service that is linked to another service, another service will have broken service link. To avoid this situtation this PR prevents deletion and returns error instead.


jakolehm:
LGTM!

PR #129, opened 2015-08-28
Fix service container stats

### LINKED ISSUE #96: Service stats net I/O wrong after re-deploy
It seems that `service stats` shows wrong net I/O stats after service is re-deployed. It should show cumulative sum of network I/O (also from old containers), yes?

jakolehm:
Latest improvements did broke stats. This PR fixes that and also fixes how network/memory percentage is calculated. Fixes #96 .


jakolehm:
@nevalla ping

miskun:
@jakolehm pong looking so good!

PR #130, opened 2015-08-30
Registry 2.1 with authentication option

jakolehm:
This PR upgrades private registry to 2.1 and adds optional authentication command line option.


nevalla:
Looks good!

PR #131, opened 2015-08-30
Improve cli service commands

jakolehm:
- show all possible information about service
- fix service containers list output
- update command should not allow all the same options as in create


jakolehm:
@nevalla @miskun should we include this in 0.8.0?

nevalla:
Looks good! No reason to exclude this.

PR #132, opened 2015-08-30
Add grid uri (master uri) to grid show output

jakolehm:
This will make users life easier when setting up nodes/agents.


miskun:
yes, we need this! +1

PR #142, opened 2015-08-31
Do not add dash when deploy prefix is empty

### LINKED ISSUE #139: kontena deploy -p switch not working with empty string
```
$ kontena deploy -p ""
DEPRECATION WARNING: Support for 'kontena deploy' will be dropped. Use 'kontena app deploy' instead.
creating -wordpress
creating -mysql
```

The name of services should be "wordpress" and "mysql" in this case. Also,

```
$ kontena service show "-wordpress"
ERROR: Unrecognised option '-w'

See: 'kontena service show --help'
```

It's impossible to do anything with these services.

jakolehm:
Fixes #139

### DIFF
--- a/cli/lib/kontena/cli/apps/common.rb
+++ b/cli/lib/kontena/cli/apps/common.rb
@@ -19,6 +19,8 @@ module Kontena::Cli::Apps
     end
 
     def prefixed_name(name)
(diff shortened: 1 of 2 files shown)


nevalla:
LGTM

PR #144, opened 2015-09-01
Fix digitalocean dns configuration

### LINKED ISSUE #143: Internal DNS not working on nodes provisioned to Digital Ocean
When provision nodes to Digital Ocean with Kontena Cli, internal DNS is not working correctly on provisioned nodes. That prevents using Kontena's built-in Docker registry because agent can't pull any images from registry.

jakolehm:
Fixes #143


nevalla:
:+1:

PR #155, opened 2015-09-07
Fix empty multivalue options handling on service commands

### LINKED ISSUE #154: Affinity and port options do not work together
When ever service is updated with port information the service spec loses it's affinity setup. And when affinity is set the service loses its port configuration.

Steps to reproduce:
```
kontena service create --affinity node==xyz nginx nginx:latest
kontena service deploy nginx
kontena service show nginx # Should show affinity in action
kontena service update -p 80:80 nginx
kontena service deploy nginx
kontena service show nginx # Affinity setup lost!!!
```

Seems to work ok if one specifies BOTH options at the same time. :)

nevalla:
This PR fixes multi-value options handling when user has not set option on `kontena service create` or `kontena service update` commands. 

Fixes #154

### DIFF
--- a/cli/lib/kontena/cli/services/update_command.rb
+++ b/cli/lib/kontena/cli/services/update_command.rb
@@ -38,7 +38,7 @@ module Kontena::Cli::Services
       data[:memory_swap] = parse_memory(memory_swap) if memory_swap
       data[:cpu_shares] = cpu_shares if cpu_shares
       data[:affinity] = affinity_list unless affinity_list.empty?
-      data[:env] = env_list if env_list.empty?
+      data[:env] = env_list unless env_list.empty?
       data[:container_count] = instances if instances
       data[:cmd] = cmd.split(" ") if cmd
       data[:user] = user if user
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/services/update_command.rb
```diff
@@ -33,18 +33,18 @@ def execute
...
       data = {}
-      data[:ports] = parse_ports(ports_list) if ports_list
+      data[:ports] = parse_ports(ports_list) unless ports_list.empty?
       data[:memory] = parse_memory(memory) if memory
       data[:memory_swap] = parse_memory(memory_swap) if memory_swap
       data[:cpu_shares] = cpu_shares if cpu_shares
-      data[:affinity] = affinity_list if affinity_list
-      data[:env] = env_list if env_list
+      data[:affinity] = affinity_list unless affinity_list.empty?
+      data[:env] = env_list if env_list.empty?
```
jakolehm: this should be `unless env_list.empty?`, right?

jakolehm:
:+1:

PR #158, opened 2015-09-08
Fix service network calculation

jakolehm:
Fix calculation of network stats.


jakolehm:
@nevalla ping

miskun:
looking good!

PR #159, opened 2015-09-08
Fix kontena registry create command

nevalla:
S3 and Azure options are not working currently on `kontena registry create` command. This PR fixes this.


jakolehm:
LGTM!

PR #160, opened 2015-09-08
Update weave to official 1.1.0

jakolehm:
Update weave to official 1.1.0 release (we were using unofficial release from git)


nevalla:
Looking good!

PR #162, opened 2015-09-09
Use host networking in cadvisor 

jakolehm:
.. so that we don't reserve any ports from other containers.


jakolehm:
@nevalla @miskun ping

miskun:
perfect!

PR #166, opened 2015-09-14
Use optional name parameter instead of --name switch on node provision

### LINKED ISSUE #157: Consistent node provision commands in CLI
The usage of node provision commands should be made consistent in CLI. At the moment, create command will take node name with `--name` switch and for rest of the commands it is given as argument. Therefore, the create command usage should be changed to something like this:

```
kontena node aws create [OPTIONS] [NAME]
```

nevalla:
This PR changes node provision format so that instead  --name switch, optional node name is given as command parameter as on other node commands.

Also on DigitalOcean node provision given region is stored to node labels.

Fixes #157


jakolehm:
LGTM!

PR #171, opened 2015-09-17
Fix automatic collection indexing

### LINKED ISSUE #170: Database indexes are not created automatically
Rake task seems to do nothing.. I think we should create these directly from code.

jakolehm:
Fixes #170


nevalla:
Looks good!

PR #173, opened 2015-09-17
Add latest tag to image if tag is not specified

jakolehm:
This pr adds "latest" tag to image if image tag is not specified.


nevalla:
Looks good!

PR #176, opened 2015-09-19
Fix duplication on app deploy command

nevalla:
`kontena app deploy` deploys linked services twice unintentionally. This PR fixes that problem and services are deployed only once.


jakolehm:
Look good!

PR #179, opened 2015-09-20
Allow to define weave image names through env variables

jakolehm:
Fixes hardcoded weave image names in agent (other images are already taken from env).


miskun:
Looking good!

PR #180, opened 2015-09-20
Handle agent connection errors on boot

jakolehm:
This PR adds better error handling to etcd/weave launchers when connection to master cannot be established.


nevalla:
Looks good!

PR #189, opened 2015-09-28
Add support for external_links key in kontena.yml

nevalla:
This PR add support for `external_links` key. External links are like normal service links but their name won't be prefixed.


cli/lib/kontena/cli/apps/deploy_command.rb
```diff
@@ -57,6 +57,12 @@ def create_or_update_service(name, options)
         end
       end
 
+      if options['external_links']
```
jakolehm: I think we should move this to separate method

jakolehm:
:+1:

PR #165, opened 2015-09-12
Heroku style deployment

nevalla:
This PR updates `kontena app init` command to support [app.json Application Manifest](https://devcenter.heroku.com/articles/app-json-schema).

When Kontena detects Procfile and app.json, it generates corresponding services, addons and environment variables for docker-compose.yml and kontena.yml.

Functionality is very similar to Heroku's [Build and Deploy with Docker](https://devcenter.heroku.com/articles/docker)

When deploying application Kontena builds Docker image and pushes it to registry if image is not found locally. Image can be built and pushed again with `kontena app build` command.

**Example**

When having the following files:

**app.json**

```
{
  "name": "Small Sharp Tool",
  "description": "This app does one little thing, and does it well.",
  "env": {
    "SECRET_TOKEN": {
      "description": "A secret key for verifying the integrity of signed cookies.",
      "generator": "secret"
    },
    "WEB_CONCURRENCY": {
      "description": "The number of processes to run.",
      "value": "5"
    }
  },
  "addons": ["heroku-redis","mongolab:shared-single-small"]
}
```

**Procfile**

```
web: rackup -p $PORT -E production
worker: RACK_ENV=production ruby ./worker.rb run
```

`kontena app init` produces:

**docker-compose.yml**

```
web:
  build: "."
  command: "/start web"
  env_file: ".env"
  links:
  - redis:redis
  - mongolab:mongolab
  environment:
  - REDIS_URL=redis://redis:6379
  - MONGOLAB_URI=mongolab:27017
redis:
  image: redis:latest
mongolab:
  image: mongo:latest
worker:
  build: "."
  command: "/start worker"
  env_file: ".env"
  links:
  - redis:redis
  - mongolab:mongolab
  environment:
  - REDIS_URL=redis://redis:6379
  - MONGOLAB_URI=mongolab:27017
```

**.env**

```
SECRET_TOKEN=41977498daabf8b0f0ffbae63ce8bb4cd137b4a5f89d28248abe27e49724da274794e6bcc7df60a2b2c54ecaf16487830eb6c65e1fcf5aabc6f9cab8d3e34d83
WEB_CONCURRENCY=5
```

**kontena.yml**

```
---
web:
  extends:
    file: docker-compose.yml
    service: web
  image: registry.kontena.local/docker-twitter-stream:latest
redis:
  extends:
    file: docker-compose.yml
    service: heroku-redis
mongolab:
  extends:
    file: docker-compose.yml
    service: mongolab
worker:
  extends:
[...]

PR #193, opened 2015-10-01
Add support for env dictionaries

jakolehm:
```
serviceA:
  environment:
    FOO: bar
    TEXT: |
      Lorem ipsum dolor sit amet, consectetur adipiscing elit. 
      Phasellus non feugiat magna. Nam nec nulla a velit porttitor aliquam. 
      Donec interdum urna et tellus euismod laoreet. Ut egestas eu quam vitae 
      tristique. Ut eu lobortis arcu, vitae convallis magna.
```

### DIFF
--- a/cli/lib/kontena/cli/apps/deploy_command.rb
+++ b/cli/lib/kontena/cli/apps/deploy_command.rb
@@ -22,8 +22,6 @@ module Kontena::Cli::Apps
       deploy_services(deploy_queue)
     end
 
-    private
-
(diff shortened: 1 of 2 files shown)


nevalla:
This is a very cool addition. However current implementation lacks environment variable inheritance when extending services in yml. Environment variable inheritance is implemented in #165 and since this feature is very crucial in order to get inheritance work correctly I have implemented this functionality also in that branch.

PR #196, opened 2015-10-04
Pick url/token/grid from env variables

### LINKED ISSUE #114: CLI should pick credentials from env variables
This is handy when using cli from CI.

jakolehm:
Fixes #114


nevalla:
Looks good!

PR #201, opened 2015-10-06
Auto-update agents to match master version

jakolehm:
This PR implements automatic version sync between master and agent nodes. If major or minor number is different, master will close the connection (and agent will shutdown). If only patch number is different connection is accepted. In any case master will send information to agent that can persist this information.

Optionally if agent has mount to `/etc/kontena.env` it will write `KONTENA_VERSION=X.Y.Z` there so that next start can pull same agent version that master has.

agent

### DIFF
--- a/agent/lib/kontena/agent.rb
+++ b/agent/lib/kontena/agent.rb
@@ -1,6 +1,8 @@
 module Kontena
   class Agent
 
+    VERSION = File.read('./VERSION').strip
+
     def initialize(opts)
       @opts = opts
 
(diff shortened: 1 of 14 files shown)


jakolehm:
Note to self: include `/etc/kontena.env` mount to coreos cloudconfigs

jakolehm:
@nevalla @miskun ping

nevalla:
Looks good! Do we need to make any changes to node provision scripts / cloudinit?

PR #208, opened 2015-10-08
Tab-complete service names on kontena app commands

nevalla:
This PR adds tab-complete for service names on `kontena app` commands.


jakolehm:
LGTM

PR #211, opened 2015-10-10
Use optional name parameter instead of --name switch on AWS node provision

nevalla:
This PR updates AWS node provision to follow node provision conventions introduced in #166.


jakolehm:
Looks good!

PR #216, opened 2015-10-10
Add/remove service env commands

jakolehm:
This PR adds `kontena service add-env` and `kontena service remove-env` subcommands for easier service env variable manipulation.


miskun:
looking awesome!

PR #217, opened 2015-10-11
Better db migrations

jakolehm:
This PR refactores current db indexing job to separate database migrations. Current logic is hitting large databases too hard and this new migration way should help with that (each migration is executed only once).


nevalla:
Great! Looking so good!

PR #227, opened 2015-10-18
Node ssh subcommand

jakolehm:
This PR adds helper command for accessing host nodes through ssh.

```
$ kontena node ssh -h
Usage:
    kontena node ssh [OPTIONS] NODE_ID

Parameters:
    NODE_ID                       Node id

Options:
    -i, --identity-file IDENTITY_FILE Ssh private key to use
    -u, --user USER               Login as a user (default: "core")
    --private-ip                  Connect to node using private ip
    --internal-ip                 Connect to node through VPN
```


miskun:
Looking good!

PR #232, opened 2015-10-19
Add possibility to define custom ca.pem file per kontena master

jakolehm:
So that it's easier & safer to use self-signed certificates with master.


nevalla:
Looks good!

PR #233, opened 2015-10-20
Set addon services stateful by default on kontena app init command

nevalla:
This PR sets Heroku addon services stateful by default on `kontena app init` command.

### DIFF
--- a/cli/lib/kontena/cli/apps/common.rb
+++ b/cli/lib/kontena/cli/apps/common.rb
@@ -122,7 +122,7 @@ module Kontena::Cli::Apps
           },
           'memcachedcloud' => {
               'image' => 'memcached:latest',
-              'envirnoment' => ["MEMCACHEDCLOUD_SERVERS=#{prefix}memcachedcloud:11211"]
+              'environment' => ["MEMCACHEDCLOUD_SERVERS=#{prefix}memcachedcloud:11211"]
           }
       }
     end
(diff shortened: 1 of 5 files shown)


jakolehm:
What happens if user deploys another app to the same grid? Does it throw error because load balancer exists or something else?

nevalla:
With this current implementation loadbalancer is app specific so if there is another loadbalancer service on the same grid, apparently the public port 80 is already reserved and deploy fails.

nevalla:
I have removed load balancer at this moment and let's figure out proper way to do it later.

cli/lib/kontena/cli/apps/common.rb
```diff
@@ -80,5 +80,51 @@ def dockerfile_exist?
...
+              'image' => 'postgres:latest',
+              'environment' => ["DATABASE_URL=postgres://#{prefix}postgres:@postgresql:5432/postgres"]
+          },
+          'mongolab' => {
+              'image' => 'mongo:latest',
+              'environment' => ["MONGOLAB_URI=#{prefix}mongolab:27017"]
+          },
+          'memcachedcloud' => {
+              'image' => 'memcached:latest',
+              'enviroment' => ["MEMCACHEDCLOUD_SERVERS=#{prefix}memcachedcloud:11211"]
```
jakolehm: let's fix this typo

nevalla:
ping @jakolehm

jakolehm:
I think all the typos are gone.. so LGTM!

PR #239, opened 2015-10-25
Destroy containers that have been marked for deletion

jakolehm:


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #245, opened 2015-10-25
Update container by unique id

jakolehm:
When container is not found by docker container id, we should use our internal unique id (not name) to update container information.


nevalla:
LGTM

PR #260, opened 2015-10-29
Do not generate MAINTAINER to Dockerfile

nevalla:
This PR removes maintainer generation from Dockerfile. It is not needed in Dockerfile and if user is not logged in to Kontena Master when running `kontena app init` it triggers error.


jakolehm:
LGTM

miskun:
Looking good, merge please.

PR #262, opened 2015-10-30
Fix app init command when app.json is not present

### LINKED ISSUE #253: `kontena app init` products "undefined method `[]' for nil:NilClass"
Dockerfile gets created but nothing else does
Ubuntu 14.10

nevalla:
Fixes #253

### DIFF
--- a/cli/lib/kontena/cli/apps/common.rb
+++ b/cli/lib/kontena/cli/apps/common.rb
@@ -90,6 +90,8 @@ module Kontena::Cli::Apps
     def app_json
       if !@app_json && File.exist?('app.json')
         @app_json = JSON.parse(File.read('app.json'))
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/apps/common.rb
```diff
@@ -90,6 +90,8 @@ def create_yml(services, file='kontena.yml')
     def app_json
       if !@app_json && File.exist?('app.json')
         @app_json = JSON.parse(File.read('app.json'))
+      else
+        @app_json = {}
```
jakolehm: Is it ok to initialize `@app_json` to empty hash on every call (if file is missing)?

miskun:
Should work. Merge please.

PR #265, opened 2015-10-31
Change docker bip to non-default value

### LINKED ISSUE #263: VPN docker bridge conflict
Looks like kontena nodes are using the same ip as the default docker bridge (at least on Ubuntu workstations).
Workaround for now is adding a fix to `/etc/default/docker`
```
DOCKER_OPTS="--bip 10.200.42.1/16"
```
It looks like at least on Ubuntu machines you ALSO have to reset the ip on the existing bridge:
```
sudo ip addr del 172.17.42.1/16 dev docker0
sudo ip addr add 10.200.42.1/16 dev docker0
```

jakolehm:
Current docker bridge ip causes conflict with linux users because local docker bridge has the same ip. Fixes #263 .

### DIFF
--- a/cli/lib/kontena/machine/aws/cloudinit.yml
+++ b/cli/lib/kontena/machine/aws/cloudinit.yml
@@ -12,7 +12,7 @@ write_files:
   - path: /etc/systemd/system/docker.service.d/50-kontena.conf
     content: |
         [Service]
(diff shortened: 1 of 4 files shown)


jakolehm:
ping @nevalla

miskun:
Looking good!

PR #267, opened 2015-10-31
Fix vpn/dns config in linux

jakolehm:
We need to append resolv-conf lines to vpn config when cli is used from linux workstations.


miskun:
Looking good!

PR #274, opened 2015-11-02
Append node_number to HostNode#name if it's not unique

### LINKED ISSUE #273: Append instance number to node name when master sees naming conflict
Master should append instance number to node name if it sees naming conflict.

jakolehm:
Fixes #273


jakolehm:
ping @miskun @nevalla

nevalla:
LGTM

PR #275, opened 2015-11-02
Small fixes to quick-start.md

vmj:
Three small corrections:

 * there's no global `-v` option, use `version` command instead
 * text speaks about `mygrid` but example used `testing` as the grid name
 * logs do not form user actions, I believe - user actions form logs


jakolehm:
Good catches, thanks!

PR #282, opened 2015-11-09
Add separate subcommand for service envs

jakolehm:
Long environment variables pollute service show output. This PR adds separate subcommand `kontena service envs <name>` that shows all env variables and truncates env output on `kontena service show <name>`.


nevalla:
LGTM!

PR #272, opened 2015-11-01
Dynamic service rescheduling/failover on grid change

jakolehm:
This PR introduces dynamic rescheduling when nodes join/leave grid. This PR also refactors a lot of communication between server and agents.

Fixes #110

PR #286, opened 2015-11-13
Daemon strategy

jakolehm:
Many users have requested for a way to run a daemon on every node in a Kontena Grid, or on a certain set of nodes in a grid. This is essential for use cases such as building a sharded datastore, or running a logger on every node. In comes the `daemon` strategy, a way to conveniently create and manage daemon-like workloads in Kontena.

Depends on #272 

daemon


miskun:
Looking good!

PR #307, opened 2015-11-23
Dont validate deploy possibility

jakolehm:
We are setting desired state so it's not good idea to block deploy command before hand.


jakolehm:
@nevalla ping

nevalla:
:+1:

PR #313, opened 2015-11-24
Added pid option for service to put it in host pid namespace

jnummelin:
Pid option allows to put the service containers into host pid namespace. Needed for example for Elastic Topbeat agent to pull host level process stats.


jakolehm:
thank you

PR #328, opened 2015-11-25
add env option to grid

hans-d:
optional -e flag to prefix with 'export'
lists KONTENA_URI and KONTENA_TOKEN either for:
- `eval "$(kontena grid env -e)"`
- or `kontena grid env > kontena.env`
to use it for further node deployment


jakolehm:
Looks good, thanks!

PR #331, opened 2015-11-27
Cleanup only non-initial stateless nodes

jakolehm:
This PR changes automatic node cleanup logic a bit. From now on only non-initial stateless nodes are automatically removed from master.

stateless node: node that does not have any stateful services


nevalla:
LGTM

PR #348, opened 2015-12-02
Map container to node always on update

jakolehm:
Master should always try to update container node when it receives updated container information from agent. Reason: in some (really rare) cases node might have been gone so long that master has removed node information.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #357, opened 2015-12-08
Secrets management aka Vault

### LINKED ISSUE #354: Secrets management
We need to have convenient way for developers to manage their secrets. The proposed solution: **Kontena Vault**.

The idea is that users can manage their secrets with kontena cli command `kontena vault`. The sub commands should include at least:

* list - list all secrets (keys)
* add - add a secret to vault
* remove - remove secret from vault
* update - update secret in vault
* get - retrieve secret from vault
* export - retrieve all secrets from vault

All vault operations should naturally use audit trail so it is possible to see who has been accessing vault.

The secrets stored in vault could be used in `kontena.yml` file like this:

```
app:
  image: some/image
  vault:
    - env:USERNAME
    - env:PASSWORD 
  environment:
    FOO: bar
```

With the above example, the service is started with 3 environment variables: `USERNAME`, `PASSWORD` and `FOO`. `USERNAME` and `PASSWORD` environment variable values are fetched from vault.

The above example is just an idea and may be improved. For example, we might want to have possibility to rename the env vars to something else.

jakolehm:
This PR implements Kontena Vault (aka secrets). Fixes #354 

**cli**
```
kontena vault
Usage:
    kontena vault [OPTIONS] SUBCOMMAND [ARG] ...

Parameters:
    SUBCOMMAND                    subcommand
    [ARG] ...                     subcommand arguments

Subcommands:
    write                        Write a secret
    list, ls                      List secrets
    read                          Read a secret
    remove, rm                    Remove secret

```

**kontena.yml**

````
api:
  image: example/api:latest
  secrets:
    - secret: mysql_rds_password
      name: MYSQL_PASSWORD
      type: env
  environment:
    - MYSQL_USER=admin
```


hans-d:
besides code, also provide documentation around if/how secrets are protected.

jakolehm:
I would like to ensure that secrets are encypted while in flight and at rest. But I also would like to have those secrets as a env variables (not as secrets in a volume) because it's so much easier to consume those. This is not so easy to accomplish with docker atm...

One option for encryption: http://rocketjob.github.io/symmetric-encryption/index.html

jakolehm:
First implementation will encrypt values to master database. Secrets are exposed to services as a environment variables (without any special encryption).

jakolehm:
@nevalla @miskun ping

miskun:
Looking good!

PR #358, opened 2015-12-08
Multiple masters on cli

jnummelin:
This PR implements support for working with multiple masters on cli.
Commands:
```
$ kontena master
...
    list, ls                      List masters where client has logged in
    use                           Switch to use selected master
```
```
$ kontena login --help
Usage:
    kontena login [OPTIONS] URL

Parameters:
    URL                           Kontena Master URI

Options:
    -n, --name NAME               Local alias name for the master. Default default

```
The -n flag specifies the local name for the master when logging in.

The local config file has the following model:
```
{
  "current_server": "test",
  "servers": [
    {
      "url": "https://1.2.3.4:8443",
      "token": "kontena-xyz",
      "grid": "test-grid",
      "name": "test"
    },
    {
      "url": "https://other.master.com:8443",
      "token": "kontena-zyx",
      "grid": "test-grid",
      "name": "other"
    }
  ]
}
```
The new functionality takes care of "migrating" the client config on first use and configures the currently know only master with a alias name `default`and naturally sets that as current_server.

There's few things I still intend to work on:
- Clean up the common.rb code, it's a bit messy atm
- Some more testing, especially on the common functionality

What about logout, should logout actually forget the master totally? Now it just removes the token but leaves the master info into the config file. Or should there be some command like: `kontena mster forget my-master`which then fully removes the info?

Signed-off-by: Jussi Nummelin <jussi.nummelin@gmail.com>

### DIFF
--- a/cli/lib/kontena/cli/common.rb
+++ b/cli/lib/kontena/cli/common.rb
@@ -9,7 +9,7 @@ module Kontena
       end
 
       def require_token
-        token = ENV['KONTENA_TOKEN'] || settings['server']['token']
+        token = ENV['KONTENA_TOKEN'] || current_master['token']
         unless token
           raise ArgumentError.new("Please login first using: kontena login")
         end
@@ -25,7 +25,6 @@ module Kontena
 
           @client = Kontena::Client.new(api_url, headers)
         end
-
         @client
       end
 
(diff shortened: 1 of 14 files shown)


jnummelin:
Althought this is a bit WIP stuff I wanted to push this PR to get some comments already. :)

jakolehm:
This is so handy!

thank you

nevalla:
Looks great! There is also `kontena whoami` command that displays current user and master server. It would be nice if alias name could be shown there too. Or what do you think?

hans-d:
Kontena master is now used for two things:
- master machine management, eg kontena master vagrant ...

hans-d:
And master login management

jakolehm:
I think it's not a big problem to have multiple different subcommands for `master`. Other option is to add new root level command... Imho we should avoid that unless there is really good reason.

jnummelin:
@nevalla I'll add the info to the `whoami`command also. What about the logout, should the cli forget the whole master or just the token?

jakolehm:
@jnummelin I would prefer that `logout` only wipes the token.

jnummelin:
@jakolehm @nevalla `whoami`command updated now to show current master and logout nullifies the token. Please take a look.

jakolehm:
Looks good!

jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #366, opened 2015-12-16
node list --all to show all nodes for all grids

hans-d:


jakolehm:
What kind of use case is this solving?

hans-d:
As the person responsible responsible for all IT infrastructure

1) I want to verify that all nodes are connected / online
and I don't care about the grids a certain node belongs to
and want to a have a single command to list all nodes
instead of having to write a script to get me all nodes
so I can use a simple `kontena <list all nodes> | grep offline`

2) I want to verify that a certain node is present
and I don't care or don't know about the grid that node belongs to
and want to a have a single command to list all nodes
instead of having to write a script to get me all nodes
so I can use a simple `kontena <list all nodes> | grep <mynode>`

3) I want to lookup to what grid certain node is attached to
and want to a have a single command to list all nodes
instead of having to write a script to get me all nodes
so I can use a simple `kontena <list all nodes> | grep <mynode>`

jakolehm:
Ok, sounds good to me. @nevalla @miskun what do you think?

nevalla:
Looks good to me too. I see a real use case for this feature.

miskun:
I'm wondering what kind of implications this will have on user access control model? I think we want to limit visibility of grid(s) to selected users only. Therefore, the access control scope for this kind of operation must be system wide admin access. I hope this feature don't break this idea.

hans-d:
It should follow the scope for the user - only showing the nodes for the grids they have access to.

For sysadmins, they should have (read) access to the grids they are responsible for - eg in the case of a shared master.

miskun:
Alright, sounds good. Another issue I'm having with this feature is the current grid scope. Now, everything in `kontena node` namespace is in the scope of current grid. This switch is breaking the scope. So, I'm thinking if this command could be `kontena master nodes` or something similar... What do you think?

jakolehm:
I think we should move this feature to `kontena master nodes` (or similar). Reason: `kontena node list` is already scoped to current grid and this option violates that pattern.

miskun:
Perfect timing @jakolehm :)

hans-d:
would be less logical for me, as `kontena master` is managing the master (creationg, logins - see #358).
And for listing all grids on a master it's not `kontena master grids` but `kontena grid list`. 

I would argue that `kontena node` is by default scoped to the current grid, but expanded by this option.

miskun:
Alright, I think this argument is good enough. In fact, I'm preferring this over my previous idea. I'd say we merge this.

jakolehm:
@hans-d good point.

hans-d:
https://www.youtube.com/watch?v=NmPhaG1ud38

PR #370, opened 2015-12-19
Introduce app scale subcommand

nevalla:
App scale command was previously implemented but never introduced as a subcommand. This PR will fix this issue.


jakolehm:
LGTM

PR #372, opened 2015-12-20
Add --secret option to service create/update

jakolehm:
```
$ kontena service create --secret <secret>:<name>:<type> myapi my/api:latest
```


miskun:
Just realized we can't allow ":" in the secret/name/type. But I guess that's ok?

jakolehm:
Currently we allow only letter, number, underscore and hyphen chars.

miskun:
Alright! Sounds good to me!

PR #373, opened 2015-12-20
Append values to env if secrets have same name

jakolehm:
```
lb:
  image: kontena/lb:latest
  secrets:
    - secret: DOMAIN_NET_CERT
      name: SSL_CERTS
      type: env
    - secret: DOMAIN_COM_CERT
      name: SSL_CERTS
      type: env
```
creates environment variable `SSL_CERTS` that includes both secrets.


nevalla:
LGTM!

PR #377, opened 2015-12-22
[WIP] Event stream API

nevalla:
This PR implements API for grid events.

Current use case scenario is that 

1. Client requests access token for event stream (https://kontena-master/v1/grids/#{grid_name}/events)
2. Client opens WebSocket connection with given access token and optional event types that he wants to receive (wss://kontena-master/v1/grids/#{grid_name}/events?access_token=xxxyyyzzzz&event_types=grid,service,node)
3. Client receives events via WebSocket connection

At this moment the following event types are supported:

* grid [update]
* node [create, update, delete]
* service [create, update, delete]
* container [create, update, delete]


jnummelin:
Are we planning to continue the work at later date from this PR or should we create fresh start? If the latter we can close this.

PR #417, opened 2016-01-04
Do not allow to delete grid if nodes exist

### LINKED ISSUE #316: It should not be possible to remove a grid while there are still nodes on it
with 0.10.0.beta its possible to remove a/the last/the only grid while there're still nodes attached:

```
> kontena grid remove master
removed master
 > kontena node vagrant terminate node
Please select grid first using: kontena grid use <grid name>
```

nevalla:
Fixes #316


jakolehm:
LGTM

PR #421, opened 2016-01-06
Improve server performance & mem usage

jakolehm:
- [x] use worker pools
- [x] write logs & stats as fast as possible


nevalla:
LGTM!

PR #425, opened 2016-01-07
Pass external-registry add values as options

jakolehm:
This allows to add extra long tokens as a password (for example with Google Container Registry)


jakolehm:
@nevalla ping

nevalla:
LGTM!

PR #450, opened 2016-01-14
Send container information asap to master (don't wait for weave)

jakolehm:
Sometimes it might take a long time to pull new weave release and meanwhile master does not see any containers from node. This PR fixes this situation by sending container information right after websocket connection is established.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #451, opened 2016-01-14
Check that weave is running when creating a service

jakolehm:
If weave image pull takes a long time it will raise on error when master sends service pod create request. This PR fixes this problem by adding a check that weave is actually running before proceeding with request.


jakolehm:
@nevalla ping

nevalla:
LGTM!

PR #454, opened 2016-01-15
Search alternate Dockerfile from build context

nevalla:
With this PR alternate Dockerfile searched from build context instead of directory where the command is executed.


nevalla:
ping @jakolehm

jakolehm:
LGTM

PR #466, opened 2016-01-22
Dc's with private networks can't see each other

jakolehm:
Reason: grid is weaved together using private addresses. This PR adds possibility to add/remove custom peer addresses per grid (basically public ip that can be seen/connected from outside of datacenter).


nevalla:
LGTM!

PR #470, opened 2016-01-25
Update node information when connected / with interval

### LINKED ISSUE #468: performing 'node remove' before terminating the
Had a grid with three nodes, performed `kontena node remove` on each node, without first doing a `kontena node digitalocean terminate` 
When listing nodes I ended up with this:
```
$ kontena node list
Name                           OS                                       Driver          Labels                         Status    
                                ()                                                                                     offline   
                                ()                                                                                     offline   
                                ()                                                                                     offline   
$ kontena grid list
Name                           Nodes    Services     Users     
badcatlab *                    3        2            1         
```
Seems there is no way to fully remove the nodes from this point using the cli.

jakolehm:
Fixes #468

### DIFF
--- a/agent/lib/kontena/node_info_worker.rb
+++ b/agent/lib/kontena/node_info_worker.rb
@@ -16,15 +16,18 @@ module Kontena
     # @param [Queue] queue
     def initialize(queue)
       @queue = queue
+      Pubsub.subscribe('websocket:connected') do |event|
+        self.publish_node_info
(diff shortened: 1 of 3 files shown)


jakolehm:
This will bring node back to master with full information if it's still running. Basically there should not be any "zombie/empty" nodes after this fix.

nevalla:
LGTM!

PR #482, opened 2016-01-31
Role based authorization

nevalla:
This PR implements roles based authorization to Kontena Master.

Implementation relies on [Authority](https://github.com/nathanl/authority) gem.
The most notable changes are:

* new `Role` model with `name` and `description` fields
* `Master admin` and `Grid admin` roles are created.
* new API endpoints
  * `GET /v1/users/`
  * `POST /v1/users/:user/roles`
  * `DELETE /v1/users/:user/roles/:role`
* following permission rules are enabled
  * only `Master admin` can invite users to master server
  * only `Master admin` can add and remove roles to/from users
  * only `Master admin` and  `Grid admin` can add add users to grid


server/app/routes/v1/users_api.rb
```diff
@@ -8,6 +8,24 @@ class UsersApi < Roda
 
     route do |r|
       r.post do
+        r.on 'roles' do
+          validate_access_token
+          require_current_user
+
+          params = parse_json_body          
+          params[:current_user] = current_user
+          params[:role_name] = params['role']
```
jakolehm: Any reason for this? Should we just pick a good name for the "role" attribute and use that in every place.

server/app/initializers/authority.rb
```diff
@@ -0,0 +1,12 @@
+require 'authority'
+
+Authority.configure do |config|
+  config.abilities =  {
```
jakolehm: Is it possible to define abilities per model in Authority? Somehow it feels funny to have global abilities like `assignable` or `deployable`.

server/app/models/user.rb
```diff
@@ -21,4 +24,17 @@ def name
...
   end
 
+  ##
+  # @param [String] role
+  def in_role?(role)
+    self.roles.where(name: role).exists?
+  end
+
+  def master_admin?
+    self.in_role?('Master admin')
```
jakolehm: My gut feeling says that pointing to these roles by names is not good idea. At least hide them behind some variables or constants for now.

server/app/initializers/authority.rb
nevalla: Apparently no :( Authority::UserAbilities reads those verbs and creates dynamically User.can_<verb>? methods. https://github.com/nathanl/authority/blob/master/lib/authority/user_abilities.rb#L11-L18

server/app/models/user.rb
nevalla: I have changed this so that we have constants now for those roles: https://github.com/kontena/kontena/blob/feature/user_permissions/server/app/models/user.rb#L9-L10

jakolehm:
Should we change `name` -> `description` and `identifier` -> `name`. Why? Because every other model has name as an identifier.

jakolehm:
LGTM! Please squash commits before merging...

PR #492, opened 2016-02-07
Add --force-deploy option to app/service deploy commands

jakolehm:
This PR adds `--force-deploy` option to `kontena app/service deploy` subcommands. This forces master to update service timestamp and by doing this, will force agent to replace container(s).


nevalla:
LGTM!

PR #496, opened 2016-02-07
Add possibility to override current grid with --grid option

jakolehm:
Example:

```
$ kontena app deploy --grid production-grid
```


nevalla:
Looks good. I think tests are failing because 'kontena/cli/grid_options' is not required when running tests.

PR #497, opened 2016-02-07
Add %{grid} interpolation when parsing kontena.yml

jakolehm:
This enables per grid dns image push to internal registry (and probably something else):

Example:
```
my-app:
  image: registry.%{grid}.kontena.local/my-app:latest
```


nevalla:
:+1:

PR #500, opened 2016-02-08
Allow to set custom weave peers via env to agent

jakolehm:
Sometimes it's necessary to give custom weave peer ips to agent that overrides peers coming from the master. This is only necessary when network topology is complex and weave should not try to connect to all other peers.


nevalla:
LGTM

PR #513, opened 2016-02-10
more consistent remove, add short versions of commands

hans-d:


miskun:
I agree, we should be more consistent with sub commands. I suggest we do following:

* Keep `delete` as is. Add deprecation notice for this command. Suggest using `remove` instead.
* Add `remove` (with short `rm`) command that is used everywhere to delete/remove stuff.

Adding `ls` short for all `list` sub commands is also good addition for making the CLI more consistent.

hans-d:
updated as per @miskun comments

PR #526, opened 2016-02-14
Integrate examples from kontena/examples repository

jakolehm:


hans-d:
would prefer to keep it in a separate repository, esp because these target users of kontena, while the main repo is about the inner workings...

jnummelin:
I also think we should keep the examples in separate repo. @jakolehm Should we close this or do you think we should "merge" examples to the main repo?

jakolehm:
Let's close this for now.. too old branch/pr anyway.

Issue #521, opened 2016-02-13
move user related sub commands


miskun:
User related subcommands are cluttering Kontenan CLI root level commands, yet they are used very rarely. Suggestion: move all user related commands under `user` subcommand. We may consider leaving  `login` and `whoami` to root level.

Issue #512, opened 2016-02-10
grid list-users --> grid users list


hans-d:
`kontena grid` has some user specific subcommands: `list-users`, `add-users`, `remove-user`.

Suggest to make this into another subcommand:
- `kontena grid users list`
- `kontena grid users add`
- `kontena grid users remove`

and deprecate the current ones (like `kontena deploy`)

jakolehm:
I agree :)

Related PR's: 
- https://github.com/kontena/kontena/pull/505 
- https://github.com/kontena/kontena/pull/482

PR #528, opened 2016-02-15
Add service add/remove-secret sub-commands

jakolehm:


cli/lib/kontena/cli/service_command.rb
```diff
@@ -31,10 +35,14 @@ class Kontena::Cli::ServiceCommand < Clamp::Command
...
   subcommand "logs", "Show service logs", Kontena::Cli::Services::LogsCommand
   subcommand "stats", "Show service statistics", Kontena::Cli::Services::StatsCommand
+  subcommand "monitor", "Monitor", Kontena::Cli::Services::MonitorCommand
+
   subcommand "envs", "Show environment variables", Kontena::Cli::Services::EnvsCommand
   subcommand "add-env", "Add environment variable", Kontena::Cli::Services::AddEnvCommand
   subcommand "remove-env", "Remove environment variable", Kontena::Cli::Services::RemoveEnvCommand
-  subcommand "monitor", "Monitor", Kontena::Cli::Services::MonitorCommand
+
+  subcommand "add-secret", "Add secret from Vault", Kontena::Cli::Services::AddSecretCommand
```
hans-d: why not as `'kontena service secret add`  and `kontena service secret remove`? (see #521 and #512)

jakolehm: simple reason: we will refactor sub-commands to new (better) style in a one PR .. target for that is post 0.12

jakolehm:
ping @nevalla @miskun

miskun:
I'm having same thinking as @hans-d but I guess we can refactor sub-commands names also later. Since the sub commands naming plan is still "under construction" I'm ok to go ahead now with this implementation. +1

nevalla:
Yes agree with @miskun. Let's go now with this implementation.

PR #532, opened 2016-02-20
Add support for digitalocean floating ip's

### LINKED ISSUE #531: CoreOS Digital Ocean Floating IPs special configuration
Hello,

First I would like to thank for sharing Kontena work with all!

While testing Kontena, noticed that was not possible to use Digital Ocean's [Floating IP](https://www.digitalocean.com/community/tutorials/how-to-use-floating-ips-on-digitalocean) feature with Kontena.

After some initial investigation and confirmation from Digital Ocean support staff, special settings are required for CoreOS to allow usage of the Floating IP:

https://www.digitalocean.com/community/tutorials/how-to-enable-floating-ips-on-an-older-droplet#coreos

I'm not sure the best approach to automate this: either is by the usage of special option during node creation (or update) or if each droplet can be setup with the same Floating IP and that will dynamically work.

Please let me know if further details are required or possible access to DO account for testing.

Thank you.
:heart: :heart: :heart:

jakolehm:
Fixes #531

### DIFF
--- a/cli/lib/kontena/machine/digital_ocean/cloudinit.yml
+++ b/cli/lib/kontena/machine/digital_ocean/cloudinit.yml
@@ -24,6 +24,13 @@ write_files:
       nameserver 172.17.43.1
       nameserver 8.8.8.8
       nameserver 8.8.4.4
+  - path: /opt/bin/floating_ip.sh
+    permissions: 0755
+    owner: root
+    content: |
+      #!/bin/sh
+      IP=$(curl -s http://169.254.169.254/metadata/v1/interfaces/public/0/anchor_ipv4/address)
+      /usr/bin/ip addr add $IP/16 dev eth0
 coreos:
   units:
     - name: 10-weave.network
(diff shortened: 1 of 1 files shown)


nevalla:
:+1:

luislavena:
Thank you @jakolehm 

Is not clear to me when this service will be executed (it says `oneshot` so I assume is run during boot).

Since a Floating IP can be associated at any point on the droplet's life, perhaps is possible to run this periodically?

Let me know so I can contribute some documentation for this if manual execution is required.

Thank you.

jakolehm:
Yes, it is executed only during boot. It just registers anchor (private) ip to CoreOS that can be then later associated to floating ip. So I think that we need to execute this only once during boot phase and after that floating ip's should "just work" without any additional tricks (at least it worked in my tests).

mtrofimm:
Thank you @jakolehm :+1: 

btw, There is a good HA tutorial using floating IPs:
https://www.digitalocean.com/community/tutorials/how-to-create-a-high-availability-haproxy-setup-with-corosync-pacemaker-and-floating-ips-on-ubuntu-14-04.

It would be great to include such feature inside kontena stack :)

luislavena:
@mtrofimm I think that can be deployed as a service to be run on all your grid instances :wink: 

The only thing will be how to determine each `droplet_id` and be able to inform that to the script that assigns the floating IP :smile: 

Cheers.

PR #536, opened 2016-02-21
Assign first user to master admin role

nevalla:


server/app/models/role.rb
```diff
@@ -20,4 +20,8 @@ class Role
   def master_admin?
     self.name == MASTER_ADMIN_ROLE
   end
-end
\ No newline at end of file
+
+  def self.master_admin
+      self.find_by(name: MASTER_ADMIN_ROLE)
```
jakolehm: wrong indent style

nevalla: good catch and now it's fixed

jakolehm:
Please rebase/squash and this is good to merge!

PR #539, opened 2016-02-23
Implement kontena master users commands

nevalla:
This PR implements initial `kontena master users` commands:

- `kontena master users invite <email1> <email2> ...`
- `kontena master users add-role <role> <user1> <user2> ...`
- `kontena master users remove-role <role> <user1> <user2> ...`


jakolehm:
Why `users` instead of `user`?

nevalla:
I chose `users` because commands takes user/email array as parameter (updated to description). @miskun any opinions?

miskun:
I'm always happy to see plural :) :+1:

jakolehm:
:+1:

PR #540, opened 2016-02-23
Fix DigitalOcean node termination

### LINKED ISSUE #537: CLI: Terminate Digital Ocean node reports error (404)
Hello folks,

When trying to destroy all the nodes of a grid, first I ensured no app/service was defined or running on the nodes, and then proceed to use `node digitalocean terminate`:

```
$ kontena node digitalocean terminate  --token $DO_TOKEN frosty-field-48
Terminating DigitalOcean droplet frosty-field-48 ... fail
404: {"id":"not_found","message":"The resource you were accessing could not be found."}
```

This happened with every node that was terminated, but looking at Digital Ocean dashboard, I see the nodes were terminated properly.

Perhaps the CLI is attempting to grab the node status and it has already been removed from DO?

Thank you
:heart: :heart: :heart:

nevalla:
DigitalOcean deletes Droplet immediately on delete request, so we don't need to poll droplet status anymore.

Fixes #537


jakolehm:
LGTM

PR #541, opened 2016-02-23
Retry weave, etcd and cadvisor starts on docker exception

nevalla:
When using devicemapper and trying to restart  `kontena-agent`, restart of `kontena-etcd` and `kontena-cadvisor` containers often fails with error like this: 
`Docker::Error::ServerError: Cannot destroy container 7203c7163a4fbb1fff2a76b6473f684c469f07e417e83eff75ce29947cb4a24a: Driver devicemapper failed to remove root filesystem 7203c7163a4fbb1fff2a76b6473f684c469f07e417e83eff75ce29947cb4a24a: Device is Busy`

With this PR `kontena-agent` will retry to start those containers if they are raising Docker::Error::ServerError at first time.

### DIFF
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -16,11 +16,18 @@ module Kontena::Launchers
     end
 
     def start
+      retries = 0
       begin
         start_cadvisor
+      rescue Docker::Error::ServerError => exc
+        if retries < 4
+          retries += 1
+          sleep 0.25
+          retry
+        end
+        log_error(exc)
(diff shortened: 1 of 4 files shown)


jakolehm:
Just wondering should there be some kind of delay and possibly more retries? Otherwise looks good!

nevalla:
I was wondering the same thing. Any rough estimations how much time we should give devicemapper to recover?

jakolehm:
Not a devicemapper expert but I think we should wait for couple of tenths and retry few times.

jakolehm:
Needs rebase.. but otherwise looks good!

PR #549, opened 2016-03-03
Fix daemon strategy instance count with affinity

jakolehm:
If daemon strategy was used with affinity it did not result in a correct number of service instances.


nevalla:
Looks good!

PR #551, opened 2016-03-04
Fix ubuntu package missing vault config

jakolehm:
This pr also improves how ubuntu packages are built + requires more stricter docker versions.


nevalla:
:+1:

PR #552, opened 2016-03-04
Refactor cli specs

nevalla:
This PR simplifies cli spec files when operating with YAML files by introducing fixtures directory and mechanism.

Also `Dir.home` path is pointed to `/tmp`. Before that test execution could override kontena cli config file.


jakolehm:
:+1:

PR #562, opened 2016-03-12
docs: fix registry redeployment commands

msokk:
Just a docs <-> CLI mismatch I found.


msokk:
Found also that registry needed secret setup to read the password from vault.

Issue #579, opened 2016-03-23
kontena grid logs follow option to -t like in app logs


matti:
```
$ kontena grid logs --help
    -f, --follow                  Follow (tail) logs (default: false)
```

```
$ kontena app logs --help
    -t, --follow                  Follow (tail) logs (default: false)
```

matti:
while at it, also change the --follow to --tail for both.

PR #580, opened 2016-03-23
-t/--tail instead -f/--follow

matti:
`kontena app -f` specifies kontena compose file, and it should be like that.

This would also be in-line with `heroku logs -t/--tail`

https://github.com/kontena/kontena/issues/579


jakolehm:
Me likes!

PR #588, opened 2016-04-02
Pre-build hooks

### LINKED ISSUE #479: enh: add pre-build hook
Currently, images can be build directly from a Dockerfile and pushed to a registry.
Would be nice if something can be started before the `kontena app build` does the docker building

Use case:
- have a dockerfile BUILD that has various extras needed to do the build, making it quite bloated
- have a dockerfile RUN that is much smaller and picks up the results from runnnig BUILD (eg via a COPY).

Currently using make to manage this build. Happy just to have some pre-build hook to invoke this makefile.

jnummelin:
This PR adds named pre-build hooks. The hooks are executed in the order defined and the build is aborted if any of the commands fail.

resolves #479


jakolehm:
Excellent! Could you also add documentation for this (at least to kontena.yml reference)?

jnummelin:
Sure, I'll add some docs in the evening

nevalla:
Awesome :+1:

jnummelin:
Added some docs to the kontena.yml reference.

jakolehm:
Looks good! Should we include this to 0.12 release? 

/cc @miskun @nevalla

PR #589, opened 2016-04-02
Login email into local storage

### LINKED ISSUE #525: add username to .kontena_client.json
store the username along with the server details in `.kontena_client.json`. Now `kontena whoami` needs to retrieve this part from the server. when the master is down, this results in a timeout.
Also, when showing some details in the promt (eg https://gist.github.com/jakolehm/0edf6b90b205aa34d895) it would be nice to have the username there without requiring a round trip

jnummelin:
This PR adds login email into local `~/kontena_client.json`file. If the email is not known when executing the `kontena whoami` command it is added. Otherwise the email is added during login.

Fixes #525

### DIFF
--- a/cli/lib/kontena/cli/login_command.rb
+++ b/cli/lib/kontena/cli/login_command.rb
@@ -24,7 +24,7 @@ class Kontena::Cli::LoginCommand < Clamp::Command
     response = do_login(email, password)
 
     if response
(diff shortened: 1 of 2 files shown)


jakolehm:
LGTM

nevalla:
LGTM

PR #594, opened 2016-04-05
Switch from fog-aws to aws-sdk

jakolehm:
Fog-aws has nasty dependencies (nokogiri) that require user to install development packages. Aws-sdk should make installation so much easier.

This PR also adds minor tweak that generates self-signed ssl certificate for master if user does not provide any cert.


jakolehm:
@nevalla @miskun ping

nevalla:
LGTM!

PR #596, opened 2016-04-07
Authorize grid actions

nevalla:
This PR adds authorization of grid actions (create, update, delete). With this PR only `master_admin` can create and delete grids. `grid_admin` can also update grid.


jakolehm:
Looks good!

PR #599, opened 2016-04-08
Added restart flag for services

jnummelin:
This PR adds restart flag for service containers.

Currently the defaulting to `always` policy happens in agent, would it make more sense to have it on master model level?

### DIFF
--- a/server/app/mutations/grid_services/create.rb
+++ b/server/app/mutations/grid_services/create.rb
@@ -114,8 +114,8 @@ module GridServices
       if self.strategy && !self.strategies[self.strategy]
         add_error(:strategy, :invalid_strategy, 'Strategy not supported')
       end
-      @re ||= Regexp.union(/^no$/, /^on-failure/, /^always$/, /^unless-stopped$/)
-      if self.restart && self.restart.match(@re).nil?
+      re = Regexp.union(/^no$/, /^on-failure/, /^always$/, /^unless-stopped$/)
+      if self.restart && self.restart.match(re).nil?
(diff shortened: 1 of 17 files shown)


jakolehm:
Yes, default should come from master.

One thing missing from PR is how master handles service instance failures if restart-policy is something else than `always`. Now if it sees that some instances are not properly running it will try to re-redeploy them.

jnummelin:
Yes, absolutely. Kinda forgot to slab WIP flag on this. :)

jnummelin:
@jakolehm Added the handling for balancing job. I think it should play pretty well with the docker restart policy mechanism: https://docs.docker.com/engine/reference/run/#restart-policies-restart

So only services with policies `always` or `unless-stopped` should be restarted/balanced by the master.

server/app/mutations/grid_services/create.rb
```diff
@@ -113,6 +114,10 @@ def validate
       if self.strategy && !self.strategies[self.strategy]
         add_error(:strategy, :invalid_strategy, 'Strategy not supported')
       end
+      @re ||= Regexp.union(/^no$/, /^on-failure/, /^always$/, /^unless-stopped$/)
```
jakolehm: I think we don't need instance variable here?

jnummelin: True, I'll remove that.

jakolehm:
I think we need to indicate service state somehow to user so that "fully completed" services do not show as "running".

jnummelin:
I think service restarting needs anyways bigger thinking going forward. I'll close this one as it's getting dusty anyway...

PR #603, opened 2016-04-14
Fix node used memory calculation

jakolehm:
0.12.0 shows grid used memory with buffers and caches. This is not completely wrong but confuses users to think that grid is running out of memory. This PR fixes this by not including buffers/caches.


nevalla:
Looks good!

PR #606, opened 2016-04-16
Add memory filter to scheduler

jakolehm:
This PR implements new scheduler filter that calculates node memory usage and compares that to service instance memory requirements. If node does not have enough free memory it will be filtered out.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #608, opened 2016-04-16
Display public ip on node list

nevalla:
This PR displays node's public ip address on node list.

```
$ ./bin/kontena node list
Name                        Public IP            Status     Labels                                  
proud-sun-9                 8.8.8.8              online     -       
```


jakolehm:
Not completely sold on this but if we want to add these columns then we need to make column width  dynamic.

nevalla:
I will close this now. Let's figure out the ultimate solution first.

PR #616, opened 2016-04-19
larger vault list column widths

matti:
This sets total width to 80 chars.


matti:
...will be back with less ähellys.

PR #619, opened 2016-04-19
Sets kontena vault list width to be 80 characters and not cropping

matti:
See issue #615 

If the key is longer than 54, then the whole line will be longer (with the timestamp at the end). This ensures that no keys get cropped.


jakolehm:
Looks good, thanks!

PR #620, opened 2016-04-19
Do not overwrite service environment variable if value is empty

### LINKED ISSUE #614: App deploy overwrites envs that are not part of kontena.yml
We should provide some sane way to define env variables outside of kontena.yml so that they are not wiped away when user adds `environment` block to `kontena.yml`.

jakolehm:
This PR makes it possible to set environment variables outside of `kontena.yml`. User just have to document environment variable keys in `kontena.yml` and use cli to set the actual values.

**kontena.yml:** (inside folder named "demo")
```
app:
  image: registry.kontena.local/demo-app:latest
  environment:
    - MY_VAR
    - OTHER_VAR
```
**cli:**
```
$ kontena service add-env demo-app MY_VAR=foo
$ kontena service add-env demo-app OTHER_VAR=bar
```

Fixes #614


nevalla:
LGTM!

PR #626, opened 2016-04-21
Fix too aggressive overlay cidr cleanup

jakolehm:
If deploying a service instance is taking long time we might accidentally release ip that has been reserved.


nevalla:
Looks good!

PR #627, opened 2016-04-21
Fix image puller cache invalidation on new deploy using same image tag

### LINKED ISSUE #577: Agent does not always pull fresh image layers
Currently agent will cache image pulls for 10 minutes. During this cache period agent will not fetch new image layers and this might cause confusion. Cache should be tied for example to deploy revision.

jakolehm:
Fixes #577


nevalla:
Looks good!

PR #632, opened 2016-04-26
Return nil for current_grid if Kontena Master settings not present

### LINKED ISSUE #628: App build requires master configuration
`kontena app build` requires master url, grid and token information.. it should not require those.

nevalla:
Fixes #628


jakolehm:
LGTM

PR #641, opened 2016-04-27
fix (CLI) list vault secrets & service envs by name

christofferh:
Before
```
$ kontena vault list
NAME                 CREATED AT
B_SECRET        2016-01-15T14:48:21.777Z
A_SECRET        2016-02-15T14:48:38.611Z
```
After
```
$ kontena vault list
NAME                 CREATED AT
A_SECRET        2016-02-15T14:48:38.611Z
B_SECRET        2016-01-15T14:48:21.777Z
```


jakolehm:
Nice improvement, thanks!

PR #644, opened 2016-05-01
Grid trusted subnets

jakolehm:
This PR adds support for trusted subnets within a grid. 

> If some of grid nodes are colocated in a trusted network (for example within the boundary of your own datacenter) you can add subnets to grid trusted subnet list. This disables data plane encryption within trusted subnet and switches overlay to faster (near-native) mode as an optimization.


jakolehm:
@nevalla ping

nevalla:
Looks good!

PR #647, opened 2016-05-03
Fix memory parsing errors

### LINKED ISSUE #629: service mem_limit is not always a string
In https://github.com/kontena/kontena/blob/master/cli/lib/kontena/cli/services/services_helper.rb#L276 the mem_limit is assumed to be a string (`@param [String] memory`)

But if one follows the [kontena docs](https://kontena.io/docs/references/kontena-yml) and puts `mem_limit: 1000000000` then this method will explode in `kontena app deploy --no-build` with

```
undefined method `end_with?' for 1000000000:Fixnum
```

even if this case is support and handled at end end of the method (no suffix).

There are two options to fix this: 
a) change the assumption so that the string is ensured with `.to_s` - might be confusing later on and is a code smell
b) change the YAML result parsing so that memory is always parsed as a string - don't know if something else will start failing because of this.

How should this be fixed?

jakolehm:
Fixes #629 and improves a documentation.

### DIFF
--- a/cli/lib/kontena/cli/apps/deploy_command.rb
+++ b/cli/lib/kontena/cli/apps/deploy_command.rb
@@ -149,8 +149,8 @@ module Kontena::Cli::Apps
       data[:container_count] = options['instances']
       data[:links] = parse_links(options['links'] || [])
       data[:ports] = parse_ports(options['ports'] || [])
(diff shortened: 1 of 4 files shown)


nevalla:
LGTM

PR #653, opened 2016-05-04
Fix linked service deletion on app rm command

### LINKED ISSUE #314: Kontena app rm command does not remove services in correct order
When trying to remove app with `kontena app rm` linked services does not be removed first and error is returned.

nevalla:
When deleting application (`kontena app rm`) deletion of service that is linked to another service may still fail despite #576. 

This PR changes behavior so that instead of fixed wait time CLI is now polling when service is deleted on the Kontena Master side before deleting other services.

This PR also unifies the output with `kontena app deploy` command

Fixes #314


jakolehm:
LGTM

PR #657, opened 2016-05-04
New service deploy option: interval

### LINKED ISSUE #656: Erosion resistance
https://devcenter.heroku.com/articles/erosion-resistance

jakolehm:
Fixes #656

### DIFF
--- a/cli/lib/kontena/cli/services/services_helper.rb
+++ b/cli/lib/kontena/cli/services/services_helper.rb
@@ -319,6 +319,20 @@ module Kontena
           secrets
         end
 
+        # @param [String] time
+        # @return [Integer]
+        def parse_relative_time(time)
+          if time.end_with?('min')
+            time.to_i * 60
+          elsif time.end_with?('h')
+            time.to_i * 60 * 60
+          elsif time.end_with?('d')
+            time.to_i * 60 * 60 * 24
+          else
+            memory.to_i
+          end
+        end
+
         def int_to_filesize(int)
           {
             'B'  => 1000,
--- a/cli/lib/kontena/cli/services/services_helper.rb
+++ b/cli/lib/kontena/cli/services/services_helper.rb
@@ -52,6 +52,7 @@ module Kontena
           puts "  deploy_opts:"
           puts "    wait_for_port: #{service['deploy_opts']['wait_for_port'] || '-'}"
           puts "    min_health: #{service['deploy_opts']['min_health']}"
+          puts "    interval: #{service['deploy_opts']['interval'] || '-'}"
           puts "  dns: #{service['name']}.#{grid}.kontena.local"
 
           if service['affinity'].to_a.size > 0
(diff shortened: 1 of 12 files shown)


jakolehm:
I don't really like name of this feature/option (`redeploy_after`). Could we come up with something better?

/cc @nevalla @miskun @matti

cli/lib/kontena/cli/services/services_helper.rb
```diff
@@ -319,6 +319,20 @@ def parse_secrets(secret_opts)
...
+        # @return [Integer]
+        def parse_relative_time(time)
+          if time.end_with?('min')
+            time.to_i * 60
+          elsif time.end_with?('h')
+            time.to_i * 60 * 60
+          elsif time.end_with?('d')
+            time.to_i * 60 * 60 * 24
+          else
+            memory.to_i
```
nevalla: should be time.to_i

nevalla:
How about redeploy_interval or auto_deploy_interval?

cli/lib/kontena/cli/services/services_helper.rb
jakolehm: Good catch, was going to write tests for this PR :)

jakolehm:
Interval makes sense, maybe something like this:

```
deploy:
  interval: 8h
```

nevalla:
I like it. Simple and clear.

christofferh:
Isn't the semantics more of `restart` rather than `redeploy`?

jakolehm:
@christofferh, restart does not help with "erosion" so much because it just restarts the actual container. Deploy will pull latest version of image and swap container to fresh one.

christofferh:
@jakolehm, IMO this should be an idempotent action. Automatically pulling for the latest version is something I would consider dangerous and not part of 'erosion resistance'.

jakolehm:
Part of erosion resistance is that app/service should receive security updates. This can't be done without pulling new version of defined "image:tag". 

If this is not acceptable I think we should introduce separate option for image pulls (always, on_demand).

christofferh:
That's true if you're not in complete control of the underlying OS of the service. e.g. the case of Heroku.

Yes, it should to be optional with image pulls if `redeploy` is the way you guys wanna go. But I just don't think it's the right way to go 😢 

(You guys are awesome! 💯 )

matti:
This is a nice feature for somebody who is adventurous, but we really need the other feature that @christofferh talked about.

We want to test every patch version that goes out: when you are using ruby and node and other LATEST stuff by the LATEST people, it's very common that patch versions break stuff.

So another issue with title of "New service restart option: interval"

```
restart:
  interval: 2h
```

matti:
(the word `interval` is great here)

jakolehm:
Just a clarification: this will only pull fresh image of what user has defined in kontena.yml (or service definition). 

So user still has total control of versions that will be deployed. It's no different what already happens when service instance moves to new node (by some event that is not controlled by user, for example node reboot).

PR #665, opened 2016-05-09
Return 404 error code when (un)assigning nonexistent user to grid

nevalla:
This PR makes API to return 404 / "User not found" error when trying to assign or remove nonexistent user to/from grid. Previously it returned status 422 with cryptic "User can't be nil" message


jakolehm:
LGTM

PR #668, opened 2016-05-09
Update cadvisor to 0.23.0

jakolehm:
This seems to fix random cadvisor errors that are related to some systemd mystery.


jakolehm:
@nevalla ping

nevalla:
LGTM!

PR #671, opened 2016-05-10
Point user account requests to auth provider

### LINKED ISSUE #605: User should be able to reset password without master

jakolehm:
Fixes #605


nevalla:
Looks good!

PR #672, opened 2016-05-10
Generate self-signed cert for DO master if no cert provided

jakolehm:
Sometimes it's cumbersome (particularly when testing) to generate certs manually for master provisioning. This PR adds a little helper that will generate self-signed SSL certificate for master if user does not provide any. Also enhances security because user does not have option to provision master without https.


nevalla:
LGTM!

PR #674, opened 2016-05-11
Update kontena.yml reference

nevalla:


docs/references/kontena-yml.md
```diff
@@ -313,3 +354,39 @@ nginx:
...
+    - KONTENA_LB_VIRTUAL_HOSTS=www.my-app.com
+  deploy:
+    strategy: ha
+    wait_for_port: 80
+  hooks:
+    post_start:
+      - name: sleep
+        cmd: sleep 10
+        instances: *
+  log_driver: fluentd
```
jakolehm: I think we should remove log_driver/log_opt because they are not something that user normally would use.

jakolehm:
LGTM

PR #677, opened 2016-05-11
Remove container_log fulltext indexing

jakolehm:
Fulltext indexes take too much disk space. Proper way to search logs is to stream them from mongo to some log service and do queries there.


jakolehm:
@nevalla ping

PR #679, opened 2016-05-12
Strip secrets from container environment variables

### LINKED ISSUE #578: Hide/redact vault secrets in master API
Currently the master API e.g. `v1/grids/:grid_name/services` exposes all attached secrets as environment variables. This action is also not audited.

I would suggested redacting or hiding environment variables that are injected as secrets.

nevalla:
Fixes #578

### DIFF
--- a/server/app/services/agent/container_info_mapper.rb
+++ b/server/app/services/agent/container_info_mapper.rb
@@ -87,7 +87,7 @@ module Agent
       config = info['Config'] || {}
       labels = config['Labels'] || {}
       state = info['State'] || {}
-      strip_secrets_from_env(container.grid_service, config['Env'])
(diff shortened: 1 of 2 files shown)


server/app/services/agent/container_info_mapper.rb
```diff
@@ -87,6 +87,7 @@ def container_attributes_from_docker(container, info)
       config = info['Config'] || {}
       labels = config['Labels'] || {}
       state = info['State'] || {}
+      strip_secrets_from_env(container.grid_service, config['Env'])
```
jakolehm: we should check that container has `grid_service` ?

jakolehm:
LGTM

PR #681, opened 2016-05-14
Improved memory parsing in CLI

christofferh:
Allows capitalised letters (docker-compose does this) and raises error if invalid.


jakolehm:
LGTM!

Issue #683, opened 2016-05-15
Add functionality to list services from another configuration file


janivihervas:
Enable auto-complete for services listed in another configuration file

janivihervas:
PR: [#684](https://github.com/kontena/kontena/pull/684)

PR #684, opened 2016-05-15
Cli: Added functionality to auto-complete services from another configuration file

janivihervas:
Enabled auto-complete of services from another configuration files for `kontena app <sub-command> -f <configuration-file.yml>`and `kontena app <sub-command> --file <configuration-file.yml>` commands. Auto-complete current directory's entries when completing the `-f` or `--file` option.

Implements [#683](https://github.com/kontena/kontena/issues/683).

### DIFF
--- a/cli/lib/kontena/scripts/completer
+++ b/cli/lib/kontena/scripts/completer
@@ -55,8 +55,14 @@ class Helper
   end
 
   def yml_services
-    if File.exist?('kontena.yml')      
-      yaml = YAML.load(File.read('kontena.yml'))
+    # Check if using another configuration file instead of 'kontena.yml'
+    # and list the services from that. Fallback to 'kontena.yml'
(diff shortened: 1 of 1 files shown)


nevalla:
Nice addition! We added support to read services from YAML v2 and due to that this PR has some conflicts. Could you rebase and resolve conflicts?

janivihervas:
Sure! I should have time to do it at the end of the week.

SpComb:
Thanks for the contribution, but closing as this seems to have stagnated and is now out of date. Please re-open if you want to update this for Kontena 1.2.

OTOH the `kontena app` commands are less used these days, and the new `kontena stack` services will be completed naturally as part of the `kontena service ..` commands. I think `kontena stack build` is the only command that takes a service name... that would need to be completed in a similar way.

janivihervas:
That's OK :+1:

PR #690, opened 2016-05-17
Improve deploy queue

jakolehm:
This PR changes how deployments are queued. Now each deployment is queued to database and worker pools will pick them from there (before each deployment were queued in actor (celluloid) system and they did not persist between deploys).

This PR also implements checks for stale deployment states.


jakolehm:
@nevalla ping

nevalla:
LGTM

PR #695, opened 2016-05-18
Release 0.13.0

jakolehm:
Highlights for this release include:

* **Grid Trusted Subnets** - It's now possible to set trusted subnets per grid that will enable much faster network speeds within those subnets.
* **Pre-Build Hooks** - We added pre-build hooks support to kontena.yml. With these hooks user can easily build binaries or assets before docker build is executed.
* **Interval Deploy Option** - With interval deploy option user can define interval for service deployments. This will help with service erosion and also makes possible to update services automatically from image registry.
* **External Master Database** - AWS and DigitalOcean master provision commands now support external MongoDB services via `--mongodb-uri` option. With this option combined with hosted MongoDB services (Compose.io, mLab etc) it's really easy to build high-availability master setups.
* **Unified CLI Commands** - CLI sub-commands are now unified to follow same naming logic.


miskun:
+1 for release highlights description :)

PR #697, opened 2016-05-19
Allow to set agent ip's via env variables

jakolehm:


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -72,17 +72,25 @@ def publish_node_info
...
     def private_ip
-      ip = interface_ip(private_interface)
-      unless ip
-        ip = interface_ip('eth0')
+      if ENV['KONTENA_PRIVATE_IP'].to_s != ''
+        ip = ENV['KONTENA_PRIVATE_IP'].to_s.strip
+      else
+        ip = interface_ip(private_interface)
+        unless ip
+          ip = interface_ip('eth0')
```
nevalla: I would change this to single line conditional 
`ip = interface_ip('eth0') unless ip`

or it can be even
`ip = interface_ip(private_interface) || interface_ip('eth0')`

PR #698, opened 2016-05-19
Fix agent ws hang on close when connection is unstable

jakolehm:
Steps to reproduce:

* start master & agent
* pause master container
* wait for agent to notice that master does not respond to ping
* unpause master
* agent ws connection is jammed


nevalla:
👍

PR #710, opened 2016-05-20
CLI: debug help only for non Kontena StandardErrors 

christofferh:
Fixes regression in 0.13

```
$ kontena version
cli: 0.12.3
$ kontena container exec web-1 pwd
{"error":"Not found"}

$ kontena version
cli: 0.13.1
$ kontena container exec web-1 pwd
Kontena error: {"error":"Not found"}
Rerun the command with environment DEBUG=true set to get full the exception
```


jakolehm:
LGTM

Issue #673, opened 2016-05-11
Force write/upsert vault secrets


christofferh:
I'm programatically adding/updating secrets with the master API and would like to avoid to keeping track of whether to use the `update` or `write` action.

Are you up for including support to make it possible to do e.g. `POST /v1/grids/grid_id/secrets` with an optional argument for doing an upsert operation of the secret if already present?

Exposed in the CLI I imagine it could be something similar to:
`kontena vault write MY_SECRET *secret* --force` or
`kontena vault write MY_SECRET *secret* --upsert`

If yes I'm happy to submit a PR.

jakolehm:
Makes sense to me. Correct http verb might be `PUT` for `--upsert` option?

/cc @nevalla

nevalla:
Sounds good, go ahead!

+1 for `--upsert`

jnummelin:
implemented in #711

PR #711, opened 2016-05-20
Option for upserting secrets on update

christofferh:
Implements: https://github.com/kontena/kontena/issues/673

Thought about audit log: Maybe extend Mutation outcomes to generate the audit message based on the outcome? So you could do for example:

> outcome.audit_message # Successful
=> 'created secret'
> outcome.audit_message # Failure
=> 'failed to create secret'


jakolehm:
LGTM

PR #712, opened 2016-05-21
Confirm dialog on destructive commands

christofferh:
Prompts users on destructive commands either with a y/n dialog or a name to confirm.
Can be overridden with `--confirm` option.

Also introduces an `error` helper that I suggest to replace the error handling/printing in `bin/kontena`. (Related https://github.com/kontena/kontena/pull/710)

### DIFF
--- a/cli/lib/kontena/cli/apps/remove_command.rb
+++ b/cli/lib/kontena/cli/apps/remove_command.rb
@@ -8,6 +8,7 @@ module Kontena::Cli::Apps
 
     option ['-f', '--file'], 'FILE', 'Specify an alternate Kontena compose file', attribute_name: :filename, default: 'kontena.yml'
     option ['-p', '--project-name'], 'NAME', 'Specify an alternate project name (default: directory name)'
+    option '--confirm', :flag, 'Confirm remove', default: false, attribute_name: :confirmed
 
     parameter "[SERVICE] ...", "Remove services"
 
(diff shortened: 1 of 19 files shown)


jakolehm:
Nice addition! Not sure about `--confirm` option name.

@nevalla @jnummelin @kke are you ok with `--confirm` ?

jnummelin:
I'm not sold on `--confirm` either. I think I'd rather have the option as `--force` to be more inline with other cli tools.

kke:
Sounds good to me.

In addition to confirming destructives, there should probably be options or subcommands to list zones/regions/facilities (`fi-hel1`, `ewr1`, ...), server configurations (`baremetal_0`, `1xCPU-1GB`, ...) and other parameters that mostly seem to be only listable through the provider APIs, the user interfaces only show human readable names.

Something like `kontena node packet --token XXX list-facilities`

christofferh:
I agree, '--force' is more versatile and can be used for other commands.
I'll make the change once the name is decided.

jakolehm:
`--force` sounds good to me.

force

christofferh:
Force restored in the galaxy. ping @jakolehm

jakolehm:
LGTM

jakolehm:
Needs rebase...

christofferh:
You guys are too fast ;) Done! (ping @jakolehm)

PR #715, opened 2016-05-23
fix (cli) Inviting users was confusing and had a typo.

matti:
As a user I assume that an email gets sent. Also was "send".


jakolehm:
LGTM

PR #717, opened 2016-05-24
Fix service name displaying on app deploy

nevalla:
When using empty project name on app deploy (`kontena app deploy -p ""`), service name might not be displayed correctly. This PR fixes this issue.


jakolehm:
LGTM

Issue #699, opened 2016-05-19
Dynamic etcd cluster reconfiguration


jakolehm:
We should automate etcd cluster member remove/add so that user does not have to know ugly details.

https://coreos.com/etcd/docs/latest/runtime-reconf-design.html

jakolehm:
Fixed in #719

PR #719, opened 2016-05-24
etcd member replacement functionality

jnummelin:
Work-In-Progress

This PR adds functionality to handle etcd initial node replacing.

Uses the etcd members API to figure out if we are about to replace a member or if we are bootstrapping a new cluster.

### DIFF
--- a/agent/lib/kontena/launchers/etcd.rb
+++ b/agent/lib/kontena/launchers/etcd.rb
@@ -88,19 +88,20 @@ module Kontena::Launchers
       cluster_size = info['grid']['initial_size']
       node_number = info['node_number']
       cluster_state = 'new'
+      weave_ip = weave_ip(info['node_number'])
 
       container = Docker::Container.get('kontena-etcd') rescue nil
       if container && container.info['Config']['Image'] != image
         container.delete(force: true)
       elsif container && container.running?
-        info "etcd is already running"
+        info 'etcd is already running'
         @running = true
-        return
+        return container
       elsif container && !container.running?
-        info "etcd container exists but not running, starting it"
+        info 'etcd container exists but not running, starting it'
         container.start
         @running = true
-        return
+        return container
       elsif container.nil? && node_number <= cluster_size
         # No previous container exists, update previous membership info if needed
         cluster_state = update_membership(info)
@@ -113,15 +114,15 @@ module Kontena::Launchers
 
       cmd = [
         '--name', name, '--data-dir', '/var/lib/etcd',
-        '--listen-client-urls', "http://127.0.0.1:2379,http://#{weave_ip(info)}:2379,http://#{docker_ip}:2379",
+        '--listen-client-urls', "http://127.0.0.1:2379,http://#{weave_ip}:2379,http://#{docker_ip}:2379",
         '--initial-cluster', initial_cluster(cluster_size).join(',')
       ]
       if node_number <= cluster_size
         cmd = cmd + [
-          '--listen-client-urls', "http://127.0.0.1:2379,http://#{weave_ip(info)}:2379,http://#{docker_ip}:2379",
-          '--listen-peer-urls', "http://#{weave_ip(info)}:2380",
-          '--advertise-client-urls', "http://#{weave_ip(info)}:2379",
-          '--initial-advertise-peer-urls', "http://#{weave_ip(info)}:2380",
+          '--listen-client-urls', "http://127.0.0.1:2379,http://#{weave_ip}:2379,http://#{docker_ip}:2379",
+          '--listen-peer-urls', "http://#{weave_ip}:2380",
+          '--advertise-client-urls', "http://#{weave_ip}:2379",
+          '--initial-advertise-peer-urls', "http://#{weave_ip}:2380",
           '--initial-cluster-token', grid_name,
           '--initial-cluster-state', cluster_state
         ]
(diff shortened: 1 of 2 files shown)


jnummelin:
Fixes #699

jakolehm:
Works beautifully! I think we should merge this asap after some code cleanup.

success

jakolehm:
ping @nevalla

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -134,12 +143,83 @@ def create_container(image, info)
...
       info "started etcd service"
       @running = true
       container
     end
 
+    # Removes possible previous member with the same IP
+    #
+    # @param [String] node weave ip
+    # @return [String] the state of the cluster member
+    def update_membership(info)
```
jakolehm: This method is too complex.. can we split this to smaller pieces?

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -157,6 +237,12 @@ def docker_gateway
       interface_ip('docker0')
     end
 
+    ##
+    # @param [Hash] node info
+    # @return [String] weave network ip of the node
+    def weave_ip(info)
```
nevalla: just a minor thing, but can we just pass a node number as parameter?

agent/lib/kontena/launchers/etcd.rb
jnummelin: Absolutely. I'll break out at least the rescue-retry loop to find online etcd node.

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -85,38 +85,47 @@ def create_data_container(image)
     # @param [String] image
     # @param [Hash] info
     def create_container(image, info)
```
nevalla: just thinking return values when container exists, should we always return container?

jnummelin: The return value in this case is actually not used, but for consistency sake it makes sense to always return the container.

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -134,12 +143,83 @@ def create_container(image, info)
...
+    end
+
+    ##
+    # Add new peer membership
+    #
+    # @param [Excon::Connection] etcd HTTP members API connection
+    # @param [String] The peer URL of the new peer to be added to the cluster
+    def add_membership(connection, peer_url)
+      info "Adding new etcd membership info with peer URL #{peer_url}"
+      connection.post(:body => JSON.generate({"peerURLs": [peer_url]}),
```
nevalla: You don't need quotations on `JSON.generate({"peerURLs": [peer_url]})` (or even brackets):

```
JSON.generate(peerURLs: [peer_url])
```

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -134,12 +143,83 @@ def create_container(image, info)
...
+                      :headers => { "Content-Type" => "application/json" })
```
nevalla: Minor style guide issue here. If string does not contain any variables use single quotation marks

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -134,12 +143,83 @@ def create_container(image, info)
...
+      node_number = info['node_number']
+      peer_url = "http://#{weave_ip(info)}:2380"
+      client_url = "http://#{weave_ip(info)}:2379"
+
+      begin
+        etcd_host = "http://10.81.0.#{tries}:2379/v2/members"
+
+        info "connecting to existing etcd at #{etcd_host}"
+        connection = Excon.new(etcd_host)
+        members = JSON.parse(connection.get().body)
```
nevalla: redundant parenthesis on `connection.get()`

agent/lib/kontena/launchers/etcd.rb
nevalla: Yes that was in my mind also

nevalla:
Found couple of small issue, otherwise great work!

jnummelin:
Refactored and cleaned, fixed styling issues. There seems to be some flaky test on server which fails the Travis build.

ping @jakolehm @nevalla

agent/lib/kontena/launchers/etcd.rb
```diff
@@ -85,20 +85,31 @@ def create_data_container(image)
...
       elsif container && container.running?
-        info "etcd is already running"
+        info 'etcd is already running'
         @running = true
-        return
+        return container
+      elsif container && !container.running?
+        info 'etcd container exists but not running, starting it'
+        container.start
+        @running = true
```
jakolehm: Are we using this `@running` information anywhere? If not, maybe we should remove `@running` and `#running?`.

jnummelin: It's actually being used here:
https://github.com/kontena/kontena/blob/feature/etcd_dynamic_cluster/agent/lib/kontena/workers/log_worker.rb#L55

jakolehm: Good catch :)

nevalla: I was thinking the same. We could separate private methods from public ones.

nevalla:
LGTM

jakolehm:
LGTM

PR #724, opened 2016-05-26
Fix stopped services with open deploys blocking deploy queue

jakolehm:


jnummelin:
This means now that when user tries to deploy a stopped service the deploy is just silently dropped. Is that really OK or are users expecting that deploying a stopped service actually deploys AND starts it?

jakolehm:
Good point. User expects deploy to change service state because that's how it has been working in the past.

jakolehm:
Improved deploy mutation logic a bit. 

/cc @nevalla

nevalla:
LGTM

jnummelin:
👍

PR #725, opened 2016-05-26
Fix service balancer not picking up instances without any deploys

jakolehm:
Also some refactoring / cleanup...


jnummelin:
LGTM - Much more readable

PR #726, opened 2016-05-26
Packet provisioning support

kke:
The packethost gem v0.0.6 does not yet support device rebooting, the reboot script aborts if the client doesn't respond_to?(:reboot_device).


jakolehm:
Excellent!

PR #727, opened 2016-05-26
Fixed upstream removal to remove only right container instance

### LINKED ISSUE #686: Service randomly detached from loadbalancer
Service was all of a sudden detached from loadbalancer which started to server 503s. This has happen on multiple occasions. A manual restart of the service resolves it.

According to @jakolehm in chat, this was due to a LB entry for the service in ETCD that went missing.

jnummelin:
Fixes #686 

This PR makes sure that instance removal removes only correct instance from LB upstreams.


jakolehm:
LGTM

PR #731, opened 2016-05-27
Fix possible agent ws ping_timer race condition

jakolehm:
In some really rare conditions agent might drop connection and loose track of connection status. This stops connection retries and agent is "offline" until it's restarted.

### DIFF
--- a/agent/lib/kontena/websocket_client.rb
+++ b/agent/lib/kontena/websocket_client.rb
@@ -52,6 +52,7 @@ module Kontena
 
     def connect
       return if connecting?
+      @connected = false
       @connecting = true
       info "connecting to master at #{api_uri}"
       headers = {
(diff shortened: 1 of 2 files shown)


agent/lib/kontena/websocket_client.rb
```diff
@@ -77,61 +80,61 @@ def connect
...
-      error "master does not accept our token, shutting down ..."
-      EM.next_tick{ abort("Shutting down ...") }
+      error 'master does not accept our token, shutting down ...'
+      EM.next_tick { abort('Shutting down ...') }
     end
 
-    # @param [Faye::WebSocket::Api::Event] event
-    def handle_invalid_version(event)
+    # @param [Faye::WebSocket::API::Event] _
+    def handle_invalid_version(_)
```
nevalla: Why do we need parameter here?

jakolehm: We don't, cleanup coming ...

jnummelin:
Looks good, hopefully this means final byebye for the socket issue.

nevalla:
LGTM

PR #734, opened 2016-05-27
Use region as az in DO

### LINKED ISSUE #718: Add az label automatically to DigitalOcean nodes
Because DO does not have real availability zones we have to use same values as in "region" label.

jnummelin:
This PR adds automatic az labeling based on DO region as DO does not support actual az's.

Fixes #718


jakolehm:
LGTM

PR #739, opened 2016-05-27
Support Docker Compose v2 YAML format

### LINKED ISSUE #636: validate kontena.yml
Currently a `kontena.yml` with missing, but required keys will cause a painfully cryptic error message:

`undefined method `include?' for nil:NilClass``

The user needs to go deep down in the code base to know what's wrong (this time: key `image` missing from kontena.yml). The "schema" of kontena.yml should be validated.

nevalla:
This PR adds support for Docker Compose version 2 format.

```
version: '2'
services:
  app:
    image: kontena/todo-example:latest
    ports:
      - 80:9292
    depends_on:
      - mongodb
    environment:
      - MONGODB_URI=mongodb://${project}-mongodb.kontena.local/todo_production
    command: bundle exec puma -p 9292 -e production
  mongodb:
    stateful: true
    image: mongo:3.2
    command: mongod --smallfiles
```

fixes #636

### DIFF
--- a/cli/lib/kontena/cli/apps/service_generator_v2.rb
+++ b/cli/lib/kontena/cli/apps/service_generator_v2.rb
@@ -7,8 +7,8 @@ module Kontena::Cli::Apps
     def parse_data(options)
       data = super(options)
       data['net'] = options['network_mode'] if options['network_mode']
-      data['log_driver'] = options['logging']['driver'] if options['logging'] && options['logging']['driver']
-      data['log_opts'] = options['logging']['options'] if options['logging'] && options['logging']['driver'] && !options['logging']['driver'].empty?
+      data['log_driver'] = options.dig('logging', 'driver')
+      data['log_opts'] = options.dig('logging', 'options')
       if options['depends_on']
         data['links'] ||= []
         data['links'] = (data['links'] + parse_links(options['depends_on'])).uniq
--- a/cli/lib/kontena/cli/apps/service_generator_v2.rb
+++ b/cli/lib/kontena/cli/apps/service_generator_v2.rb
@@ -1,0 +1,23 @@
+require 'yaml'
+require_relative 'service_generator'
+
+module Kontena::Cli::Apps
+  class ServiceGeneratorV2 < ServiceGenerator
+
+    def parse_data(options)
+      data = super(options)
+      data['net'] = options['network_mode'] if options['network_mode']
+      data['log_driver'] = options['logging']['driver'] if options['logging'] && options['logging']['driver']
+      data['log_opts'] = options['logging']['options'] if options['logging'] && options['logging']['driver'] && !options['logging']['driver'].empty?
+      if options['depends_on']
+        data['links'] ||= []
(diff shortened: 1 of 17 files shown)


nevalla:
ping @jakolehm, @jnummelin

cli/lib/kontena/cli/apps/yaml/validator_v2.rb
```diff
@@ -0,0 +1,70 @@
...
+      mem_limit memswap_limit privileged stateful user instances hooks secrets
+      ).freeze
+
+      UNSUPPORTED_KEYS = %w(
+      cgroup_parent container_name devices dns dns_search tmpfs entrypoint
+      expose extra_hosts labels log_driver log_opt net networks security_opt stop_signal ulimits volume_driver
+      cpu_quota cpuset domainname hostname ipc mac_address
+      read_only restart shm_size stdin_open tty working_dir
+      ).freeze
+
```
jnummelin: Why are there some keys in unsupported list that are currently supported? Log opts for example?

cli/lib/kontena/cli/apps/service_generator_v2.rb
```diff
@@ -0,0 +1,23 @@
...
+require_relative 'service_generator'
+
+module Kontena::Cli::Apps
+  class ServiceGeneratorV2 < ServiceGenerator
+
+    def parse_data(options)
+      data = super(options)
+      data['net'] = options['network_mode'] if options['network_mode']
+      data['log_driver'] = options['logging']['driver'] if options['logging'] && options['logging']['driver']
+      data['log_opts'] = options['logging']['options'] if options['logging'] && options['logging']['driver'] && !options['logging']['driver'].empty?
```
jnummelin: Would it make these more readable if `dig`is used instead on key access nesting?

cli/lib/kontena/cli/apps/yaml/validator_v2.rb
nevalla: `log_opts` and `log_driver` are not supported on version 2: https://docs.docker.com/compose/compose-file/#log-driver

Are there any other that I might have missed?

cli/lib/kontena/cli/apps/service_generator_v2.rb
nevalla: +1 for `dig`. That would definitely improve the readability.

cli/lib/kontena/cli/apps/yaml/validator_v2.rb
jnummelin: Ahaa, so the whole point of the `UNSUPPORTED_KEYS` list is to check for the deprecated keys? In that case some renaming and maybe a small comment would make things more clear.

nevalla: Yes, they are either deprecated in YAML version or not supported on Kontena.

jakolehm:
Should we have docs for v2?

nevalla:
Yes we should. I will add v2 documentation to PR #674.

PR #743, opened 2016-05-28
Allow to deploy service that is already in deploying state

jakolehm:
0.13.3 has regression that blocks deploys if service is in deploying state.


jakolehm:
@nevalla ping

nevalla:
LGTM!

PR #696, opened 2016-05-18
Improve YAML parsing

nevalla:
This PR:

- implementes YAML file validation (#636)

```
$ kontena app deploy
YAML contains the following unsupported options and they were rejected:
kontena.yml:
  db:
    - statefull: ["invalid option"]
base.yml:
  db:
    - user: ["unsupported option"]
YAML validation failed!
kontena.yml:
  web:
    - hooks: {"post_start":{"0":{"cmd":["is missing"]}}}
```
- enables extending services from the same YAML file

```
web:
  extends:
    service: base
```

- adds support for environment variables substitution (both $VAR and ${VAR} syntax are supported)

```
web:
  image: myapp:$TAG
```

- refactors code base to make it possible to support docker-compose.yml version 2

Fixes #636

PR #747, opened 2016-05-30
Allow build tag in build command

### LINKED ISSUE #634: kontena build should have support for specifying tag
Currently you need to specify that in the kontena.yml `image`.

Something like `kontena app/service build --tag anytag`

jnummelin:
Fixes #634


nevalla:
#696 introduced environment variable substitution, not sure do we this also.

jnummelin:
Good point. Maybe I'll close this one as unnecessary unless someone claims otherwise. :)

jakolehm:
Let's close this because variable substitution is already merged.

PR #748, opened 2016-05-30
Upcloud provisioning support

kke:
The ssh username for ssh access on Upcloud is "root" instead of the usual "core".

To get the list of zones, use:
[Upcloud API Zone list](https://api.upcloud.com/1.2/zone) 

To get the list of available server plans, use:
[Upcloud API Plan list](https://api.upcloud.com/1.2/plan)

(Use upcloud username + password to authenticate)


nevalla:
LGTM!

jakolehm:
Looks good! We need to think a little bit those firewall rules but for now this is good enough.

PR #750, opened 2016-05-30
Fix app remove command (regression caused by #696)

nevalla:
#696 changed internal service config handling so that config hashes will have stringified keys. This PR takes that into an account also when removing linked services.


jakolehm:
LGTM

PR #751, opened 2016-05-30
Allow to define app name in kontena.yml

nevalla:
This PR allows to define app name in `kontena.yml`. It will be used if user has not given any `-p` option on app commands. If the app name is not defined in `kontena.yml` name of the current work dir is used.
```
version: '2'
name: todo-example
services:
  app:
    image: kontena/todo-example:latest
```

depends on #739

### DIFF
--- a/cli/spec/spec_helper.rb
+++ b/cli/spec/spec_helper.rb
@@ -22,7 +22,6 @@ RSpec.configure do |config|
     allow(Dir).to receive(:home).and_return('/tmp/')
   end
 end
-RSpec::Expectations.configuration.warn_about_potential_false_positives = false
 
 require_relative 'support/client_helpers'
 require_relative 'support/fixtures_helpers'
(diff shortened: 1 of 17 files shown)


cli/spec/spec_helper.rb
```diff
@@ -22,6 +22,7 @@
     allow(Dir).to receive(:home).and_return('/tmp/')
   end
 end
+RSpec::Expectations.configuration.warn_about_potential_false_positives = false
```
jakolehm: why?

nevalla: Oh that was left there. The reason was that recent version of rspec displays that warning on containers_command_spec
```
WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives, since literally any other error would cause the expectation to pass, including those raised by Ruby (e.g. NoMethodError, NameError and ArgumentError), meaning the code you are intending to test may not even get reached. Instead consider using `expect {}.not_to raise_error`. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`.
```
Maybe the right solution is to change that spec file and remove error class from `raise_error` call?

nevalla: I reverted this. We should fix it elsewhere.

jakolehm:
LGTM

PR #752, opened 2016-05-30
Notify grid nodes when node information is updated

jakolehm:
Notify all connected grid nodes when node information (currently only labels) are changed. Changing node information might have effect on how network and other stuff is configured.


jakolehm:
@jnummelin ping

jnummelin:
LGTM

PR #754, opened 2016-05-31
Take az into account in ha scheduling strategy

jakolehm:
This PR implements logic that will spread service instances across availability zones by default.


jakolehm:
Rebased

jnummelin:
Should we add some docs about this also to: docs/using-kontena/deploy

Otherwise looks really good!

jnummelin:
👍

PR #757, opened 2016-05-31
Fix build context parsing

nevalla:
In YAML version 2 build context can be either string or hash. This PR takes both options into an account.

### DIFF
--- a/cli/spec/kontena/cli/app/service_generator_v2_spec.rb
+++ b/cli/spec/kontena/cli/app/service_generator_v2_spec.rb
@@ -46,4 +46,25 @@ describe Kontena::Cli::Apps::ServiceGeneratorV2 do
       }])
     end
   end
+
+  describe '#parse_build_options' do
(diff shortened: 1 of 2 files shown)


cli/spec/kontena/cli/app/service_generator_v2_spec.rb
```diff
@@ -46,4 +46,25 @@
...
+      data = {
+        'build' => {
+          'context' => '.',
+          'dockerfile' => 'alternate-dockerfile'
+         }
+      }
+
+      result = subject.send(:parse_build_options, data)
+      expect(result).to eq(data['build'])
+    end
```
jnummelin: Minor nitpicking but there seems to be two test cases bundled into one... :)

nevalla: good point. I separated test cases.

jnummelin:
👍

PR #760, opened 2016-05-31
Fix wait service deploy to finish

nevalla:
When deploying a service/app, service state is not changed immediately to `deploying`, that's why we need to wait a bit before starting to poll service state.


jakolehm:
Sleeps are bad but I don't have any better idea for this... so LGTM!

Issue #678, opened 2016-05-12
StatsWorker: undefined method `count' for nil:NilClass


mikkopiu:
kontena-etcd logs an Error `ERROR -- Kontena::Workers::StatsWorker: error on stats fetching: undefined method 'count' for nil:NilClass` from stats_worker.rb#L71 when (I think) `::per_cpu_usage` is `nil`.

Here's the approriate line: `num_cores = current_stat[:cpu][:usage][:per_cpu_usage].count` from `send_container_stats`.

Surely not a huge issue, but seems like an issue nonetheless.

kontena-agent version is 0.12.2 and docker version is 1.8.2-el7.

jnummelin:
fixed in #769

PR #769, opened 2016-06-01
Fix/nil on cpu usage and refactor stats worker

jnummelin:
Fix possible error relating to missing per cpu stats as reported in #678.

Refactored the stats worker to fetch single container stats as cAdvisor throws errors on all containers if one single container fails when getting stats for all containers on same call.

ping @jakolehm

### DIFF
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -68,9 +68,9 @@ module Kontena::Workers
 
       current_stat = container[:stats][-1]
 
-      num_cores = current_stat[:cpu][:usage][:per_cpu_usage].count
-      raw_cpu_usage = current_stat[:cpu][:usage][:total] - prev_stat[:cpu][:usage][:total]
-      interval_in_ns = get_interval(current_stat[:timestamp], prev_stat[:timestamp])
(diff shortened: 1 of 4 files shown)


agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -1,3 +1,5 @@
+require 'pp'
```
jakolehm: probably not needed? :)

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -46,68 +48,75 @@ def start
...
-      current_stat = container[:stats][-1]
+      prev_stat = container.dig(slice_key, :stats)[-2] if container
+      return if prev_stat.nil?
 
-      num_cores = current_stat[:cpu][:usage][:per_cpu_usage].count
-      raw_cpu_usage = current_stat[:cpu][:usage][:total] - prev_stat[:cpu][:usage][:total]
-      interval_in_ns = get_interval(current_stat[:timestamp], prev_stat[:timestamp])
+      current_stat = container.dig(slice_key, :stats, -1)
+      # Need to default to something usable in calculations
+      num_cores = current_stat.dig(:cpu, :usage, :per_cpu_usage).count rescue 1
```
jakolehm: can we get rid of this rescue, imho does not look nice

jakolehm:
Added couple notes, otherwise looks good!

PR #772, opened 2016-06-02
Fix automatic scale down on too many instances

jakolehm:
This PR fixes situation where node has been down so long that master looses track of "offline" service instances.


jnummelin:
👍

PR #774, opened 2016-06-02
Verify Upcloud API access

kke:
Fail gracefully if Upcloud API access fails due to wrong username and password or if API access is not enabled for the user.


jakolehm:
Very nice!

PR #775, opened 2016-06-02
Enable security group setting for master and nodes in AWS

jnummelin:
This PR enables setting existing security groups for both master and nodes.

### DIFF
--- a/cli/lib/kontena/machine/aws/node_provisioner.rb
+++ b/cli/lib/kontena/machine/aws/node_provisioner.rb
@@ -31,9 +31,11 @@ module Kontena
 
           opts[:vpc] = default_vpc.vpc_id unless opts[:vpc]
 
-          security_group = ensure_security_group(opts[:grid], opts[:vpc])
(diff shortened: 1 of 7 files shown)


cli/lib/kontena/machine/aws/node_provisioner.rb
```diff
@@ -96,21 +98,18 @@ def run!(opts)
 
         ##
         # @param [String] grid
-        # @return [Aws::EC2::SecurityGroup]
+        # @return [Array] Security group id in array
```
jakolehm: Why array here?

jnummelin: The instance creation takes an array of group ids and we do not use the actual group info at all after creating it so I thought it would be more clean that way. :)

jakolehm: ok, maybe the method name is a bit misleading now.. but this is not a blocker for this pr :)

jakolehm:
LGTM

Issue #767, opened 2016-06-01
Etcd could not connect after first node is removed from the cluster


mtrofimm:
I have a grid with initial_size=1. When I remove the first node  from the grid  etcd daemons on the remaining nodes fail to connect to the etcd cluster.

jnummelin:
That is actually expected. When the grid initial-size is 1 only the first node run the etcd in member role. All the other nodes start in proxy mode. Now when you remove the first node you actually remove the whole etcd as you remove all member nodes.

We should probably document this more detailed how the initial-size effects the etcd clustering and what situations we can recover from.

mtrofimm:
Thank for information. I was forced to remove the first node due to failure of Digital Ocean droplet after resizing in their control panel :(
I managed to workaround this issue by updating node_number in the host_nodes mongo collection.

jnummelin:
So you "forced" some other node to take place of the first one?

mtrofimm:
yes, and then recreated all kontena related containers on all nodes

jnummelin:
cool workaround :)

I'll tag this as documentation as clearly there's some improvements needed on those regarding the etcd membership and caveats.

mtrofimm:
@jnummelin what do you think about starting one etcd per grid at kontena master?

jnummelin:
Single etcd would create rather fragile failure point for multiple grids and it would need to be clustered for HA which in turn would mean more resource needs for master etc.. It's a bit of a rabbit hole. :) Also it would mean that there should be a secure communication path from agent to masters etcd which again would require quite much tuning. Current way enables rather easy comms path from agent to master as it requires only one websocket. So I think moving etcd to master side does not really make sense and would create more issues/challenges than benefits.

jakolehm:
Documentation improved in #776

PR #776, opened 2016-06-02
Clarify initial-size and etcd dependencies

jnummelin:
Some clarifications on how the grid initial-size effect etcd membership and some genera etcd related advice. Related issue #767

### DIFF
--- a/docs/using-kontena/grids.md
+++ b/docs/using-kontena/grids.md
@@ -20,7 +20,9 @@ The [Grid](../core-concepts/architecture.md#the-grid) is top-level object in Kon
 $ kontena grid create --initial-size=3 mygrid
 ```
 
-Creates a new grid named `mygrid` with initial size of 3 nodes (grid must have at least 3 nodes that are part of etcd cluster).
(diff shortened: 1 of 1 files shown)


docs/using-kontena/grids.md
```diff
@@ -20,7 +20,9 @@ The [Grid](../core-concepts/architecture.md#the-grid) is top-level object in Kon
 $ kontena grid create --initial-size=3 mygrid
 ```
 
-Creates a new grid named `mygrid` with initial size of 3 nodes (grid must have at least 3 nodes that are part of etcd cluster).
+Creates a new grid named `mygrid` with initial size of 3 nodes (grid must have at least 3 nodes that are part of etcd cluster). Etcd member nodes create the actual (data) cluster and rest of the grid nodes act as `proxy` nodes for etcd. Typically you want to spread your grid nodes in different availability zones / regions so that losing majority of the initial nodes becomes virtually impossible. Losing majority impairs the etcd and will cause troubles within the Kontena grid for example in service load balancing.
```
jakolehm: Probably not good idea to spread etcd across regions because latency... should we just recommend different az's here?

PR #777, opened 2016-06-02
Fix kontena app build command

nevalla:


jakolehm:
I would like to have some tests for this command (smoke style test is fine)

nevalla:
Added some smoke tests.

jakolehm:
LGTM

PR #779, opened 2016-06-03
Allow to use app name defined in yaml on app config command

nevalla:
#751 implemented support to define app/stack name in kontena.yml. This PR enables this also on `kontena app config` command


nevalla:
ping @jakolehm

jakolehm:
LGTM

PR #783, opened 2016-06-03
Support for tagging master and nodes on AWS

### LINKED ISSUE #762: Allow specifying tags when creating AWS nodes with the CLI
In some setups, nodes are expected to have certain tags. It would be handy to be able to provide these tags through the CLI.

jnummelin:
This PR adds support for tagging instances (both master and nodes) on AWS. Adds following options:
```
kontena node aws create --access-key access_key --secret-key secret \
--key-pair testing --region eu-central-1 --zone a \
--tag tag1=foo --tag tag2=bar
```

Fixes #762


jakolehm:
LGTM

PR #787, opened 2016-06-06
Improved VPN creation for non-public environments

### LINKED ISSUE #786: VPN service doesn't work if AWS nodes don't have public IPs
In this setup, `ovpn_genconfig` will get `$OVPN_SERVER_URL` as `udp://:1194` which will cause it to barf [when it tries to parse the URL](https://github.com/kontena/openvpn/blob/1a92995d01a4787acaf28acc5065ff2371602a3a/bin/ovpn_genconfig#L109).

jnummelin:
This PR improves VPN creation logic for environments where there are no public IPs for nodes. If the `--node` option is provided and if node does not have any public IP the VPN creation fallbacks to nodes private IP.

Fixes #786 

ping @jakolehm

### DIFF
--- a/cli/lib/kontena/cli/vpn/create_command.rb
+++ b/cli/lib/kontena/cli/vpn/create_command.rb
@@ -6,7 +6,7 @@ module Kontena::Cli::Vpn
     include Kontena::Cli::GridOptions
 
     option '--node', 'NODE', 'Node name where VPN is deployed'
-    option '--ip', 'IP', 'Node ip-address'
+    option '--ip', 'IP', 'Node ip-address to use in VPN service configuration'
 
     def execute
       require_api_url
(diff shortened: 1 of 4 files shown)


jakolehm:
Should we have some tests for this?

cli/lib/kontena/cli/vpn/create_command.rb
```diff
@@ -54,13 +54,8 @@ def execute
...
       return ip unless ip.nil?
-
-      # vagrant
-      if api_url == 'http://192.168.66.100:8080'
-        node['private_ip']
-      else
-        node['public_ip']
-      end
+      
+      node['public_ip'].to_s.empty? ? node['private_ip'].to_s : node['public_ip'].to_s
```
jakolehm: How about that vagrant "hack"?

jnummelin: I was thinking that we could treat Vagrant as any other "private" env, the user should provide the node and IP. But now thinking that bit further, it's kind of breaking change from user expectation point of view. I'll add the hack back in some form :)

jnummelin:
Will add some test cases

jakolehm:
LGTM!

PR #789, opened 2016-06-06
Add s3-v4auth flag for registry create

### LINKED ISSUE #788: Add v4auth option to registry
Some of the newer AWS regions only support v4auth on S3 thus makes it impossible to deploy registry as v4auth is by default false on the Docker registry.

We should make the v4auth config an option in the cli.

jnummelin:
Add v4auth flag for `kontena registry create` command.

Fixes #788


jakolehm:
Looks good!

PR #794, opened 2016-06-07
Cli plugins

### LINKED ISSUE #701: Initial plugins api
Plugins could be just a ordinary rubygems that have special file (`lib/kontena_cli_plugin.rb`). Kontena CLI would pick all installed plugin gems automatically and plugin could attach new subcommands.

Cli should also have basic commands for managing these plugin (install, list, show, remove).

jakolehm:
This PR implements initial way to provide cli plugins. It also provides basic sub-commands for managing plugins (list/search/install/uninstall).

```
$ kontena plugin list # list installed plugins
$ kontena plugin search # list/search available plugins
$ kontena plugin install <name> # install a plugin
$ kontena plugin uninstall <name> # uninstall a plugin
```

Plugin is just a normal rubygem with one exception, it has `lib/kontena_cli_plugin.rb` file that will be loaded automatically by kontena cli.

Fixes #701


cli/lib/kontena/cli/plugins/install_command.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    parameter 'NAME', 'Plugin name'
+
+    def execute
+      install_plugin(name)
+    end
+
+    def install_plugin(name)
+      plugin = "kontena-plugin-#{name}"
+      gem_bin = `which gem`.strip
```
jnummelin: Portability, is `which` available on all platforms?

jakolehm: Good catch.. of course it's not available on windows. Suggestions?

jnummelin: Not sure why we need the full path anyway. Couldn't we just use plain `gem`? `which` also would require to have the `gem` in path to get the full path, so IMO we do not gain much having the full path.

jakolehm: We need to get the full path because I did not find any solid way to pass environment. This is a problem with rvm and friends that modify env variables.

kke: You can replace `which` by traversing through $PATH and picking up the first hit. That's what "which" does. This should work:

```ruby
def which(cmd_name)
  ENV["PATH"].split(File::PATH_SEPARATOR).each do |path|
    Dir[File.join(File.expand_path(path), cmd_name)].each do |file_path|
      return file_path if File.executable?(file_path) && !File.directory?(file_path)
    end
  end
end

```

jakolehm: Actually we already have this method in a another module... will refactor things and use that :)

jakolehm:
I'll merge this, we can improve logic in a new pr.

Issue #702, opened 2016-05-19
Stacks api


jakolehm:
> A stack is a collection of services that make up an application in a specific environment. A stack file is a file in YAML format, similar to a docker-compose.yml file, that defines one or more services.

Currently Kontena already has `kontena.yml` that makes up an application but Kontena Master still handles these application services independently. We should add higher level abstraction also to Master.

PR #796, opened 2016-06-08
Initial stacks API

jnummelin:
This PR brings the stack concept to the server side.

### DIFF
--- a/server/app/models/grid.rb
+++ b/server/app/models/grid.rb
@@ -28,6 +28,7 @@ class Grid
   has_many :audit_logs
   has_many :registries, dependent: :delete
   has_many :overlay_cidrs, dependent: :delete
+  has_many :stacks
   has_and_belongs_to_many :users
 
   index({ name: 1 }, { unique: true })
--- a/server/app/models/grid.rb
+++ b/server/app/models/grid.rb
@@ -28,7 +28,7 @@ class Grid
   has_many :audit_logs
   has_many :registries, dependent: :delete
   has_many :overlay_cidrs, dependent: :delete
-  has_many :stacks
+  has_many :stacks, dependent: :destroy
   has_and_belongs_to_many :users
 
   index({ name: 1 }, { unique: true })
--- a/server/app/models/stack.rb
+++ b/server/app/models/stack.rb
@@ -10,7 +10,6 @@ class Stack
   belongs_to :grid
 
   has_many :grid_services
-  #has_many :audit_logs
 
   index({ grid_id: 1 })
   index({ name: 1 })
--- a/server/app/models/stack.rb
+++ b/server/app/models/stack.rb
@@ -1,0 +1,26 @@
+class Stack
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::Enum
+
+  field :name, type: String
+  field :version, type: String
+  enum :state, [:initialized, :deployed, :terminated]
+
+  belongs_to :grid
+
+  has_many :grid_services
+  #has_many :audit_logs
+
+  index({ grid_id: 1 })
+  index({ name: 1 })
+
+  validates_presence_of :name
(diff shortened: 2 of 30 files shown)


jnummelin:
Fixes #702

server/app/models/stack.rb
```diff
@@ -0,0 +1,26 @@
...
+  include Mongoid::Enum
+
+  field :name, type: String
+  field :version, type: String
+  enum :state, [:initialized, :deployed, :terminated]
+
+  belongs_to :grid
+
+  has_many :grid_services
+  #has_many :audit_logs
```
jakolehm: Let's remove these commented code blocks... if we don't need them then they should not be here :)

server/app/models/stack.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  index({ grid_id: 1 })
+  index({ name: 1 })
+
+  validates_presence_of :name
+  #validates_uniqueness_of :name, scope: [:grid_id]
```
jakolehm: Probably we need to validate uniqueness?

server/app/mutations/grid_services/add_env.rb
```diff
@@ -7,6 +7,7 @@ class AddEnv < Mutations::Command
 
     def execute
       self.grid_service.env << env
+      grid_service.revision += 1
```
jakolehm: Can revision be nil? Should we cast revision to integer first?

jnummelin: It shouldn't be nil as the model defaults to 1

server/app/mutations/stacks/delete.rb
```diff
@@ -0,0 +1,29 @@
...
+        GridServices::Delete.validate(current_user: self.current_user, grid_service: service)
+      end
+    end
+
+    def execute
+      self.stack.state = :terminated
+      self.stack.save
+      # Remove all services of the stack
+      self.stack.grid_services.each do |service|
+        worker(:grid_service_remove).async.perform(service.id)
```
jakolehm: Why no mutation here?

server/app/routes/v1/stacks_api.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      def update_stack(stack, data)
+        data[:grid] = @grid
+        data[:current_user] = current_user
+        data[:stack] = stack
+        outcome = Stacks::Update.run(data)
+
+        if outcome.success?
+          @stack = outcome.result
+          #audit_event(request, @grid, @grid_secret, 'update', nil, [:body])
```
jakolehm: I would like to have audit log :)

server/app/models/grid.rb
```diff
@@ -28,6 +28,7 @@ def self.default_overlay_cidr
   has_many :audit_logs
   has_many :registries, dependent: :delete
   has_many :overlay_cidrs, dependent: :delete
+  has_many :stacks
```
jakolehm: Should this have `dependent: :destroy`?

server/app/models/stack.rb
jnummelin: Yes, and in the scope of grid.

server/app/routes/v1/stacks_api.rb
jnummelin: You shall have one. :D

server/app/models/grid.rb
jnummelin: yep

jnummelin:
Comments addressed in last commit

server/app/models/stack.rb
jakolehm: How about these commented out lines...?

jnummelin: removed in last push

jakolehm:
LGTM!

PR #798, opened 2016-06-09
Fix port definitions to include possibility to set bind IP

### LINKED ISSUE #795: Ability to define host ip in ports mapping
I mean something like defined here:
https://docs.docker.com/compose/compose-file/#ports
i.e. exposing container port only on the given host interface/ip, for example for a corporate VPN.

jnummelin:
This PR adds the possibility to use bind IP address in port expose settings. Scheduler will NOT make sure the address is actually available on the scheduled node so user has to use proper affinity rules to control this.

Fixes #795


jakolehm:
Looks good!

PR #803, opened 2016-06-13
Cli: Container exec command to handle whitespace and strings

christofferh:
See spec for example.

### DIFF
--- a/cli/lib/kontena/cli/containers/exec_command.rb
+++ b/cli/lib/kontena/cli/containers/exec_command.rb
@@ -1,7 +1,11 @@
+require_relative '../grid_options'
+require_relative './containers_helper'
+
 module Kontena::Cli::Containers
   class ExecCommand < Clamp::Command
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/cli/containers/exec_command.rb
```diff
@@ -10,9 +14,12 @@ def execute
       require_api_url
       token = require_token
 
-      payload = {cmd: ['sh', '-c', cmd_list.join(' ')]}
-      service_name = container_id.match(/(.+)-(\d+)/)[1]
+      cmd = build_command(cmd_list)
+      payload = {cmd: ["sh", "-c", cmd]}
+
+      service_name = container_id.match(/(.+)-(\d+)/)[1] rescue nil
```
christofferh: Not sure if we want to explicitly handle the missing match error here? This at least displays `Not found` instead of crashing hard.

jnummelin: I think it's better to show `Not found` than crashing hard

jakolehm:
Nice fix, thanks!

PR #812, opened 2016-06-17
Initial health check for remote services

### LINKED ISSUE #658: Proposal: HTTP health check for services as a platform feature
Could be used by loadbalancer to detect unhealthy instances and remove from rotation. 

First implementation could be simple and just present the info for the instance e.g. `kontena service status my-service.

```
server:
  image: christofferh/www:latest
  environment:
    - KONTENA_LB_INTERNAL_PORT=8080
  links:
    - loadbalancer
  health_check:
    url: /monitor
    timeout: 2 # Timeout seconds
    unhealthy_threshold: 5 # The number of consecutive failed health checks
```

Inspiration:

http://docs.deis.io/en/latest/using_deis/config-application/#custom-health-checks
http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-healthchecks.html

jnummelin:
This PR implements initial HTTP health checks for services. The status check results are stored per instance (container). The status is not acted upon by the scheduler it is only informatory to the user at this point.

Creating a service with health check looks like this: 
```
kontena service create --instances 2 --health-check-port 80 --health-check-interval 20 --health-check-timeout 2 --health-check-uri / nginx nginx
```

And with kontena.yml:
```
web:
  image: nginx
  stateful: false
  health_check:
    protocol: http
    port: 80
    interval: 20
    uri: /
    initial_delay: 10
    timeout: 2


mysql:
  image: mysql
  stateful: false
  deploy:
    strategy: ha
    wait_for_port: 3306
  environment:
    - MYSQL_ALLOW_EMPTY_PASSWORD=true
  health_check:
    protocol: tcp
    port: 3306
    interval: 20
    initial_delay: 10
    timeout: 2
```
Both http and tcp protocols are supported. When http is used, currently only 200 status code is translated into healthy service. When tcp is used, successful port opening test will translate into healthy service.

The initial delay parameter is used to allow decent time for the service in the container to startup after which the health checks are started.

The health status is "aggregated" into service listing as color code:
screen shot 2016-06-17 at 14 12 13

Code needs still some polishing and tuning thus the WIP flag. :)

fixes #658

### DIFF
--- a/agent/lib/kontena/helpers/port_helper.rb
+++ b/agent/lib/kontena/helpers/port_helper.rb
@@ -1,0 +1,20 @@
+  module Kontena
+    module Helpers
+      module PortHelper
+
+        def container_port_open?(ip, port, timeout = 2.0)
+          Timeout::timeout(timeout) do
+            begin
+              TCPSocket.new(ip, port).close
+              true
+            rescue Errno::ECONNREFUSED, Errno::EHOSTUNREACH
+              false
+            end
+          end
+        rescue Timeout::Error
+          false
+        end
+
+      end
+    end
+  end
--- a/cli/lib/kontena/cli/services/services_helper.rb
+++ b/cli/lib/kontena/cli/services/services_helper.rb
@@ -170,6 +170,22 @@ module Kontena
             puts "  pid: #{service['pid']}"
           end
 
+          if service['health_check']
+            puts "  health check:"
+            puts "    protocol: #{service['health_check']['protocol']}"
+            puts "    uri: #{service['health_check']['uri']}" if service['health_check']['protocol'] == 'http'
+            puts "    port: #{service['health_check']['port']}"
+            puts "    timeout: #{service['health_check']['timeout']}"
+            puts "    interval: #{service['health_check']['interval']}"
+            puts "    initial_delay: #{service['health_check']['initial_delay']}"
+          end
+          
+          if service['health_status']
+            puts "  health status:"
+            puts "    healthy: #{service['health_status']['healthy']}"
+            puts "    total: #{service['health_status']['total']}"
+          end
+
           puts "  instances:"
           result = client(token).get("services/#{parse_service_id(service_id)}/containers")
           result['containers'].each do |container|
--- a/server/app/services/agent/message_handler.rb
+++ b/server/app/services/agent/message_handler.rb
@@ -60,6 +60,8 @@ module Agent
           self.on_container_log(grid, message['node_id'], data['data'])
         when 'container:stats'.freeze
           self.on_container_stat(grid, data['data'])
+        when 'container:health'.freeze
+          self.on_container_health(grid, data['data'])
         else
           error "unknown event: #{message}"
       end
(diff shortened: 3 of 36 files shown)


jnummelin:
Last commit added support for tcp checks. What's still missing is the automatic translation into http checks on LB. I'll work on that in the next commit.

jnummelin:
The last commit adds support for http checks in LB, from the agent side. This needs obviously the corresponding support from LB: https://github.com/kontena/kontena-loadbalancer/issues/17

ping @jakolehm @nevalla

agent/lib/kontena/helpers/port_helper.rb
```diff
@@ -0,0 +1,20 @@
+module Kontena
```
jakolehm: Indentation is way off here?

cli/lib/kontena/cli/services/services_helper.rb
```diff
@@ -362,6 +378,10 @@ def int_to_filesize(int)
             'TB' => 1000 * 1000 * 1000 * 1000 * 1000
           }.each_pair { |e, s| return "#{(int.to_i / (s / 1000))}#{e}" if int < s }
         end
+
+        def parse_health_check(health_check_uri, health_check_port, health_check_timeout, health_check_interval, health_check_initial_delay)
```
nevalla: useless method?

server/app/views/v1/grid_services/_grid_service.json.jbuilder
```diff
@@ -33,3 +33,15 @@ json.instances do
...
+		json.protocol grid_service.health_check.protocol
+		json.uri grid_service.health_check.uri
+		json.port grid_service.health_check.port
+		json.timeout grid_service.health_check.timeout
+		json.initial_delay grid_service.health_check.initial_delay
+		json.interval grid_service.health_check.interval
+	end
+	json.health_status grid_service.health_status
+end
+json.health_status grid_service.health_status if grid_service.health_check
```
nevalla: What kind of json response we will get if grid_service has health_check defined? Will there be duplicated health_status?

server/app/services/agent/message_handler.rb
```diff
@@ -164,6 +166,26 @@ def on_container_stat(grid, data)
     ##
     # @param [Grid] grid
     # @param [Hash] data
+    def on_container_health(grid, data)
+      #return if @queue.length > 100
+
+      #container = grid_container(grid.id, data['id'])
```
nevalla: Some cleaning needed

server/app/views/v1/grid_services/_grid_service.json.jbuilder
jnummelin: Yep, the other one should be removed.

jnummelin:
Code cleaned. I squashed the commits into one so it was easier to rebase :)

jakolehm:
Not completely sure about the cli "layout", otherwise looks good to me! (and we can improve the layout later in separate pr if needed)

jnummelin:
Do you mean the layout of the commands or the output on service list and service details?

jakolehm:
I mean the output of list/show.

jnummelin:
I'm open to suggestions on the layout. Or as you said, we can improve them later on when we've had a spin around the block with this feature.

jakolehm:
I think we can merge this after rebase.

jnummelin:
"All your rebase are belong to us" :D

PR #822, opened 2016-06-29
Initial cli commands for stack API

jnummelin:
This PR adds initial CLI commands for stacks API.

This also fixes a deploy revision related regression and now separates the deployment revision from the service version. Naturally there can be multiple deployments (re-balancing, interval deploy etc.) for a given service version and thus we need to separate these concepts.

We need to further work on the versioning when we actually enable the capability to deploy older version of service/stack.

### DIFF
--- a/cli/lib/kontena/cli/stack_command.rb
+++ b/cli/lib/kontena/cli/stack_command.rb
@@ -12,7 +12,7 @@ class Kontena::Cli::StackCommand < Clamp::Command
   subcommand "show", "Show stack details", Kontena::Cli::Stacks::ShowCommand
   subcommand "update", "Update stack", Kontena::Cli::Stacks::UpdateCommand
   subcommand "deploy", "Deploy Kontena stack", Kontena::Cli::Stacks::DeployCommand
-  subcommand ["remove","rm"], "Remove services", Kontena::Cli::Stacks::RemoveCommand
+  subcommand ["remove","rm"], "Remove stack", Kontena::Cli::Stacks::RemoveCommand
   
   def execute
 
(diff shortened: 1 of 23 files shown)


cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -0,0 +1,44 @@
+#require 'yaml'
+
+#require_relative '../services/services_helper'
+#require_relative '../apps/common'
```
jakolehm: We should remove these lines...

jnummelin:
rebased and cleaned

cli/lib/kontena/cli/stack_command.rb
```diff
@@ -0,0 +1,20 @@
...
+require_relative 'stacks/show_command'
+
+class Kontena::Cli::StackCommand < Clamp::Command
+
+  subcommand "create", "Create stack", Kontena::Cli::Stacks::CreateCommand
+  subcommand ["ls", "list"], "List stacks", Kontena::Cli::Stacks::ListCommand
+  subcommand "show", "Show stack details", Kontena::Cli::Stacks::ShowCommand
+  subcommand "update", "Update stack", Kontena::Cli::Stacks::UpdateCommand
+  subcommand "deploy", "Deploy Kontena stack", Kontena::Cli::Stacks::DeployCommand
+  subcommand ["remove","rm"], "Remove services", Kontena::Cli::Stacks::RemoveCommand
```
jakolehm: "Remove services" -> "Remove stack"

jnummelin:
I will still work on the service_rev vs. deploy_rev thingy before merging this

Issue #814, opened 2016-06-23
CLI log tailing should retry on connection error


christofferh:
If connection to master is cut when tailing a log (`kontena service logs -t`), CLI will terminate. Suggestion is to retry if I EOFError is raised.

jnummelin:
should be fixed with #624 as they are pretty much same thing

jnummelin:
fixed in #835

PR #835, opened 2016-07-05
Log tailing retry in EOF case

### LINKED ISSUE #624: Grid log tailing crashes with EOFError after a minute
Running `kontena grid logs -t` crashes at some point (usually a minute or two) with message
```
kontena-agent | I, [2016-04-20T13:51:52.163798 #1]  INFO -- Kontena::Workers::NodeInfoWorker: publishing node information
end of file reached (EOFError)
```

Looks like something in publishing node stats contains invalid character.

jnummelin:
This PR adds retry logic into log tailing when EOF is received. Also refactored the service log to use streaming based tailing

fixes #624 and #814

### DIFF
--- a/cli/lib/kontena/cli/grids/logs_command.rb
+++ b/cli/lib/kontena/cli/grids/logs_command.rb
@@ -55,14 +55,25 @@ module Kontena::Cli::Grids
           @buffer << chunk
         end
         if log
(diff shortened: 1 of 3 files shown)


jakolehm:
LGTM!

PR #838, opened 2016-07-07
Restart event handling for weave

jnummelin:
This PR adds specific handling for weave container restart event. This enables the agent to do full container re-attach and DNS handling in case weave router has to be restarted.

ping @jakolehm

### DIFF
--- a/agent/lib/kontena/workers/event_worker.rb
+++ b/agent/lib/kontena/workers/event_worker.rb
@@ -39,7 +39,14 @@ module Kontena::Workers
 
     # @param [Docker::Event] event
     def publish_event(event)
+
       return if Actor[:network_adapter].adapter_image?(event.from)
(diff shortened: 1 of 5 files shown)


agent/lib/kontena/workers/event_worker.rb
```diff
@@ -39,8 +39,15 @@ def stream_events
 
     # @param [Docker::Event] event
     def publish_event(event)
+
       return if Actor[:network_adapter].adapter_image?(event.from)
 
+      if Actor[:network_adapter].router_image?(event.from)
+        if event.status == 'restart'
+          publish('network_adapter:restart', event)
```
jakolehm: Why EventWorker should care about this?

jakolehm:
LGTM

PR #850, opened 2016-07-12
Fix docker event stream filter params

jakolehm:
Docker >= 1.10 accepts filters as json encoded value.


jakolehm:
@nevalla ping

nevalla:
LGTM!

PR #859, opened 2016-07-20
Fix agent not reconnecting to master

### LINKED ISSUE #840: Agent goes offline and does not try to reconnect

jakolehm:
This PR fixes race condition in agent websocket client that is triggered by client trying to send message via ws client that does not exist (or is in some weird state).

Fixes #840


jakolehm:
@miskun ping

PR #489, opened 2016-02-07
Refactor registry to use vault for secrets management

jakolehm:
Depends on:
- https://github.com/kontena/registry/pull/1
- https://github.com/kontena/registry/pull/2

PR #863, opened 2016-07-23
Fix deploying registry on azure

nooulaif:
I believe there was a small mistake in #489 which makes deploying registry on azure impossible.

Also it should be `REGISTRY_STORAGE_AZURE_CONTAINER` and not like was before `REGISTRY_STORAGE_AZURE_CONTAINERNAME`.


jakolehm:
Thanks!

PR #878, opened 2016-08-05
Fix possible docker event stream lockups

### LINKED ISSUE #867: Docker event stream sometimes hangs
This has been started to pop-up since Docker 1.10. Agent thinks that it has connection to docker event stream but no events comes through.

jakolehm:
Fixes #867


jnummelin:
Code looks good!

Just thinking that should we update the docker-api gem at the same time as it has some event related fixes also:
https://github.com/swipely/docker-api/issues/422

PR #880, opened 2016-08-05
Discard empty lines in env_file

### LINKED ISSUE #877: Empty lines not supported in env_file
I noticed that empty lines are not allowed in the env_file file. I don't see why it shouldn't support it, and I didn't see any mention about this in the documentation.

The error message could be more clear:
```
{"env":[null,null,null,null,null,null,null,null,"Env[8] can't be blank",null,null]}
```

nevalla:
Fixes #877


jakolehm:
LGTM!

PR #893, opened 2016-08-09
Hide experimental stack commands

jnummelin:
As stacks is still in flux, let's hide the command behind KONTENA_EXPERIMENTAL env flag.


jakolehm:
LGTM!

PR #900, opened 2016-08-10
Show health-check with icon in service list

jakolehm:
screen shot 2016-08-10 at 15 08 50


jnummelin:
`ServicesHelper` seems to swell quite a bit but that should be addressed in different PR.

LGTM

PR #901, opened 2016-08-10
Handle env_file on YAML file parsing

nevalla:
Currently environment variables are merged with variables from env_file after the YAML file is read and parsed. It means that the path of the env_file is related to directory where the command is executed not the directory of YAML file. For example this will fail if some service has `env_file:  .env` definition: 

`$ kontena app config -f ./subfolder/kontena.yml`

This PR moves env_file parsing to YAML reading phase and env_files are read correctly.


jakolehm:
LGTM!

Issue #902, opened 2016-08-11
Add console command to master for debugging


kke:
Like rails console that you can use if you connect to the host and run a command inside the container.

Practically it's `require irb; require server.rb`.

jakolehm:
Maybe even better if we require `app/boot.rb` instead of `server.rb`. (boot does not trigger background workers)

jnummelin:
fixed in #903

PR #903, opened 2016-08-11
Added kontena-console command to master for local debugging

kke:
```
$ MONGODB_URI=mongodb://mongodb:27017/kontena_server bin/kontena_console

irb(main):001:0> Role.first
=> #<Role _id: 57ac1f8a699fb77ac1000006, created_at: 2016-08-11 06:47:38 UTC, updated_at: 2016-08-11 06:47:38 UTC, user_ids: nil, name: "master_admin", description: "Master admin">
```

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -15,7 +15,7 @@ RUN apk --update add --virtual build-dependencies ruby-dev build-base openssl-de
 ADD . /app
 RUN chown -R nobody:nogroup /app
 USER nobody
-
+ENV PATH="/app/bin:${PATH}"
 ENV RACK_ENV production
 EXPOSE 9292
 
--- a/server/bin/kontena-console
+++ b/server/bin/kontena-console
@@ -1,0 +1,16 @@
+#!/usr/bin/env ruby
+# encoding: UTF-8
+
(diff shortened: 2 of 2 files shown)


kke:
Implements #902

jakolehm:
LGTM!

SpComb:
The explicit `ENV["MONGODB_URI"]` overrides any `.env` and `config/mongoid.yml` configuration, so it's a little difficult to use in the development environment.

Won't the production Docker environment have an explicit `MONGODB_URI` configured anyways, so the default value here is a bit of a guess?

kke:
With #1621 it should keep from crashing

Not sure about the path and command name.

It would be nice to do:
```
$ docker exec -it kontena-server-api console
# or
$ docker exec -it kontena-server-api kontena-console

```

instead of

```
$ docker exec -it kontena-server-api ./bin/kontena-console
```

Maybe we can put `app/bin/` to `$PATH`? Maybe use rake to start the console like the admin reset:

```
$ docker exec -it kontena-server-api rake kontena:console
```
 
@ping jakolehm

jakolehm:
> Maybe we can put app/bin/ to $PATH?

Sounds good.

PR #913, opened 2016-08-12
Reschedule service if instance deploy_rev does not match

jakolehm:


jakolehm:
@jnummelin @nevalla ping

server/app/services/grid_service_deployer.rb
```diff
@@ -60,10 +60,9 @@ def deploy
     info "starting to deploy #{self.grid_service.to_path}"
     ping_subscription = self.subscribe_to_ping
     creds = self.creds_for_registry
-    self.grid_service.set_state('deploying')
-    self.grid_service.set(:deployed_at => Time.now.utc)
-
     deploy_rev = Time.now.utc.to_s
+    self.grid_service.set_state('deploying')
+    self.grid_service.set(:deployed_at => deploy_rev)
```
SpComb: This assigns a String to the `GridService.deployed_at` `type: DateTime` attribute?

server/app/jobs/service_balancer_job.rb
```diff
@@ -118,6 +118,7 @@ def pending_deploys?(service)
   def lagging_behind?(service)
     return true if service.deployed_at && service.updated_at > service.deployed_at
     return true if service.deployed_at && service.deploy_requested_at && service.deploy_requested_at > service.deployed_at
+    return true if service.deployed_at && service.containers.where(:deploy_rev.ne => service.deployed_at).count > 0
```
SpComb: This compares the `Container.deploy_rev` String with the `GridServer.deployed_at` DateTime?

jnummelin:
Looks good

server/app/jobs/service_balancer_job.rb
jakolehm: Ok, have to agree that something is off here... will investigate / add better tests.

jakolehm:
Let's close this one, with stacks we need to change the logic anyway.

PR #921, opened 2016-08-16
Fix build arguments normalizing

### LINKED ISSUE #920: Application build fails with Ruby 2.3.x if build option is string
Ruby 2.3 introduced native `Hash#dig` which is not fully compatible with functionality provided by `ruby_dig` gem. At this moment an error is returned when build option is string in v2 styled YAML file (e.g `build: .`)

Workaround is to use the following syntax in YAML file
```
build:
  context: .
```

nevalla:
Fixes #920


cli/lib/kontena/cli/apps/yaml/safe_dig.rb
```diff
@@ -0,0 +1,13 @@
+module SafeDig
```
jakolehm: I think we should move this module...?

jakolehm:
LGTM!

PR #924, opened 2016-08-22
Support for callbacks when running cli commands.

kke:
Adds callback support to Kontena::Command

Types of callbacks:
 - after_load runs after command is loaded. With this you can for
   example add parameters. Example:
```ruby
   def after_load
     command.class_eval do
       option ["--auth-token"], '[AUTH_TOKEN]', 'Use authentication
       token'
     end
   end
```
 - before runs before command's #execute method
 - after runs after command's #execute method
 - before_parse runs before command line parameters are parsed.

Here is an example of a callback that runs after provisioning a master and suggests inviting yourself as a regular user if you are currently using the master's local "admin" account:

```ruby
class SuggestInvitingYourself < Kontena::Callback

  include Kontena::Cli::Common

  # Runs for all commands that are something like:
  # Kontena::Cli::Vagrant::Master::CreateCommand
  command_types :master =>  :create

  # Runs after the original command's #execute method
  def after
    # .command is an accessor to the original command instance.
    # command.exit_code is 1 when the original command exited with "exit(1)"
    return nil unless command.exit_code == 0

    if current_master && current_master.username.to_s == 'admin'
      puts
      puts "Protip:"
      puts "  You are currently using Kontena Master administrator account."
      puts "  Consider inviting yourself as a regular user and using the"
      puts "  returned invite code to authenticate. Use:"
      puts "    kontena master users invite your@email.address.example.com"
      puts "    kontena auth master --join <invite_code>"
      puts
    end
  end
end
```

### DIFF
--- a/cli/lib/kontena/callback.rb
+++ b/cli/lib/kontena/callback.rb
@@ -1,0 +1,47 @@
+class Kontena::Callback
+
+  attr_reader :command
+
+  def initialize(command)
+    @command = command
+  end
+
+  # Register callback for command types it is supposed to run with.
+  def self.command_types(*cmd_types)
+    return @command_types if cmd_types.empty?
+
+    # In case you call command_types([array, items, here]) instead of command_types(array, items, here)
+    cmd_types = cmd_types.first if cmd_types.first.kind_of?(Array)
+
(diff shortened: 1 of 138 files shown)


kke:
Included in #1035

PR #925, opened 2016-08-22
New more object oriented config handler

kke:
You can now do things like `config.current_master.grid = 'foo'` or
`config.find_server('foo').url = 'https://foo'`

Config handling is now done in it's own Kontena::Cli::Config instead of in Kontena::Cli::Common.


kke:
Included in #1035

PR #927, opened 2016-08-23
Show container exit code

### LINKED ISSUE #792: Store docker exit code for container
Would be nice to see the exit code of a container in the `kontena service show` listing. There are some specific cases where docker sets the [exit code](https://github.com/docker/docker/blob/master/daemon/start.go#L138-L150) and those would be handy to spot. And maybe these might be also useful in determining the container state.

jnummelin:
fixes #792


jakolehm:
LGTM!

PR #928, opened 2016-08-23
Add user-agent for http healthcheck

### LINKED ISSUE #896: Agent http health check should set User-Agent
It would be nice to set user agent to something like "Kontena/<VERSION>".

jnummelin:
fixes #896


jakolehm:
I think we should add version number to user agent? Otherwise looks good!

jnummelin:
I'll add the agent version, good idea!

PR #930, opened 2016-08-23
CLI HTTP API Client can now handle token expiration

kke:
The CLI initialize now takes a token object as argument.

It can be a hash or a Kontena::Cli::Config::Token object from the new configuration handler.

If the server responds with HTTP 401 to a request, it tries to refresh the access token and retry the request.


kke:
Included in #1035

PR #931, opened 2016-08-23
Master auth renewal

kke:
- External OAuth2 providers can be used as authentication provider
- Internal OAuth2 provider now supports refresh token and authorization
  code grant flow
- API endpoint /authenticate is used for performing authentication
- API endpoint /v1/auth no longer available, there is no more username
  and password login
- API endpoint /v1/auth_provider is used to configure external oauth2
  provider settings
- Only encryption hashes are stored for local access and refresh tokens
- Token validation is done in RACK middleware instead of a helper module
- There's a RACK middleware that restricts access to Kontena CLI clients
  prior to version X (currently set to 0.15.0 in server.rb). The error
  message contains a hint about how to upgrade and the old CLI will show
  it when trying to perform requests.

Authentication flow:

Upon deployment a local user "admin" is created. The administrator gets
an access token with a code, generated by the CLI before deployment.

This code is then used by the CLI to perform initial authentication.

The usage of "admin" account does not require that an oauth2 provider is
configured, it's a local only account.

To configure an auth provider you can use the CLI's "kontena master
auth-provider config" command.

Authenticating a regular user:

A) Auth provider not configured:
  1) Client requests /authenticate?redirect_uri=http://localhost:xxx/cb
  2) Master responds with HTTP 500 - Authentication provider not
  configured

B) Auth provider configured
  1) Client requests
     /authenticate?redirect_uri=http://localhost:xxx/cb
     or if the user has an invitation code:
     /authenticate?redirect_uri=..&invite_code=<invite_code>
  2) Master responds with a HTTP 302 redirect to auth provider
  authorization URL.
  3) Client opens the link in a browser
  4) Once the user permits the authorization, the auth provider
  redirects back to master's /cb with an authorization code
  5) Master exchanges this code to an access token from AP's token
  endpoint.
  6) Master requests user info from AP's user info endpoint. Using the
  access token.
  7) Local user is determined by matching the external_id field.
  For first time users an invitation code is required.
  8) Master creates a local access token with an authorization code
  9) Master responds with a redirect to
  http://localhost:xxx/cb?code=<authorization_code>
  10) Client exchanges the authorization code to an access token using
  master's /oauth2/token

The Oauth2 consumer used to authenticate from external providers should
be configurable to work with almost any provider that follows the Oauth2
specification at some level. It requires that you have created an Oauth2
application on the provider and have the client_id and client_secret.

The external access tokens are currently not stored on master as it
serves no purpose because the token is not used after the authentication.
In the future it might be used to sync user data or for example to send
notifications. At that point it would be a good idea to store them in
some encrypted format or in kontena vault.

The server's oauth2 application client secret should also be stored in
such way.

TODO/Questions:
 - how to migrate existing installations?
 - the authentication related things probably need to write something to audit logs?
 - creating service users and tokens for automation and headless operation


server/app/models/access_token.rb
```diff
@@ -1,18 +1,202 @@
...
+  validates :refresh_token, uniqueness: { scope: :internal, allow_nil: true }
+
+  field :token_type, type: String, default: 'bearer'
   field :token, type: String
   field :refresh_token, type: String
-  field :expires_at, type: Time
+  field :expires_at, type: DateTime
   field :scopes, type: Array
+  field :deleted_at, type: BSON::Timestamp, default: nil
+  field :internal, type: Boolean, default: true
```
jakolehm: What is the purpose of `internal` ?

server/app/mutations/users/invite.rb
```diff
@@ -2,24 +2,50 @@ module Users
   class Invite < Mutations::Command
     required do
       model :user
+    end
+
+    optional do
```
jakolehm: What happens if we don't receive any of the optional params?

server/app/routes/v1/audit_logs_api.rb
```diff
@@ -0,0 +1,21 @@
...
+    include RequestHelpers
+    include Auditor
+
+    plugin :streaming
+    
+    route do |r|
+      r.get do
+        halt(403, 'Access denied') unless current_user.master_admin?
+
+        limit = request.params['limit'] || 500
```
jakolehm: We should check this limit better so that we don't fetch too many items from db.

server/app/services/memory_store.rb
```diff
@@ -0,0 +1,55 @@
+# Simple memory store for Rack::Attack throttling
+class MemoryStore
```
jakolehm: Not a thread safe implementation. One option is to use [MonitorMixin](http://ruby-doc.org/stdlib-2.3.1/libdoc/monitor/rdoc/MonitorMixin.html) (or plain old mutex).

server/app/boot_jobs.rb
```diff
@@ -5,6 +5,7 @@
 unless ENV['RACK_ENV'] == 'test'
   MongoPubsub.start!(PubsubChannel.collection)
   JobSupervisor.run!
+  Mongodb::Migrator.new.migrate
```
kke: Puma hooks may not be a good spot if you want to do something like `irb -r ./server.rb`

server/app/initializers/symmetric_encryption.rb
```diff
@@ -1,10 +1,15 @@
 require 'symmetric-encryption'
 require 'symmetric_encryption/extensions/mongoid/encrypted'
 
+if ENV['VAULT_INIT']
```
kke: To init random vault keys when running locally.

server/app/models/access_token.rb
kke: internal: false is for auth provider's tokens. 
internal: true is for authenticating to this master

server/app/mutations/users/invite.rb
kke: Then errors are added through model validations:

```
        end_user.errors.each do |key, message|
          add_error(key, :invalid, message)
        end
```

server/app/routes/v1/audit_logs_api.rb
kke: The thing from container logs that @SpComb made should probably be integrated into the audit logs too. With tailing.

server/app/boot_jobs.rb
kke: Probably most obvious place would be the mongo initializer.

server/app/services/memory_store.rb
kke: Or even easier, [ActiveSupport::Cache::MemoryStore](http://apidock.com/rails/v4.2.7/ActiveSupport/Cache/MemoryStore) since we already have active support in gems.

server/app/boot_jobs.rb
kke: moved to boot.rb

server/app/helpers/token_authentication_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  include OAuth2Api::Common
+
+  def validate_access_token(*scopes)
+    # These only happen when in a "soft exclude" path where
+    # the headers are processed but request is not halted
+    # by the token authentication middleware.
+    
+    unless current_user
+      Server.logger.debug "No current user"
```
kke: fixed by using the existing Logging module for logging purposes.

server/app/initializers/symmetric_encryption.rb
kke: removed. dotenv can be used locally.

server/app/boot.rb
```diff
@@ -30,3 +30,5 @@
 Dir[__dir__ + '/jobs/**/*.rb'].each {|file| require file }
 
 Dir[__dir__ + '/services/**/*.rb'].each {|file| require file }
+
+Mongodb::Migrator.new.migrate
```
jakolehm: Still not a correct place for this.. this problem is already fixed in the master so imho this PR should not try to re-implement it differently? :)

See #999

/cc @jnummelin @nevalla

server/app/helpers/token_authentication_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+    # the headers are processed but request is not halted
+    # by the token authentication middleware.
+    
+    unless current_user
+      debug "No current user"
+      mime_halt(403, 'access_denied')
+    end
+
+    unless current_access_token
+      Server.logger.debug "No current token"
```
jakolehm: Maybe just `debug`

server/app/helpers/config_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+      @salt
+    end
+
+    def digest(string)
+      BCrypt::Engine.hash_secret(string, salt)
+    end
+
+    def config
+      return @config if @config
+      if Object::const_defined?('Server')
```
jakolehm: Why we need this if/elsif/else ?

server/app/routes/v1/audit_logs_api.rb
jakolehm: Yes, but at least we need to plug this "bug" so that server won't crash if too many logs are requested? Refactor to @SpComb implementation should be in a separate PR and not here.

server/db/migrations/13_recreate_access_tokens.rb
```diff
@@ -0,0 +1,7 @@
+class RecreateAccessTokens < Mongodb::Migration
+  def self.up
+    AccessToken.collection.drop
```
jakolehm: Is there any sane way to migrate old tokens to new ones? If yes, we should implement that in a separate PR. This is ok implementation for this PR :)

server/server.rb
```diff
@@ -13,7 +19,82 @@ class Server < Roda
...
+  def self.root_url
+    @url ||= config[:root_url]
+  end
+
+  def self.name
+    @name ||= config[:server_name]
+  end
+
+  # Global logger
+  def self.logger
```
jakolehm: Why not include `Logging` module? This seems a little bit duplicate implementation to me.

kke: Yes, that is no longer used and is now removed.

server/app/boot.rb
nevalla: I agree, let's rely on fix on master branch on this.

server/app/helpers/token_authentication_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+  include Logging
+
+  def validate_access_token(*scopes)
+    # These only happen when in a "soft exclude" path where
+    # the headers are processed but request is not halted
+    # by the token authentication middleware.
+    
+    unless current_user
+      debug "No current user"
+      mime_halt(403, 'access_denied')
```
nevalla: How this `mime_halt` differs from `halt_request`?

server/db/migrations/13_recreate_access_tokens.rb
kke: Yes. 

Something like:

```
AccessToken.each |at|
  unless at.token.start_with?('$')
    at.token_plain = at.token
    at.refresh_token = nil
    at.expires_at = nil
    at.internal = true
    at.save
  end
end
```

server/app/mutations/users/invite.rb
```diff
@@ -2,24 +2,50 @@ module Users
...
+      string :grid
```
nevalla: I think this grid option is redundant here(?)

server/app/routes/oauth2/authenticate_api.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        redirect_uri = URI.parse(params['redirect_uri'])
+
+        # Only allow redirect to client's localhost or back to same server's /code'
+        unless (redirect_uri.host && redirect_uri.host == 'localhost') || (redirect_uri.host.nil? && redirect_uri.path == '/code')
+          mime_halt(400, 'invalid_request', 'Invalid redirect_uri') and return
+        end
+
+        invite_code = params['invite_code']
+        if invite_code.kind_of?(String) && invite_code.length > 4
```
nevalla: What happens if `invite_code` is less than 5 chars?

server/app/helpers/config_helper.rb
kke: We don't. Overgeneralization.

server/app/routes/oauth2/callback_api.rb
```diff
@@ -0,0 +1,117 @@
...
+  # the AP.
+  #
+  # If the AuthorizationRequest object includes a redirect uri
+  # then the final step will redirect the browser to that uri.
+  class CallbackApi < Roda
+    include RequestHelpers
+    include Logging
+
+    route do |r|
+      r.get do
```
nevalla: This is quite a huge route. Any chance to refactor this into smaller pieces, maybe use some helper methods?

server/app/routes/v1/auth_provider.rb
```diff
@@ -0,0 +1,65 @@
...
+        @auth_provider = AuthProvider.instance
+        render('auth_provider/show')
+      end
+
+      r.post do
+        @auth_provider = AuthProvider.instance
+
+        validate_access_token('master_admin') unless current_user_admin?
+
+        halt_request(403, 'Access denied. You must be the local administrator.') unless current_user.email == 'admin'
```
nevalla: How about implementing `User#internal_admin?` (or similar) method instead of comparing email here?

server/app/boot.rb
kke: #999 doesn't run migrations except when running puma, it's not running them before tests either.

server/app/routes/v1/audit_logs_api.rb
kke: done, limited to max 3000.

server/app/helpers/token_authentication_helper.rb
kke: It responds with correct mime type and format and adds [WWW-Authenticate header](http://self-issued.info/docs/draft-ietf-oauth-v2-bearer.html#authn-header).

server/app/routes/oauth2/authenticate_api.rb
kke: The code is ignored. The callback will try to find the user by using email. But it should probably halt if you try an invalid code. Fixed now.

server/app/routes/v1/auth_provider.rb
kke: yes.

kke: there already is `is_local_admin?`

server/app/mutations/users/invite.rb
kke: Correct, removed.

kke:
Included in #1035

PR #933, opened 2016-08-24
Fix command to create master in vagrant

muhh:
Two words need to be switched.


muhh:
sorry, reverted the first fix by accident again in the second commit.

jakolehm:
Thanks!

PR #935, opened 2016-08-25
Local development with docker-compose

matti:
This workflow works for me on Docker for Mac

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,8 +10,8 @@ services:
     environment:
       - RACK_ENV=production
       - MONGODB_URI=mongodb://mongodb:27017/kontena_development
-      - VAULT_KEY=asdfasdfasdfasdfasdfasfdasdf
-      - VAULT_IV=kljhlkjhlkjhlkhjkljhkljhlkjhlkjh
+      - VAULT_KEY=asdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdf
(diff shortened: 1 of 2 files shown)


jakolehm:
I like this, especially the "dontena" part :)

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    ports:
+      - 9292:9292
+    environment:
+      - RACK_ENV=production
+      - MONGODB_URI=mongodb://mongodb:27017/kontena_development
+      - VAULT_KEY=asdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdf
+      - VAULT_IV=kljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljhkljh
+      - ACME_ENDPOINT=https://acme-staging.api.letsencrypt.org/
+    volumes:
+      - ./server:/app
```
jakolehm: Are these volume mappings really needed? Build should not take too long if Gemfile.lock does not change.

matti: No not really needed, but I think that the local volume mappings are revolutionary in the way that docker can be used to develop software locally: all that is needed is a simple restart of the service (that can be combined with fswatch etc) -- also exec works as "expected" --> once the container is running it will get it's files updated even during an exec etc.

Now only if the #936 would be resolved so that the restart would actually be fast (and not have to use kill)

PR #936, opened 2016-08-25
PoC: shutdown gracefully on SIGTERM

matti:
When developing locally for example with #935, shutting down api and agent take time because api or agent don't shutdown gracefully on SIGTERM -- docker-compose will eventially SIGKILL them.

I don't know enough of the codebase to make a real pull-request of this, but something like this patch should be done?

As it is now:
ezgif com-video-to-gif 1

With this PoC patch:
ezgif com-video-to-gif 2

### DIFF
--- a/agent/bin/kontena-agent
+++ b/agent/bin/kontena-agent
@@ -35,4 +35,5 @@ agent = Kontena::Agent.new(
 )
 agent.connect!
 
-sleep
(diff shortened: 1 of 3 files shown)


jakolehm:
Thanks for spotting this! Closing this one because we have a little bit different implementation in #937

matti:
Cool.

PR #940, opened 2016-08-29
Feature/cli authentication renewal

kke:
Big changes to authentication from CLI and improvements to the Kontena::Command that add support for running callbacks after commands.


kke:
Included in #1035

PR #945, opened 2016-08-30
Update puma to 3.6.0

jakolehm:
Older pumas seem to lock in some circumstances under ruby 2.3.x.


jakolehm:
@nevalla @jnummelin should we actually change milestone to 0.15.1?

jnummelin:
I think we should

jakolehm:
@nevalla, waiting for your comment on this...

nevalla:
Yes we should

Issue #943, opened 2016-08-30
Certificates not trusted by all client software


jnummelin:
There are some client software (wget, some Rubies, Docker, ...) that doesn't seem to trust the LE issued certificates if they are used as standalone. All browsers seem to trust LE intermediaries directly. Instead we should provide option for user to use fullchain certificates so that the whole trust chain can be established. Maybe the best option is to by default use fullchain certificates and if user opts for plain certificate use that.

matti:
let's encrypt is especially great for SSLing kontena registry with the kontena loadbalancer in less than 20 seconds. This usecase doesn't work right now (docker push fails).

jnummelin:
fixed in #946

PR #946, opened 2016-08-30
Default to fullchain certificate with options for cert only or chain

jnummelin:
This PR changes certificate management to default to using fullchain certificates as plain certs of LE are not overly trusted with all client libraries. (some rubies, docker, wget, ...). Also adds cli options to get either chain, fullchain or plain cert.


matti:
#943

nevalla:
Looks good!

jakolehm:
@jnummelin tagged this to 0.15.1, is it ok? .. and should we add documentation for this new option?

jnummelin:
version is ok. I'll add some docs for the options.

jnummelin:
docs updated

PR #962, opened 2016-09-02
kontena.yml reference improvements

jakolehm:
- separate sections for kontena/compose keys
- misc fixes

### DIFF
--- a/docs/references/kontena-yml.md
+++ b/docs/references/kontena-yml.md
@@ -15,6 +15,124 @@ You can use environment variables in configuration values with a Bash-like ${VAR
 ## Service configuration reference
 > **Note:** Kontena supports both Docker Compose file versions respectively. Volumes and networks introduced in version 2 are not supported by Kontena yet. See more details about versioning on [Docker Compose documentation](https://docs.docker.com/compose/compose-file/#versioning)
 
+### Kontena specific keys
+
+#### instances
+
+Number of instences (replicas) to run for the service (default: 1).
+
+```
+instances: 1
+```
+
+#### stateful
+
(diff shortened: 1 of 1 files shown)


matti:
So much clearer!

jakolehm:
@nevalla ping

docs/references/kontena-yml.md
```diff
@@ -471,33 +473,36 @@ db:
...
+services:
+  loadbalancer:
+    image: kontena/lb:latest
+    ports:
+      - 80:80
+  app:
+    build: .  
+    image: registry.kontena.local/example-app:latest
+    instances: 2
+    privileged: true
```
nevalla: While editing this example, could we just remove this `privileged` option. It's not the most common option.

nevalla:
Definitely looks better now!

docs/references/kontena-yml.md
```diff
@@ -15,6 +15,124 @@ You can use environment variables in configuration values with a Bash-like ${VAR
 ## Service configuration reference
 > **Note:** Kontena supports both Docker Compose file versions respectively. Volumes and networks introduced in version 2 are not supported by Kontena yet. See more details about versioning on [Docker Compose documentation](https://docs.docker.com/compose/compose-file/#versioning)
```
matti: Could these "volumes and networks not supported" be moved under `#### Not supported in Kontena` heading -- I remember misinterpreting this somehow in the past.

Also, `working_dir:` is not supported in kontena.yml, that would fit under this category.

jakolehm: Maybe we don't need that "not supported" section because we list all supported keys.. everything else is not supported?

matti: Not required. Just saying that I've assumed twice that working_dir is valid kontena.yml.

But yes, there is validation that tells you that it is not.

PR #965, opened 2016-09-02
Allow to use strings as value of extends option in kontena.yml

### LINKED ISSUE #956: Service '' not found in kontena.yml error when extending in compose style
https://github.com/matti/kontena-extends-repro

nevalla:
Fixes #956


matti:
🍺  deserved.

jakolehm:
LGTM!

PR #972, opened 2016-09-05
Fix cli log stream buffer mem leak

### LINKED ISSUE #963: tailing service logs - memory consumption
When I run:

`kontena service logs -t fancyname`

The CLI immediately begins to eat memory, eventually crashing. During the times that ive had issues there's been no output happening on the service.

memoryabused
 

Version: kontena-cli (0.15.1)

jakolehm:
Memory leak happens only when kontena cli gets partial json chunks from a log stream. This PR fixes partial json buffering and adds failsafety so that buffer cannot ever grow infinitely.

Fixes #963

### DIFF
--- a/cli/lib/kontena/cli/grids/logs_command.rb
+++ b/cli/lib/kontena/cli/grids/logs_command.rb
@@ -1,6 +1,9 @@
+require_relative '../helpers/log_helper'
+
 module Kontena::Cli::Grids
   class LogsCommand < Clamp::Command
     include Kontena::Cli::Common
+    include Kontena::Cli::Helpers::LogHelper
 
     option ["-t", "--tail"], :flag, "Tail (follow) logs", default: false
     option "--lines", "LINES", "Number of lines to show from the end of the logs"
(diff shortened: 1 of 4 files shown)


jakolehm:
@kke ping

SpComb:
LGTM on the `@buffer << chunk` vs `@buffer = chunk` logic fix.

The larger the log messages or the slower the network, the more likely it is to happen, as excon is less likely to read the entire HTTP chunk in 1-2 socket reads: https://github.com/excon/excon/blob/v0.49.0/lib/excon/response.rb#L115

With a bit of logging of the buffer + chunk sizes on the current implementation, I see it happening at approx ~15KB with a vagrant master:

```
stream_logs buffer 0 + chunk='{... ' 6366 bytes
stream_logs buffer 6366 + chunk='2...:' 1448 bytes
stream_logs buffer 14180 + chunk='3...}' 313 bytes
stream_logs buffer 28673 + chunk='{...6' 7234 bytes
stream_logs buffer 64580 + chunk='T...}' 920 bytes
stream_logs buffer 130080 + chunk='{...6' 7234 bytes
stream_logs buffer 267394 + chunk='T...}' 947 bytes
stream_logs buffer 535735 + chunk='{...:' 7733 bytes
stream_logs buffer 1079203 + chunk='3...}' 475 bytes
```

I don't think the buffer size limit at 10KB is a good idea though. Imagine a JSON-encoded log message that is larger than 10KB. If you have a >10KB `@buffer = "{..."` and you discard it, then the next chunk will contain a partial "...}" JSON message. This leaves the JSON stream parsing desynchronized, and would kill all further log messages.

jakolehm:
I agree that buffer size was a bad idea, removed it and added test for bigger chunks.

jakolehm:
And hopefully the final fix is in: skip whitespace chunks if buffer is empty.

jakolehm:
@kke @SpComb all good now? should we hit merge button?

SpComb:
LGTM, I've tested this + #973 

Only remaining issue is the reliance on HTTP `Transfer-Encoding: chunked` for JSON message boundries, which could potentially lead to similar issues with a `@buffer` containing multiple JSON messages. I haven't been able to figure out any sane case that would triggering that, though, so it's probably worth a separate issue to consider and address that. Hitting the exponential `@buffer` growth bug is the most likely explanation for #963.

SpComb:
Separate issue for the chunked message boundaries filed:  #974

PR #973, opened 2016-09-05
Fix server log stream thread leak

jakolehm:
Thread leak happens only when stream is closed from the client side and no log items are coming in from the database. This PR fixes this situation by appending empty strings to the output that will trigger socket close eventually.


jakolehm:
@kke ping

SpComb:
LGTM.

Having some kind of keepalive mechanism to trigger TCP send()s and the resulting timeouts is probably a good idea. Just dealing with disconnecting clients that close() their TCP connection is not enough. If a client dies without closing the TCP connection, it will take a long time to notice unless one or the other side is actively sending TCP frames.

PR #977, opened 2016-09-07
Use correct IP address on quick start guide

### LINKED ISSUE #976: Docs: getting-started IP for connecting to the worpress app after deploy is incorrect on the website
Hi,

from: https://kontena.io/docs/getting-started/quick-start
The section:
After the deploy is finished you can verify it using

$ kontena app show wordpress

It should show details of the service. You can pick the private ip address of the service and verify in a browser that application is responding.

This should be the private IP of the node as per conversation with Lauri in the slack channel

nevalla:
This PR fixes Quick Start guide where currently user is asked to pick up the private IP address of the service. He/she should use private IP address of the node instead.

Fixes #976

### DIFF
--- a/docs/getting-started/quick-start.md
+++ b/docs/getting-started/quick-start.md
@@ -143,8 +143,8 @@ After the deploy is finished you can verify it using
 $ kontena app show wordpress
 ```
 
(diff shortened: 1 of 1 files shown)


jakolehm:
LGTM!

PR #984, opened 2016-09-07
Test and bugfix cli app logs, with spec tests

SpComb:
Some bugfix work on `kontena app logs` to fix some issues, with spec tests and refactoring: 

* Add spec tests using a mock client implementation
* Refactor into separate `tail_logs()`, `show_logs()`, `get_logs()` and `show_log()` for easier testing
* Change `--tail` option to be an Integer
* Fix the incorrect use of `?limit=` even when tailing logs with `?from=`

    This would lead to skipped log lines if any service was logging more than `--limit` lines per the 2 second loop iteration

* Track per-service `?from=` offset when tailing

    This addresses a *potential* race condition where new service logs appear while looping through multiple `services/..../container_logs` requests for different services. Imagine the following scenario:

    * Service A has log messages 1...3
    * Service B has log messages 4..5
    * We request service A log messages -> 1..3
    * Service A gets a log message 6
    * Service B gets a log message 7
    * We request service B log messages -> 4..5,7
    * We enter the next loop iteration with `last_id=7`
    * Service A gets a log message 8
    * We request Service A logs with `?from=7` -> 8

    We have lost the Service A log message 6.

    Note that I have not actually tried to trigger such a race condition, this is just theoretical :)

Observe some `rspec spec/kontena/cli/app/logs_command_spec.rb -e 'tails'` mock client trace with the correct request options:

```
Randomized with seed 19971

GET services/testgrid/test-wordpress/container_logs {"limit"=>100}
GET services/testgrid/test-mysql/container_logs {"limit"=>100}
GET services/testgrid/test-wordpress/container_logs {"from"=>"57cff2e8cfee65c8b6efc8bf"}
GET services/testgrid/test-mysql/container_logs {"from"=>"57cff2e8cfee65c8b6efc8bd"}
GET services/testgrid/test-wordpress/container_logs {"from"=>"57cff2e8cfee65c8b6efc8c2"}
GET services/testgrid/test-mysql/container_logs {"from"=>"57cff2e8cfee65c8b6efc8c1"}
.

Finished in 0.01271 seconds (files took 0.30931 seconds to load)
1 example, 0 failures
```

If the commits look sane, I'll do a bit more integration testing against the server to make sure this works correctly.

### DIFF
--- a/cli/lib/kontena/cli/apps/logs_command.rb
+++ b/cli/lib/kontena/cli/apps/logs_command.rb
@@ -8,7 +8,9 @@ module Kontena::Cli::Apps
 
     option ['-f', '--file'], 'FILE', 'Specify an alternate Kontena compose file', attribute_name: :filename, default: 'kontena.yml'
     option ['-p', '--project-name'], 'NAME', 'Specify an alternate project name (default: directory name)'
-    option ["-l", "--lines"], "LINES", "How many lines to show", default: '100'
+    option ["-l", "--lines"], "LINES", "How many lines to show", default: 100 do |s|
+      Integer(s)
+    end
     option "--since", "SINCE", "Show logs since given timestamp"
     option ["-t", "--tail"], :flag, "Tail (follow) logs", default: false
     parameter "[SERVICE] ...", "Show only specified service logs"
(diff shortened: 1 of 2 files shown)


SpComb:
And naturally you only realize it once you've submitted the PR, but the `kontena app logs` should probably just be reimplemented on top of `v1/grids/:name/container_logs?service=...`. It would be less complicated and probably perform better, and all of the cli logs commands could use the same `buffered_log_json` implementation.

I'll update this PR with that approach, unless there's some particular reason to stick with this. I suppose it would change the behavior of `--tail` to limit the total number of lines, rather than per-service but... yeah :)

matti:
That --tail change sounds initally bad, I'm not sure if I got it correctly..

Would service logs --tail N give you N lines of logs or not?

jakolehm:
Good clean code! But I agree that it would be better to implement this on top of streaming server side api... it accepts multiple services as parameters so implementation should be pretty straightforward.

SpComb:
> That --tail change sounds initally bad, I'm not sure if I got it correctly.. Would service logs --tail N give you N lines of logs or not?

Implementing `kontena app logs --tail` on top of `v1/grids/:name/container_logs?follow=1&limit=100&service=....` would give up to 100 lines of the most recent logs across all services combined.

The current `kontena app logs --tail` gives up to 100 lines of logs *per service*, i.e. there may be more than 100 lines of output.

I doubt it's a behavior that's worth preserving, and the current behavior is probably somewhat confusing, as you would get a long history of output from quiet services, intermingled with only very recent output from chatty services.

matti:
Agreed

SpComb:
Replaced with  #987

SpComb:
Note that the fix for the `?limit=` dropping log mesages here is presumably invalid, because omitting the `?limit=` parameter will simply use the default `limit = 100` on the server api...

PR #985, opened 2016-09-07
Experimental completer script

kke:
Scans the commands instead of relying on hardcoded lists.
Also completes options (if you start with - and then tab)

It may be a lot slower than the previous one, but it could be made a lot faster by building a cache when running `gem build` or something like that.

Completers will need to be defined to the command defininitions.
To complete for example grids in "kontena grid use", your grid use_command
needs a method like:
def self.complete(query=nil)
  if query.nil?
    get('grids')['grids'].map{|g| g['name']}
  else
    get('grids')['grids'].select{|g| g['name'].start_with?(query)}.map{|g| g['name']}
  end
end

or if you have an option like --grid GRID_NAME
then your command needs something like:

def self.complete_option(opt, query=nil)
  if opt == '--grid'
    # same stuff here
  end
end


kke:
New version, quite nice now in my opinion.

Example command with completer support:

```
class Kontena::Cli::Nodes::UpdateCommand < Kontena::Command
  option ['--grid'], '[GRID]', "Grid name"
  parameter "NODE_ID", "Node id"

  def complete_grid(query = nil)
    require_api_url
    names = client.get("grids")["grids"].map{|grid| grid['name']}
    if query.nil?
      names
    else
      names.select{ |name| name.start_with?(query) }
    end
  end

  def complete(query = nil)
    # list + filter nodes like grids above
  end

  def execute
    puts "..."
  end
end
```

kke:
Better implementation in kontena-cli-completer repo.

PR #987, opened 2016-09-08
CLI: Reimplement app logs, with spec tests

SpComb:
Reimplement `kontena app logs` using `Kontena::Cli::Helpers::LogHelper` with the server `/v1/grids/.../container_logs?services=...` API. This is version 2 of #984.

This resolves several issues with the earlier implementation: 

* The incorrect use of `?limit=` even when tailing logs with `?from=` would lead to skipped log lines if any service was logging more than `--limit` lines per the 2 second loop iteration
* A *potential* race condition where new service logs appear while looping through multiple `services/..../container_logs` requests for different services.

This changes the behavior of `--lines`, which now limits the total number of output lines across all services. Previously `--lines` applied per service, which could give up to (limit)*(number of services) lines of output. The previous behavior is probably somewhat confusing, as you would get a long history of output from quiet services, intermingled with only very recent output from chatty services.

Summary of changes:

* Add spec tests using a mock client implementation
* Add `LogHelper#stream_logs()` method
* Refactor `LogsCommand` into separate `tail_logs()`, `show_logs()`, `show_log()` for easier testing
* Change `--tail` option to be an Integer
* Change `kontena grid logs` to also use the `LogHelper#stream_logs()` method

TODO is to verify that the various server-side logs endpoints all behave similarly enough to use the same CLI `LogHelper` implementation for all cli logs commands, and add tests for `LogHelper#stream_logs`.

### DIFF
--- a/cli/lib/kontena/cli/apps/logs_command.rb
+++ b/cli/lib/kontena/cli/apps/logs_command.rb
@@ -1,9 +1,12 @@
 require_relative 'common'
+require_relative '../helpers/log_helper'
 
 module Kontena::Cli::Apps
   class LogsCommand < Clamp::Command
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
+    include Kontena::Cli::Helpers::LogHelper
+
     include Common
 
     option ['-f', '--file'], 'FILE', 'Specify an alternate Kontena compose file', attribute_name: :filename, default: 'kontena.yml'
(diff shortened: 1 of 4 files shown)


jakolehm:
LGTM

ping @nevalla

jnummelin:
👍

cli/lib/kontena/cli/apps/logs_command.rb
```diff
@@ -1,72 +1,65 @@
...
     include Kontena::Cli::GridOptions
+    include Kontena::Cli::Helpers::LogHelper
+
     include Common
 
     option ['-f', '--file'], 'FILE', 'Specify an alternate Kontena compose file', attribute_name: :filename, default: 'kontena.yml'
     option ['-p', '--project-name'], 'NAME', 'Specify an alternate project name (default: directory name)'
-    option ["-l", "--lines"], "LINES", "How many lines to show", default: '100'
+    option ["-l", "--lines"], "LINES", "How many lines to show", default: 100 do |s|
+      Integer(s)
```
nevalla: I think more common Ruby way to do this is `s.to_i`

nevalla:
Just a minor comment, otherwise looks great. Let's include this in 0.15.x since this is more bug fix than new feature.

jnummelin:
yep, in 0.15.2 it shall go :)

cli/lib/kontena/cli/apps/logs_command.rb
SpComb: I used the `Integer(s)` syntax from the clamp examples.

I dropped this change from this PR, it's not really relevant. If I end up refactoring all the CLI logs commands, then I'll fix this for all of them :)

SpComb:
Updated PR without the `--tail` option integer type conversion (definition and specs). No other changes.

PR #989, opened 2016-09-08
Catch service remove timeout error and rollback to prev state

### LINKED ISSUE #828: Occasionally unable to remove services
We have occasionally bumped into the problem of not being able to remove services.
After a `service remove`, the service state is stuck in `deleting`. In some of the cases executing `remove` again will successfully remove the service. In other nothing happens and the state remains in `deleting`. It has taken hours for some of the services to be removed once stuck in that state.

jakolehm:
In some cases service remove might timeout and this leaves service to "deleting" state. This PR fixes these situations by dynamically adjusting timeout / rollbacking service state.

Fixes #828

### DIFF
--- a/server/app/jobs/grid_service_remove_worker.rb
+++ b/server/app/jobs/grid_service_remove_worker.rb
@@ -10,19 +10,25 @@ class GridServiceRemoveWorker
 
   # @param [GridService] grid_service
   def remove_grid_service(grid_service)
-    prev_state = grid_service.state
-    grid_service.set_state('deleting')
-    grid_service.containers.scoped.each do |container|
-      terminate_from_node(container.host_node, container.name)
+    begin
+      prev_state = grid_service.state
+      grid_service.set_state('deleting')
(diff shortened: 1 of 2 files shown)


server/app/jobs/grid_service_remove_worker.rb
```diff
@@ -10,19 +10,25 @@ def perform(grid_service_id)
...
+        terminate_from_node(container.host_node, container.name)
+      end
 
-    wait_instance_removal(grid_service)
+      wait_instance_removal(grid_service, grid_service.containers.scoped.count * 30)
 
-    grid_service.destroy
+      grid_service.destroy
+    rescue Timeout::Error
+      grid_service.set_state(prev_state)
```
jnummelin: should we log something to see when this happens?

jakolehm: good point, will add

jnummelin:
other than the log thingy, LGTM

PR #990, opened 2016-09-08
Always add etcd dns address when agent boots

### LINKED ISSUE #988: Etcd dns address is sometimes lost

jakolehm:
This PR fixes situation where etcd has started by docker (probably on host node boot) before agent.

Fixes #988


nevalla:
LGTM

PR #993, opened 2016-09-09
Allow to use numeric version value in kontena.yml

### LINKED ISSUE #986: Invalid version number yields confusing error message
If you have in your kontena.yml version defined as an integer as appose to string the error message kontena gives is:
```
kontena app build
Kontena error: no implicit conversion of Hash into String
```

Compared to docker-compose which in similar situation gives much more clearer error message:
```
docker-compose build
ERROR: Version in "./docker-compose.yml" is invalid - it should be a string.
```

nevalla:
This PR fixes error when using numeric value of version in kontena.yml, like the following

```
version: 2
name: foo

services:
  bar:
    image: konttisample
    build:
      context: .
      dockerfile: Dockerfile
```

Previously version was not recognised correctly when generating service config for the YAML file and build options get corrupted. This PR fixes this situation by casting version value to integer before selecting the service generator class.

Fixes #986

### DIFF
--- a/cli/lib/kontena/cli/apps/common.rb
+++ b/cli/lib/kontena/cli/apps/common.rb
@@ -33,7 +33,7 @@ module Kontena::Cli::Apps
     # @return [Hash]
     def generate_services(yaml_services, version)
       services = {}
-      if version == '2'
+      if version.to_i == 2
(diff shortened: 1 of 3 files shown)


jakolehm:
LGTM

PR #994, opened 2016-09-09
Handle malformed YAML files in a sane way

nevalla:
When YAML files are malformed Kontena does not handle them very well atm. This PR will catch those errors and outputs error message instead:

When YAML file is not well structured for example:
```
version: '2'
services:
  wordpress:
    build: .
      context: .
      dockerfile: Dockerfile.alternative
```

Instead of this output
```
Kontena error: (<unknown>): mapping values are not allowed in this context at line 5 column 14
```

The following output is displayed:
```
Error while parsing ./kontena/cli/spec/fixtures/kontena-malformed-yaml.yml (<unknown>): mapping values are not allowed in this context at line 5 column 14
```

Also when service level config is not Hash previously the following output was displayed
```
Kontena error: undefined method `dig' for "service config cannot be string":String
```

With this PR the output will look like this:
```
YAML validation failed! Aborting.
./kontena/cli/spec/fixtures/kontena-not-hash-service-config.yml:
  wordpress:
    - options: "must be a mapping not a string"
```

### DIFF
--- a/cli/lib/kontena/cli/apps/yaml/reader.rb
+++ b/cli/lib/kontena/cli/apps/yaml/reader.rb
@@ -214,11 +214,7 @@ module Kontena::Cli::Apps
             options['build']['args'][k] = v
           end
         end
-      end
-
-      def abort_on_malformed_options(file, service_name)
-        #abort("Service '#{service_name}' contains malformed options #{file}".colorize(:red))
-      end
+      end      
     end
   end
 end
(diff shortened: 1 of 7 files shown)


jakolehm:
Nice!

cli/lib/kontena/cli/apps/yaml/reader.rb
```diff
@@ -199,6 +215,10 @@ def normalize_build_args(options)
           end
         end
       end
+
+      def abort_on_malformed_options(file, service_name)
```
jakolehm: This method does nothing... remove?

nevalla: Good catch, it was some leftover and now removed

jakolehm:
Tests fail...

nevalla:
yes I have to do some refactoring and clean harderer.

jakolehm:
LGTM, but is the hardened cleaning complete?

jakolehm:
@nevalla ping

nevalla:
Cleaning complete.

PR #995, opened 2016-09-09
Server: Refactor API container_logs endpoints, fix limit behavior

SpComb:
The current server API implementation contains four separate endpoints for container logs, each of which is implemented in a different way:

* [`GET /v1/services/:grid_name/:service_name/container_logs`](https://github.com/kontena/kontena/blob/v0.15.1/server/app/routes/v1/services_api.rb#L73)
* [`GET /v1/grids/:name/container_logs`](https://github.com/kontena/kontena/blob/v0.15.1/server/app/routes/v1/grids/grid_container_logs.rb#L4)
* [`/v1/grids/:name/container_logs`](https://github.com/kontena/kontena/blob/v0.15.1/server/app/routes/v1/grids_api.rb#L56) (old broken implementation!)
* [`GET /v1/containers/:grid_name/:service_name/:name/logs`](https://github.com/kontena/kontena/blob/v0.15.1/server/app/routes/v1/containers_api.rb#L43) (not exposed in the CLI, see #1001)

This PR refactors all of those API endpoints to use a single `LogsHelper#render_container_logs` method, with rspec tests. This PR includes the following changes:

* Uses the correct mongodb ordering to avoid any unnecessary `Array#reverse` operations when following logs.
* The `?limit=` parameter is bounded to a `1..10000` range.  Invalid integer values are ignored, using the default value of 100. It should not be possible to request an unlimited batch of logs, which the server would not be happy about rendering.
* The `?follow=1` implementation queries at most `LOGS_LIMIT_MAX` lines at a time. This should not be a user-visible change.
* ***Minor API change***: The non-following `?from=...`  implementation is changed to return the *next* limit lines after the given id, not the *last* limit lines. IMO this makes more sense for anything using `?from=`, and would actually have avoided the log-skipping issue with `kontena app logs`.
* The same change also applies to `?follow=1&from=...`. This should mean that the CLI `kontena service/grid logs` behavior to retry on SocketErrors should not drop any logs while retrying. Using a very small `?from=` value will return a lot of logs, but in batches of 10k lines..


SpComb:
Rebased against 0.15.2 master

jakolehm:
LGTM!

PR #997, opened 2016-09-09
cli: handle invalid master name gracefully and improved formatting

christofferh:
Before:
```
➜  ~ kontena master use invalid_name
Kontena error: no implicit conversion of String into Integer

```
After:
```
➜  ~ kontena master use invalid_name
Could not resolve master with name: invalid_name

You have access to following grids:

  * gcp-1
  * gcp-2
➜  ~ kontena master use gcp-1
Using master: gcp-1 (https://master.kontena.com)
Using grid: gcp-1
```

### DIFF
--- a/cli/spec/kontena/cli/master/use_command_spec.rb
+++ b/cli/spec/kontena/cli/master/use_command_spec.rb
@@ -26,13 +26,13 @@ describe Kontena::Cli::Master::UseCommand do
       subject.run(['some_master'])
     end
 
-    it 'should fetch grid list from master' do
+    it 'should fetch grid list from master if invalid master name given' do
       allow(subject).to receive(:require_token).and_return('token')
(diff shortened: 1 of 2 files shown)


cli/spec/kontena/cli/master/use_command_spec.rb
```diff
@@ -26,13 +26,13 @@
       subject.run(['some_master'])
     end
 
-    it 'should fetch grid list from master' do
+    it 'should fetch grid list from master if invalid master name given' do
```
jakolehm: Should this be a new test?

christofferh: I reverted this behaviour to the original. (But also added a new test for the failure case)

jakolehm:
LGTM

PR #999, opened 2016-09-11
Fix boot process / migration race condition

jakolehm:
Master bootup process has a race condition related to how database migrations are handled. This PR moves migrations to `on_worker_boot` puma hook that will block boot until migrations have completed. In a multi-master scenarios migration will acquire db lock before running migrations (same logic as before).


nevalla:
LGTM!

PR #1000, opened 2016-09-11
Fix HostNode#initial_member? error when node_number is nil

jakolehm:
`HostNode#initial_member?` throws error if `HostNode#node_number` is not yet set. This PR fixes this by returning false is node_number is not set.

### DIFF
--- a/server/app/models/host_node.rb
+++ b/server/app/models/host_node.rb
@@ -108,6 +108,7 @@ class HostNode
   end
 
   def initial_member?
+    return false if self.node_number.nil?
     return true if self.node_number <= self.grid.initial_size
     false
   end
--- a/server/spec/models/host_node_spec.rb
+++ b/server/spec/models/host_node_spec.rb
@@ -70,6 +70,10 @@ describe HostNode do
     it 'returns false if not initial_member' do
       expect(node_2.initial_member?).to be_falsey
     end
+
+    it 'returns false if node_number is not set' do
+      expect(subject.initial_member?).to be_falsey
+    end
   end
 
   describe '#attributes_from_docker' do


jakolehm:
@nevalla @jnummelin ping

jnummelin:
I'm just wondering in what cases can the node number be actually `nil`? Shouldn't it get set always when the agent calls home.

jakolehm:
There is a very small time window when node number is nil (because it's done `after_save`)... this happens pretty frequently if you spin up many nodes (I tested with 20 parallel node provisions).

matti:
I was wondering the same thing. To me it sounds like this should be handled in some other level and not here...?

jakolehm:
Okay, it seems that my fix is too confusing... I still think that this method should not blow up when `node_number` is nil ;) But then again, maybe we can improve how node_number is set. Will push some additional fixes for that later.

jakolehm:
Nope, did not find any better way to set node_number. Should we merge this fix or not?

jnummelin:
the code itself is good, just "smelled" a bit wrong place to fix this. but as there's no other alternative: LGTM :)

nevalla:
I think we can fix this now and improve the node number assigning later.

kke:
Create a NodeNumber model (and a migration) and use find_and_modify to mark numbers used with find_and_modify which finds and updates the model in single query (or returns nil)

```ruby
class NodeNumber
  include Mongoid::Document
  include Mongoid::DateStamps

  belongs_to :node
  field :node_number, type: Fixnum
  index({ node_number: 1 }, { unique: true})
  index({ node_id: 1 }, { sparse: true, unique: true})

  #@param [Node] instance_of_node 
  # @return [NodeNumber, NilClass] returns nil if no free numbers
  def self.reserve_first_available(node)
    HostNode.where(node: nil).find_and_modify({ '$set' => { node_id: node.id } })
  end

end
```

Migration:
```
   def self.up
     NodeNumber.create_indexes
     (1..9999).each |num|
       NodeNumber.create(node_number: num)
     end
  end
```

jakolehm:
@kke thanks! I think that will be a good way to handle this.. but let's create a separate PR for that.

kke:
a bit hacky to prepopulate a table with a running number..

PR #1001, opened 2016-09-12
CLI: Refactor logs commands, add container logs command

SpComb:
This PR continues the work from #995 to also refactor the cli to use a single `LogHelper` implementation, matching the server `LogsHelpers`.

* Moves as much of the implementation into the common `LogsHelpers` as possible for consistency, including the command-line option definitions.
* Changes the `--lines` option to use `{|s| Integer(s)}`, which provides syntax validation:

        $ kontena container logs -t --lines=100x test-1
        ERROR: option '--lines': invalid value for Integer(): "100x"

    Note that using `s.to_i` would silently return `0`.
* The output format used by the various logs commands should remain identical.
* Adds a `kontena container logs service-1` command, with "end to end" args-get-output specs.

AFAIK this refactoring should preserve the exact behavior of the existing commands...


SpComb:
Updated for ruby 2.x compat to replace indented heredoc `<<~LOG` syntax introduced in ruby 2.3.

SpComb:
Updated for 0.15.2 master.

jakolehm:
LGTM

PR #1005, opened 2016-09-12
Retry when unknown exception occurs while streaming docker events

jakolehm:
This PR fixes a possible bug that might cause event stream to stop by adding catch-all exception handler that will retry event streaming if processing flag is still enabled.


nevalla:
LGTM!

PR #1011, opened 2016-09-13
Add docs for master/agent/cli environment variables

jakolehm:


docs/references/environment-variables.md
```diff
@@ -0,0 +1,33 @@
...
+# Environment Variables
+
+## Kontena Master
+
+- `MONGODB_URI`: MongoDB connection uri (required)
+- `VAULT_KEY`: secret key for the Kontena Vault (required)
+- `VAULT_IV`: initialization vector for the Kontena Vault (required)
+- `WEB_CONCURRENCY`: number of forked master api processes (default: 1)
+- `MAX_THREADS`: number of threads inside single master api process (default: 8)
+- `LOG_LEVEL`: logging level
```
jnummelin: Possible values?

docs/references/environment-variables.md
```diff
@@ -0,0 +1,33 @@
...
+- `AUTH_API_URL`: specifies authentication server url (default: https://auth.kontena.io)
+
+## Kontena Agent
+
+- `KONTENA_URI`: Kontena Master websocket uri (required)
+- `KONTENA_TOKEN`: Kontena Grid token (required)
+- `KONTENA_PEER_INTERFACE`: network interface for peer/private communication (default: eth1)
+- `KONTENA_PUBLIC_IP`: specify node public ip, overrides default resolving
+- `KONTENA_PRIVATE_IP`: specify node private ip, overrides default resolving
+- `LOG_LEVEL`: logging level
```
jnummelin: Possible values?

docs/references/environment-variables.md
```diff
@@ -0,0 +1,33 @@
...
+toc_order: 2
+---
+
+# Environment Variables
+
+## Kontena Master
+
+- `MONGODB_URI`: MongoDB connection uri (required)
+- `VAULT_KEY`: secret key for the Kontena Vault (required)
+- `VAULT_IV`: initialization vector for the Kontena Vault (required)
```
jnummelin: Document also the requirements for these, min length etc.

docs/references/environment-variables.md
```diff
@@ -0,0 +1,33 @@
...
+- `WEB_CONCURRENCY`: number of forked master api processes (default: 1)
+- `MAX_THREADS`: number of threads inside single master api process (default: 8)
```
jnummelin: Should we document some "rules of thumb" for these? Something like: "For 2 cpu machines use `WEB_CONCURRENCY=2` && `MAX_THREADS=16`"

matti: Definitely. For a new user who doesn't know how the process behaves (or has no experience from running Ruby app servers) having some "most likely to be fine" initial values is great.

docs/references/environment-variables.md
matti: And also if there can be multiple values or not

kke:
Also, clamp allows you to set `environment_variable:' to options, we should add more of those.

docs/references/environment-variables.md
jakolehm: I agree.. but master should check number of cpus and adjust this value automatically if user has not set env. (celluloid already has this kind of helper so we could use it)

jnummelin: Setting these automatically is a good idea.

SpComb:
kontena-agent 0.16.3 adds:

* `WEAVE_DEBUG=`

kontena-agent 0.17 also adds:

* `IPAM_IMAGE=kontena/ipam-plugin`
* `IPAM_VERSION=latest`

kke:
Needs rebase

kke:
How about the variables inside containers? @jakolehm @nevalla @jnummelin 

```ruby
 92       env << "KONTENA_SERVICE_ID=#{grid_service.id.to_s}"
 93       env << "KONTENA_SERVICE_NAME=#{grid_service.name}"
 94       env << "KONTENA_GRID_NAME=#{grid_service.grid.try(:name)}"
 95       env << "KONTENA_STACK_NAME=#{grid_service.stack.try(:name)}"
 96       env << "KONTENA_NODE_NAME=#{host_node.name}"
 97       env << "KONTENA_SERVICE_INSTANCE_NUMBER=#{instance_number}"
```

PR #1017, opened 2016-09-16
Add support for building cli docker image

### LINKED ISSUE #1016: Package cli also as Docker container
Would be nice to have cli also packaged as a Docker container. Would minimize Ruby/gem version issues on environments where docker is present. And enable to use it eg on Drone.

jakolehm:
Fixes #1016


nevalla:
LGTM

PR #1019, opened 2016-09-16
Cli installer for OSX (using omnibus)

jakolehm:


jakolehm:
Will create a new PR after I clean the commits...

PR #1020, opened 2016-09-18
Reconnect event stream always if it stops without error

jakolehm:


agent/lib/kontena/workers/event_worker.rb
```diff
@@ -34,9 +34,11 @@ def stream_events
...
       defer {
         begin
-          Docker::Event.stream({filters: filters}) do |event|
-            raise "stop event stream" unless processing?
-            @event_queue << event
+          while processing?
+            Docker::Event.stream({filters: filters}) do |event|
+              raise "stop event stream" unless processing?
+              @event_queue << event
+            end
```
jnummelin: Maybe log something when the stream closes. Just so that we see how often this happens.

jakolehm: Added info level logging

jakolehm:
@jnummelin ok for merge?

PR #1024, opened 2016-09-19
Change service container restart policy to unless-stopped

### LINKED ISSUE #881: Docker restart policy and service state clashes
Now all service containers are started with `restart-always` option. Now when service is stopped by Kontena and Docker engine restarts, it still restarts the containers as the policy says.

We should change the restart policy upon service stop as it is now possible with Docker > 1.11

jakolehm:
Fixes #881


jnummelin:
AFAIK `unless-stopped` was introduced in Docker 1.9 so this basically drops support for anything older than that. Shouldn't be an issue, just wanted to point out. :) Nobody should be running anything older that that. :D

PR #1033, opened 2016-09-22
Improve travis automation

jakolehm:
- use bundler cache
- install mongodb stuff only when `$TEST_DIR=server`
- deploy gem & docker images on tags
  - if a pre-release, tag image also as `:edge`
 - if not a pre-release, tag with semver and `:latest`

travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,26 @@
 language: ruby
+sudo: required
 rvm:
   - 2.0.0-p648
   - 2.1.8
   - 2.2.3
   - 2.3.1
 cache: bundler
-addons:
-  apt:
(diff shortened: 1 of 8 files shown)


jakolehm:
@kke @SpComb ping, should we merge?

kke:
LGTM

SpComb:
Unfortunately I really don't know enough about Travis to say.

jakolehm:
Ok, let's do this...

kke:
If it's broken, it's probably just going to break tests.
Wurst case scenario is that broken stuff gets deployed, but we should realize that soon enough.

PR #1035, opened 2016-09-23
Feature/auth update

kke:
Combined #931 and #940


jakolehm:
LGTM!

cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -0,0 +1,130 @@
...
+
+module Kontena::Cli::Cloud
+  class LoginCommand < Kontena::Command
+    include Kontena::Cli::Common
+
+    option ['-t', '--token'], '[TOKEN]', 'Use a pre-generated access token'
+    option ['-c', '--code'], '[CODE]', 'Use an authorization code'
+    option ['-v', '--verbose'], :flag, 'Increase output verbosity'
+
+    def execute
```
nevalla: I would like to see this method to split into smaller methods

cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -0,0 +1,130 @@
...
+      #puts
+      #puts "If you are running this command over an ssh connection or it's"
+      #puts "otherwise not possible to open a browser from this terminal"
+      #puts "then you must use a pregenerated access token using the --token"
+      #puts "option : kontena cloud login --token <access_token>"
+      puts
+      puts "Once the authentication is complete you can close the browser"
+      puts "window or tab and return to this window to continue."
+      puts
+      any_key_to_continue
```
nevalla: Should we add some timer after browser is opened automatically. Just in case user does not notice that some action is needed.

cli/lib/kontena/cli/cloud/master/show_command.rb
```diff
@@ -0,0 +1,23 @@
+require 'pp'
```
nevalla: do we need `pp` here?

cli/lib/kontena/cli/cloud/login_command.rb
kke: Yes I was also thinking about adding some kind of timeout for the press any key thing. Something like `Press any key to continue or press ctrl-c (automatically continuing in 8...7...6.. seconds)`

cli/lib/kontena/cli/cloud/master/show_command.rb
kke: no

kke:
TODO: 

- Some refactoring
- Increase coverage
- Unify the way how to require_api_url etc in all of the commands not changed in this PR
- Add optional timeout for any_key_to_continue method
- Add more `environment_variable:` definitions to options in all commands
..to be continued

cli/lib/kontena/cli/cloud/login_command.rb
kke: The master login and cloud login commands have a lot of commonalities that should be extracted into some  AuthCommon thing.

cli/lib/kontena/cli/common.rb
```diff
@@ -79,96 +185,132 @@ def ensure_custom_ssl_ca(url)
...
+        print pastel.bright_white("#{msg}")
+        char = STDIN.getch
+        print "\r#{' ' * msg.length}\r"
+        if char == "\u0003"
+          error "Canceled"
+        end
+      end
+
+      def display_login_info(only: nil)
+        unless only == :account
```
nevalla: these can be refactored into own methods for example `display_account_master_info` etc.

server/spec/spec_helper.rb
```diff
@@ -56,6 +58,7 @@ def app
   end
 
   config.before(:suite) do
+    Mongodb::Migrator.new.migrate
```
nevalla: Please revert this back, we don't want to run migrations for each test suite.

nevalla:
Found couple of minor issues. After removing migration run from spec helper we are ready to hit the green button

PR #1036, opened 2016-09-25
Do not clip service env output

### LINKED ISSUE #1014: kontena service show clips over 60 char env variables
e.g.

```
kontena service show localgyver-registry
localgyver/localgyver-registry:
--SNIP--
  env: 
    - KONTENA_LB_VIRTUAL_HOSTS=kontena-registry.localgyve...
```

Would be nice to have the envs printed out in full so that one don't need to issue yet another command to check the values. Or even a bit larger say 80 char limit would solve this in this particular case, but really I don't understand why must there be a limit at all..

jakolehm:
Fixes #1014


jnummelin:
LGTM

PR #1041, opened 2016-09-27
Replace dry-validation gem with hash_validator gem

nevalla:
This PR replaces `dry-validation` (and its dependencies) with `hash_validator` gem. With hash_validator gem all service configs are also validated that they do not contain any extra options than those that are supported in `kontena.yml`. Because of this strict validation `--skip-validation` option is added to `app config`, `app build` and `app deploy` commands. This could be take in place when extending docker-compose.yml that contains some non-supported options, but they don't break Kontena deploys.

Rest of the `kontena app` commands skip the validation, because they are using only service names from the YAML file.


jakolehm:
Needs rebase

PR #1042, opened 2016-09-27
Add instance number to container env

### LINKED ISSUE #1039: Populate an INSTANCE_NUM env var
Please populate an environment variable with the instance number of the service.  Some software, like Apache ZooKeeper and Apache Kafka, for example, want to know which instance they are to know who's who.

jnummelin:
As instance number is handy for some apps to know it should be populated in the container env.

fixes #1039

### DIFF
--- a/server/app/services/docker/service_creator.rb
+++ b/server/app/services/docker/service_creator.rb
@@ -54,7 +54,7 @@ module Docker
         log_opts: grid_service.log_opts,
         pid: grid_service.pid
       }
-      spec[:env] = build_env
+      spec[:env] = build_env(instance_number)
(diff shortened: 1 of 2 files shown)


jakolehm:
Just wondering the env variable name, maybe it should be `KONTENA_SERVICE_INSTANCE_NUMBER` ?

matti:
KONTENA_SERVICE_INSTANCE is enough?

jnummelin:
@jakolehm I actually had that but last second before push changed it. :) Wrong bet. I think your proposal is more explicit as the name.

jakolehm:
LGTM

Issue #1045, opened 2016-09-28
Possibility to give predefined grid token on grid creation


matti:
Needed for automated setups (like Terraform)

PR #1046, opened 2016-09-28
Allow to set a grid token manually

kke:
Implements #1045

```
$ kontena grid create --grid-token abcd1234 testing
```


jakolehm:
Looks good! But I would change the option name to just `--token`.

matti:
Yep, like the --force-deploy in deploy was changed to --force (main command was deploy, here grid)

kke:
Changed

jakolehm:
@kke needs rebase

PR #1055, opened 2016-10-03
Allow to install plugins in cli docker image

jakolehm:


jakolehm:
@kontena/core-platform ping

cli/Dockerfile
```diff
@@ -9,9 +9,12 @@ RUN apk update && \
   ca-certificates libssl1.0 openssl libstdc++ && \
   gem install kontena-cli --no-rdoc --no-ri -v ${CLI_VERSION} && \
   adduser kontena -D -h /home/kontena -s /bin/sh && \
+  echo "gem: --user-install" > /home/kontena/.gemrc && \
   chown -R kontena.kontena /home/kontena && \
   chmod +sx /usr/local/bin/docker
 
+ENV PATH=$PATH:/home/kontena/.gem/ruby/2.3.0/bin
```
nevalla: any chance that this path will change?

jakolehm: it's defined at line 11 .. if someone changes user home directory within this image it does not work anymore.. but still better than broken plugin install?

nevalla: I was thinking about the ruby version but now when thinking harder it won't change.

nevalla:
LGTM!

PR #1056, opened 2016-10-03
Add docs for needed open ports

### LINKED ISSUE #756: Add documentation on needed open ports
We should document what ports needs to be open and for which purpose.

jnummelin:
fixes #756


jnummelin:
@saana-kontenacm Could you check this out so that it's written clearly enough? :)

PR #1061, opened 2016-10-04
Link CHANGELOG version heading to the release

matti:
For faster access [and as a reminder] that the github releases page exists


jakolehm:
Ok, makes sense. Thanks!

PR #1078, opened 2016-10-05
Do not send Content-Type header with GET requests

nevalla:
Some API gateways might crash when sending Content-Type header with GET requests.


jakolehm:
LGTM!

PR #1082, opened 2016-10-05
Implement cloud logout command

nevalla:


jakolehm:
Should we have also `master logout`?

nevalla:
Yes, let's do it in separate PR?

jakolehm:
Yes, a separate PR is better. This one LGTM!

PR #1092, opened 2016-10-06
Unified the way current master / grid is required

kke:
Instead of:

``` ruby
def execute
  require_api_url
  token = require_token
  client(token).get(foofoo)
end
```

There now is:

``` ruby
class FooCommand < Kontena::Command
  requires_current_master

  def execute
    client.get(foofoo)
  end
end
```

Which will also add this to the help:

```
Usage:
    kontena service list [OPTIONS]

  Requires current grid: This command requires that you have selected a grid as the current grid using 'kontena grid use', by setting KONTENA_GRID environment variable or using the --grid option.

  Requires current master: This command requires that you have selected a current master using 'kontena master login' or 'kontena master use'. You can also use the environment variable KONTENA_URL to specify the master address or KONTENA_MASTER=master_name to override the current_master setting.
```

The help text may be a bit much, perhaps it should only show "Requires current master" and no further explanation.

Also modified GridOptions to use env KONTENA_GRID as an alternative way of passing '--grid'.

Also changed a bunch of `client(token).post...` lines to `client.post...` (client defaults to current master + current_master.token)

### DIFF
--- a/cli/lib/kontena/cli/apps/common.rb
+++ b/cli/lib/kontena/cli/apps/common.rb
@@ -86,7 +86,7 @@ module Kontena::Cli::Apps
     # @param [String] name
     # @return [Boolean]
     def service_exists?(name)
-      get_service(token, prefixed_name(name)) rescue false
+      get_service(prefixed_name(name)) rescue false
     end
 
     # @param [Hash] services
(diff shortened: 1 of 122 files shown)


SpComb:
:+1: on this refactoring, the current CLI is a mixture of explicit-token vs implicit-token client etc. See also #1747

SpComb:
Assigning for conflict fixes.

SpComb:
Please close this PR if it's too far out of date to update.

SpComb:
Too far out of date.

PR #1093, opened 2016-10-06
Speed up CLI launching by lazy-loading subcommands

kke:
Current implementation loads all commands when running any command.

This implementation loads only the command that is being run.

Spin up time improvement on an SSD disk is about 2x.

### DIFF
--- a/cli/lib/kontena/cli/master_command.rb
+++ b/cli/lib/kontena/cli/master_command.rb
@@ -1,39 +1,19 @@
-require_relative '../main_command'
-require_relative 'master/use_command'
-require_relative 'master/remove_command'
-require_relative 'master/list_command'
-require_relative 'master/users_command'
-require_relative 'master/current_command'
-require_relative 'master/config_command'
-require_relative 'master/login_command'
-require_relative 'master/logout_command'
-require_relative 'master/join_command'
-require_relative 'master/audit_log_command'
-require_relative 'master/token_command'
-require_relative 'master/init_cloud_command'
-require_relative 'master/ssh_command'
-
 class Kontena::Cli::MasterCommand < Kontena::Command
   include Kontena::Util
 
-  subcommand ["list", "ls"], "List masters where client has logged in", Kontena::Cli::Master::ListCommand
-  subcommand ["remove", "rm"], "Remove a master from configuration file", Kontena::Cli::Master::RemoveCommand
-  subcommand ["config", "cfg"], "Configure master settings", Kontena::Cli::Master::ConfigCommand
-  subcommand "use", "Switch to use selected master", Kontena::Cli::Master::UseCommand
-  subcommand "users", "Users specific commands", Kontena::Cli::Master::UsersCommand
-  subcommand "current", "Show current master details", Kontena::Cli::Master::CurrentCommand
-  subcommand "login", "Authenticate to Kontena Master", Kontena::Cli::Master::LoginCommand
-  subcommand "logout", "Log out of Kontena Master", Kontena::Cli::Master::LogoutCommand
-  subcommand "token", "Manage Kontena Master access tokens", Kontena::Cli::Master::TokenCommand
-  subcommand "join", "Join Kontena Master using an invitation code", Kontena::Cli::Master::JoinCommand
-  subcommand "audit-log", "Show master audit logs", Kontena::Cli::Master::AuditLogCommand
-  subcommand "init-cloud", "Configure current master to use Kontena Cloud services", Kontena::Cli::Master::InitCloudCommand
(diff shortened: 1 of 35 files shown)


SpComb:
I like the idea. Please fix the conflicts so we can test.

SpComb:
Testing with `/usr/bin/time -f '%U' kontena version --cli`, this seems to cut CPU user time to about 50% compared to 1.1.2:

```
old     new
0.66	0.37
0.60	0.38
0.60	0.29
0.62	0.33
0.63	0.40
0.59	0.37
```

SpComb:
Comparing wall-clock time, the difference varies a lot more, presumably due to all the IO effects... but the lazy loader is still significantly faster

```
before  after
0.88	0.54
0.84	0.49
0.80	0.52
0.67	0.49
0.69	0.44
0.69	0.53
0.62	0.47
0.70	0.51
0.58	0.60
0.76	0.60
```

cli/lib/kontena/cli/service_command.rb
```diff
@@ -1,49 +1,27 @@
...
+  subcommand "create", "Create a new service", load_subcommand('services/create_command')
+  subcommand "show", "Show service details", load_subcommand('services/show_command')
+  subcommand "update", "Update service configuration", load_subcommand('services/update_command')
+  subcommand "deploy", "Deploy service", load_subcommand('services/deploy_command')
+  subcommand "stop", "Stop service", load_subcommand('services/stop_command')
+  subcommand "start", "Start service", load_subcommand('services/start_command')
+  subcommand "restart", "Restart service", load_subcommand('services/restart_command')
+  subcommand "scale", "Scale service", load_subcommand('services/scale_command')
+  subcommand ["remove", "rm"], "Remove service", load_subcommand('services/remove_command')
+  subcommand "delete", "[DEPRECATED] Delete service", load_subcommand('services/delete_command')
```
SpComb: Did this rise from the dead? #1759

cli/lib/kontena/cli/master_command.rb
```diff
@@ -1,39 +1,19 @@
...
-require_relative 'master/audit_log_command'
-require_relative 'master/token_command'
-require_relative 'master/init_cloud_command'
-require_relative 'master/ssh_command'
-
 class Kontena::Cli::MasterCommand < Kontena::Command
   include Kontena::Util
 
-  subcommand ["list", "ls"], "List masters where client has logged in", Kontena::Cli::Master::ListCommand
-  subcommand ["remove", "rm"], "Remove a master from configuration file", Kontena::Cli::Master::RemoveCommand
```
SpComb: Did we lose `kontena master rm`?

cli/lib/kontena/main_command.rb
```diff
@@ -37,24 +18,24 @@ class Kontena::MainCommand < Kontena::Command
...
+  subcommand "node", "Node specific commands", load_subcommand('node_command')
+  subcommand "master", "Master specific commands", load_subcommand('master_command')
+  subcommand "vpn", "VPN specific commands", load_subcommand('vpn_command')
+  subcommand "registry", "Registry specific commands", load_subcommand('registry_command')
+  subcommand "container", "Container specific commands", load_subcommand('container_command')
+  subcommand "etcd", "Etcd specific commands", load_subcommand('etcd_command')
+  subcommand "external-registry", "External registry specific commands", load_subcommand('external_registry_command')
+  subcommand "whoami", "Shows current logged in user", load_subcommand('whoami_command')
+  subcommand "plugin", "Plugin related commands", load_subcommand('plugin_command')
+  subcommand "version", "Show version #{Kontena::Cli::VERSION}", load_subcommand('version_command')
```
SpComb: Subtle :P

cli/lib/kontena/cli/subcommand_loader.rb
```diff
@@ -0,0 +1,81 @@
...
+        raise ArgumentError, "Can not load #{path} or #{Kontena.cli_root(path)}"
+      end
+      @subcommand_class = get_class
+    end
+
+    def new(*args)
+      klass.new(*args)
+    end
+
+    def method_missing(meth, *args)
```
SpComb: Should this use some `Delegator` support or such?

kke: Delegator does the same thing except in addition it hides the delegator class from backtraces when exceptions arise.

cli/lib/kontena/cli/master_command.rb
kke: Yes, brought back.

kke:
A lot of the require_relative things are fishy anyway.

The lib is in load path so something like: `require 'kontena/cli/foofoo'` should always work without some crazy `require_relative '../../lib/kontena/cli/foofoo'` magic.

PR #1094, opened 2016-10-06
Require authenticated user on grid list command

nevalla:
With this PR a proper error message is shown if user is not logged in and he/she is trying to list grids.

Before:
```
$ kontena grid list
 [error] Expected([200, 201]) <=> Actual(400 Bad Request)
```

After:
```
$ kontena grid list
 [error] You are not logged into a Kontena Master. Use: kontena master login
```


jakolehm:
LGTM

PR #1105, opened 2016-10-07
Remove CLI version limitation, move error message to login endpoint

kke:


server/app/routes/v1/auth_api.rb
```diff
@@ -4,7 +4,7 @@ class AuthApi < Roda
 
     route do |r|
       r.post do
-        halt_request(403, 'This version of Kontena Master does not support user credential authentication') and return
+        halt_request(403, 'This version of Kontena Master does not support user credential authentication. Use: "gem install kontena-cli" to upgrade.')
```
jakolehm: Let's remove this "gem install" message from the server.. user might have alternate installation method.

PR #1107, opened 2016-10-07
Add initial admin code for ubuntu packages

### LINKED ISSUE #1099: Ubuntu package is missing new oauth2 config
Ubuntu server(master) package is missing new oauth2 related envs (initial admin token). Maybe we should prompt those (or non-interactive, generate them if possible).

jnummelin:
fixes #1099


jnummelin:
ping @jakolehm

server/packaging/ubuntu/kontena-server/DEBIAN/templates
```diff
@@ -0,0 +1,4 @@
+Template: kontena-server/initial_admin_code
+Type: string
+Default:
+Description: Initial admin code for Kontena Master
```
jakolehm: Should we say that it can be anything (random string)?

jnummelin: 👍  I'll add that

jakolehm: And we need to update Ubuntu doc page ...

jnummelin: updated.

PR #1112, opened 2016-10-09
Omnibus

jakolehm:
Omnibus based automation provides native cli packages for mac (linux & windows might come later).


cli/omnibus/package-scripts/kontena/postinstall
```diff
@@ -0,0 +1,12 @@
...
+#
+# Perform necessary kontena setup steps
+# after package is installed.
+#
+chown $USER /opt/kontena/embedded/bin
+chown $USER -R /opt/kontena/embedded/lib/ruby/gems/2.1.0
+
+mkdir -p /usr/local/bin
+ln -s /opt/kontena/bin/kontena /usr/local/bin/kontena
+
```
kke: We could detect rbenv and remove all `bin/kontena`'s from `/usr/local/opt/rbenv/versions/` but then again maybe people with rbenv don't need this installer.

The brew recipe could do that or give a post install message if `which kontena` points to the wrong place.

PR #1113, opened 2016-10-10
Ignore vendor files when building docker images

jakolehm:


SpComb:
Can you add `.bundle` as well while you're at it?

```
$  bundle 
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/..


Your user account isn't allowed to install to the system Rubygems.
  You can cancel this installation and run:

      bundle install --path vendor/bundle

  to install the gems into ./vendor/bundle/, or you can enter your password
  and install the bundled gems to Rubygems using sudo.

  Password: 
$ bundle --path vendor/bundle
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/..
Installing ...
Bundle complete! 8 Gemfile dependencies, 31 gems now installed.
Bundled gems are installed into ./vendor/bundle.

$ docker build -t ... . && docker run ...
Sending build context to Docker daemon 23.08 MB
...
env: can't execute 'ruby2.3': No such file or directory
```

jakolehm:
@SpComb why `.bundle` ?

jakolehm:
Ok, I think I got it... will add `.bundle` also.

PR #1115, opened 2016-10-10
Add health status actions on agent and master

### LINKED ISSUE #911: Use new health status in scheduling services
The new health status should be used in service scheduling. I.e. un-healthy services should be re-scheduled automatically.

jnummelin:
This PR adds health status actions. Agent restart unhealthy containers and master will schedule full re-deploy if overall health goes under threshold.

fixes #911

### DIFF
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -47,6 +47,24 @@ module Kontena::Workers
         msg = check_tcp_status(ip, port, timeout)
       end
       @queue << msg
+
+      if msg.dig(:data, 'status') == 'unhealthy'
+        name = @container.labels['io.kontena.container.name']
+        # Restart the container, master will handle re-scheduling logic
+        info "About to restart container #{name} as it's reported to be unhealthy"
+        log = {
+            event: 'container:log',
+            data: {
+                id: @container.id,
+                time: Time.now.utc.xmlschema,
+                type: 'stderr',
+                data: "*** [Kontena/Agent] Restarting service as it's reported to be unhealthy."
+            }
+        }
+        @queue << log
+        Kontena::ServicePods::Restarter.perform_async(name)
+      end
+
     end
 
     def check_http_status(ip, port, uri, timeout)
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -48,6 +48,11 @@ module Kontena::Workers
       end
       @queue << msg
 
+      handle_action(msg)
+
+    end
+
+    def handle_action(msg)
       if msg.dig(:data, 'status') == 'unhealthy'
         name = @container.labels['io.kontena.container.name']
         # Restart the container, master will handle re-scheduling logic
@@ -64,7 +69,6 @@ module Kontena::Workers
         @queue << log
         Kontena::ServicePods::Restarter.perform_async(name)
       end
-
     end
 
     def check_http_status(ip, port, uri, timeout)
--- a/server/app/jobs/grid_service_health_monitor_job.rb
+++ b/server/app/jobs/grid_service_health_monitor_job.rb
@@ -1,0 +1,37 @@
+class GridServiceHealthMonitorJob
+  include Celluloid
+  include Logging
+
+  PUBSUB_KEY = 'service:health_status_events'
+
+  def initialize
+    async.subscribe_health_events
+  end
+
+  def subscribe_health_events
+    MongoPubsub.subscribe(PUBSUB_KEY) do |message|
+      self.handle_event(message)
+    end
+  end
+
+  def handle_event(event)
+    service = GridService.find_by(id: event['id'])
+    if deploy_needed?(service)
+      info "service health too low, triggering full deploy for #{service.to_path}"
+      GridServiceDeploy.create(grid_service: service)
+    end
+  end
+
+  def deploy_needed?(service)
+    health_status = service.health_status
+    health_percent = health_status[:healthy].to_f / health_status[:total].to_f
+    min_health = service.deploy_opts.min_health || 0.8
+    expected_health = 1 - min_health
+    if health_percent < expected_health
+      service.running? && !service.deploy_pending?
+    else
+      false
+    end
+  end
+
+end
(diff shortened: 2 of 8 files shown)


agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -47,6 +45,24 @@ def check_status
         msg = check_tcp_status(ip, port, timeout)
       end
       @queue << msg
+
+      if msg.dig(:data, 'status') == 'unhealthy'
+        name = @container.labels['io.kontena.container.name']
```
jakolehm: Imho this code block deserves own method?

server/app/jobs/grid_service_health_monitor_job.rb
```diff
@@ -0,0 +1,37 @@
+class GridServiceHealthMonitorJob
+  include Celluloid
+  include Logging
+
+  PUBSUB_KEY = 'service:health_status_events'
+
+  def initialize
+    async.subscribe_health_events
```
jakolehm: How do we handle multi-master scenarios?

agent/lib/kontena/workers/container_health_check_worker.rb
jnummelin: Will do.

server/app/jobs/grid_service_health_monitor_job.rb
jnummelin: I think we should only make actions on the leader, will change in few minutes

colinmollenhour:
Is it possible to enable health checks but not do re-deploys? E.g. in a Galera cluster it could take a while for a node to sync initially and you wouldn't want it to re-deploy during this time but it might still be useful to have health checking for monitoring purposes..

Also, what about supporting the Docker 1.12 "HEALTHCHECK" command in the Dockerfile? I like how their implementation uses an arbitrary command instead of an explicit tcp/http check and it also uses the command output to give details about the health. For example with Galera cluster the command could report not-healthy with detail stating "Syncing with master" which is much more informative than a binary status alone.

jakolehm:
@colinmollenhour could you use `initial_delay` option to delay health checks for Galera? (might not be a good idea if initial sync is dependent of the database size)

I think we should add support for Dockerfile `HEALTHCHECK` , but that is a topic for an another PR :)

jnummelin:
@colinmollenhour I added separate issue to track the Docker `HEALTHCHECK` support, see #1121 

I think it would make sense for Galera type of use cases to have a flag to disable re-deploys based on health status. Something like:
```
version: '2'
services:
  mysql:
    image: mysql
    stateful: true
    deploy:
      strategy: ha
      wait_for_port: 3306
    environment:
      - MYSQL_ALLOW_EMPTY_PASSWORD=true
    health_check:
      protocol: tcp
      port: 3306
      interval: 10
      initial_delay: 10
      timeout: 2
      deploy_on_failure: false
```

@jakolehm What do you think, should we add that kind of flag with default `true`?

colinmollenhour:
Yeah the initial delay for a Galera cluster can vary widely and is hard to predict. I suppose just specifying a large `initial_delay` like one day would be a valid workaround, though.

Does the health_check influence the service discovery DNS records directly?

jnummelin:
@colinmollenhour The DNS is not affected by service health.

jakolehm:
@jnummelin flag makes sense.

colinmollenhour:
I've been spending a *lot* of time working on Galera on Kontena. I've got the startup script so that it recovers the cluster in most cases but one issue is that to do an update there doesn't seem to be an easy way to take down just one node at a time.. Perhaps this is a good case study to consider for the planning of this feature.

Right now I am doing the following:

```
deploy:
  strategy: daemon
  min_health: 0.34
instances: 1
affinity:
  - label==db-node
```

In this way the service deploys only one instance per server and only on servers designated 'db-node'. This works pretty well, but if I want to update my container image I should be able to do that without any downtime, but I'm not sure it is possible given that when a node is shutdown it takes some time to become SYNCED again.

If I update the image and run `kontena app deploy galera-node` then what I would like to happen is:

* At some point the updated images are pulled to each server, ideally before nodes are shutdown to minimize downtime of each node (maybe it already does this, I dunno).
* Based on some kontena.yml option the deploy will happen only on one node at a time.
* The node being updated will be gracefully shutdown (btw, it would be great if the timeout was configurable as 10 seconds may not be enough on a busy server).
* The container re-launched for this single node and before Kontena goes to re-deploy the next container either some configurable timeout or a very specific health check is waited on.

For the health check I want to make sure the node is not only up but "synced" since with a 3-node cluster to avoid downtime you really need two nodes up because when the third re-joins it needs a donor node which will be de-synced for a period of time (very long period if Galera decides a full state snapshot transfer is needed on a large database which it might with an unclean shutdown).

However, when deploying three nodes from scratch Galera really needs *all* nodes to boot at once so that they can communicate and figure out who has the most advanced state so a slow startup is actually not desirable when starting completely stopped services.

I don't fully understand "min_health" but I don't see that it is sufficient to cover a wide range of needs. I was using Docker Swarm briefly before switching over to Kontena and I have to say I did like their options for rolling upgrades:

```
      --restart-condition string       Restart when condition is met (none, on-failure, or any)
      --restart-delay value            Delay between restart attempts (default none)
      --restart-max-attempts value     Maximum number of restarts before giving up (default none)
      --restart-window value           Window used to evaluate the restart policy (default none)
      --stop-grace-period value        Time to wait before force killing a container (default none)
      --update-delay duration          Delay between updates
      --update-failure-action string   Action on update failure (pause|continue) (default "pause")
      --update-parallelism uint        Maximum number of tasks updated simultaneously (0 to update all at once) (default 1)
```

So for my scenario I'd use something like:

* --update-delay 15m         // No rush!
* --update-failure-action pause   // If something is wrong it is best to keep remaining nodes up
* --update-parallelism 1      // Very important to avoid multiple nodes down
* --stop-grace-period 120    // Better to take a while than have unclean shutdown

nevalla:
@jnummelin This needs rebase

jnummelin:
@colinmollenhour Thanks for writing up your use case, it really helps us to see how users would use these kind of features.

I think your use case would require executing some health check script within the container to figure out the proper state of Galera cluster. Then that info could be used in the deployment process when deciding if it is OK to deploy the next instance. This should cover your use case pretty well, right?

For the script based health check we do already have an open issue: #1121 

Not sure if we should address the health check status during the deployment process in this PR or create another for it. There are improvements coming to the deployment tracking so might be easier to have that before using health status during deployment. Without that users might see some unxpected "deployment stuck" features when the deployment is actually waiting for health statuses. @jakolehm @nevalla Any thoughts?

jakolehm:
@jnummelin I think enhancements to the deployment tracking should be done in a separate PR.

PR #1118, opened 2016-10-11
Fixes  to handling colorization during tests

kke:


jakolehm:
Let's not expose enable/disable color api... I took best parts from here and integrated to #1117

PR #1119, opened 2016-10-11
Add retriable for retrying hooks during deploy

kke:


jakolehm:
Why we need to retry hooks?

kke:
Robustness.

During testing I encountered a situation where the first step of auth provider config passed and the second one gave some weird timeout error and so the rest of the steps were not run since one of them raised.

Of course it would be possible to just retry X times using plain rescue.

jakolehm:
Ok, just thought if there are any side-effects. It seems that retrying these hooks is safe.

jakolehm:
Rebase, pls.

jakolehm:
LGTM

PR #1120, opened 2016-10-11
Use OAuth bearer token when updating master information to cloud

nevalla:
With this PR Kontena Cloud's OAuth2 access token is used (instead of BasicAuth) when updating Kontena Master's redirect url to Kontena Cloud.

### DIFF
--- a/server/app/services/auth_provider.rb
+++ b/server/app/services/auth_provider.rb
@@ -3,7 +3,7 @@
 # Userinfo parsing in done through jsonpath :
 #   http://goessner.net/articles/JsonPath/
 #
-# You can define multiple optional jsonpaths and separate them with a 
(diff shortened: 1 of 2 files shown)


server/app/services/auth_provider.rb
```diff
@@ -89,18 +88,46 @@ def update_kontena
...
         'Content-Type' => 'application/json',
-        'Accept' => 'application/json'
+        'Accept' => 'application/json',
+        'Authorization' => "Bearer #{master_access_token}"
       },
       body: body.to_json
     )
   end
 
+  def master_access_token
```
kke: this should probably be in config, something like @master_access_token ||= config['cloud.master_token_secret'] ||= get_master_access_token 

(anything ending with _secret is stored encrypted in the db)

PR #1131, opened 2016-10-13
Interactive server deletion from config or cloud

kke:
Introduces:

```
$ kontena master rm [master_name]
```

If you don't supply a master name, you'll get a multi select menu where you can select multiple masters you want to remove from config.

Same goes with 
```
$ kontena cloud master rm
```

But Cloud API does not support DELETE yet so the command currently reports "Not found"


jakolehm:
LGTM

PR #1138, opened 2016-10-14
Upgrade CLI hint when doing legacy login from cli

kke:


server/app/routes/v1/auth_api.rb
```diff
@@ -4,7 +4,11 @@ class AuthApi < Roda
 
     route do |r|
       r.post do
-        halt_request(403, 'This version of Kontena Master does not support user credential authentication.')
+        if r.user_agent.to_s.start_with?('kontena-cli')
+          halt_request(403, 'This version of Kontena Master does not support user credential authentication. You need to upgrade your Kontena CLI, use: gem install kontena-cli')
```
jakolehm: Could we remove the method for upgrading the cli? There are too many ways to upgrade the cli ... might confuse? /cc @jnummelin @nevalla

jnummelin:
I agree, remove the method.

kke:
Removed the gem install hint.

PR #1139, opened 2016-10-14
Implement 'kontena master init-cloud' command

kke:
```
$ kontena master init-cloud
Proceeding will:
 * Register the Kontena Master kontena-master-5 to Kontena Cloud
 * Configure the Kontena Master to use Kontena Cloud as the
   authentication provider

After this:
 * Users will not be able to reauthenticate without authorizing the
   Master to access their Kontena Cloud user information
 * Users that have registered a different email address to Kontena
   Cloud than the one they currently have as their username in the
   master will not be able to authenticate before an administrator
   of the Kontena Master creates an invitation code for them
   (kontena master users invite old@email.example.com)
> Proceed? Yes
 [done] Registering current Kontena Master 'kontena-master-5' to Kontena Cloud
 [done] Loading Kontena Cloud auth provider base configuration to Kontena Master
 [done] Updating OAuth2 client-id and client-secret to Kontena Master
```

Also the URL's to production cloud have been updated.

### DIFF
--- a/cli/lib/kontena/callbacks/master/deploy/05_before_deploy_configuration_wizard.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/05_before_deploy_configuration_wizard.rb
@@ -45,55 +45,18 @@ module Kontena
         end
       end
 
-      def cloud_masters
-        masters = []
-        begin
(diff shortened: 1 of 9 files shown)


jakolehm:
Docs need update

kke:
Done

PR #1141, opened 2016-10-14
Moved the "auth provider not configured" error to right endpoint

kke:
And removed the old user invite endpoint.


jakolehm:
We should probably have an api test for this?

jakolehm:
Ok, there actually is an api test for this.. missed it somehow.

PR #1147, opened 2016-10-16
Restart weave if trusted-subnets change

jakolehm:


agent/spec/lib/kontena/network_adapters/weave_spec.rb
```diff
@@ -47,6 +47,56 @@
...
+      config = {
+        'grid' => {
+          'trusted_subnets' => []
+        }
+      }
+      weave_config = {
+        'Image' => valid_image,
+        'Cmd' => ['--trusted-subnets', '']
+      }
+      weave = spy(:weave, config: weave_config)
```
SpComb: Why `spy` instead of just `double`? Using `double` means that you get an rspec error if you forgot to mock something out, whereas `spy` just silently returns nil for everything..

Asking because I'm trying to fix `spy` -> `double` elsewhere :)

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -233,6 +233,16 @@ def post_start(info)
       Celluloid::Notifications.publish('network_adapter:start', info)
     end
 
+    # @param [Docker::Container] weave
+    # @param [Hash] config
+    def config_changed?(weave, config)
+      return true if weave.config['Image'].split(':')[1] != WEAVE_VERSION
+      cmd = Hash[*weave.config['Cmd'].flatten(1)]
```
SpComb: This is very fragile, but it works. Just don't ever add any extra flag options to the command line :D

SpComb: Maybe add a separate env/label for the `trusted_subnets` instead, and compare that?

jakolehm: Please tell if you know a better way :)

SpComb: Looks like `weaver --trusted-subnets=192.168.66.0/24` also works, that might be easier to parse from the cmd?

agent/spec/lib/kontena/network_adapters/weave_spec.rb
jakolehm: Double might be a slightly better (overall).. in this case it does not really matter (imho).

agent/lib/kontena/network_adapters/weave.rb
SpComb: A bit of reverse-engineering also gives this: `$ curl -s http://127.0.0.1:6784/report -H 'Accept: application/json' | vgrep TrustedSubnets`

```
{
    "Version": "1.7.2",
    ...
    "Router": {
        ...
        "TrustedSubnets": [
            "192.168.66.0/24"
        ],
```

SpComb:
Seems to work just fine, apart from the hairy command-line parsing :P

```
        Version: 1.7.2 (up to date; next check at 2016/10/17 17:50:50)

        Service: router
       Protocol: weave 1..2
           Name: c6:3f:e0:48:25:8d(core-01)
     Encryption: enabled
  PeerDiscovery: enabled
        Targets: 1
    Connections: 1 (1 established)
          Peers: 2 (with 2 established connections)
 TrustedSubnets: none

        Service: dns
         Domain: kontena.local.
       Upstream: 10.0.2.3
            TTL: 1
        Entries: 7
```

`kontena grid trusted-subnet add development 192.168.66.0/24`

```
        Version: 1.7.2 (up to date; next check at 2016/10/17 16:06:03)

        Service: router
       Protocol: weave 1..2
           Name: c6:3f:e0:48:25:8d(core-01)
     Encryption: enabled
  PeerDiscovery: enabled
        Targets: 1
    Connections: 1 (1 established)
          Peers: 2 (with 2 established connections)
 TrustedSubnets: 192.168.66.0/24

        Service: dns
         Domain: kontena.local.
       Upstream: 10.0.2.3
            TTL: 1
        Entries: 7
```

agent/spec/lib/kontena/network_adapters/weave_spec.rb
jakolehm: Changed.

Issue #1143, opened 2016-10-15
Fix Ubuntu master installation / auth docs


jakolehm:
Docs in the master branch are missing few critical pieces:

- configuring initial token
- login with initial token
- how to configure external authentication

jnummelin:
The initial admin code setup is now documented in the packaging update PR.

External auth is IMO pretty well already documented in master, see 
https://github.com/kontena/kontena/blob/master/docs/using-kontena/authentication.md

Issue #730, opened 2016-05-26
Ubuntu docker.io package compatibility


jalberto:
Ubuntu renamed docker-engine to docker.io in 16.04

d2s:
This would be important for our future testing environment,
because it would enable usage of more recent Ubuntu version.

jakolehm:
Ubuntu 16.04 packages are now merged with #1150 (will be part of 0.16 release). They still depend on "official" docker packages so docker.io package is not supported.

SpComb:
#1173 added Ubuntu xenial `docker.io=1.12*` support.

jnummelin:
Close via #1150 & #1173

PR #1150, opened 2016-10-17
Add packaging for Ubuntu Xenial (16.04)

### LINKED ISSUE #738: Ubuntu 16.04 deb packages
We should have packages for Ubuntu 16.04

Related: #730

jnummelin:
This PR adds packaging for Ubuntu 16.04 (Xenial) packages. Xenial is based on systemd so the packages add proper systemd units for all components plus the config file.

I wasn't able yet to test the upload to bintray as I do not posses proper creadentials for that.

Fixes #738, #730, #1143


agent/lib/tasks/release.rake
```diff
@@ -63,6 +77,9 @@ namespace :release do
     raise ArgumentError.new('You must define REV') if rev.blank?
     sh('rm -rf release && mkdir release')
     sh('cp build/ubuntu/*.deb release/')
+    sh('rm -rf release_xebial && mkdir release_xenial')
```
SpComb: xebial

jnummelin: yeah, I just renamed the ubuntu release :D

agent/lib/tasks/release.rake
```diff
@@ -40,6 +40,20 @@ namespace :release do
...
   end
 
+  desc 'Build ubuntu Xenial packages'
+  task :build_ubuntu_xenial => :environment do
+    rev = ENV['REV'] || '1'
+    sh('mkdir -p build')
+    sh('rm -rf build/ubuntu_xenial/')
+    sh('cp -ar packaging/ubuntu_xenial build/')
+    sh("sed -i \"s/VERSION/#{VERSION}-#{rev}/g\" build/ubuntu_xenial/#{NAME}/DEBIAN/control")
+    sh("sed -i \"s/VERSION/#{VERSION}/g\" build/ubuntu_xenial/#{NAME}/DEBIAN/postinst")
```
SpComb: The `postinst` doesn't use VERSION? should this be for the `/etc/kontena-agent.env `instead?

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
```diff
@@ -0,0 +1,11 @@
...
+#KONTENA_URI=wss://localhost:8443
+
+# Set kontena token
+#KONTENA_TOKEN=
+
+# kontena peer interface (overlay network interface)
+KONTENA_PEER_INTERFACE=eth1
+
+# kontena version
+KONTENA_VERSION=latest
```
SpComb: This should match the package version?

agent/packaging/ubuntu_xenial/kontena-agent/etc/systemd/system/kontena-agent.service
```diff
@@ -0,0 +1,24 @@
...
+[Service]
+Restart=always
+RestartSec=5
+EnvironmentFile=/etc/kontena-agent.env
+ExecStartPre=-/usr/bin/docker stop kontena-agent
+ExecStartPre=-/usr/bin/docker rm kontena-agent
+ExecStartPre=/usr/bin/docker pull kontena/agent:${KONTENA_VERSION}
+ExecStart=/usr/bin/docker run --name kontena-agent \
+    -e KONTENA_URI=${KONTENA_URI} \
+    -e KONTENA_TOKEN=${KONTENA_TOKEN} \
```
SpComb: Could we use `docker run --env-file=/etc/kontena-agent.env` to avoid leaking the secret tokens on the command line?

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
```diff
@@ -0,0 +1,11 @@
+# Set URL of your kontena server
+#KONTENA_URI=wss://localhost:8443
+
+# Set kontena token
+#KONTENA_TOKEN=
```
SpComb: Hmm.. should this file be `chmod 0600` as it contains secrets?

server/packaging/ubuntu_xenial/kontena-server/DEBIAN/conffiles
```diff
@@ -0,0 +1,5 @@
+/etc/kontena-server.env
+/opt/bin/kontena-haproxy.sh
+/etc/systemd/system/kontena-server.service
```
SpComb: Hmm.. package units should be under `/lib/systemd/system`, but eh :)

server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
```diff
@@ -0,0 +1,20 @@
...
+VAULT_IV=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 64 | head -n 1)
+sed -i -r "s#\#VAULT_KEY=.*#VAULT_KEY=${VAULT_KEY}#" /etc/kontena-server.env
+sed -i -r "s#\#VAULT_IV=.*#VAULT_IV=${VAULT_IV}#" /etc/kontena-server.env
+
+# Fetching configuration from debconf
+db_get kontena-server/initial_admin_code
+sed -i -r "s#\#INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
+
+systemctl daemon-reload
+systemctl start kontena-server-mongo
```
SpComb: You probably also want to `systemctl install` to enable them on boot?

agent/packaging/ubuntu_xenial/kontena-agent/etc/systemd/system/kontena-agent.service
```diff
@@ -0,0 +1,24 @@
...
+    -e KONTENA_PEER_INTERFACE=${KONTENA_PEER_INTERFACE} \
+    -v=/var/run/docker.sock:/var/run/docker.sock \
+    -v=/etc/kontena-agent.env:/etc/kontena.env \
+    --net=host \
+    kontena/agent:${KONTENA_VERSION}
```
SpComb: Missing the `[Install]` section?

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
jnummelin: That's a good question. I was thinking that using latest would enable smooth upgrade with just `systemctl restart kontena-agent` as it works on CoreOS that are provisioned with our cli plugins. @jakolehm What do you think, should we pin the version to the package version?

agent/packaging/ubuntu_xenial/kontena-agent/etc/systemd/system/kontena-agent.service
jnummelin: We haven't been using the `[Install]` section with our cli plugins for CoreOS. What should there be in?

server/packaging/ubuntu_xenial/kontena-server/DEBIAN/conffiles
jnummelin: doh, of course. I'll change these.

agent/packaging/ubuntu_xenial/kontena-agent/etc/systemd/system/kontena-agent.service
SpComb: Hmm.. you need `[Install] WantedBy=multi-user.target` for `systemctl enable` to work, and start the systemd service at boot.

Not sure what's going on in the plugins then, AFAIK `systemctl enable` without any `[Install]` is a no-op?

```
$ sudo systemctl cat kontena-agent.service | grep Install
$ sudo systemctl is-enabled kontena-agent.service
static
$ sudo systemctl enable kontena-agent.service
$ sudo systemctl is-enabled kontena-agent.service
static
```

> Static units are those which cannot be enabled/disabled, but it doesn't mean they are always executed. They will only if another unit depends on them, or if they are manually started.

SpComb: I suspect CoreOS may invoke the cloudinit unit `command: start` at each boot. Systemd itself won't start any units at boot unless they are enabled.

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
jakolehm: We should pin the version because otherwise these packages cannot be used properly with configuration management systems (chef/ansible etc).

server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
jnummelin: yes

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
jnummelin: roger, I'll add version pinning

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
jnummelin: I think it is already, at least in the latest commit
```
$ stat -f "%A" agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env 
600
```

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/kontena-agent.service
```diff
@@ -0,0 +1,25 @@
...
+EnvironmentFile=/etc/kontena-agent.env
+ExecStartPre=-/usr/bin/docker stop kontena-agent
+ExecStartPre=-/usr/bin/docker rm kontena-agent
+ExecStartPre=/usr/bin/docker pull kontena/agent:${KONTENA_VERSION}
+ExecStart=/usr/bin/docker run --name kontena-agent \
+    --env-file /etc/kontena-agent.env \
+    -v=/var/run/docker.sock:/var/run/docker.sock \
+    -v=/etc/kontena-agent.env:/etc/kontena.env \
+    --net=host \
+    kontena/agent:${KONTENA_VERSION}
```
jakolehm: I think we should add also `ExecStop`?

agent/lib/tasks/release.rake
```diff
@@ -11,7 +11,7 @@ namespace :release do
   end
 
   desc 'Build all'
-  task :build => [:build_docker, :build_ubuntu] do
+  task :build => [:build_docker, :build_ubuntu, :build_ubuntu_xenial] do
```
jakolehm: I think it would make sense to have `build_ubuntu` task that runs `build_ubuntu_trusty` & `build_ubuntu_xenial` .. now it sounds a little bit strange?

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/kontena-agent.service
jnummelin: Yes, I'll add that

agent/lib/tasks/release.rake
dacod: Hi guys, sorry for asking, any forecast for release this fix or should I go back to the ubuntu trusty?
If you need some tests, i'm available ok

jnummelin:
@dacod current plan is to get these into 0.16 release in a week or so. testing help would be highly appreciated.

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/docker.service.d/50-kontena.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+Environment='DOCKER_OPTS=--insecure-registry="10.81.0.0/19"'
```
SpComb: Seems like the systemd service unit in the `docker-engine=1.11*` package doesn't know about `DOCKER_OPTS`:

`systemctl cat docker.service`
```
# /lib/systemd/system/docker.service
...
[Service]
...
ExecStart=/usr/bin/docker daemon -H fd://
```


`apt policy docker-engine`
```
docker-engine:
  Installed: 1.11.2-0~xenial
  Candidate: 1.12.2-0~xenial
```

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/kontena-agent.service
```diff
@@ -0,0 +1,26 @@
...
+Requires=network-online.target
+Requires=docker.service
+
+[Service]
+Restart=always
+RestartSec=5
+EnvironmentFile=/etc/kontena-agent.env
+ExecStartPre=-/usr/bin/docker stop kontena-agent
+ExecStartPre=-/usr/bin/docker rm kontena-agent
+ExecStartPre=/usr/bin/docker pull kontena/agent:${KONTENA_VERSION}
```
SpComb: This could maybe be prefixed with - to continue trying to run the service, even if the Docker hub is down, or using a local image version...

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/docker.service.d/50-kontena.conf
jnummelin: annoying :/ the only way is then to override the ExecStart. any idea if there's some easy way to grab the current ExecStart and dynamically add the `DOCKER_OPTS` into it? Not sure if we would want to just do plain override.

server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
```diff
@@ -0,0 +1,25 @@
...
+sed -i -r "s#\#VAULT_IV=.*#VAULT_IV=${VAULT_IV}#" /etc/kontena-server.env
+
+# Fetching configuration from debconf
+db_get kontena-server/initial_admin_code
+sed -i -r "s#\#INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
+
+systemctl daemon-reload
+
+systemctl enable kontena-server-mongo
+systemctl enable kontena-server
```
jnummelin: good catch

agent/packaging/ubuntu_xenial/kontena-agent/lib/systemd/system/docker.service.d/50-kontena.conf
jnummelin: we discussed this with @jakolehm and decided that we'll create drop-in unit to override the `ExecStart` with proper opts

SpComb:
Can we get a CI build of this out ASAP for testing?

PR #1153, opened 2016-10-18
Show container health status in service show

### LINKED ISSUE #1152: Show instance health check details in `service show`
Instance health status information should be visible in `service show`output

jnummelin:
fixes #1152

### DIFF
--- a/cli/lib/kontena/cli/services/services_helper.rb
+++ b/cli/lib/kontena/cli/services/services_helper.rb
@@ -192,6 +192,10 @@ module Kontena
             puts "      dns: #{container['name']}.#{grid}.kontena.local"
             puts "      ip: #{container['overlay_cidr'].to_s.split('/')[0]}"
             puts "      public ip: #{container['node']['public_ip'] rescue 'unknown'}"
+            if container['health_status']
+              puts "      health: #{container['health_status']}"
+              puts "      health updated: #{container['health_status_at']}"
+            end
             if container['status'] == 'unknown'
               puts "      status: #{container['status'].colorize(:yellow)}"
             else
--- a/cli/lib/kontena/cli/services/services_helper.rb
+++ b/cli/lib/kontena/cli/services/services_helper.rb
@@ -193,7 +193,8 @@ module Kontena
             puts "      ip: #{container['overlay_cidr'].to_s.split('/')[0]}"
             puts "      public ip: #{container['node']['public_ip'] rescue 'unknown'}"
             if container['health_status']
-              puts "      health: #{container['health_status']}"
+              health_time = Time.now - Time.parse(container['health_status_at'])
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/services/services_helper.rb
```diff
@@ -192,6 +192,10 @@ def show_service(token, service_id)
             puts "      dns: #{container['name']}.#{grid}.kontena.local"
             puts "      ip: #{container['overlay_cidr'].to_s.split('/')[0]}"
             puts "      public ip: #{container['node']['public_ip'] rescue 'unknown'}"
+            if container['health_status']
+              puts "      health: #{container['health_status']}"
```
jakolehm: Could we combine these ..? output is taking already too much space.

server/app/views/v1/containers/_container.json.jbuilder
```diff
@@ -41,3 +41,7 @@ else
   json.overlay_cidr nil
 end
 json.network_settings container.network_settings
+if container.health_status
+  json.health_status container.health_status
+  json.health_status_at container.health_status_at
```
jakolehm: Maybe `health_status.updated_at` instead of `health_status.health_status_at`?

cli/lib/kontena/cli/services/services_helper.rb
jnummelin: Like this?
```
health: unhealthy@2016-10-18T06:07:34.081Z
```

server/app/views/v1/containers/_container.json.jbuilder
jnummelin: Nope, the health status is not a model itself, just attributes on container:
```
field :health_status, type: String
field :health_status_at, type: Time
```

cli/lib/kontena/cli/services/services_helper.rb
jakolehm: A bit cryptic but better, how about `health: unhealthy (30s ago)`

server/app/views/v1/containers/_container.json.jbuilder
jakolehm: Ok, got it (need more coffee). I misread that in a way that json has `health_status` object.. somehow it sounds more natural to me. Is it a bad idea? (internal db model does not have to reflect directly to json api)

jnummelin: I've been thinking of having it as a model in a way that also can have some level of history available. This would also means that it would be easy to have more details on the health check, such as return codes, returned data etc.. That's maybe a separate PR.

jakolehm: Ok, so then we should return health json with a future proof format (as an object)?

PR #1155, opened 2016-10-18
CLI: Improve request error handling

SpComb:
Fix the client HTTP request error handling to give nicer error messages for all unexpected HTTP response codes.

Uses `Excon::Errors::HTTPStatusError` to raise a `Kontena::Errors::StandardError`, using either:

* response body string
* response body JSON `{"error": ...}` data
* response reason phrase

Adds rspec tests for the `Client#request` handling using WebMock.

## Preserves existing HTTP 404 errors

`bundle exec bin/kontena grid user add --grid asdfasdfasdf foobar@cloud.kontena.io`
```
 [error] Not found
```

vs existing:

```
 [error] Not found
```

## Nicer output for previously unhandled errors

`bundle exec bin/kontena grid create wat && echo yay`

```
 [warn] Option --initial-size=1 is only recommended for test/dev usage
 [fail] Creating wat grid      
 [error] Internal Server Error
```

vs existing:

```
 [warn] Option --initial-size=1 is only recommended for test/dev usage
 [fail] Creating wat grid      
 [error] Expected([200, 201]) <=> Actual(500 InternalServerError)
```

## Nicer output for detailed errors returned by the server:

```ruby
  # POST /v1/grids/:name/users
  r.post do
    data = parse_json_body
    user = User.find_by(email: data['email'])
    halt_request(404, {error: 'User not found'}) unless user
```

`kontena grid user add foobar@cloud.kontena.io`

```
 [error] User not found
```

vs existing:

```
 [error] Not found
```

### DIFF
--- a/cli/lib/kontena/client.rb
+++ b/cli/lib/kontena/client.rb
@@ -319,7 +319,7 @@ module Kontena
       # Store the response into client.last_response
       @last_response = http_client.request(request_options)
 
-      parse_response
+      parse_response(@last_response)
     rescue Excon::Errors::Unauthorized
       if token
         logger.debug 'Server reports access token expired'
@@ -332,13 +332,10 @@ module Kontena
         retry if refresh_token
       end
       raise Kontena::Errors::StandardError.new(401, 'Unauthorized')
-    rescue Excon::Errors::NotFound
-      raise Kontena::Errors::StandardError.new(404, 'Not found')
-    rescue Excon::Errors::Forbidden
-      raise Kontena::Errors::StandardError.new(403, 'Access denied')
-    rescue
-      logger.debug "Request exception: #{$!} - #{$!.message}\n#{$!.backtrace.join("\n")}"
-      handle_error_response
+    rescue Excon::Errors::HTTPStatusError => error
+      logger.debug "Request #{error.request[:method].upcase} #{error.request[:path]}: #{error.response.status} #{error.response.reason_phrase}: #{error.response.body}"
+
+      handle_error_response(error.response)
     end
 
     # Build a token refresh request param hash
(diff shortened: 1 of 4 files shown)


jnummelin:
This is IMO a must have for 0.16. Without this many errors are too cryptic for end users

SpComb:
I'm working on some rspec tests for this ATM

jnummelin:
Ok. Let's wait for those and then merge asap.

SpComb:
Added `Client#request` tests using webmock.

jnummelin:
ping @nevalla @kke You know the new cli stuff way better, could you have a look also.

cli/lib/kontena/client.rb
```diff
@@ -332,13 +332,10 @@ def request(http_method: :get, path:'/', body: nil, query: {}, headers: {}, resp
...
-    rescue Excon::Errors::NotFound
-      raise Kontena::Errors::StandardError.new(404, 'Not found')
-    rescue Excon::Errors::Forbidden
-      raise Kontena::Errors::StandardError.new(403, 'Access denied')
-    rescue
-      logger.debug "Request exception: #{$!} - #{$!.message}\n#{$!.backtrace.join("\n")}"
-      handle_error_response
+    rescue Excon::Errors::HTTPStatusError => error
+      logger.debug "Request #{error.request[:method].upcase} #{error.request[:path]}: #{error.response.status} #{error.response.reason_phrase}: #{error.response.body}"
+
```
SpComb: Hmm... or include it in the error?

```ruby

begin 
  foo = client.get("foofoo")
rescue Kontena::Errors::RequestError => error
  puts "Server said: #{error.response.body}"
end
```

jnummelin: The latter looks pretty good IMO

jakolehm: `@last_response` is a bad practice because it keeps the state and because of that it will have a wrong value if first request is successful and next one throws an error.

SpComb: Only remaining reference to `Kontena::Client#last_response` is in `Kontena::Cli::Master::LoginCommand#get_authorization_url`, and that uses `client.request(expects: [...])` so AFAIK it shouldn't be affected by this change.

SpComb:
Fixed specs. The `kontena master login` command seems to still work: `bundle exec bin/kontena master login --token ... --name development http://localhost:9292/`
```
 [error] Authentication provider not configured
```

PR #1157, opened 2016-10-18
Remove Content-Type request header when sending empty request body

nevalla:
This PR removes Content-Type request header when sending empty request body. By default `application/json` content type is sent and that is causing problems.

Also 204 response status is added to accepted response status codes.

### DIFF
--- a/cli/lib/kontena/cli/cloud/master/remove_command.rb
+++ b/cli/lib/kontena/cli/cloud/master/remove_command.rb
@@ -14,7 +14,7 @@ module Kontena::Cli::Cloud::Master
     def delete_server(id)
       spinner "Deleting server #{id} from Kontena Cloud" do |spin|
         begin
-          cloud_client.delete("user/masters/#{id}", nil, {}, { 'Content-Type' => 'text/plain' })          
+          cloud_client.delete("user/masters/#{id}")
         rescue
           spin.fail
         end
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/cloud/master/remove_command.rb
```diff
@@ -14,7 +14,7 @@ class RemoveCommand < Kontena::Command
     def delete_server(id)
       spinner "Deleting server #{id} from Kontena Cloud" do |spin|
         begin
-          cloud_client.delete("user/masters/#{id}")
+          cloud_client.delete("user/masters/#{id}", nil, {}, { 'Content-Type' => 'text/plain' })
```
jakolehm: Too hacky

nevalla:
Updated the implementation to sniff length of the request body and remove Content-Type header based on that information

jakolehm:
Much better!

cli/lib/kontena/client.rb
```diff
@@ -286,9 +286,11 @@ def request(http_method: :get, path:'/', body: nil, query: {}, headers: {}, resp
...
-      body_content = body.nil? ? '' : encode_body(body, request_headers[CONTENT_TYPE])
-      if http_method == :get
-        request_headers.delete('Content-Type')
+      if body.nil?
+        body_content = ''
+        request_headers.delete(CONTENT_TYPE)
+      else
+        body_content =  encode_body(body, request_headers[CONTENT_TYPE])
       end
       request_headers.merge!('Content-Length' => body_content.bytesize)
```
nevalla: Thanks for the tip, changed implementation to set `Content-Length` only if body is present

nevalla:
ping @jakolehm

PR #1160, opened 2016-10-18
Cli: improve app services sorting

jakolehm:


jakolehm:
Sorting is improved in the stacks functionality.

PR #1162, opened 2016-10-18
Add validation for existing grid in grid create

### LINKED ISSUE #1079: kontena grid create returns JSON internal server error if the grid already exists
```
$ kontena grid create --initial-size=3 wat
$ kontena grid create --initial-size=3 wat
{"message":"Internal server error"}
```

jnummelin:
Grid create mutation now validates if grid already exists with given name

fixes #1079

### DIFF
--- a/server/app/mutations/grids/create.rb
+++ b/server/app/mutations/grids/create.rb
@@ -17,6 +17,8 @@ module Grids
 
     def validate
       add_error(:user, :invalid, 'Operation not allowed') unless user.can_create?(Grid)
+      existing = Grid.find_by(name: self.name)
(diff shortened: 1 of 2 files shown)


jnummelin:
Hmm, I'm always somewhat against using internal codes of other systems to figure out problems. In the past I've seen too many times them changing and causing really hard-to-debug issues. :) If we want to go to the direction of figuring out specific things from moped errors we really need one common module/helper for that.

SpComb:
Yeah, it would really need something like specific `Grids::DuplicateError` exception classes to handle them sanely.

Issue #1142, opened 2016-10-14
Update completion script


kke:
The autocompleter script is currently out of sync in master, some removed commands are still there and any new commands are not.

Either:

A) Update the script

or

B) Replace the script with something automatic that stays up to date, for example some variation of #985 

Perhaps a good idea would be to add the lazy loader in #1093 while at it to make it snappier (about 2x faster cli spin up).

jakolehm:
For 0.16.0 my vote goes to option A.

jakolehm:
Fixed in #1168 .. for B we can open a new issue.

Issue #1111, opened 2016-10-07
Add zsh auto completion


kke:
Mostly it should work as is but the installation command line needs to be changed somehow.

kke:
1) Change completer init script to look like:
```bash
#!/usr/bin/env bash

_kontena_complete() {
  COMPREPLY=()
  local src="${BASH_SOURCE[0]}"
  if [ "$source" == "" ]
  then
    src="${(%):-%x}"
  fi
  DIR=$( cd "$( dirname "$src" )" && pwd )
  echo $DIR
  local word="${COMP_WORDS[COMP_CWORD]}"
  local completions="$(${DIR}/completer ${COMP_WORDS[*]})"
  COMPREPLY=( $(compgen -W "$completions" -- "$word") )
}

complete -F _kontena_complete kontena
```

2) Run:

```shell
$ autoload -U +X compinit && compinit
$ autoload -U +X bashcompinit && bashcompinit
$ source `kontena whoami --bash-completion-path`
```

3) Profit

kke:
Anyway, the current autocomplete script is not that great and needs to be manually kept up to date between releases.

PR #1168, opened 2016-10-20
Updated the completion scripts for current command set.

kke:
Also works now with zsh (but without any of the fancy stuff).


kke:
Fixes  #1111

jakolehm:
Does this also fix #1142 ?

kke:
For the A part yes.

PR #1171, opened 2016-10-20
Fix Ubuntu xenial server/agent Docker image version pinning

SpComb:
* Move the version out of the `/etc/kontena-*.env` conffile

  Upgrading the `KONTENA_VERSION=` within the conffile while preserving the
  initial `KONTENA_TOKEN=` etc configuration would have required manual editing
  on every upgrade.

* Use a systemd service drop-in to provide the `KONTENA_VERSION=`

  Using a `/lib/systemd/system/kontena-*.service.d/kontena-version.conf` unit allows
  using the same customized `/etc/systemd/system/kontena-*.service` unit file across version
  upgrades.

* Disable automatic agent version upgrades

  Removes the `-v etc/kontena-agent.env:/etc/kontena.env` bind mount, disabling
  the automatic `KONTENA_VERSION=` upgrades on server_info version conflicts.

* Drop the extraneous `docker pull` step

  Assuming that the tagged images never change, the `docker run kontena/*:${KONTENA_VERSION}`
  should be sufficient to pull the image on the initial run, or use any existing local image.

### DIFF
--- a/agent/lib/tasks/release.rake
+++ b/agent/lib/tasks/release.rake
@@ -51,7 +51,7 @@ namespace :release do
     sh('rm -rf build/ubuntu_xenial/')
     sh('cp -ar packaging/ubuntu_xenial build/')
     sh("sed -i \"s/VERSION/#{VERSION}-#{rev}/g\" build/ubuntu_xenial/#{NAME}/DEBIAN/control")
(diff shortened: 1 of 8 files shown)


SpComb:
Do we need to consider upgrades from existing installs that have a `KONTENA_VERSION=` in the `/etc/kontena-*.env` file?

jakolehm:
LGTM

PR #1175, opened 2016-10-21
Known plugin subcommands will now suggest installing the plugin 

### LINKED ISSUE #1087: Cli should suggest plugin install if subcommand is not found
Something like:

```
$ kontena vagrant master create
The plugin 'vagrant' is currently not installed. You can install it by typing: 'kontena plugin install vagrant'
```

kke:
Upgraded clamp and got #subcommand_missing

Fixes #1087


kke:
(Upgrading clamp will drop support for ruby <= 2.0.0)

jakolehm:
👍

PR #1177, opened 2016-10-21
Reintroduce kontena register with a link to signup page

kke:
Some tutorials may have "kontena register" as one of the first steps, it may scare some people away if the first command in the tutorial just reports "unknown subcommand".


jakolehm:
LGTM

PR #1179, opened 2016-10-21
Server telemetry / anon stats (with possibility to opt-out)

jakolehm:
> To continuously improve the Kontena experience, Kontena Master reports anonymous usage data to Kontena, Inc. This data is used to monitor the reliability of Kontena internal components, installations and to find out which features are most popular. This telemetry data is very helpful to us, so please leave this enabled.


jakolehm:
@nevalla ping

PR #1183, opened 2016-10-24
Skip init-cloud when not using KC

kke:
Auth provider config was not skipped properly because a question mark was missing from the method name in `respond_to?`


jakolehm:
LGTM

PR #1185, opened 2016-10-24
Fix vpn remove error

jakolehm:


jakolehm:
@kontena/core-platform ping

Issue #1165, opened 2016-10-19
Agent Kontena::Workers::NodeInfoWorker: publish_node_info: Address not available


SpComb:
On a DigitalOcean instance without private networking, the default `/etc/kontena-agent.env` `KONTENA_PEER_INTERFACE=eth1` fails if the `eth1` interface exists, but does not have any address configured:

`grep KONTENA_PEER_INTERFACE /etc/kontena-agent.env`

```
KONTENA_PEER_INTERFACE=eth1
```

`ip addr show dev eth1`

```
3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 1e:4f:db:dd:5d:bf brd ff:ff:ff:ff:ff:ff
```

I guess `Kontena::Helpers::IfaceHelper#interface_ip` does not rescue on `Errno::EADDRNOTAVAIL`, and that causes `Kontena::Workers::NodeInfoWorker` `private_ip` and `publish_node_info` to fail completely, and not fallback to `interface_ip('eth0')`.

Workaround is to configure explicit `KONTENA_PEER_INTERFACE=eth0` or `KONTENA_PRIVATE_IP=...` to use the node's public IP address.

SpComb:
Symptom of this is that the host node agent does not publish the node_info, leaving weird node info on the master:

` kontena node ls`
```
Name                                                                   Status     Initial    Labels                                  
                                                                       online     yes                                          
```

PR #1186, opened 2016-10-24
 Fix Ubuntu xenial kontena-server HTTP/HTTPS ports

### LINKED ISSUE #1182: Xenial package uses different ports
Ubuntu Xenial packages uses ports 80/443, Trusty uses ports 8080/8443.

SpComb:
Fixes #1182

* Revert to the same default 8080/8443 default ports used in the existing Ubuntu packages

* Use environment variables to document and allow configuration of different ports via `/etc/kontena-server.env`

*  Fix Ubuntu xenial `/etc/kontena-agent.env` example `KONTENA_URI=ws://127.0.0.1:8080` to match the default value configured by debconf.

   The current `wss://localhost:8443` example does not work unless a `SSL_CERT=` is configured.

### DIFF
--- a/agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
+++ b/agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
@@ -1,5 +1,5 @@
 # Set URL of your kontena server
-#KONTENA_URI=wss://localhost:8443
+#KONTENA_URI=ws://127.0.0.1:8080
 
 # Set kontena token
 #KONTENA_TOKEN=
--- a/agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
+++ b/agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
@@ -1,5 +1,5 @@
 # Set URL of your kontena server
-#KONTENA_URI=ws://127.0.0.1:8080
+#KONTENA_URI=ws://localhost:8080
 
 # Set kontena token
 #KONTENA_TOKEN=
--- a/server/packaging/ubuntu_xenial/kontena-server/etc/kontena-server.env
+++ b/server/packaging/ubuntu_xenial/kontena-server/etc/kontena-server.env
@@ -13,8 +13,8 @@
 
 # TCP ports used by clients and agents to connect to the kontena server on the master node
 # Configured as published ports on the Docker kontena-server-haproxy container
-HTTP_PORT=8080
-HTTPS_PORT=8443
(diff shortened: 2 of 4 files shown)


agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
```diff
@@ -1,5 +1,5 @@
 # Set URL of your kontena server
-#KONTENA_URI=wss://localhost:8443
+#KONTENA_URI=ws://127.0.0.1:8080
```
jakolehm: why not localhost?

server/packaging/ubuntu_xenial/kontena-server/etc/kontena-server.env
```diff
@@ -7,5 +7,14 @@
 # Kontena initial admin code
 #INITIAL_ADMIN_CODE=
 
+## These settings are used by the kontena-server-haproxy.service
+# To reload these settings, use:
+#   systemctl restart kontena-server-haproxy.service
+
+# TCP ports used by clients and agents to connect to the kontena server on the master node
+# Configured as published ports on the Docker kontena-server-haproxy container
+HTTP_PORT=8080
```
jakolehm: Should we comment out these because they are defaults?

agent/packaging/ubuntu_xenial/kontena-agent/etc/kontena-agent.env
SpComb: That's what's in the debconf... I'll change this to localhost, seems like docker-proxy listens on both IPv4 and IPv6:

```
LISTEN     0      128                                       :::8443                                                  :::*                   users:(("docker-proxy",pid=5114,fd=4))
LISTEN     0      128                                       :::8080                                                  :::*                   users:(("docker-proxy",pid=5147,fd=4))
```

server/packaging/ubuntu_xenial/kontena-server/etc/kontena-server.env
SpComb: Sure, the `/opt/bin/kontena-haproxy.sh` script should work either way.

SpComb:
Travis spec failure for `Kontena::Workers::EventWorker#stream_events` seems unrelated.

Tested to work with a fresh Ubuntu 16.04 install per the docs + dpkg -i kontena-server + kontena-agent.

Had to still tweak the `/etc/kontena-agent.env` `KONTENA_PEER_INTERFACE=eth1` default because of #1165

jakolehm:
LGTM

PR #1187, opened 2016-10-24
Refuse to start unless mongodb version = 3.0

### LINKED ISSUE #1161: server: refuse to start if mongodb is not 3.0
Kontena master will start (and migrate) with 3.2 if instances=1 -- So it's very confusing that things go south after scaling master to >1

Master should refuse to start if mongodb is not in supported version.

See https://github.com/matti/kontena-mongodb32-bomb

kke:
Fixes #1161


jakolehm:
LGTM

PR #1190, opened 2016-10-24
Fixed some typos and tried to make it read more nicely

kke:
Maybe @saana-kontenacm can make it even less robotic.


jakolehm:
LGTM!

PR #1196, opened 2016-10-25
Don't use force when removing containers

jakolehm:
Fixes (partially) #828


jnummelin:
here's some explanation why using force creates `dead` containers sometimes: https://github.com/docker/docker/issues/26372

PR #1203, opened 2016-10-25
Dont use existing token when --force login to master

kke:
This PR should make master login not pick up the existing token from config when doing `kontena master login --force`


kke:
Still going to make master not accept admin as user for regular login.

jakolehm:
LGTM

Issue #1200, opened 2016-10-25
There is no "kontena master ssh"


kke:
But there is `kontena node ssh`.

jnummelin:
closed in #1205

PR #1205, opened 2016-10-25
Ssh command improvements

kke:
```
$ kontena master ssh docker logs kontena-server-api
192.168.66.1 - - [25/Oct/2016:12:29:17 +0000] "GET /v1/config/server.provider HTTP/1.1" 200 29 0.0678
192.168.66.1 - - [25/Oct/2016:12:32:19 +0000] "GET /v1/config/server.provider HTTP/1.1" 200 29 0.0754
...
Connection to 127.0.0.1 closed.
```

```
$ kontena node ssh echo foo
foo
Connection to 127.0.0.1 closed.
```

Node ssh uses vagrant plugin if labels include `provider=vagrant`.
Master ssh uses vagrant plugin if master config has `server.provider=vagrant`. (which it doesn't unless you put it there manually)

Gets master host from `URI.parse(current_master.url).host` <-- maybe there is a more correct way?

### DIFF
--- a/cli/lib/kontena/cli/master/ssh_command.rb
+++ b/cli/lib/kontena/cli/master/ssh_command.rb
@@ -1,0 +1,40 @@
+module Kontena::Cli::Master
+  class SshCommand < Kontena::Command
+
+    include Kontena::Cli::Common
+
+    parameter "[COMMANDS] ...", "Run command on host"
+
+    option ["-i", "--identity-file"], "IDENTITY_FILE", "Path to ssh private key"
+    option ["-u", "--user"], "USER", "Login as a user", default: "core"
+
+    requires_current_master
+
+    def master_host
+      require 'uri'
+      URI.parse(current_master.url).host
+    end
+
+    def master_provider
+      Kontena.run('master config get --return server.provider', returning: :result)
+    end
+
+    def execute
+      if master_provider == 'vagrant'
+        cmd = "ssh"
+        cmd << " #{self.commands_list.join(' ')}" unless self.commands_list.empty?
+        Kontena.run("vagrant master #{cmd}")
+      else
+        cmd = ['ssh']
+        cmd << "#{user}@#{master_host}"
+        cmd << "-i #{identity_file}" if identity_file
+        if self.commands_list && !self.commands_list.empty?
+          cmd << '--'
+          cmd += self.commands_list
+        end
+        exec(cmd.join(' '))
+      end
+    end
+  end
+end
+
--- a/cli/lib/kontena/cli/master/ssh_command.rb
+++ b/cli/lib/kontena/cli/master/ssh_command.rb
@@ -20,19 +20,22 @@ module Kontena::Cli::Master
     end
 
     def execute
+
+      commands_list.insert('--') unless commands_list.empty?
+
       if master_provider == 'vagrant'
-        cmd = "ssh"
-        cmd << " #{self.commands_list.join(' ')}" unless self.commands_list.empty?
-        Kontena.run("vagrant master #{cmd}")
+        unless Kontena::PluginManager.instance.plugins.find { |plugin| plugin.name == 'kontena-plugin-vagrant' }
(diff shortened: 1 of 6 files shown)


kke:
Fixes #1200

kke:
Updated + rebased

image

cli/lib/kontena/cli/master/ssh_command.rb
```diff
@@ -0,0 +1,40 @@
...
+        Kontena.run("vagrant master #{cmd}")
+      else
+        cmd = ['ssh']
+        cmd << "#{user}@#{master_host}"
+        cmd << "-i #{identity_file}" if identity_file
+        if self.commands_list && !self.commands_list.empty?
+          cmd << '--'
+          cmd += self.commands_list
+        end
+        exec(cmd.join(' '))
```
SpComb: [`exec`](https://ruby-doc.org/core-2.3.3/Kernel.html#method-i-exec) takes a shell-less command in the form `exec(cmdname, *args)`, so to avoid any issues with shellwords and interpolation, just use something like:

```ruby
    exec(*cmd)
```

cli/lib/kontena/cli/nodes/ssh_command.rb
```diff
@@ -4,28 +4,47 @@ class SshCommand < Kontena::Command
...
+    requires_current_master
+    requires_current_grid
+
     def execute
-      require_api_url
-      require_current_grid
-      token = require_token
+      node = client.get("nodes/#{current_grid}/#{node_id}")
+
+      provider = node["labels"].find{ |l| l.start_with?('provider=')}.to_s.split('=').last
```
SpComb: What if there is no `provider=` label? I suppose this just ends up being nil by some magic.

```
irb(main):005:0> [].find{ |l| l.start_with?('provider=')}.to_s.split('=').last
=> nil
```

kke: if find returns nil, `nil.to_s` results in `""`, `"".split('=')` results in `[]`, ´[].last` results in nil.

a bit magical yes.

SpComb: Looks like recent kontena-server is also returning nodes with `"labels": null"`:

```
NoMethodError: undefined method `find' for nil:NilClass
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/nodes/ssh_command.rb:20:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:187:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.1/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:187:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.1/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:187:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.1/lib/clamp/command.rb:132:in `run'
  bin/kontena:16:in `<top (required)>'
```

kke: Null is now taken care of.

kke:
The vagrant stuff needs https://github.com/kontena/kontena-plugin-vagrant/pull/23

cli/lib/kontena/cli/master/ssh_command.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    requires_current_master
+
+    def master_host
+      require 'uri'
+      URI.parse(current_master.url).host
+    end
+
+    def master_provider
+      Kontena.run('master config get --return server.provider', returning: :result)
```
jakolehm: Repeating myself but there is something here that I don't like (related to return)... but let's discuss this more in a separate issue/pr.

PR #1209, opened 2016-10-25
Reload config after cloud login

kke:
When performing cloud login the config needs to be reloaded by the caller or `cloud_client` will use nil/old token.


kke:
Rebased

kke:
what. something downcased everything?

PR #1210, opened 2016-10-25
Reload config after cloud login

kke:
Config needs to be reloaded by the caller when doing Kontena.run('cloud login')


jakolehm:
LGTM

PR #1229, opened 2016-10-26
Fix ubuntu/bintray repo issues

jakolehm:
Changes ubuntu repo location because old one does not support multiple ubuntu versions properly.

**From bintray docs:**

There are two supported repository metadata layouts:

“Trivial”
This is the deprecated repository layout, however it is still used quite commonly. In this layout all the metadata files are stored in the root directory of the repository, so it can be useful when you are only hosting a bunch of files with no well defined structure in the repository.

“Automatic”
This is the default layout, and it contains information on the included distributions. Each distribution can include multiple components (e.g. “main”, “free”, …), each component can include multiple architectures (e.g. “amd64”, “i386”, …).

Old one is "trivial"....

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,7 @@ Dotenv.load
 
 VERSION = File.read('./VERSION').strip
 UBUNTU_IMAGE = 'kontena-ubuntu-build'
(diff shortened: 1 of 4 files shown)


jakolehm:
@kontena/core-platform ping

nevalla:
LGTM!

PR #1232, opened 2016-10-26
Installing via Docker Compose docs improvements

colinmollenhour:
Made some changes to help new users like myself:
- Update minimum version
- Changed container names to be consistent with other install methods so that docs on other pages are consistent regardless of install method used.
- Added "restart: always" to make sure daemons are restarted on reboot by default.
- Added tip on setting host config since figuring out how to do this in a scripted manner on multiple platforms is _hard_ and the Docker docs don't advertise this feature prominently enough.


jakolehm:
Nice improvements, thanks!

PR #1236, opened 2016-10-27
Username now shown when logging in using auth code

### LINKED ISSUE #1234: kontena login is missing the username of the authenticated user when using the initial admin code
"authenticated to master .... as [clips here]"

kke:
Fixes #1234


kke:
This version of config parsing has a lot of trickery to make it compatible with pre 0.16 .kontena_client.json format. We can probably get rid of all that before 1.0.

jakolehm:
LGTM

PR #1238, opened 2016-10-27
Duplicate masters now renamed during config load

### LINKED ISSUE #1226: Local CLI config allows multiple entries with same name
This can cause quite a headache since the commands will actually go to wrong master which might not exist anymore.

kke:
Fixes #1226


jakolehm:
LGTM

Issue #1240, opened 2016-10-27
Cli can't register first master, because no current master can be found


james-nesbitt:
With empty configuration, trying to register my first master, I get a complaint that no current master can be found, followed by instructions to try to register to a master.

```
> % kontena master login http://X.X.X.X 
Running #<Kontena::MainCommand:0x0056302f3ed720> -- callback matcher = 'nil'
Running #<Kontena::Cli::MasterCommand:0x0056302f3d2a38> -- callback matcher = 'nil'
Running #<Kontena::Cli::Master::LoginCommand:0x0056302f41bf30> -- callback matcher = 'master login'
D, [2016-10-27T10:28:47.938554 #1] DEBUG -- CONFIG: Configuration file not found, using default settings.
D, [2016-10-27T10:28:47.938729 #1] DEBUG -- CONFIG: Configuration loaded with 0 servers.
D, [2016-10-27T10:28:47.938753 #1] DEBUG -- CONFIG: Current master: default
D, [2016-10-27T10:28:47.938788 #1] DEBUG -- CONFIG: Current grid: (not selected)
/usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/config.rb:313:in `require_current_master': You are not logged into a Kontena Master. Use: kontena master login (ArgumentError)
```

```
-> % kontena version
cli: 0.16.0
```

Full error stack:
```
/usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/config.rb:313:in `require_current_master': You are not logged into a Kontena Master. Use: kontena master login (ArgumentError)
	from /usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/common.rb:139:in `require_current_master'
	from /usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/common.rb:179:in `client'
	from /usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/master/login_command.rb:170:in `block in get_authorization_url'
	from /usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/common.rb:80:in `vspinner'
	from /usr/lib/ruby/gems/2.3.0/gems/kontena-cli-0.16.0/lib/kontena/cli/master/login_command.rb:169:in `get_authorization_url'
[...]

PR #1245, opened 2016-10-28
Strip possible trailing arguments from remote code display

kke:
`http://foo/code#code=abcd&server_name=foofoo` resulted in the code being shown as `abcd&server_name=foofoo` when performing login with --remote.


kke:
Fixes issue mentioned in #1240

jakolehm:
LGTM

PR #1248, opened 2016-10-28
Add auth references in installation guides

jnummelin:
Installation guides didn't have references to auth guides so users missed the `kontena master init-cloud` step when setting up master with e.g. compose.


jakolehm:
LGTM

PR #1250, opened 2016-10-28
Master provider was still not sent to cloud

kke:
The provider and version were still not actually sent to the cloud during master registration

Added specs for the cloud master add command.


jakolehm:
LGTM

PR #1251, opened 2016-10-28
Call weave attach-router

### LINKED ISSUE #1249: Weave expose fails randomly on node reboot
Weave bridge creation is sometimes really slow and this causes expose to fail. 

Related: https://github.com/weaveworks/weave/issues/2479

jakolehm:
Fixes #1249

### DIFF
--- a/agent/lib/kontena/network_adapters/weave.rb
+++ b/agent/lib/kontena/network_adapters/weave.rb
@@ -201,6 +201,7 @@ module Kontena::NetworkAdapters
         end
       end
 
+      attach_router
       connect_peers(peer_ips)
       info "using trusted subnets: #{trusted_subnets.join(',')}" if trusted_subnets && !already_started?
 
@@ -211,6 +212,10 @@ module Kontena::NetworkAdapters
     rescue => exc
       error "#{exc.class.name}: #{exc.message}"
       debug exc.backtrace.join("\n")
+    end
+
+    def attach_router
+      self.exec(['--local', 'attach-router'])
     end
 
     # @param [Array<String>] peer_ips


SpComb:
This might also work using some kind of

```ruby
if weave.running?
    attach_router
else
    launch_router
end
```

logic?

jnummelin:
Would the `attach-router` call do any undesired side-effects if there's already properly running weave?

SpComb:
> Would the attach-router call do any undesired side-effects if there's already properly running weave?

Dunno... Looking at the [`attach-router` implementation](https://github.com/weaveworks/weave/blob/v1.7.2/weave#L1983), it most does exactly the same stuff that `launch-router` does, but there's an additional [`fetch_router_args`](https://github.com/weaveworks/weave/blob/v1.7.2/weave#L1983) that does some command-line parsing for the `-ipalloc-range`... as long as that works correctly and it doesn't try and do any `ipam_reclaim` stuff?

SpComb:
Yeah no, the `fetch_router_args` parsing is broken for the `['--ipalloc-range', '']` used by the kontena agent:

```
+ fetch_router_args
+ docker inspect -f {{.Args}} weave
+ CONTAINER_ARGS=[--port 6783 --name ba:7a:6c:b1:e8:55 --nickname coreos-dev-02 --datapath datapath --ipalloc-range  --dns-effective-listen-address 172.18.0.1 --dns-listen-address 172.18.0.1:53 --http-addr 127.0.0.1:6784 --resolv-conf /var/run/weave/etc/resolv.conf --dns-domain kontena.local 172.17.8.101]
+ grep -o -e -ipalloc-range [^ ]*
+ echo [--port 6783 --name ba:7a:6c:b1:e8:55 --nickname coreos-dev-02 --datapath datapath --ipalloc-range --dns-effective-listen-address 172.18.0.1 --dns-listen-address 172.18.0.1:53 --http-addr 127.0.0.1:6784 --resolv-conf /var/run/weave/etc/resolv.conf --dns-domain kontena.local 172.17.8.101]
+ IPRANGE=-ipalloc-range --dns-effective-listen-address
+ grep -o -e --no-dns
+ echo [--port 6783 --name ba:7a:6c:b1:e8:55 --nickname coreos-dev-02 --datapath datapath --ipalloc-range --dns-effective-listen-address 172.18.0.1 --dns-listen-address 172.18.0.1:53 --http-addr 127.0.0.1:6784 --resolv-conf /var/run/weave/etc/resolv.conf --dns-domain kontena.local 172.17.8.101]
+ NO_DNS_OPT=
+ true
+ wait_for_status weave http_call 127.0.0.1:6784
+ container=weave
+ shift
+ true
+ http_call 127.0.0.1:6784 GET /status
+ return 0
+ setup_awsvpc
+ [ -n  ]
+ populate_router
+ [ -n -ipalloc-range --dns-effective-listen-address ]
+ [ -n  ]
+ with_container_addresses ipam_reclaim_no_check_alive weave:expose
```

Is this harmful?

jakolehm:
Bad idea

SpComb:
> It messes up the IPRANGE=-ipalloc-range --dns-effective-listen-address bit and then proceeds to the ipam_reclaim stuff.

Reported upstream: https://github.com/weaveworks/weave/issues/2567

It doesn't seem like the spurious ipam_reclaim stuff would necessarily be harmful. It appears to do some HTTP requests against the Weave API, which just return HTTP 404 with the IPAM disabled.

PR #1241, opened 2016-10-27
[WIP] Agent weaveexec error handling

SpComb:
TODO on things like testing all the error cases. It would probably be a good idea to split out the harmless logging stuff from the actual error handling changes...

Fixes #1178
- Raise on weaveexec errors (including output logs in the message)
- Only launch weave router if not already running
- Remove error handling/retry logic, and let the weave actor crash if start fails
  
  Rely on the celluloid supervisor restarting the actor instead. 
- Trap exec errors on `container_attach` (don't want the weave actor to crash)
- Minimize dependencies on the weave actor, using helper methods instead
- Limit wait-for-weave-to-be-running to a 10/30s timeout, and log debug messages showing which class is spinning in wait.
- Log weave actions and errors
  - Each `weavexec` error is logged at **_warn**_ on retry, and **_error**_ when raising.
  - Trapped errors are logged at **_error**_ (using `error_exception` with backtrace).
  - Ignored errors are logged at `warn`.
  - Weave actions like start, attach are logged at `info`
  - Each weaveexec and `wait_running` sleep is logged at `debug`
  - Use `WEAVE_DEBUG=1` for extreme `weaveexec` details.
## Fixes
- The current [Weave#start sleep loop](https://github.com/kontena/kontena/blob/v0.16.0/agent/lib/kontena/network_adapters/weave.rb#L195) is broken if `Docker::Container.get('weave')` raises any exception after the `weaveexec launch-router`:

``` ruby
        weave = Docker::Container.get('weave') rescue nil
        wait = Time.now.to_f + 10.0
        sleep 0.5 until (weave && weave.running?) || (wait < Time.now.to_f)
```

If the `weave = ... rescue nil` is nil then it will remain nil regardless if how long you sleep.
## TODO
- Fix `Kontena::NetworkAdapters::Weave` to retry `start()` with the existing `node_info` after crashing...

PR #1252, opened 2016-10-28
Disable weave restart policy

### LINKED ISSUE #1249: Weave expose fails randomly on node reboot

jakolehm:
Fixes #1249

### DIFF
--- a/agent/lib/kontena/network_adapters/weave.rb
+++ b/agent/lib/kontena/network_adapters/weave.rb
@@ -13,12 +13,14 @@ module Kontena::NetworkAdapters
     WEAVE_VERSION = ENV['WEAVE_VERSION'] || '1.7.2'
     WEAVE_IMAGE = ENV['WEAVE_IMAGE'] || 'weaveworks/weave'
     WEAVEEXEC_IMAGE = ENV['WEAVEEXEC_IMAGE'] || 'weaveworks/weaveexec'
+    CHECK_EVENTS = ['kill', 'destroy']
 
     def initialize(autostart = true)
       @images_exist = false
       @started = false
       info 'initialized'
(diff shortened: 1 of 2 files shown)


jakolehm:
@jnummelin @SpComb ping

SpComb:
Hmm... I suppose this would be ideal, but testing the restart logic is a bit hairy.

Maybe the smaller `attach_router` change in #1251 would be simpler for 0.16.1, and then go for this + #1241 in 0.17?

jakolehm:
I prefer this PR over #1251 because this is simpler and does not bring any new weave calls into the mix (are we really really sure that `attach-router` is ok to run).

jakolehm:
Maybe we need to add backoff logic? This healing might be very resource intensive if weave does not want to stay alive (bad config etc). 

I'll look #1251 with fresh eyes and check if we can minimize the `attach-router` calls.

jakolehm:
Is this a good idea anymore? Should we close?

SpComb:
I think this ties into the broader theme of #1241 and the new `Kontena::Workers::ContainerStarterWorker`, maybe we should be doing something similar for all containers, not just weave?

None of the Docker restart-after-reboot policies really work, the only container that can actually do anything useful after a reboot is the `kontena-agent`. Having Docker start the etcd and kontena-ipam containers just leads to longer restart cycles once the agent does get everything up and running.

jakolehm:
@jnummelin what do you think, close or continue?

jnummelin:
I would close this and maybe (at some point) work on the broader restart-policy scope as @SpComb also mentions.

PR #1254, opened 2016-10-31
Login no longer raises when SERVER_NAME is null. 

kke:
Added master login specs.


jakolehm:
LGTM

jakolehm:
Oops, wrong button.

PR #1257, opened 2016-10-31
Add instructions how to setup DNS for overlay network

jnummelin:


colinmollenhour:
Note, this works if not using a local resolver already but if using a local resolver the better solution is to add the DNS as a forward zone. E.g. for 'unbound' I use:

~~~
    cat > /etc/unbound/unbound.conf.d/kontena.conf <<CONF
server:
  private-domain: "kontena.local"
  domain-insecure: "kontena.local"

forward-zone:
  name: "kontena.local."
  forward-addr: $DOCKER_GW_IP
CONF
~~~

jnummelin:
Thanks for that info. I'll update that also to the docs. Of course we cannot add every possible setup but having couple of different examples covered should give enough info for people to figure out the correct setup for their environment.

PR #1261, opened 2016-11-01
Removing ruby versions between 2.0 and 2.3 from travis

kke:
If something works on 2.0.0 and 2.3.1 it's hard to think of something that could break in 2.1 or 2.2.


kke:
Of course there could have been a temporary bug in 2.1 or 2.2

kke:
Bad idea.

Issue #1266, opened 2016-11-01
CLI appears to require ruby 2.1


mikeyyuen:
When running

```
$ DEBUG=true kontena cloud login
... after login from browser ...
  * Parsing params: "code=<omitted>"
/Library/Ruby/Gems/2.0.0/gems/kontena-cli-0.16.1/lib/kontena/cli/localhost_web_server.rb:85:in `serve_one': undefined method `to_h' for [["code", "<omitted>"]]:Array (NoMethodError)
	from /Library/Ruby/Gems/2.0.0/gems/kontena-cli-0.16.1/lib/kontena/cli/cloud/login_command.rb:85:in `block in execute'
```

it seems that to_h is a 2.1 feature

http://stackoverflow.com/questions/31683993/undefined-method-to-h-on-ruby-array

However the doc's say 2.0 should work

https://www.kontena.io/docs/getting-started/quick-start

kke:
Well spotted!

jakolehm:
Fixed in #1267

PR #1267, opened 2016-11-01
Removes use of to_h for ruby 2.0 compatibility

mikeyyuen:
Proposed fix for https://github.com/kontena/kontena/issues/1266


kke:
LGTM

mikeyyuen:
Full disclosure this is the first time I ever tried to write ruby...

kke:
It could have been made a bit prettier (for example by just calling flatten without (1)) but the original code was quite ugly (and lacking tests) anyway so for now it will do just fine.

PR #1269, opened 2016-11-02
Option to check only cli version.

matti:
running `kontena version` pings the server every time, this PR introduces a `--cli` option to check only cli version faster.


kke:
I would like to have a standard global / top level `--version`.  Too bad the `--version` has been used in some commands.

nevalla:
Looks good!

PR #1273, opened 2016-11-02
Agent: use docker-api 1.32 Docker::Network#connect body_opts

SpComb:
Uses the new docker-api 1.32 `Docker::Network#connect` parameter `body_opts`: https://github.com/swipely/docker-api/commit/0c95907eaf9d1d6ac8ac3cbe10989450ddd6d57f

We don't need to monkey-patch `Docker::Network` anymore.


SpComb:
Disclaimer: I have not actually tested this yet.

jakolehm:
Travis is not happy

jakolehm:
Docker-api has been already updated.

SpComb:
The unnecessary `Docker::Network#connect` monkey-patching is still there https://github.com/kontena/kontena/blob/master/agent/lib/docker/network.rb

Issue #1277, opened 2016-11-02
Server start order issue


mikeyyuen:
There seems to be a start order issue where `/app/app/initializers/symetric_encryption.rb` is not initialised before its been used in `/app/app/initializers/00_seed.rb`

```
[1] Puma starting in cluster mode...
[1] * Version 3.6.0 (ruby 2.3.1-p112), codename: Sleepy Sunday Serenity
[1] * Min threads: 8, max threads: 8
[1] * Environment: production
[1] * Process workers: 1
[1] * Phased restart available
[1] * Listening on tcp://0.0.0.0:9292
[1] Use Ctrl-C to stop
[6] + Gemfile in context: /app/Gemfile
/usr/lib/ruby/gems/2.3.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/symmetric_encryption.rb:109:in `decrypt': Call SymmetricEncryption.load! or SymmetricEncryption.cipher= prior to encrypting or decrypting data (SymmetricEncryption::ConfigError)
	from /app/app/models/configuration.rb:37:in `decrypt'
	from /app/app/models/configuration.rb:59:in `get'
	from /app/app/models/configuration.rb:17:in `block in seed'
	from /app/app/models/configuration.rb:16:in `each'
	from /app/app/models/configuration.rb:16:in `seed'
	from /app/app/initializers/00_seed.rb:6:in `<top (required)>'
	from /usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require'
	from /usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require'
	from /app/app/boot.rb:20:in `block in <top (required)>'
	from /app/app/boot.rb:20:in `each'
	from /app/app/boot.rb:20:in `<top (required)>'
	from config/puma.rb:5:in `require_relative'
	from config/puma.rb:5:in `block in _load_from'
	from /usr/lib/ruby/gems/2.3.0/gems/puma-3.6.0/lib/puma/configuration.rb:271:in `block in run_hooks'
	from /usr/lib/ruby/gems/2.3.0/gems/puma-3.6.0/lib/puma/configuration.rb:271:in `each'
	from /usr/lib/ruby/gems/2.3.0/gems/puma-3.6.0/lib/puma/configuration.rb:271:in `run_hooks'
[...]

PR #1278, opened 2016-11-02
Moved symetric_encryption up in app initialize order

mikeyyuen:
Proposed fix for https://github.com/kontena/kontena/issues/1277


mikeyyuen:
Ran a quick test of this change and it does seem to fix my issue.

kke:
The fix works by chance. The initializer loader was not sorting the files, so the prefix doesn't do anything. New fix in #1280 - releasing asap.

mikeyyuen:
Ah, good catch. Will close this one.

PR #1283, opened 2016-11-03
Refactor login commands and improve coverage

kke:
Cloud + master login now at 100% coverage.

Both now a bit more pleasant for the eye too.

### DIFF
--- a/cli/lib/kontena/cli/cloud/login_command.rb
+++ b/cli/lib/kontena/cli/cloud/login_command.rb
@@ -4,54 +4,55 @@ module Kontena::Cli::Cloud
   class LoginCommand < Kontena::Command
     include Kontena::Cli::Common
 
-    option ['-t', '--token'], '[TOKEN]', 'Use a pre-generated access token'
+    option ['-t', '--token'], '[TOKEN]', 'Use a pre-generated access token', environment_variable: 'KONTENA_ACCOUNT_TOKEN'
     option ['-c', '--code'], '[CODE]', 'Use an authorization code'
     option ['-v', '--verbose'], :flag, 'Increase output verbosity'
+    option ['-f', '--force'], :flag, 'Force reauthentication'
 
     def execute
-      require_relative '../localhost_web_server'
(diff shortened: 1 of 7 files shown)


cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -4,55 +4,56 @@ module Kontena::Cli::Cloud
...
-      elsif self.token.nil?
-        token = kontena_account.token ||= Kontena::Cli::Config::Token.new(parent_type: :account, parent_name: ENV['KONTENA_ACCOUNT'] || 'kontena')
-      elsif self.token
-        kontena_account.token = Kontena::Cli::Config::Token.new(access_token: self.token, parent_type: :account, parent_name: ENV['KONTENA_ACCOUNT'] || 'kontena')
+      if self.token
+        exit_with_error "Can't use --token and --force together" if self.force?
+        exit_with_error "Can't use --token and --code together"  if self.code
+      end
+
+      if !kontena_account.token || !kontena_account.token.access_token || self.token || self.force?
```
nevalla: This could be refactored into method `create_access_token?`, would make code cleaner.
What happens if `self.token` is nil?

cli/lib/kontena/cli/master/login_command.rb
```diff
@@ -239,44 +206,127 @@ def response_from_web_flow
...
+    # An URL starting with 'http' provided: try to find a server by url from config
+    # An URL not starting with 'http' provided: try to find a server by name
+    # An URL and a name provided
+    #  - If a server is found by name: use entry and update URL to the provided url
+    #  - Else create a new entry with the url and name
+    #
+    # @param name [String] master name
+    # @param url [String] master url or name
+    # @return [Kontena::Cli::Config::Server]
+    def select_a_server(name, url)
```
nevalla: Lets implement dedicated methods to select server with name and url or only with name or url. Would make this method smaller (and maybe cleaner when hiding those dirty details).

PR #1289, opened 2016-11-05
Watch & notify when dead containers are gone

### LINKED ISSUE #828: Occasionally unable to remove services

jakolehm:
This PR implements `ContainerCoroner` actor that is created when `ContainerInfoWorker` gets event from a container that is marked as dead. `ContainerCoroner` investigates dead container and sends `destroy` event when container has gone from the docker engine. After an event is sent, coroner terminates itself.

Should fix #828

coroner

### DIFF
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -52,6 +52,20 @@ module Docker
     end
 
     # @return [Boolean]
+    def dead?
+      self.state['Dead']
+    rescue
+      false
+    end
+
+    # @return [Boolean]
+    def suspiously_dead?
+      self.state['Dead'] && self.state['ExitCode'].to_i == 137
+    rescue
+      false
+    end
+
+    # @return [Boolean]
     def service_container?
       self.labels['io.kontena.container.type'] == 'container'
     rescue
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -3,6 +3,8 @@ require 'docker'
 module Docker
   class Container
 
+    SUSPICIOUS_EXIT_CODES = [137]
+
     def name
       cached_json['Name']
     end
@@ -59,8 +61,8 @@ module Docker
     end
 
     # @return [Boolean]
-    def suspiously_dead?
-      self.state['Dead'] && self.state['ExitCode'].to_i == 137
+    def suspiciously_dead?
+      self.state['Dead'] && SUSPICIOUS_EXIT_CODES.include?(self.state['ExitCode'].to_i)
     rescue
       false
     end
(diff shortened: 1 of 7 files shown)


agent/lib/kontena/actors/container_coroner.rb
```diff
@@ -0,0 +1,48 @@
...
+    def initialize(container, autostart = true)
+      @container = container
+      async.start if autostart
+    end
+
+    def start
+      @started = Time.now.to_i
+      info "starting to investigate #{@container.name}"
+      every(5) {
+        if @started >= (Time.now.to_i - 300) # 5 minutes
```
matti: pro-tip `(Time.now.to_i - (5*60))`

agent/lib/docker/container.rb
```diff
@@ -52,6 +52,20 @@ def restarting?
...
     # @return [Boolean]
+    def dead?
+      self.state['Dead']
+    rescue
+      false
+    end
+
+    # @return [Boolean]
+    def suspiously_dead?
+      self.state['Dead'] && self.state['ExitCode'].to_i == 137
```
nevalla: could `137` be a constant with reasonable name?

agent/lib/docker/container.rb
```diff
@@ -52,6 +52,20 @@ def restarting?
...
 
     # @return [Boolean]
+    def dead?
+      self.state['Dead']
+    rescue
+      false
+    end
+
+    # @return [Boolean]
+    def suspiously_dead?
```
jnummelin: I think the spelling should be `suspiciously`

agent/lib/kontena/workers/container_info_worker.rb
```diff
@@ -39,6 +39,7 @@ def on_container_event(topic, event)
       container = Docker::Container.get(event.id)
       if container
         self.publish_info(container)
+        self.notify_coroner(container) if container.suspiously_dead?
```
jnummelin: Do we want to send the coroner after all suspiciously dead containers or just service containers?

agent/lib/docker/container.rb
jakolehm: Oh, good catch.

agent/lib/kontena/workers/container_info_worker.rb
jakolehm: I think we want to investigate all.. otherwise master will have ghosts in the db.

agent/lib/docker/container.rb
jakolehm: Yes.

jakolehm:
Fixed grammar & cleaned up a bit.

PR #1291, opened 2016-11-07
Add global --version and --debug

kke:
Alternative to #1269

```
$ kontena master list --version
0.16.2
```

```
$ kontena master list --debug
D, [2016-11-07T09:23:10.260333 #68766] DEBUG -- CONFIG: Loading configuration from /Users/kimmo/.kontena_client.json
D, [2016-11-07T09:23:10.261007 #68766] DEBUG -- CONFIG: Configuration loaded with 8 servers.
D, [2016-11-07T09:23:10.261057 #68766] DEBUG -- CONFIG: Current master: (not selected)
D, [2016-11-07T09:23:10.261071 #68766] DEBUG -- CONFIG: Current grid: (not selected)
weathered-river-41-admin http://192.168.66.100:8080
purple-dream-52-admin    http://192.168.66.100:8080
```

The problem is that the provision plugins use --version to define which version to install.

### DIFF
--- a/cli/lib/kontena/command.rb
+++ b/cli/lib/kontena/command.rb
@@ -2,6 +2,15 @@ require 'clamp'
 
 class Kontena::Command < Clamp::Command
 
+  option '--version', :flag, "Output Kontena CLI version #{Kontena::Cli::VERSION}" do
+    puts Kontena::Cli::VERSION
+    exit 0
+  end
+
+  option ['-D', '--debug'], :flag, "Enable debug", environment_variable: 'DEBUG' do
+    ENV['DEBUG'] = 'true'
+  end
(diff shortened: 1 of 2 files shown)


matti:
I see some semantical problems, like `kontena master --version` gives out CLI version.

I think that the `--debug` should be a separate PR, that's definitely something I'd really want.

kke:
Now `--version` is only for the top level: `kontena --version`

matti:
is it a good idea to steal `-D` also for debug..?

kke:
It's pretty usual. Currently there doesn't seem to be any other commands that define -D. Not even any of the plugins.

Issue #1294, opened 2016-11-08
Service deploy --force to deprecate --force-deploy


mattdavenport:
Similar to #969, for API consistency `kontena service deploy --force-deploy` should be changed to simply `kontena service deploy --force`. Also similar, this should deprecate --force-deploy for current backwards compatibility until fully removed in a major version bump or otherwise.

matti:
nice, good catch

nevalla:
Fixed in #1295

PR #1295, opened 2016-11-08
Deprecate service force deploy

mattdavenport:
Clone of #969. See issue #1294.

### DIFF
--- a/cli/lib/kontena/cli/services/deploy_command.rb
+++ b/cli/lib/kontena/cli/services/deploy_command.rb
@@ -17,8 +17,7 @@ module Kontena::Cli::Services
       data = {}
       data[:force] = true if force? || force_deploy? # deprecated
       if force_deploy?
-        print "[WARNING]".colorize(:yellow)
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/services/deploy_command.rb
```diff
@@ -7,14 +7,19 @@ class DeployCommand < Kontena::Command
...
 
     def execute
       require_api_url
       token = require_token
       service_id = name
       data = {}
-      data[:force] = true if force_deploy?
+      data[:force] = true if force? || force_deploy? # deprecated
+      if force_deploy?
+        print "[WARNING]".colorize(:yellow)
```
nevalla: `Kontena::CLI::Common` module (included already) includes `warning` method you can use here: https://github.com/mattdavenport/kontena/blob/master/cli/lib/kontena/cli/common.rb#L93
It will output the warning message to STDERR with the correct formatting and color, so you don't need to worry about that.

mattdavenport:
Fixed, thanks for the heads up!

Issue #1292, opened 2016-11-07
Add multiple node label argument support


mattdavenport:
As a convenience/enhancement, the cli tool should accept multiple labels to add and remove from nodes. For example: `kontena node label add node1 nginx firewall db` should work as opposed to 3 commands to remove labels from the same node.

The same should also be supported for `kontena node label rm`

mattdavenport:
Just noticed that this appears to be possible with `kontena node update --label <label1> --label <label2> .... `, but that appears to clobber any existing previous values. Still making a case for adding multiple label functionality to `kontena node label` add/rm to provide finer grained functionality.

mattdavenport:
For further enhancement, this argument globbing could persist throughout cli commands where currently not present (e.g. `kontena service <cmd> <service_1> <service_2> ...`)

nevalla:
I have opened new issue for service commands (#1318). This issue is fixed in #1296.

PR #1296, opened 2016-11-09
Add / Rm multiple node labels. Added label list command.

kke:
Implements #1292

Also makes this is possible:

```
$ kontena node label ls bold-sunset-91 | xargs kontena node label add happy-spring-44
```


nevalla:
LGTM!

PR #1312, opened 2016-11-12
Add quiet option to service list command

### LINKED ISSUE #1306: CLI: for scriptability kontena service ls -q | xargs kontena service ...
or similar to print out just service names for scripting:

`kontena service ls -q | xargs kontena service ...`

nevalla:
This PR adds `-q , --quiet` option to `kontena service ls` command. With that option only service names are shown in the output.

Fixes #1306


jakolehm:
Nice! I think we should also tune `service deploy/scale/rm` etc commands to support multiple service names (separate PR).

PR #1313, opened 2016-11-13
Remove previous version of a plugin on install

jakolehm:
This PR improves plugin install logic a bit: now install will first remove previous plugin versions and then install the requested plugin. Reason for doing this is that multiple versions of the same plugin might have conflict in a load path and because of this wrong libraries/files might get loaded.


kke:
LGTM

PR #1322, opened 2016-11-15
Call IPAM cleanup using API enpoints

jnummelin:
Moved the IPAM cleanup loop into IPAM launcher as it's really a task for Weave adapter.

Needs: https://github.com/kontena/kontena-docker-ipam-plugin/pull/61

As soon as we have IPAM release with the cleanup endpoint we should also pin the version and not to use `latest`.

### DIFF
--- a/agent/lib/kontena/launchers/ipam_plugin.rb
+++ b/agent/lib/kontena/launchers/ipam_plugin.rb
@@ -7,12 +7,11 @@ module Kontena::Launchers
     include Kontena::Logging
     include Kontena::Helpers::ImageHelper
 
-    CLEANUP_INTERVAL = (3*60) # Run cleanup every 3mins
-    CLEANUP_DELAY = 30
+
 
     IPAM_SERVICE_NAME = 'kontena-ipam-plugin'.freeze
 
-    IPAM_VERSION = ENV['IPAM_VERSION'] || 'latest'
+    IPAM_VERSION = ENV['IPAM_VERSION'] || '0.2'
     IPAM_IMAGE = ENV['IPAM_IMAGE'] || 'kontena/ipam-plugin'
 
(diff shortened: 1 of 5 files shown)


agent/lib/kontena/launchers/ipam_plugin.rb
```diff
@@ -21,6 +24,26 @@ def initialize(autostart = true)
       async.ensure_image if autostart
     end
 
+    def cleanup_ipam
```
jakolehm: Improvement but I still think that this cleanup should be a separate actor.

jakolehm:
Should we include this to 0.17?

jnummelin:
I think this should be part of 0.17, otherwise there's nothing that would cleanup stale address reservations.

agent/lib/kontena/network_adapters/ipam_cleaner.rb
```diff
@@ -0,0 +1,50 @@
...
+        debug "invoking cleanup with #{addresses.size} known addresses"
+        ipam_client.cleanup_network('kontena', addresses, cleanup_index)
+      end
+    end
+
+    def collect_local_addresses
+      debug "starting to collect locally known addresses"
+      local_addresses = []
+      Docker::Container.all(all: true).each do |container|
+        local_addresses << container.overlay_cidr
```
SpComb: Prefer `container.overlay_ip`

SpComb: 1. The kontena-ipam ignores the CIDR suffix in the cleanup addresses
2. The only place in the agent that really needs the full CIDR address is the `weave attach` part
3. For #1336, the `container.overlay_cidr` may still have the old CIDR suffix. The `weave attach` logic translates it

PR #1323, opened 2016-11-15
Configurable grid subnet, supernet

### LINKED ISSUE #1304: Configurable grid subnet/supernet
The kontena cli, server and agent originally used the hardcoded `10.81.0.0/19` subnet for the grid overlay network. With the kontena-IPAM in master, this is being expanded to `10.81.0.0/16` to allow the use of `docker network create --ip-range` for dynamic allocations (https://github.com/kontena/kontena-docker-ipam-plugin/issues/15) vs statically allocated host node addresses. With future support for multiple distinct networks, the kontena IPAM will currently use the default `KONTENA_IPAM_SUPERNET=10.80.0.0/12` (https://github.com/kontena/kontena-docker-ipam-plugin/issues/17).

However, the grid subnet (`10.81.0.0/16`) in particular must not overlap with any other routes on the host nodes, which is an issue for platforms which lay claim to the entire `10.0.0.0/8` address space for infrastructure use, such as [Packet](https://www.packet.net/help/kb/creating-an-overlay-network-on-packet/):

> For private addressing, we strongly reccommend using the 192.168.0.0/16 network for any shared back-end networks; the “10 network” (10.0.0.0/8) is used for numbering the physical interfaces on Packet bare-metal servers, as as well as shared services like iSCSI block storage, and is likely to cause collisions.

The solution for this is a per-grid configurable subnet (`10.81.0.0/16`) and supernet (`10.80.0.0/12`). These would be configured using `kontena grid create --subnet 192.168.0.0/16 --supernet 172.16.0.0/12 testgrid`, and presumably cannot be changed later

Note that any choice of grid subnet also comes with additional implications when using the Kontena VPN... e.g. using `192.168.0.1` for the initial kontena node address would be ***highly*** likely to cause issues when using the Kontena VPN..

Ideally this would also include support for IPv6 Unique Local Addresses ([ULA](https://en.wikipedia.org/wiki/Unique_local_address)), which can be [automatically generated](http://unique-local-ipv6.com/) at grid creation time. However, this would require additional support for IPv6 within both the kontena-ipam (https://github.com/kontena/kontena-docker-ipam-plugin/issues/10) and the overlay network (weave).

SpComb:
Fixes #1304

# Server
* Add `Grid` `subnet`, `supernet` to model, JSON views
* Add `HostNode` `overlay_ip` to models, JSON views

    Uses `IPAddr` to compute the offset, allowing the use of `node_number > 255` once any existing `10.81.1.X` service containers are migrated over to the new `10.81.128.X` range.

* Support optional `POST /grids {"subnet": "...", "supernet": "..."}`
* Migrate existing girds to use the current default subnet/supernet

# CLI
* Render grid subnet, host node overlay_ip from server JSON
* Render `kontena grid cloud-config` `--insecure-registry=` from the grid subnet
* Support `kontena grid create --subnet= --supernet=`

# Agent
* Add `IPAddr` helpers
* Use `node_info['grid']['subnet']` and `node_info['overlay_ip']` for etcd, weave launchers
* Use `node_info['grid']['supernet']` for the IPAM launcher
* Use `node_info['grid]['subnet']` for the default kontena network

    Automatically compute the iprange from the upper half of the grid subnet.


SpComb:
Rebased for 0.17 git master

SpComb:
Rebased for 1.0 git master.

TODO on fixing up the `WeaveWorker#attach_overlay` 0.16 container migration logic; we can't attach containers until we have the grid configuration, which changes the  WeaveWorker` startup ordering again.

SpComb:
Assigning to me as I still need to work on the configurable subnet migration.

jakolehm:
Is this progressing?

SpComb:
> Is this progressing?

This is blocked on the hardcoded `/16` subnet mask used for <1.0 container network migration in the WeaveWorker. Each attach event needs access to the grid/node info from the server, which only exists in the Celluloid notifications atm.

jakolehm:
But this can be only configured when grid is created, right? I can't see how this affects <1.0 migration?

SpComb:
The WeaveWorker needs to know if the `io.kontena.container.overlay_cidr` label on the container contains an old pre-migration subnet mask. If it does, then it rewrites it to the desired grid overlay subnet mask. For that to happen, it needs to know what the current actual grid overlay subnet mask is, and you can't update that into the pre-existing container labels.

For new service containers being deployed, it's indeed not an issue; the server assigns them an `overlay_cidr` and the agent just sets and uses that label.

SpComb:
Rebased to current git master.

Travis is failing on the bad server specs.

SpComb:
The CLI/server parts of this are mostly okay, but the agent bits are pending on a major refactoring to clean up the `agent:node_info` notifications and weave actors.

SpComb:
Rebased over git 1.1-dev master.

@jakolehm we could avoid the need for the agent `node_info` refactoring and include this in 1.1 if we enforce that the `kontena grid --subnet` is always a `/16`.

SpComb:
The server now validates that the `Grid.subnet` is a `/16`, so this should be compatible with the hardcoded `OVERLAY_SUFFIX = '16'` that the agent uses for 0.16 container migration.

SpComb:
Hmm... alternatively the agent could use some other container labels to detect 0.16 containers, and only do the `overlay_cidr` mangling for `/12`-> `/16` for those? Then we wouldn't need any `node_info` state or `/16` network restrictions..

SpComb:
Okay, changed the agent weave container overlay network migration logic to only apply to 0.16 containers that don't yet have the `io.kontena.container.overlay_network` label, where we can assume that the grid subnet is going to be `10.81.0.0/16`.

For 1.0 grids, the agent just uses the `io.kontena.container.overlay_cidr` label without any migration fixup logic.

agent/lib/ipaddr_helpers.rb
```diff
@@ -0,0 +1,48 @@
+class IPAddr
```
jnummelin: We're doing IPAddr patching in many places now, maybe it's time to roll these as a separate gem? (not maybe part of this PR, but overall)

SpComb: Yes, the stdlib `IPAddr` is really inadequate. I figure these should be replaced with the [`ipaddress`](https://github.com/ipaddress-gem/ipaddress) gem.

The worst part is the crazy-looking `IPAddr` operations on the server, like

```ruby
(IPAddr.new(self.grid.subnet) | self.node_number).to_s
```

This would be much nicer with the agent IPAddr patching (`IPAddr.new(node_info['grid']['subnet'])[node_number]`), but the way the kontena repo is structured, there's no easy way to share code between across all three sub-components...

agent/spec/lib/kontena/launchers/etcd_launcher_spec.rb
```diff
@@ -343,8 +357,10 @@
       expect(subject.wrapped_object).to receive(:add_membership).with(excon, 'http://10.81.0.3:2380')
       node_info = {
         'node_number' => 3,
+        'overlay_ip' => '10.81.0.3',
         'grid' => {
-          'initial_size' => 3
+          'initial_size' => 3,
+          'subnet' => '10.81.0.0/16',
```
SpComb: Refactored specs to use context `node_info`

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -109,10 +109,9 @@ def on_container_destroy(event)
     # @param [String] container_id
     # @param [String] overlay_cidr
     def attach_overlay(container)
-      if container.overlay_suffix == OVERLAY_SUFFIX
-        network_adapter.attach_container(container.id, container.overlay_cidr)
-      else
-        # override label for existing containers that may need to be migrated
+      if container.overlay_network.nil?
+        # overlay network migration for 0.16 compat
+        # override overlay network /12 -> /16 suffix for existing containers that may need to be migrated
```
SpComb: Fixed. I thought I rgrep'd for all of these; I confused the grid supernet with the old subnet :)

agent/lib/ipaddr_helpers.rb
SpComb: Should I fix all the server and agent stuff to start using `ipaddress` in this PR, or later?

SpComb:
The plugins would also need to fixed to use the `grid['subnet']` for the docker `--insecure-registry=`

SpComb:
> Documentation should be updated, especially regarding the supernet/subnet relation. That might not be really intuitive for all the Kontena users.

Added Kontena Network Model docs. Not sure what other docs should mention this? It's a pretty niche thing... until we have to start dealing with platforms/providers using `10.81.0.0/16` (like Packet, see #1304).

SpComb:
@kke IMO this should be mostly okay for 1.1, if we want to. Need to fix the plugins to support this after merging, though.

agent/lib/ipaddr_helpers.rb
jnummelin: I think we should do it later as separate exercise.

docs/core-concepts/networking.md
```diff
@@ -26,20 +26,38 @@ While a Kontena Master can manage multiple Grids, each Grid is an isolated overl
...
 
 Each Grid may also include a set of Trusted Subnets, which are used for the Overlay Networking between Host Node public and private IP addresses as described further below.
 
+### Subnet and Supernet
+
+Each grid is configured with a single subnet (default `10.81.0.0/16`) and supernet (default `10.80.0.0/12`).
+Both of these private RFC1918 IP address spaces are used for internal overlay networking within each grid.
+These overlay network addresses are only unique within a grid; different grids can (and will) use the same overlay network IP addresses.
+
+The grid subnet (`10.81.0.0/16`) is used to provide overlay networking addresses for both host nodes and service contaienrs.
```
jnummelin: Small typo

jnummelin:
I think the grid command reference should be updated to reflect this in `http://kontena.io/docs/using-kontena/grids.md`

SpComb:
> I think the grid command reference should be updated to reflect this in /docs/using-kontena/grids.md

Fixed several other things at the same time, see #1727

SpComb:
Rebased for latest 1.1.x master, pruned the /16 valiation stuff, and re-added the grid subnet docs left out of #1727

Issue #1320, opened 2016-11-15
Agent weave detach call is unnecessary


jakolehm:
Because weave will call detach automatically when container is removed.

jnummelin:
Closed via #1324

PR #1324, opened 2016-11-15
Remove launching of weave plugin and unnecessary weave detach call

### LINKED ISSUE #1319: Agent starts weave docker plugin (without using it)

jnummelin:
The weave docker plugin launch was left over and now unnecessary as docker plugin networking proved to be not usable.

Weave detach is also "unnecessary" since at the time we get to call the `detach` the container is already gone.

fixes #1319 and #1320


SpComb:
I suppose the "unnecessary" means that "by the time we get to `detach_network`, the container has already been removed and the weave veth no longer exists"?

jnummelin:
Yes, updated the PR description with some more details. :)

PR #1326, opened 2016-11-16
Call weave attach async

jnummelin:
Refactor weave execs into separate actor pool and do network attach asynchronously. This improves event handling since weave attach will not block anymore.

### DIFF
--- a/agent/lib/kontena/network_adapters/weave.rb
+++ b/agent/lib/kontena/network_adapters/weave.rb
@@ -16,6 +16,8 @@ module Kontena::NetworkAdapters
 
     DEFAULT_NETWORK = 'kontena'.freeze
 
+    finalizer :finalizer
+
     def initialize(autostart = true)
       @images_exist = false
       @started = false
@@ -25,6 +27,14 @@ module Kontena::NetworkAdapters
       subscribe('agent:node_info', :on_node_info)
       subscribe('ipam:start', :on_ipam_start)
       async.ensure_images if autostart
+      @executor_pool = WeaveExecutor.pool(args: [autostart])
+      # ^ Default size of pool is number of CPU cores, 2 for 1 core machine
+    end
+
+    def finalizer
+      @executor_pool.terminate if @executor_pool.alive?
+    rescue
+      # If Celluloid manages to terminate the pool (through GC or by explicit shutdown) it will raise
     end
 
     # @return [String]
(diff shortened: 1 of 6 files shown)


jakolehm:
Should we include this to 0.17?

jakolehm:
Needs rebase.

agent/lib/kontena/agent.rb
```diff
@@ -82,6 +82,10 @@ def supervise_launchers
     end
 
     def supervise_network_adapter
+      @supervisor.pool(
```
jnummelin: I understood from Celluloid docs that you shouldn't really do that: https://github.com/celluloid/celluloid/wiki/Pools#gotcha-dont-make-pools-inside-workers

jakolehm: I "think" the doc says that you should not create pool inside a worker that is part of a pool?

confusing

agent/lib/kontena/network_adapters/weave_executor.rb
```diff
@@ -0,0 +1,94 @@
...
+    def initialize(autostart = true)
+      @images_exist = false
+      #$CELLULOID_DEBUG = true
+      info 'initialized'
+      async.ensure_images if autostart
+    end
+
+    # @param [Array<String>] cmd
+    def execute(cmd)
+      sleep 0.5 until @images_exist
```
SpComb: Are these still guaranteed to execute consecutively and in order?

Particularly for the startup sequence, the `launch-router` must happen before the `expose` etc.

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -252,13 +192,13 @@ def start(info)
           '--password', ENV['KONTENA_TOKEN']
         ]
         exec_params += ['--trusted-subnets', trusted_subnets.join(',')] if trusted_subnets
-        self.exec(exec_params)
+        Actor[:weave_executor].execute(exec_params)
         weave = Docker::Container.get('weave') rescue nil
```
SpComb: I suspect this `Docker::Container.get('weave')` is broken with the async `launch-router` execute.

Particularly since the sleep-until loop below never re-gets it.

SpComb:
NVM, seems like only the `attach` itself is done with `Actor[:weave_executor].async.execute`, I suppose the remaining ones will remain synchronous.

jnummelin:
Yes, only `attach` is done async, all others remain synchronous.

agent/lib/kontena/agent.rb
jnummelin: Yes, you are right sir. :D

SpComb:
If the async stuff only happens for `attach`, then could this be some kind of `ContainerAttachWorker` pool?

jnummelin:
There is no weave related handling for detaching as weave does it by itself. So there should not be any races related to that.

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -25,7 +25,8 @@ def initialize(autostart = true)
       subscribe('agent:node_info', :on_node_info)
       subscribe('ipam:start', :on_ipam_start)
       async.ensure_images if autostart
-      #async.cleanup_ipam if autostart
+      Actor[:weave_executor] = WeaveExecutor.pool
```
jnummelin: Good point. Celluloid terminates pools when garbage collected, but when it's in the Actor registry it will not get collected. As the executor pool is not used outside of `WeaveAdapter` I'll move it out of the registry and store as instance variable. That way it will also get cleaned via garbage collection.

SpComb:
> There is no weave related handling for detaching as weave does it by itself. So there should not be any races related to that.

Yes, the current bit with `@executor_pool.async.execute-attach` is fine, if the container is gone by the time the pool executes the async attach call, it will just fail without any adverse side-effects.

I was speculating about my proposed `ContainerAttachWorker` that would also handle e.g. the Weave `add_dns` asynchronously, because then that might race with the `remove_dns`... although it seems like we actually dropped that functionality, and the `WeaveWorker` does not perform any `remove_dns` anymore?

* https://github.com/kontena/kontena/commit/7cb81c7c84f6488505bd681927c46f7309e6cfe9#diff-eb2e3e985ee446ad2303e5d23b346fd8L47
* https://github.com/kontena/kontena/commit/8771debab8ef0a48d378dee893b23a2240bbb8a0#diff-eb2e3e985ee446ad2303e5d23b346fd8L87

Is that intentional?

jakolehm:
LGTM

SpComb:
Conflict on #1333

jnummelin:
Rebased

jakolehm:
LGTM

PR #1333, opened 2016-11-16
Container dns registration fixes (stack related)

jakolehm:


agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -55,35 +55,9 @@ def on_container_event(topic, event)
 
     # @param [Docker::Container] container
     def weave_attach(container)
-      config = container.info['Config'] || container.json['Config']
```
SpComb: :+1: for fixing this `.info || .json` thing, because this is called with both `Docker::Container.get` and `Docker::Container.all.each`

jakolehm:
@jnummelin @SpComb ping

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -95,5 +69,51 @@ def weave_attach(container)
...
+      if container.default_stack?
+        hostname = container.labels['io.kontena.container.name']
+        dns_names = default_stack_dns_names(hostname, service_name, domain_name)
+        dns_names = dns_names + stack_dns_names(hostname, service_name, domain_name)
+
+      else
+        hostname = container.config['Hostname']
+        dns_names = stack_dns_names(hostname, service_name, domain_name)
+      end
+      dns_names.each do |name|
```
jakolehm: You mean that container should know what dns names it should have?

SpComb: Yes. The naming policy could be expressed nicely using something along the lines of `yield #{self.hostname}.#{self.domain}` and it would be easier to test.

Nothing we can't refactor later if we want to.

jakolehm: Yes, makes sense in a way.. but is it clear that if I call `container.each_dns_name` that it will return dns names that container _should_ have, but not necessarily have them before they are set to weave dns? 

Anyways not all containers have dns names so it might be a bad idea to bundle them to the Container class.

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -72,7 +72,7 @@ def weave_attach(container)
     # @param [String] container_id
     # @param [String] overlay_cidr
     def attach_overlay(container_id, overlay_cidr)
-      Actor[:network_adapter].exec(['--local', 'attach', overlay_cidr, '--rewrite-hosts', container.id])
+      Actor[:network_adapter].exec(['--local', 'attach', overlay_cidr, '--rewrite-hosts', container_id])
```
SpComb: For testing, it might be better to have something like this?

```ruby
   def network_adapter
      Actor[:network_adapter]
   end
   def attach_overlay(container)
       network_adapter.attach_container(container.id, container.overlay_cidr)
   end
```

And then spec with 

```
    subject do
        described_class.new
    end
    let :network_adapter
        instance_double(Kontena::NetworkAdapters::Weave)
    end
    before
       allow(subject).to receive(:network_adapter).and_return(network_adapter)
    end
    it '...' do
        expect(network_adapter).to receive(:attach_container).with('aaaaabbbcc', '10.81.128.0/16')
    end
```

agent/spec/lib/kontena/workers/weave_worker_spec.rb
```diff
@@ -92,6 +92,23 @@
...
+      spy(:adapter)
+    end
+
+    before(:each) do
+      allow(subject.wrapped_object).to receive(:network_adapter).and_return(adapter)
+    end
+
+    it 'calls network_adapter' do
+      expect(adapter).to receive(:exec) {|*args|
+        expect(args[0]).to include('aaa', '10.81.1.1/16')
```
SpComb: Note to self: refactor this in #1336 :)

Issue #1228, opened 2016-10-26
Generated self signed cert for master should be installed locally


kke:
The provisioning plugins place the cert into cloudinit.

To install the cert locally and get rid of the SSL cert errors you need to dig that file from the host over ssh.

The .pem should be installed automatically to `~/.kontena/certs/<hostname>.pem`

kke:
It would also be possible to generate the cert on master if one was not supplied and have a command on CLI that would install it magically (`kontena master ssl-certificate install`)

aleksijohansson:
This would really be a good feature.

nevalla:
Fixed in #1337

PR #1337, opened 2016-11-16
Install self-signed cert locally

kke:
Implements #1228 if the provisioining plugins return the cert in the result response hash. 

It also fixes a problem with having multiple instances of Kontena::Client. The current version pollutes Excon.defaults which will have effect on subsequently created instances of excon. This version only sets configuration for the current instance.

It is a bit questionable though:

The `generate_self_signed_cert` method, which I have not touched here, sets the certificate hostname for all generated certs to "Test". If your master url is not `https://Test/` the cert will never work without disabling ssl peer verification.

Because of this, it's required to fake the hostname of the master using `:ssl_verify_peer_host => "Test"`  when a cert file with a subject-string that looks like one generated by `generate_self_signed_cert` is used, otherwise you will get an error saying: `hostname "10.0.0.1" does not match the server certificate (OpenSSL::SSL::SSLError)`

With the current version, even if you are crafty enough to ssh into the master and copy the certificate from there, it's not possible to actually use the certificate unless you trick the server hostname to be "Test" using `/etc/hosts` or something like that. The only way to utilize the cli-generated self signed certs is to use `IGNORE_SSL_ERRORS=true`.

Another issue with this implementation is that it doesn't work if you use something else than the public ip to access your master.

Securitywise it feels a bit fishy to trust all certificate files that belong to "Test", but I don't know how you could end up having a file called `master_ip.pem` in your `~/.kontena/certs/` directory, it's only done for files loaded from there.

The change required for plugins looks like this:

```ruby
# master_provisioner.rb:
    ssl_cert = generate_self_signed_cert
    public_cert = certificate_public_key(ssl_cert)
# ...
      {
        name: name.sub('kontena-master-', ''),
        public_ip: public_ip['address'],
        code: opts[:initial_admin_code],
        provider: 'packet',
        version: master_version,
        ssl_certificate: public_cert
      }
```

### DIFF
--- a/cli/lib/kontena/client.rb
+++ b/cli/lib/kontena/client.rb
@@ -51,14 +51,20 @@ module Kontena
       @logger.progname = 'CLIENT'
 
       @options[:default_headers] ||= {}
-      Excon.defaults[:ssl_verify_peer] = false if ignore_ssl_errors?
+
+      cert_file = File.join(Dir.home, "/.kontena/certs/#{uri.host}.pem")
+      unless File.exist?(cert_file) && File.readable?(cert_file)
+        cert_file = nil
+      end
 
       @http_client = Excon.new(
         api_url,
         omit_default_port: true,
         connect_timeout: ENV["EXCON_CONNECT_TIMEOUT"] ? ENV["EXCON_CONNECT_TIMEOUT"].to_i : 5,
         read_timeout:    ENV["EXCON_READ_TIMEOUT"]    ? ENV["EXCON_READ_TIMEOUT"].to_i    : 30,
-        write_timeout:   ENV["EXCON_WRITE_TIMEOUT"]   ? ENV["EXCON_WRITE_TIMEOUT"].to_i   : 5
+        write_timeout:   ENV["EXCON_WRITE_TIMEOUT"]   ? ENV["EXCON_WRITE_TIMEOUT"].to_i   : 5,
+        ssl_verify_peer: ignore_ssl_errors? ? false : true,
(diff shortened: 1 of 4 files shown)


jakolehm:
@kke can you explain what happens when plugin returns that `ssl_certificate` key? I don't see any implementation for handling that in this pr?

kke:
It wasn't there.

cli/lib/kontena/client.rb
```diff
@@ -51,15 +51,28 @@ def initialize(api_url, token = nil, options = {})
...
-      )
+        write_timeout:   ENV["EXCON_WRITE_TIMEOUT"]   ? ENV["EXCON_WRITE_TIMEOUT"].to_i   : 5,
+        ssl_verify_peer: ignore_ssl_errors? ? false : true
+      }
+
+      cert_file = File.join(Dir.home, "/.kontena/certs/#{uri.host}.pem")
+      if File.exist?(cert_file) && File.readable?(cert_file)
+        excon_opts[:ssl_ca_file] = cert_file
+        key = OpenSSL::X509::Certificate.new(File.read(cert_file))
+        if key.issuer.to_s == "/C=FI/O=Test/OU=Test/CN=Test"
```
jakolehm: I think we should change this self-signed cert issues to something better.. how about `/C=FI/O=Kontena/OU=Test/CN=Test` ?

jakolehm: And maybe it should be constant in the ´CertHelper` so that we don't need c&p it to here?

PR #1340, opened 2016-11-17
Cloud integration 

nevalla:


server/Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'mongoid', '~> 4.0.0'
 gem 'symmetric-encryption'
 gem 'activesupport', '~> 4.1.14'
 gem 'faye-websocket', '~> 0.10.0'
+gem 'eventmachine', '>= 0.12.0'
```
jakolehm: why?

nevalla: `CloudWebsocketConnectJob` and `Cloud::WebSocketClient` are using it. Should we get rid of it or just rely on `faye-websocket` will include it?

jakolehm: Server has been using eventmachine from the beginning.. I just thought that 0.12.0 version is somewhat misleading here.

nevalla: Okay, just randomly selected the same as faye-websocket is using. But let's remove that dependency as redundant(?))

server/app/jobs/cloud_websocket_connect_job.rb
```diff
@@ -0,0 +1,51 @@
+require 'celluloid'
+require_relative '../services/logging'
+
+class CloudWebsocketConnectJob
+  include Celluloid
+  include Logging
+  include ConfigHelper # adds a .config method
+
+  def initialize(perform = true)
+    @root_url = ENV['CLOUD_WS_URL']
```
jakolehm: This env should be passed as a parameter.

nevalla:
@jakolehm improved websocket client so that config/seed.yml reads socket uri from env variable and initializer script use that to configure client. Any better that way?

PR #1341, opened 2016-11-17
Subcommand only needs to be typed until it's unique

kke:
Not sure if useful:

```sh
# kontena master current:
$ kontena m cu  
broken-waterfall-8 https://10.75.194.77
# kontena upcloud master create:
$ kontena u m c
ERROR: option '--username' is required
#  suggestions:
$ kontena c
Suggestions: cloud, certificate, container
ERROR: No such sub-command 'c'

See: 'kontena --help'
```

### DIFF
--- a/cli/lib/kontena/command.rb
+++ b/cli/lib/kontena/command.rb
@@ -188,6 +188,16 @@ class Kontena::Command < Clamp::Command
     exit(@exit_code) if @exit_code.to_i > 0
     @result
   end
+
+  def subcommand_missing(name)
+    guesses = self.class.recognised_subcommands.flat_map(&:names).select {|sc_name| sc_name.start_with?(name)}
+    if guesses.size == 1
+      return self.class.find_subcommand(guesses.first).subcommand_class
+    elsif guesses.size > 1
+      puts "Suggestions: #{guesses.join(', ')}"
+    end
+    super(name)
+  end
 end
 
 require_relative 'callback'


cli/lib/kontena/command.rb
```diff
@@ -188,6 +188,16 @@ def run(arguments)
     exit(@exit_code) if @exit_code.to_i > 0
     @result
   end
+
+  def subcommand_missing(name)
+    guesses = self.class.recognised_subcommands.flat_map(&:names).select {|sc_name| sc_name.start_with?(name)}
+    if guesses.size == 1
+      return self.class.find_subcommand(guesses.first).subcommand_class
+    elsif guesses.size > 1
+      puts "Suggestions: #{guesses.join(', ')}"
```
SpComb: `STDERR.puts`

SpComb: This could just throw a nicer error directly: `ERROR: Non-unique subcommand prefix 'l' of list, ls, label`

instead of the current:

```
Suggestions: list, ls, label
ERROR: No such sub-command 'l'
```

kke:
Probably quite useless anyway.

SpComb:
Maybe... working tab completion is probably > lazy subcommand lookups.

kke:
The kontena-cli-completer repo has a fine automatic tab completion magic thing.

It would be nice if there was some nice way to keep prototype PR's around without closing and without cluttering the open pr's list.

SpComb:
Maybe this should be a Clamp PR instead?

kke:
It will never be accepted so it's just as pointless there :)

kke:
Not very useful.

SpComb:
I suppose a closed PR is just as fine a place as any to leave prototype PR's that aren't going anywhere.

We can still have @matti show up and get a :100: :+1:'s on this, then we can re-open it :)

kke:
Maybe some label. Close + label "interesting" :)

SpComb:
Do you mean a "maybe" label?

matti:
not interested thank you for remembering me. bash/zsh completition serves better.

Issue #1275, opened 2016-11-02
Services get stuck when node label is not defined


mattdavenport:
If a service is defined in `kontena.yml` with affinities such as `label==db` and there are no nodes with such label, the service gets stuck in a deployment state. It seems that this will randomly block further deployment steps, but sometimes deployment will continue leaving instances at 0 for the particular service. However there is no helpful message printed during deployment stating there is no node matching the affinity for the service.

mattdavenport:
Furthermore, app logs for the service provide no output, and logs from the `kontena_master` container only show constant GET requests to the `/containers` endpoint for the service.

colinmollenhour:
It seems "deploy" is expecting some condition which doesn't need to happen. Using a config like:

```yaml
  scaling: 1
  strategy: daemon
  deploy_opts:
    min_health: 0.0
  affinity: 
    - label==db-node
```

with nodes:

```
$ k node ls
Name                                                                   Status     Initial    Labels                                  
test2.local.lan                                                        online     yes        region=packet-ewr             
test3.local.lan                                                        online     no         region=packet-ewr,db-node
test4.local.lan                                                        online     no         region=packet-ewr,db-node
test5.local.lan                                                        online     no         region=packet-ewr,db-node
```

I run the deploy to add a node (two already were running, added db-node label to third, ran deploy). It spins for a long time and finally times out:

```
$ k app deploy galera-node
 [done] Updating galera-node      
 [fail] Deploying galera-node      
 [error] deploy timed out
```

However, while it was spinning the master was getting hit with requests almost every second:

```
"GET /v1/services/mytest/magento-galera-node/containers HTTP/1.1" 200 6539 0.1671
"GET /v1/services/mytest/magento-galera-node/containers HTTP/1.1" 200 6539 0.1382
"GET /v1/services/mytest/magento-galera-node/containers HTTP/1.1" 200 6539 0.1372
"GET /v1/services/mytest/magento-galera-node/containers HTTP/1.1" 200 6539 0.1378
```

Checking the containers via CLI *while the deploy was still spinning* I got:

```
$ k service containers magento-galera-node
mytest/magento-galera-node/magento-galera-node-1:
  rev: 2016-11-04 17:23:22 UTC
  node: test4.local.lan
  dns: magento-galera-node-1.mytest.kontena.local
  ip: 10.81.1.165
  public ip: 147.75.***.***
  status: running

mytest/magento-galera-node/magento-galera-node-2:
[...]

PR #1348, opened 2016-11-17
Enhanced deployment tracking

jnummelin:
This PR enhances the service deployment tracking to catch also errors. This way the deployment waiting on the cli side can be terminated and proper error shown to users.

fixes (partially) #1275

### DIFF
--- a/server/app/models/grid_service_deploy.rb
+++ b/server/app/models/grid_service_deploy.rb
@@ -1,10 +1,15 @@
 class GridServiceDeploy
   include Mongoid::Document
   include Mongoid::Timestamps
+  include Mongoid::Enum
+
+
 
   field :started_at, type: DateTime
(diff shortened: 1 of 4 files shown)


jakolehm:
@jnummelin @nevalla I think we should include this to 0.17?

jnummelin:
At least the related issue #1275 is tagged for 0.17

nevalla:
Yes we should

server/app/models/grid_service_deploy.rb
```diff
@@ -1,10 +1,15 @@
 class GridServiceDeploy
   include Mongoid::Document
   include Mongoid::Timestamps
+  include Mongoid::Enum
+
```
jakolehm: Too many empty lines.

PR #1359, opened 2016-11-21
kontena node ssh --any: default to first connected node

SpComb:
This is convenient when debugging a grid with plugin-generated names for nodes, and using `kontena node ssh ...` to run commands. You are not going to remember the node names, so you always have to first run `kontena node ls` to copy-paste some node name for `kontena node ssh`.

Instead, just run `kontena node ssh --any`.

#### `kontena node ls`
```
Name                                                                   Status     Initial    Labels                                  
billowing-sky-30                                                       online     yes        region=ams2,az=ams2,provider=digitalocean
snowy-night-25                                                         online     no         region=ams2,az=ams2,provider=digitalocean
lively-pond-83                                                         online     no         region=ams2,az=ams2,provider=digitalocean
```

#### `kontena node ssh`
```
Last login: Mon Nov 21 13:35:06 UTC 2016 from 82.181.116.101 on pts/1
CoreOS stable (1185.3.0)
core@billowing-sky-30 ~ $ 
```

### DIFF
--- a/cli/lib/kontena/cli/nodes/ssh_command.rb
+++ b/cli/lib/kontena/cli/nodes/ssh_command.rb
@@ -3,9 +3,8 @@ module Kontena::Cli::Nodes
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
 
-    parameter "NODE_ID", "Node id"
+    parameter "[NODE_ID]", "SSH to Grid node, or first connected node"
     parameter "[COMMANDS] ...", "Run command on host"
-
     option ["-i", "--identity-file"], "IDENTITY_FILE", "Path to ssh private key"
     option ["-u", "--user"], "USER", "Login as a user", default: "core"
     option "--private-ip", :flag, "Connect to node's private IP address"
@@ -15,7 +14,12 @@ module Kontena::Cli::Nodes
     requires_current_grid
 
     def execute
-      node = client.get("nodes/#{current_grid}/#{node_id}")
+      if node_id
+        node = client.get("grids/#{current_grid}/nodes/#{node_id}")
+      else
+        nodes = client.get("grids/#{current_grid}/nodes")['nodes']
+        node = nodes.select{ |node| node['connected'] }.first
+      end
 
       provider = Array(node["labels"]).find{ |l| l.start_with?('provider=')}.to_s.split('=').last
 
--- a/cli/lib/kontena/cli/nodes/ssh_command.rb
+++ b/cli/lib/kontena/cli/nodes/ssh_command.rb
@@ -3,8 +3,9 @@ module Kontena::Cli::Nodes
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
 
-    parameter "[NODE_ID]", "SSH to Grid node, or first connected node"
+    parameter "[NODE_ID]", "SSH to Grid node, or --any"
     parameter "[COMMANDS] ...", "Run command on host"
+    option ["-a", "--any"], :flag, "Connect to first connected node"
     option ["-i", "--identity-file"], "IDENTITY_FILE", "Path to ssh private key"
     option ["-u", "--user"], "USER", "Login as a user", default: "core"
     option "--private-ip", :flag, "Connect to node's private IP address"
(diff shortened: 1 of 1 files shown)


SpComb:
This probably conflicts with #1205, in that the optional `[NODE_ID]` param messes with the following optional command.

SpComb:
Rebased for git 1.1 master

SpComb:
@kke PTAL

SpComb:
I can close this if nobody likes the idea of having `kontena node ...` commands default to a random connected initial node.

kke:
I don't know why you would ssh to a random node. Maybe if you just want to be inside the overlay network. 

Perhaps `kontena node ssh --any`?

SpComb:
If you have a plugin-deployed grid then the node names are some random strings that you can never remember. When debugging, you need to execute some commands within the grid, and so you first have to `kontena node ls` to copy-paste some random node name into your `kontena node ssh ...` command.

SpComb:
Added an explicit `--any` flag to be used if omitting the node name.

cli/lib/kontena/cli/nodes/ssh_command.rb
```diff
@@ -3,9 +3,9 @@ class SshCommand < Kontena::Command
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
 
-    parameter "NODE_ID", "Node id"
+    parameter "[NODE_ID]", "SSH to Grid node, or --any"
```
kke: No need for the "or --any" IMHO.

 Also maybe it should be `[NODE_NAME]`? The fact that the id==name is a design choice the end user should not care about and "id" is usually associated with a funny string of random characters or numbers.

cli/lib/kontena/cli/nodes/ssh_command.rb
```diff
@@ -15,7 +15,14 @@ class SshCommand < Kontena::Command
...
 
     def execute
-      node = client.get("nodes/#{current_grid}/#{node_id}")
+      if node_id
+        node = client.get("grids/#{current_grid}/nodes/#{node_id}")
+      elsif any?
+        nodes = client.get("grids/#{current_grid}/nodes")['nodes']
+        node = nodes.select{ |node| node['connected'] }.first
+      else
+        exit_with_error "No host node given, nor --any"
```
kke: perhaps something like `Missing node name. Use --any to connect to the first available node`

cli/lib/kontena/cli/nodes/ssh_command.rb
SpComb: > Also maybe it should be [NODE_NAME]?

All of the existing `kontena node ...` commands use `NODE_ID`... which is indeed odd, because the "node id" is like `44C7:P5OM:NBJT:WXHV:6EDU:67T5:YDMX:4YPU:PF6D:VUH5:7LE7:`

cli/lib/kontena/cli/nodes/ssh_command.rb
SpComb: Used the same string here + for the `[NODE_ID]` parameter description

cli/lib/kontena/cli/nodes/ssh_command.rb
```diff
@@ -3,9 +3,9 @@ class SshCommand < Kontena::Command
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
 
-    parameter "[NODE_ID]", "SSH to Grid node, or --any"
+    parameter "[NODE_ID]", "SSH to Grid node. Use --any to connect to the first available node"
```
kke: It's still there, just different wording.

`!name.nil? && any?` should maybe exit_with_error "Node name and --any can't be used together"

SpComb: > It's still there, just different wording.

Yes... I reckon it's an important part of the usage and should be apparent:

```
Usage:
    kontena node ssh [OPTIONS] [NODE_ID] [COMMANDS] ...

Parameters:
    [NODE_ID]                     SSH to Grid node. Use --any to connect to the first available node
    [COMMANDS] ...                Run command on host

```

kke: Hmm, how does it work with commands anyway? Doesn't the first command get recognized as the node name when using --any?

SpComb: Yeah, it will fail now with the check for `node_id && any?`:

```
$ kontena node ssh --any echo test
 [error] Cannot combine --any with a node name
```

kke: So commands can't be used with --any

A fix for this could be checking if `name == '--'`. Then `kontena node ssh --any -- echo foo` should work.

kke: This can be done like:
```ruby
parameter '[NODE_ID]', "Foofoo" do |name|
  @name = nil if name == '--'
end
```

SpComb: That feels like it's perhaps a bit too much magic. If we really want something like that, then it should be `kontena grid ssh ...`

PR #1363, opened 2016-11-21
Refactor containers api endpoint

jakolehm:
This PR changes containers api endpoint logic a bit (because previous logic had many flaws). Now each container is always scoped in to a node within a grid. Basically this means that container id is composed with `<grid>/<node_name>/<container_name>` inside a grid. Because user might not know exact location of a single container, helper command `kontena container ls` is added. This will list all containers within a grid and shows usable container is for the user.

## Container list

```
$ kontena container ls
CONTAINER ID                       IMAGE                               COMMAND                        CREATED              STATUS    
core-01/default-nginx-77           nginx:latest                        "nginx -g daemon off;"         7 seconds ago        running   
core-01/default-nginx-32           nginx:latest                        "nginx -g daemon off;"         7 seconds ago        running   
core-01/default-nginx-83           nginx:latest                        "nginx -g daemon off;"         10 seconds ago       running   
core-01/default-nginx-33           nginx:latest                        "nginx -g daemon off;"         22 seconds ago       running   
core-01/default-nginx-75           nginx:latest                        "nginx -g daemon off;"         26 seconds ago       running   
core-01/default-nginx-36           nginx:latest                        "nginx -g daemon off;"         26 seconds ago       running   
core-01/default-nginx-78           nginx:latest                        "nginx -g daemon off;"         1 minutes ago        running 
```

## Container exec 

```
$ kontena container exec core-01/default-nginx-78 export

```


nevalla:
VPN fixes seem to be included in also.

jakolehm:
Yes, vpn uses exec so had to refactor that one also.

PR #1367, opened 2016-11-22
docs (kontena.yml) reference health checks.

matti:
kontena.yml reference is missing docs for health check. This PR is a quick fix to reference existing docs on those.


jakolehm:
Nice, thanks!

PR #1372, opened 2016-11-22
Null is the new default (stack)

jakolehm:


SpComb:
Having any possibility of ever seeing a JSON `{"stack": "null"}` just seems WTF to me.

PR #1378, opened 2016-11-22
Integrate Opto to stacks yaml parser

kke:
TODO: 
- Gemfile links to github (publish gem?)
- More testing required


kke:
To play around:

```
$ cd cli
$ bundle
$ TAG=foo MYSQL_IMAGE=bar DEBUG=true irb -r ./lib/kontena_cli.rb
> r = Kontena::Cli::Stacks::YAML::Reader.new('spec/fixtures/stack-with-prompted-variables.yml')
> r.execute
```

kke:
You may need to run `bundle update opto`

Services can now have stuff like:
```yaml
services:
   mysql:
    only_if:
      db: mysql  # match if option db has value "mysql"
   wordpress:
    skip_if: foo # skips if foo has a truthy value
   galera:
     only_if: 
       - foo
       - bar   # require both to have a truthy value
```

see `spec/fixtures/stack-with*` for inspiration.

kke:
@jakolehm @nevalla ping, tests pass now.

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -98,5 +104,45 @@ def abort_on_validation_errors(errors)
       display_notifications(errors, :red)
       abort
     end
+
+    def valid_addons(prefix=nil)
```
jakolehm: Remove.

kke:
Maybe opto should be rubygemsified before merge. Now it's in Gemfile's :development section and won't be installed if you do gem install kontena-cli --pre.

jakolehm:
At least we need to cut a gem before we make a new pre-release. I'll check opto repo rights.

kke:
Now using gem from rubygems.

PR #1399, opened 2016-11-24
Execute cloud RPC requests only in leader instance

nevalla:


server/app/services/cloud/websocket_client.rb
```diff
@@ -129,17 +130,21 @@ def on_open(event)
...
-        }
-      elsif notification_message?(data)
-        EM.defer {
-          rpc_server.handle_notification(data)
-        }
-      end
+        data = MessagePack.unpack(event.data.pack('c*'))
+        if request_message?(data)
+          EM.defer {
+            if leader?
```
jakolehm: This should be checked before we unpack event... we can do this later if there are problems with that.

PR #1404, opened 2016-11-26
Rename services api container_count attribute to instances

jakolehm:


server/app/mutations/grid_services/delete_instance.rb
```diff
@@ -0,0 +1,21 @@
+module GridServices
```
jnummelin: Is this really relevant in this PR? I couldn't find anything using this.

jakolehm: Good catch, not relevant to this PR.

Issue #1413, opened 2016-11-27
Stack variable with default and prompt does not prompt


jakolehm:
It should prompt user with the defined default value.

jnummelin:
This requires https://github.com/kontena/opto/issues/2, right?

jakolehm:
Yes, I think it will need some changes to opto.

jnummelin:
closed via #1417

PR #1417, opened 2016-11-28
Option default is now suggestion for prompts

kke:
Updated opto to use default as last resort.

Enums now have "(Other)" which prompts for manual answer if enum's `can_be_other: true` setting is enabled.

Also the cursor will be on the selection that has the option's default value in enum selection menu.


jnummelin:
This obsoletes #1410 ?

kke:
Yes and fixes #1413

PR #1419, opened 2016-11-28
Assign external id to user on master create

nevalla:


cli/lib/kontena/cli/master/users/invite_command.rb
```diff
@@ -20,10 +21,12 @@ def execute
...
         roles = []
       end
-
+      if email_list.size == 1 && self.external_id
+        external_id = self.external_id
+      end
       email_list.each do |email|
         begin
-          data = { email: email, response_type: 'invite' }
+          data = { email: email, external_id: external_id, response_type: 'invite' }
```
jnummelin: Maybe check that it's really set before assigning it into the data. Also should this be only set for the first email, in case someone invites multiple emails and sets the external-id? Now this would try to set the same external id for all emails.

kke:
LGTM but not entirely sure what's the benefit. 

The email address came from cloud so I would guess the callback thing should get the same email in auth response (they both ask it from the same endpoint on cloud).

nevalla:
The problem is that when user tries to send RPC messages to freshly created master, user object does not have external_id and RPC requests will fail. This PR will add external_id to user.

PR #1426, opened 2016-11-29
Fix vpn exec api endpoints

jakolehm:


cli/lib/kontena/cli/vpn/config_command.rb
```diff
@@ -7,7 +7,8 @@ def execute
       require 'rbconfig'
       require_api_url
       payload = {cmd: ['/usr/local/bin/ovpn_getclient', 'KONTENA_VPN_CLIENT']}
-      stdout, stderr = client(require_token).post("services/#{current_grid}/vpn/server/containers/1/exec", payload)
+      service = client(require_token).get("services/#{current_grid}/vpn/server/containers")['containers'][0]
+      stdout, stderr = client(require_token).post("containers/#{service['id']}/exec", payload)
```
jnummelin: Is this really right? Shouldn't the path be something like `containers/<node>/vpn-server-1/exec`

cli/lib/kontena/cli/vpn/create_command.rb
```diff
@@ -60,9 +60,10 @@ def execute
     def wait_for_configuration_to_finish(token)
       finished = false
       payload = {cmd: ['/usr/local/bin/ovpn_getclient', 'KONTENA_VPN_CLIENT']}
+      service = client(require_token).get("services/#{current_grid}/vpn/server/containers", payload)['containers'][0]
       until finished
         sleep 3
-        stdout, stderr = client(require_token).post("services/#{current_grid}/vpn/server/containers/1/exec", payload)
+        stdout, stderr = client(require_token).post("containers/#{service['id']}/exec", payload)
```
jnummelin: Is this really right? Shouldn't the path be something like `containers/<node>/vpn-server-1/exec`

PR #1428, opened 2016-11-29
Fix stack reg show command. 

### LINKED ISSUE #1423: kontena stack registry show --versions, :version
The `kontena stack registry show --versions` flag to only output the version numbers got clobbered in #1403 

It also ignores any `kontena stack registry show foo/bar:version` -> `stack_version`. The "latest_version" is probably a bit of a misnomer in this case.

#### `kontena stack registry show terom/redis:0.10.1`
```yaml
terom/redis:
  latest_version: 0.11.0
  expose: -
  description: This is version 0.11
  available_versions:
      ...
```

kke:
Include "semantic" for comparing/sorting stack versions.

Fixes #1423

### DIFF
--- a/cli/lib/kontena/cli/stacks/registry/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/registry/show_command.rb
@@ -13,15 +13,19 @@ module Kontena::Cli::Stacks::Registry
     requires_current_account_token
 
     def execute
-      stack = ::YAML.load(stacks_client.show(stack_name))
-      puts "#{stack['stack']}:"
-      puts "  latest_version: #{stack['version']}"
-      puts "  expose: #{stack['expose'] || '-'}"
-      puts "  description: #{stack['description'] || '-'}"
+      require 'semantic'
+      unless versions?
+        stack = ::YAML.load(stacks_client.show(stack_name, stack_version))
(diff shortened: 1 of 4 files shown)


jnummelin:
A bit broken atm, search actually does not show any stacks even if the registry really returns some stacks

cli/lib/kontena/cli/stacks/registry/show_command.rb
```diff
@@ -13,15 +13,19 @@ class ShowCommand < Kontena::Command
...
+        puts "  expose: #{stack['expose'] || '-'}"
+        puts "  description: #{stack['description'] || '-'}"
 
-      puts "  available_versions:"
-      stacks_client.versions(stack_name).map { |s| s['version']}.sort.reverse_each do |version|
-        puts "    - #{version}"
+        puts "  available_versions:"
+      end
+
+      stacks_client.versions(stack_name).reject {|s| s['version'].nil? || s['version'].empty?}.map { |s| Semantic::Version.new(s['version'])}.sort.reverse_each do |version|
```
SpComb: Empty versions should not happen anymore after https://github.com/kontena/kontena-stack-registry/pull/16

PR #1429, opened 2016-11-29
Dont query registry for every search result

kke:
Probably better to try to get the latest version into stack search response.


cli/lib/kontena/cli/stacks/registry/search_command.rb
```diff
@@ -16,8 +16,7 @@ def execute
       columns = "%-40s %-10s %-40s"
       puts columns % titles
       results.each do |stack|
-        stack = ::YAML.load(stacks_client.show(stack['name'])) rescue nil
-        puts columns % [stack['stack'], stack['version'], stack['description'] || '-'] if stack
+        puts columns % [stack['stack'] || stack['name'], stack['version'] || '?', stack['description'] || '-'] if stack
```
SpComb: The `stack['name']` should be irrelevant for the current stack registry API

PR #1432, opened 2016-11-29
Cli: log client requests/responses at DEBUG level

SpComb:
Helps a lot when figuring out what a command is actually doing:

#### `DEBUG=1 bundle exec bin/kontena stack registry remove --force test/redis`
```
D, [2016-11-29T12:20:47.655272 #6982] DEBUG -- CLIENT: DELETE /stack/test/redis: {} -> 204 No Content: ""
```

#### `DEBUG=1 kontena stack registry remove --force test/redis:1.0`
```
D, [2016-11-29T12:20:51.917360 #6989] DEBUG -- CLIENT: DELETE /stack/test/redis/version/1.0: {} -> 204 No Content: ""
```

PR #1436, opened 2016-11-29
Custom instrumentor for debugging http client requests when DEBUG=true

kke:
Less noisy than EXCON_DEBUG. Less intrusive than #1432 

image


kke:
@nevalla could be in 1.0.3 imho, only affects when DEBUG=true.

PR #1441, opened 2016-11-29
Fix legacy node put api (for plugins)

jakolehm:


server/app/routes/v1/nodes_api.rb
```diff
@@ -31,6 +30,9 @@ def load_grid_node(grid_name, node_id)
       end
 
       r.on ':grid_name/:node_id' do |grid_name, node_id|
+        validate_access_token
```
kke: validate_access_token does nothing in addition to require_current_user unless you provide scopes, such as `validate_access_token('owner')`

but doesn't break anything anyway.

kke:
LGTM

PR #1447, opened 2016-11-29
Fix ssh command to use new nodes api endpoint

jnummelin:


SpComb:
Oh, I noticed this while testing #1205, but assumed it was that PR: https://github.com/kontena/kontena/pull/1205#pullrequestreview-10566632

PR #1455, opened 2016-11-30
Resolve volumes_from dependency correctly when service has created with < 1.0.0

jakolehm:


jnummelin:
**Legacy services**
This seems to resolve volumes_from correctly with legacy services:
```
# Mimic and <1.0.0 container
$ docker create --name test-1 -l io.kontena.container.name=test-1 alpine sh
$ k service create --volumes-from test-1 web nginx
 [done] Creating web service 
$ k service deploy web
 [done] Deploying service web
```

**Stack service using legacy volume**
Now with a stack that refers to a legacy container:
```
stack: jussi/volumes
description: Volumes testing
version: 0.0.1
services:
  nginx:
    image: nginx
    volumes_from:
      - test-1

$ k stack install --deploy  ~/code/stacks/volumes.yml 
 [done] Creating stack volumes      
 [fail] Deploying stack volumes 
```

This seems to stem from the check here:
https://github.com/kontena/kontena/pull/1455/files#diff-a8f72421149035bce4bd5a08ed3782f9R33

Essentially the check makes `volumes_from` work only within a stack.

**Stack using volumes from other services within the stack**

Given with a stack:
```
stack: jussi/volumes
description: Volumes testing
version: 0.0.1

services:
  web:
    image: nginx

  nginx:
    image: nginx
    depends_on:
      - web
    volumes_from:
      - web-1
```

Deployment fails with error on master: `Cannot find applicable node for service instance test/volumes/nginx-1`

The volumes-from needs "fully qualified" container name as this works:
```
stack: jussi/volumes
description: Volumes testing
version: 0.0.1

services:
  web:
    image: nginx

  nginx:
    image: nginx
    depends_on:
      - web
    volumes_from:
      - volumes-web-1
```

Of course one can be made more usable with the fix from #1475, so one could define it like:
```
volumes_from:
  - ${stack}-web-1
```


Should `volumes_from` really work only within a stack or was it un-intended side-effect? 

Also I think it would make sense to have `volumes_from` within a stack be able to be defined only with plain service names, but that's outside of tis PR.

jakolehm:
Ok, this should work within a stack without any prefixed names... will fix it. And using volumes_from across stacks, what would be the use case for those? It somehow breaks the namespace/isolation if we can "mount" everything/everywhere.

jnummelin:
I was mainly thinking of running some "ad-hoc" services which could mount random volumes. One specific use case I was thinking is DB backups but that should propably be part of a given DB stack. I'm totally fine for being able to do `volumes-from` only within a stack, just wanted to make sure that this was really intended that way. :)

PR #1458, opened 2016-12-01
Say role not found instead of role can not be nil in role add

### LINKED ISSUE #1457: CLI: kontena master users role add master_adminTYPO says "Role can't be nil"
The error message is somewhat confusing if role name has typo in it

kke:
Fixes #1457

### DIFF
--- a/server/app/mutations/users/add_role.rb
+++ b/server/app/mutations/users/add_role.rb
@@ -1,7 +1,7 @@
 module Users
   class AddRole < Mutations::Command
 
-    attr_accessor :role_instance
+    attr_reader :role_instance
 
     required do
       model :current_user, class: User
@@ -9,14 +9,10 @@ module Users
       string :role
     end
 
-    optional do
-      model :role_instance, class: Role
-    end
-
     def validate
-      self.role_instance = Role.find_by(name: role)
+      @role_instance = Role.find_by(name: role)
 
-      unless role_instance
+      if role_instance.nil?
         add_error(:role, :not_found, "Role '#{role}' not found")
         return false
       end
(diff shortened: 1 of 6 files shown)


server/app/routes/v1/users/user_roles.rb
```diff
@@ -4,18 +4,23 @@
...
+          response.status = 201
+          render('users/show')
+        else
+          response.status = 422
+          {error: outcome.errors.message}
+        end
       else
         response.status = 422
-        {error: outcome.errors.message}
+        {error: 'Role not found'}
```
SpComb: This should be something like `{error: {role: "Not Found"}}` to match the Mutations `errors.messages` structure.

server/app/routes/v1/users/user_roles.rb
```diff
@@ -4,18 +4,23 @@
...
-      options = {
-        current_user: current_user,
-        user: @user,
-        role: role
-      }
-      outcome = Users::AddRole.run(options)
-      if outcome.success?
-        response.status = 201
-        render('users/show')
+      if role
```
jakolehm: Imho this validation should be done inside the mutation?

kke: Moved the role searching by name inside the mutations.

server/app/mutations/users/add_role.rb
```diff
@@ -1,24 +1,34 @@
...
+
     required do
       model :current_user, class: User
       model :user
-      model :role
+      string :role
+    end
+
+    optional do
+      model :role_instance, class: Role
```
jakolehm: Do we really need this because `role` is passed as a string?

jakolehm: @kke ping

kke: No

PR #1463, opened 2016-12-01
fix (docker-compose) cadvisor requires a custom image.

matti:


jakolehm:
This is only needed when using Docker for Mac (or Alpine Linux).

jnummelin:
Actually not needed anymore as we switched back to official cAdvisor image in 1.0.3. see #1569

PR #996, opened 2016-09-09
cli: display agent version in node list

christofferh:
```
Name                                                                   Version    Status     Initial
5.kontena.internal                                                     0.15.1     online     yes
4.kontena.internal                                                     0.15.2     online     yes
3.kontena.internal                                                     0.15.0     online     yes
2.kontena.internal                                                     0.15.2     online     yes
1.kontena.internal                                                     0.15.1     online     yes
```

ping @matti

PR #1468, opened 2016-12-01
CLI: Show grid initial nodes health

SpComb:
With the Kontena IPAM based on etcd, you will not be able to deploy services to a grid with fewer initial nodes created/connected than required for an etcd quorum.

This PR adds grid nodes configuration health checks to the CLI.

* Show the configured `initial_size` in `kontena grid current`  / `kontena grid show`

* Improve `kontena node ls` to show the initial node states

* Add `kontena node health` to show the connected state for a given node

* Add `kontena grid health` to show disconnected and missing initial nodes

See  #1470 for server/agent health checks

## Example

Creating a new  `--initial-size=3` grid one node at a time.

Starting from an empty grid, the `kontena node ls` is empty, and `kontena grid health` is red:

step0a
step0b

Adding the first node, the `kontena node ls` and `kontena grid health` are red:

step1

Adding the second node, the `kontena node ls` and `kontena grid health` are yellow:

step2

Adding the third node, the `kontena node ls` and `kontena grid health` are green:

step3

Adding the fourth node, everything is green:

step4

If there are fewer initial nodes connected than configured for the grid, but still a sufficient majority for etcd quorum, show a warning health status as etcd operations will still succeed:

step5

If there are fewer initial nodes connected than required for etcd quorum, then show an error health status as etcd operations will fail:

step6


SpComb:
This should probably be included in 1.0.x, because it's a CLI-only fix, and it's really easy to forget what the initial-nodes configuration/state for the grid actually is.

The agent-side stuff in #1470 could then be for 1.1.0

SpComb:
Improved the `kontena grid health` messages; CBA to update the screenshots; from the specs

* Grid does not have any initial nodes, and requires at least 1 of 1 nodes for operation
* Grid only has 1 initial nodes, and is not high-availability
* Grid only has 1 initial nodes, and is not high-availability
* Grid has all 1 of 1 initial nodes connected
* Grid only has 1 of 2 initial nodes required for operation
* Grid only has 2 initial nodes, and is not high-availability
* Grid has all 2 of 2 initial nodes connected
* Grid only has 1 of 2 initial nodes required for operation
* Grid only has 2 of 3 initial nodes required for high-availability
* Grid only has 2 of 3 initial nodes required for high-availability
* Grid only has 2 of 3 initial nodes required for high-availability
* Grid has all 3 of 3 initial nodes connected

SpComb:
Squashed onto git `1.1.dev` master.

SpComb:
Ruby <2.3 rspecs failing on `<<~END` syntax.

SpComb:
Added some specs for the `kontena node ls` show functions.

SpComb:
PTAL @jakolehm @jnummelin

jakolehm:
Where did that version column come? :)

SpComb:
> Where did that version column come? :)

 #996

cli/lib/kontena/cli/grids/health_command.rb
```diff
@@ -0,0 +1,69 @@
...
+    def show_grid_health(grid, nodes)
+      initial_size = grid['initial_size']
+      minimum_size = grid['initial_size'] / 2 + 1 # a majority is required for etcd quorum
+
+      grid_health = grid_health(grid, nodes)
+      initial_nodes = nodes.select{|node| node['initial_member']}
+      online_nodes = initial_nodes.select{|node| node['connected']}
+
+      # configuration and status
+      if initial_nodes.length == 0
```
jakolehm: Maybe we could refactor this if/elsif thingie into separate method?

SpComb: This is the separate method that it's refactored into :P

kke: Well, each of the conditions could be a separate method that would return or set the error message and do nothing or return true when the validation passes and then you would call them in order.

Like:
```ruby
def validate_foobar
  return true if nodes.foo > bars
  add_error "Foo needs to be more than bars"
  false
end
```

SpComb: The logic here is that it only ever logs one thing, and it's a progression from error -> warning -> ok

PR #1475, opened 2016-12-01
Set GRID and STACK environment variables for stack files

nevalla:


jnummelin:
These envs are not available for the variables so one cannot do something like this:
```
variables:
  node_password:
    type: string
    from:
      vault: ${STACK}_REDIS_SENTINEL_PASSWORD
      prompt: Redis password
    to:
      vault: ${STACK}_REDIS_SENTINEL_PASSWORD
```
Instead it gets the `filler` prefix.

Not sure if this should be tackled in this PR or another.

jakolehm:
LGTM

cli/lib/kontena/cli/stacks/yaml/service_extender.rb
```diff
@@ -34,12 +34,15 @@ def extend_from(from)
...
-        to   ||= []
-        from ||= []
-        to_hash   = Hash[*to.flat_map {|t| t.split('=') }]
-        from_hash = Hash[*from.flat_map {|f| f.split('=') }]
-
-        from_hash.merge(to_hash).map {|k,v| "#{k}=#{v}"}
+        env_vars = to || []
+        if from
+          from.each do |env|
+            env_vars << env unless to && to.find do |key|
```
kke: This monster came back.

To handle stuff with only variable, but no value, it could be:
```
  to_hash   = Hash[*to.flat_map {|t| key, value = t.split('=',2); [key, value] }]
```

from the previous version.

kke:
Opto could now perform its own interpolation and simple calculations like `($master_instances / 2) + 1`   https://github.com/kontena/opto/pull/9

PR #1477, opened 2016-12-02
Interpolate STACK and GRID inside variable definitions

### LINKED ISSUE #1473: stacks yml variables result always in value "filler"
Given following:

```
variables:
  node_password:
    type: string
    from:
      vault: ${PROJECT}_REDIS_SENTINEL_PASSWORD
      prompt: Redis password
    to:
      vault: ${PROJECT}_REDIS_SENTINEL_PASSWORD
```

and command `kontena stack install -n myname stackfile.yml` results says:

```
> Redis password : asdf
Value for PROJECT is not set. Substituting with an empty string.
Value for PROJECT is not set. Substituting with an empty string.
```

And in kontena vault key is:

```
filler_REDIS_SENTINEL_PASSWORD
```

So it's not "empty" and shouldn't be: I think that the install / deploy should be aborted.

kke:
This is possible now:

```
variables:
  foo:
    type: string
    from:
      vault: ${STACK}-foo-password
```

These two envs are always set before yaml parsing by the CLI in #1475

Fixes #1473

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -236,7 +236,7 @@ module Kontena::Cli::Stacks
             if val
               val
             else
-              puts "Value for #{var} is not set. Substituting with an empty string." unless skip_validation?
+              puts "Value for #{var} is not set. Substituting with #{@replace_missing.to_s == '' ? "an empty string" : "'#{@replace_missing}'"}." unless skip_validation?
               @replace_missing.to_s
             end
           end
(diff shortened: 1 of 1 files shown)


kke:
Hold on

SpComb:
Needs rebase

kke:
Needs close I think. The other PR interpolates all ENVs into variables.

SpComb:
git master already seems to interpolate `${STACK}` within the variables? Unfortunately, it only interpolates the `stack: ...` name from the YAML, which is almost useless. It should interpolate the `kontena stack install --name ...` :( #1504

SpComb:
Closing. See #1475 + #1505

PR #1482, opened 2016-12-05
Deploy stack services one-by-one

jakolehm:
Fixes stack deployments so that stack services are deployed one-by-one and with correct order.

### DIFF
--- a/server/app/jobs/stack_deploy_worker.rb
+++ b/server/app/jobs/stack_deploy_worker.rb
@@ -1,0 +1,49 @@
+class StackDeployWorker
+  include Celluloid
+  include Logging
+  include Stacks::SortHelper
+
+  def perform(stack_deploy_id)
+    stack_deploy = StackDeploy.find_by(id: stack_deploy_id)
+    if stack_deploy
+      deploy_stack(stack_deploy)
+    end
+  end
+
+  def deploy_stack(stack_deploy)
+    stack = stack_deploy.stack
+    stack_deploy.ongoing!
+
+    services = sort_services(stack.grid_services.to_a)
+    services.each do |service|
+      deploy_service(service, stack_deploy)
+    end
+    stack_deploy.success!
+
+    stack_deploy
+  rescue
+    stack_deploy.error!
+  end
+
+  # @param [GridService] service
+  # @param [StackDeploy] stack_deploy
+  def deploy_service(service, stack_deploy)
+    outcome = GridServices::Deploy.run(grid_service: service)
+    unless outcome.success?
+      error "failed to deploy #{service.to_path}"
+    else
+      service_deploy = outcome.result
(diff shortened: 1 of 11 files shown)


jakolehm:
@jnummelin ping

server/app/jobs/stack_deploy_worker.rb
```diff
@@ -0,0 +1,76 @@
...
+    remove_services(stack, stack_rev)
+    services = sort_services(stack.grid_services.to_a)
+    services.each do |service|
+      deploy_service(service, stack_deploy)
+    end
+    stack_deploy.success!
+
+    stack_deploy
+  rescue
+    stack_deploy.error!
```
jnummelin: Could we somehow get the possibile service deployment error messages appended here?

server/app/jobs/stack_deploy_worker.rb
```diff
@@ -0,0 +1,76 @@
...
+  # @param [StackDeploy] stack_deploy
+  # @param [StackRevision] stack_rev
+  def deploy_stack(stack_deploy, stack_rev)
+    stack = stack_deploy.stack
+    stack_deploy.ongoing!
+
+    remove_services(stack, stack_rev)
+    services = sort_services(stack.grid_services.to_a)
+    services.each do |service|
+      deploy_service(service, stack_deploy)
```
jnummelin: Just thinking, but should the failure in one service cause others not to be deployed?

jakolehm:
@SpComb also ping :)

server/app/jobs/stack_deploy_worker.rb
```diff
@@ -0,0 +1,76 @@
...
+            info "waiting for #{service.to_path} deploy to finish..."
+            sleep 1
+            if service_deploy.reload && (service_deploy.success? || service_deploy.error?)
+              finished = true
+            end
+          end
+        end
+      rescue => exc
+        error "#{exc.class.name}: #{exc.message}"
+      end
```
jnummelin: Maybe return the deployment object to be able to collect failure reasons at stack level.

server/app/models/stack_deploy.rb
```diff
@@ -1,9 +1,12 @@
 class StackDeploy
   include Mongoid::Document
   include Mongoid::Timestamps
+  include Mongoid::Enum
 
   index({ stack_id: 1 })
 
+  enum :deploy_state, [:created, :ongoing, :success, :error], default: :created
+
```
jnummelin: Add also a `reason`field as the CLI is expecting that on `wait_deploy_to_finish`

jakolehm:
Stack deploy will now halt when first service errors. And reason is displayed to the user:

```
$ kontena stack install --deploy test.yml
 [done] Creating stack test      
 [fail] Deploying stack test     
        Cannot find applicable node for service instance test/test/nginx-1
```

PR #1483, opened 2016-12-05
Garbage collect orphan service containers

jakolehm:
There is possibility to remove a service so that nodes still have service containers in place and they miss a remove. This PR fixes these situations by doing a garbage collection for service containers that still exist in the master db and their service definition is gone.

### DIFF
--- a/server/app/jobs/container_cleanup_job.rb
+++ b/server/app/jobs/container_cleanup_job.rb
@@ -12,12 +12,12 @@ class ContainerCleanupJob
 
   def perform
     info 'starting to cleanup stale containers'
-    sleep 0.1
     loop do
+      sleep 1.minute.to_i
       if leader?
         destroy_deleted_containers
+        terminate_ghost_containers
       end
-      sleep 1.minute.to_i
     end
   end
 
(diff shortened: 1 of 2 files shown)


server/app/jobs/container_cleanup_job.rb
```diff
@@ -26,4 +26,28 @@ def destroy_deleted_containers
...
+    end
+  end
+
+  # @param [Docker::Container] container
+  def terminate_ghost_container(container)
+    if container.host_node
+      info "terminating #{container.to_path}"
+      terminator = Docker::ServiceTerminator.new(container.host_node)
+      terminator.request_terminate_service(
+        container.grid_service_id.to_s, container.instance_number, {lb: true}
```
SpComb: Thinking about the `{lb: true}`, which will look at the container labels on the agent, and recursively remove the etcd entries for the service if the container is `instance_number=1`....

Could this ghost container be e.g. a leftover container from an earlier service deployment, such that the container's labeled service is still active with other containers, and this would cause their LB configs to be wiped out?

jakolehm: Yes, that is a possibility... maybe we should detect it somehow.

jnummelin: I think we'll hit similar case on the container level also. Say you have re-created a service `web` (new service from master point-of-view) and you terminate orphaned instance. That will also now remove the service instance (`web-1`) from the LB since it acts only based on names.

jakolehm: This GC runs once a minute.. so this disaster to happen user must first remove a service so that it will leave containers to nodes and then spin up a new service that has exactly the same name? While testing this might happen but with more serious workloads.. not so sure?

jnummelin: Maybe won't happen too often on more serious workloads but when it does causes downtime of services for users.

jakolehm: Ok, so ultimate solution is to check if service with the same name/stack exists. If it does and it has service instances -> do not remove from the lb?

SpComb:
LGTM in principle, but I'm having difficulty actually testing this, in terms of reproducing the scenario where this GC code kicks in.

jakolehm:
@SpComb you can test this easily if you stop agent in one of the nodes and then remove stack/service. After service is gone, start agent and there should be a zombie to catch.

SpComb:
> @SpComb you can test this easily if you stop agent in one of the nodes and then remove stack/service. After service is gone, start agent and there should be a zombie to catch.

I tried that and didn't see any GC... maybe I need to add more debug logging to see if I'm waiting long enough.

jnummelin:
Smoke tested with one service (and one instance) and works. I can see the termination messages in both server and agent logs and container is removed as expected.
```
api_1      | I, [2016-12-07T10:24:15.586346 #8]  INFO -- ContainerCleanupJob: terminating test/ipam/null-nginx-1

kontena-agent | I, [2016-12-07T10:24:15.605628 #1]  INFO -- Kontena::ServicePods::Terminator: terminating service: /null-nginx-1
```

PR #1493, opened 2016-12-07
Document the variable definition syntax

### LINKED ISSUE #1471: document stack file in references
Currently stacks documentation only mentions `expose:` key and does not mention anything about `variables:`

kke:
Fixes #1471

### DIFF
--- a/docs/references/kontena-yml-variables.md
+++ b/docs/references/kontena-yml-variables.md
@@ -11,9 +11,9 @@ Since Kontena version 1.0.0 and the introduction of [Stacks](../using-kontena/st
 
 ```
 ---
-stack: user/wordpress
+stack: user/mysql
 version: 1.0.0
-description: Wordpress with an optional database
+description: A MySQL server
 variables:
   mysql_root_pw: # variable name
     type: string  # type (string, integer, boolean, uri, enum)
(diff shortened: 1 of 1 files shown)


docs/references/kontena-yml-variables.md
```diff
@@ -0,0 +1,395 @@
...
+
+Define what to do with the value. The default behavior is to set it to local environment using the variable name. You can also use another environment variable name or write the value to the Vault on Kontena Master.
+
+```
+  to:
+    env: MYSQL_USER # set to local env
+    vault: foo-mysql-user # also send to vault
+```
+
+### Only_if
```
SpComb: These look a bit odd, maybe it would be better to just use fixed-width for these titles

#### `only_if`

kke:
Hmm yes maybe interpolator should escape the output so that it gets through yaml parsing.

jakolehm:
Markdown seems to be broken (at least when viewed with github)?

docs/references/kontena-yml-variables.md
```diff
@@ -1,19 +1,17 @@
 ---
 title: kontena.yml variables
-toc_order: 2
----
```
SpComb: Lost the YAML header

SpComb:
Replaced by #1531, include the new opto 1.5.3 docs in #1528 ?

nevalla:
@kke needs rebase

PR #1519, opened 2016-12-09
Moved some log messages to debug level

### LINKED ISSUE #1518: Move agent log messages to debug level
From slack:

```
dsmiley [9:16 PM]  
It'd be nice of kontena agent's INFO log messages for "starting collection" and "publishing node information" moved to a DEBUG level.
```

kke:
Fixes #1518


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -24,7 +24,7 @@ def initialize(queue, autostart = true)
       @statsd = nil
       subscribe('websocket:connected', :on_websocket_connected)
       subscribe('agent:node_info', :on_node_info)
-      info 'initialized'
+      debug 'initialized'
```
SpComb: This should stay as `info`, it only happens once at startup and then again after a worker crash.

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -13,7 +13,7 @@ def initialize(queue, autostart = true)
       @queue = queue
       @statsd = nil
       @node_name = nil
-      info 'initialized'
+      debug 'initialized'
```
SpComb: This should stay as `info`, it only happens once at startup and then again after a worker crash.

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -34,9 +34,9 @@ def on_node_info(topic, info)
     end
 
     def start
-      info 'waiting for cadvisor'
+      debug 'waiting for cadvisor'
       sleep 1 until cadvisor_running?
```
SpComb: Not necessary for this PR, but these debug+sleep should be refactored to `wait(message: "waiting for cadvisor to be running") { cadvisor_running? }`

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -24,7 +24,7 @@ def on_node_info(topic, info)
       @node_name = info['name']
       statsd_conf = info.dig('grid', 'stats', 'statsd')
       if statsd_conf
-        info "exporting stats via statsd to udp://#{statsd_conf['server']}:#{statsd_conf['port']}"
+        debug "exporting stats via statsd to udp://#{statsd_conf['server']}:#{statsd_conf['port']}"
```
jakolehm: Imho this is valuable information.. and it is only logged when agent starts or when statsd configuration changes (does not spam logs).

agent/lib/kontena/workers/stats_worker.rb
jakolehm: Imho this is valuable information.. and it is only logged when agent starts (does not spam logs).

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -34,9 +34,9 @@ def on_node_info(topic, info)
     end
 
     def start
-      info 'waiting for cadvisor'
+      debug 'waiting for cadvisor'
       sleep 1 until cadvisor_running?
-      info 'cadvisor is running, starting stats loop'
+      debug 'cadvisor is running, starting stats loop'
```
jakolehm: Imho this is valuable information.. and it is only logged when agent starts (does not spam logs).

kke:
Changed some messages back to INFO. Not sure if this PR does anything anymore.

jakolehm:
@kke it does remove recurring spam.

PR #1522, opened 2016-12-09
Fix ServiceBalancer greediness

jakolehm:
This PR fixes a situation where stack deploy is in progress and ServiceBalancerJob decides to fix stack services because desired state does not match. Now balancer checks if stack has a deploy in progress and does not touch services until deploy has finished.


jakolehm:
greedy

PR #1536, opened 2016-12-10
Fix VPN create

nevalla:
This PR will fix VPN create command. Past VPN failed to execute command on VPN container because VPN container was not yet registered to Master. With this PR cli waits until container is present and continues execution after that.

### DIFF
--- a/cli/lib/kontena/cli/vpn/create_command.rb
+++ b/cli/lib/kontena/cli/vpn/create_command.rb
@@ -45,25 +45,36 @@ module Kontena::Cli::Vpn
 
       client(token).post("grids/#{current_grid}/stacks", data)
       deployment = client(token).post("stacks/#{current_grid}/#{name}/deploy", {})
+      container = nil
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/vpn/create_command.rb
```diff
@@ -45,25 +45,36 @@ def execute
 
       client(token).post("grids/#{current_grid}/stacks", data)
       deployment = client(token).post("stacks/#{current_grid}/#{name}/deploy", {})
+      container = nil
       spinner "Deploying #{pastel.cyan(name)} service " do
         deployment['service_deploys'].each do |service_deploy|
           wait_for_deploy_to_finish(token, service_deploy)
         end
+        container = wait_for_container
```
jakolehm: Shouldn't `wait_for_deploy_to_finish` ensure that container exists?

jakolehm:
Imho this is not the right way to fix this bug... I'll push an alternative fix soon.

nevalla:
Closing in favor of #1537

PR #1537, opened 2016-12-10
Fix registry & vpn stack deploy tracking

jakolehm:
Kontena cli 1.0.1 has a regression that causes cli to error on vpn/registry deploy tracking. This PR fixes these error by tracking a stack deployment object (was service deployment object).

Replaces #1536

### DIFF
--- a/cli/lib/kontena/cli/stacks/stacks_helper.rb
+++ b/cli/lib/kontena/cli/stacks/stacks_helper.rb
@@ -1,12 +1,11 @@
 
 module Kontena::Cli::Stacks
   module StacksHelper
-    
+
     # @param [Hash] deployment
     # @return [Boolean]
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/stacks/stacks_helper.rb
```diff
@@ -0,0 +1,33 @@
+
+module Kontena::Cli::Stacks
+  module StacksHelper
+    
+    # @param [Hash] deployment
+    # @return [Boolean]
+    def wait_for_deploy_to_finish(deployment, timeout = 600)
+      deployed = false
+      progress = []
```
nevalla: This seems to be redundant

jakolehm: Fixed.

PR #1540, opened 2016-12-11
Add stack commands to tab completer

nevalla:
This PR add stack related commands to tab completer. Also service completer returns services in new format (stack/service).

### DIFF
--- a/cli/lib/kontena/scripts/completer
+++ b/cli/lib/kontena/scripts/completer
@@ -223,8 +223,10 @@ if words.size > 0
         if words[1] == 'registry'
           registry_sub_commands = %(push pull search show rm)
           completion.push registry_sub_commands
-        elsif %w(install upgrade).include?(words[1])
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/scripts/completer
```diff
@@ -197,6 +215,23 @@ if words.size > 0
...
       end
+    when 'stack'
+      completion.clear
+      sub_commands = %w(build install upgrade deploy start stop remove rm ls list
+                        logs monitor show registry)
+      if words[1]
+        if words[1] == 'registry'
+          registry_sub_commands = %(push pull search show rm)
+          completion.push registry_sub_commands
+        elsif words[1] == 'install'
```
jakolehm: Should this also include upgrade?

nevalla:
Actually stack upgrade tab complete needs still some fine tuning because first parameter is stack name and yaml file is the second parameter.

PR #1543, opened 2016-12-12
Move debug output to STDERR

### LINKED ISSUE #1542: Debug info should go to stderr
```
$ DEBUG=true kontena vpn config > foo.ovpn
$ head foo.ovpn
D, [2016-12-12T10:30:25.756673 #51676] DEBUG -- CONFIG: Loading configuration from /Users/kimmo/.kontena_client.json
D, [2016-12-12T10:30:25.758316 #51676] DEBUG -- CONFIG: Configuration loaded with 18 servers.
D, [2016-12-12T10:30:25.758356 #51676] DEBUG -- CONFIG: Current master: dry-dream-74
...
```

kke:
Fixes #1542

### DIFF
--- a/cli/lib/kontena/cli/common.rb
+++ b/cli/lib/kontena/cli/common.rb
@@ -12,7 +12,7 @@ module Kontena
 
       def logger
         return @logger if @logger
-        @logger = Logger.new(STDOUT)
+        @logger = Logger.new(ENV["DEBUG"] ? STDERR : STDOUT)
         @logger.level = ENV["DEBUG"].nil? ? Logger::INFO : Logger::DEBUG
         @logger.progname = 'COMMON'
         @logger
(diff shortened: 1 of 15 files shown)


cli/lib/kontena/cli/common.rb
```diff
@@ -12,7 +12,7 @@ module Common
 
       def logger
         return @logger if @logger
-        @logger = Logger.new(STDOUT)
+        @logger = Logger.new(ENV["DEBUG"] ? STDERR : STDOUT)
```
SpComb: Should the Logger just always use `STDERR`?

SpComb:
Could all of the `ENV["DEBUG"] && STDERR.puts(...)` be replaced with `logger.debug(...)` (or `logger.debug { ... }` to keep the lazy evaluation)?

cli/lib/kontena/client.rb
```diff
@@ -46,7 +46,7 @@ def initialize(api_url, token = nil, options = {})
       uri = URI.parse(@api_url)
       @host = uri.host
 
-      @logger = Logger.new(STDOUT)
+      @logger = Logger.new(ENV["DEBUG"] ? STDERR : STDOUT)
```
SpComb: At least here this can just be `STDERR`, since there are only `logger.debug ...` calls in the class?

kke: All the logging should be unified anyways at some point.

timolehto:
Just a thought, but couldn't you just provide an option for outputting the configs to a file instead of stdout i.e. something like: `kontena vpn config -f kontena.vpn.config`?

kke:
This will also help with piping to other commands when debug=true.

timolehto:
True

PR #1550, opened 2016-12-13
Fix agent to avoid DNS lookups for localhost with a missing /etc/hosts

SpComb:
Apparently some versions/installs of CoreOS do not have any `/etc/hosts` file, which leads to an empty `/etc/hosts` file in any Docker `--net=host` container. This may leave the `kontena-ipam-plugin` unable to resolve the etcd `localhost` name.

Fix this by using `127.0.0.1` instead, both for the `kontena-ipam` `ETCD_ENDPOINT` as well as the agent's `IpamClient` connections.

### DIFF
--- a/agent/lib/kontena/launchers/ipam_plugin.rb
+++ b/agent/lib/kontena/launchers/ipam_plugin.rb
@@ -83,7 +83,8 @@ module Kontena::Launchers
         'Cmd' => ["bundle", "exec", "thin", "-a", "127.0.0.1", "-p", "2275", "-e", "production", "start"],
         'Env' => [
           "NODE_ID=#{info['node_number']}",
-          "LOG_LEVEL=#{ENV['LOG_LEVEL'] || 1}"
+          "LOG_LEVEL=#{ENV['LOG_LEVEL'] || 1}",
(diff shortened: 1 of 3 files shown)


SpComb:
It seems like with an empty `/etc/hosts` on CoreOS, processes will actually query the external DNS resolver for the localhost name, so it may or may not work depending on the external DNS resolver..

```
10:01:28.104559 IP core-01.40070 > 10.0.2.3.domain: 2618+ A? localhost. (27)
10:01:28.105253 IP 10.0.2.3.domain > core-01.40070: 2618 1/0/0 A 127.0.0.1 (43)
10:01:28.105954 IP core-01.40070 > 10.0.2.3.domain: 2805+ AAAA? localhost. (27)
10:01:28.106336 IP 10.0.2.3.domain > core-01.40070: 2805 NotImp 0/0/0 (27)
```

SpComb:
Tested to no longer send out any DNS queries for `localhost` on agent startup with a missing `/etc/hosts`.

SpComb:
This may affect CoreOS versions earlier than [1081.1.0](https://coreos.com/releases/#1081.1.0)

> Populate /etc/hosts with an entry for localhost when the file is absent

Issue #1450, opened 2016-11-29
Current stable Ruby is 2.3.3


kke:
Add (or switch 2.3.1) to 2.3.3 to travis.

Also upgrade Ruby versions on agent & master images.

Or wait for 2.4.0.

SpComb:
Travis seems to use 2.3.3 now?

Close?

SpComb:
The server server and agent images are also Ruby 2.3.3:

```
$ docker run --rm kontena/agent:1.1.2 ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux-musl]
$ docker run --rm kontena/server:1.1.2 ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux-musl]
```

PR #1290, opened 2016-11-06
[WIP] Update Mongoid to 5.1.x

jakolehm:
https://docs.mongodb.com/ruby-driver/master/tutorials/5.1.0/mongoid-upgrade/

PR #1558, opened 2016-12-14
Add ruby 2.4.0 to test matrix

jakolehm:


kke:
Ref #1450

jakolehm:
It seems that ruby 2.4 requires `json ~> 2.0` .. and that requires `activesupport ~> 5.0`. Which is doable for the agent but updating activesupport in the server requires also latest version of mongoid (see #1290). 

Don't know how this affects cli.. maybe it will "just work" as specs are indicating.

kke:
Packet plugin requires activesupport.

PR #1560, opened 2016-12-14
Use liquid templating for stack ymls

kke:
```yaml
---
stack: user/foo
variables:
  foo:
    type: string
    value: hello

services:
  bar:
    environment:
      # {% for num in (0...10) %}
      - FOO{{num}}={{foo}}/{{forloop.index}}
      # {% endfor %}
```

```ruby
# $ irb -r ./lib/kontena_cli.rb
>> Kontena::Cli::Stacks::YAML::Reader.new('./foo.yml').execute[:services]["BAR"]["environment"]
=> ["FOO0=hello/1", "FOO1=hello/2", "FOO2=hello/3", "FOO3=hello/4", "FOO4=hello/5", "FOO5=hello/6", "FOO6=hello/7", "FOO7=hello/8", "FOO8=hello/9", "FOO9=hello/10", "FOO10=hello/11"]
```

Current implementation status:

1. Variables section is read independently, liquid sees ENV
2. Rest of the YAML can use the variables inside liquid

TODO:

- [ ] 1. Some magic for things like `{{ service.mysql.hostnames | join ',' }}`
- [ ] 2. What should happen if a service is scaled using `kontena service scale`?
- [ ] 3. The reader is messy, complete rewrite wouldn't hurt (cleaned up a bit now)
- [X] 4. It's a bit weird that you have a different set of variables in different sections of the YAML. (there's not much that can be done about that)
- [ ] 5. Unknown variables will output an empty string, even with the "strict_variables: true" setting in liquid.
- [X] 6. There's no way to debug the interpolation/parsing output from the CLI.

### DIFF
--- a/cli/lib/kontena/cli/stacks/validate_command.rb
+++ b/cli/lib/kontena/cli/stacks/validate_command.rb
@@ -1,0 +1,47 @@
+require_relative 'common'
+
+module Kontena::Cli::Stacks
+  class ValidateCommand < Kontena::Command
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+    include Common
+
+    banner "Validates a YAML file"
+
+    include Common::StackFileOrNameParam
+    include Common::StackNameOption
+
+    option '--values-to', '[FILE]', 'Output variable values as YAML to file'
+
+    include Common::StackValuesFromOption
+
+    requires_current_master # the stack may use a vault resolver
+    requires_current_master_token
+
+    def execute
+
+      if !File.exist?(filename) && filename =~ /\A[a-zA-Z0-9\_\.\-]+\/[a-zA-Z0-9\_\.\-]+(?::.*)?\z/
+        from_registry = true
+      else
+        from_registry = false
+        require_config_file(filename)
+      end
+
+      reader = reader_from_yaml(filename, from_registry: from_registry, name: name, values: values)
+      outcome = reader.execute
+      hint_on_validation_notifications(outcome[:notifications]) if outcome[:notifications].size > 0
+      abort_on_validation_errors(outcome[:errors]) if outcome[:errors].size > 0
+
+      if values_to
+        vals = reader.variables.to_h(values_only: true).reject {|k,_| k == 'STACK' || k == 'GRID' }
+        File.write(values_to, ::YAML.dump(vals))
+      end
+      result = reader.fully_interpolated_yaml.merge(
+        # simplest way to stringify keys in a hash
+        'variables' => JSON.parse(reader.variables.to_h(with_values: true, with_errors: true).to_json)
+      )
+      puts ::YAML.dump(result)
+    end
+  end
+end
+
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -6,7 +6,7 @@ module Kontena::Cli::Stacks
       include Kontena::Util
       include Kontena::Cli::Common
 
-      attr_reader :file, :raw_content, :result, :errors, :notifications, :variables, :yaml, :sections
+      attr_reader :file, :raw_content, :result, :errors, :notifications, :variables, :yaml
 
       def initialize(file, skip_validation: false, skip_variables: false, replace_missing: nil, from_registry: false)
         require 'yaml'
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -126,9 +126,18 @@ module Kontena::Cli::Stacks
 
       def load_yaml(interpolate = true)
         if interpolate
-          @yaml = ::YAML.safe_load(replace_dollar_dollars(interpolate(interpolate_liquid(raw_content, variables.to_h(values_only: true)))))
+          @yaml = ::YAML.safe_load(
+            replace_dollar_dollars(
+              interpolate(
+                interpolate_liquid(
+                  raw_content,
+                  variables.to_h(values_only: true)
+                )
+              )
+            )
+          )
         else
-          @yaml = ::YAML.safe_load(interpolate_liquid(raw_content, ENV.to_h))
+          @yaml = ::YAML.safe_load(raw_content, ENV.to_h)
         end
       rescue Psych::SyntaxError => e
         raise "Error while parsing #{file}".colorize(:red)+ " "+e.message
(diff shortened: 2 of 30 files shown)


jakolehm:
Excellent! We need to add specs for this (and also test invalid liquid syntax).

kke:
https://github.com/Shopify/liquid/pull/832

Doesn't work with ruby 2.0 anymore.

kke:
# Use cases

Just thinking what you would actually use it for.

## Create conditional services

Already possible using the opto `skip_if` / `only_if` syntax

## Iterate to generate multiple secrets

Opto could have something like:

```yaml
variables:
  host_key:
    type: string
    instances: 5
    from:
      vault: ${STACK}-host-key-${INSTANCE_NUM}
      random_string: 128
    to:
      vault: ${STACK}-host-key-${INSTANCE_NUM}

services:
  foodb:
    instances: 5
    secrets:
      - secret: ${STACK}-host-key-%{INSTANCE_NUM} # server would have to eval % syntax
         name: HOST_KEY
         type: instance_env
```

## Join instance hostnames

```yaml
environment:
   UNICAST_HOSTS: {{ hostnames | join ',' }}
```

This could be done without liquid if such magical hostname variable (or opto resolver) was implemented.

This however would be difficult without liquid:

```yaml
environment:
  # {% for instance in (1..foosvc_instances) %}
  - "UNICAST_HOST{{ instance }}=foosvc-{{instance}}"
  # {% endfor %}
```

If something requires such syntax, you could create a script in the container that would split a comma separated variable (or performs a dns lookup)

## Iterate to generate hooks

This will (or already is) break service scaling unless some kind of instance hooks are created anyway. For something like building a fixed size mongo replica set this can work, but to me it doesn't seem like a very flexible practice.

## Other

I haven't run into or can't think of any other use cases. Any ideas?

jakolehm:
Not compatible with Ruby 2.0.0 is bad thing.. what can we do about it?

jakolehm:
Needs rebase...

docs/getting-started/installing/cli.md
```diff
@@ -4,7 +4,7 @@ title: CLI
 
 ## Installing Kontena CLI
 
-> Prerequisites: You'll need Ruby version 2.0 or later installed on your system. For more details, see the official [Ruby installation docs](https://www.ruby-lang.org/en/documentation/installation/).
+> Prerequisites: You'll need Ruby version 2.1 or later installed on your system. For more details, see the official [Ruby installation docs](https://www.ruby-lang.org/en/documentation/installation/).
```
jakolehm: We need to change also quick start.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -102,9 +126,9 @@ def yaml_substitutables
 
       def load_yaml(interpolate = true)
         if interpolate
-          @yaml = ::YAML.safe_load(replace_dollar_dollars(interpolate(raw_content)))
+          @yaml = ::YAML.safe_load(replace_dollar_dollars(interpolate(interpolate_liquid(raw_content, variables.to_h(values_only: true)))))
```
jakolehm: Can we split this into multiple lines (a bit hard to read/debug)?

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -83,14 +101,20 @@ def execute(service_name = nil)
...
+          yaml['stack'].split('/').last.split(':').first if yaml['stack']
+        end
       end
 
       def stack_version
         yaml['version'] || yaml['stack'].to_s[/:(.*)/, 1] || '1'
       end
 
+      def yaml_section(section)
+        raw_content[/^(#{section}:.+?)^\S/m, 1]
```
jakolehm: How safe is this? We should add more tests for this method If we really need this kind of parsing.

docs/getting-started/installing/cli.md
kke: Quick start updated

kke:
ping @jakolehm 

[cli/lib/kontena/cli/stacks/yaml/reader.rb](https://github.com/kontena/kontena/blob/b9ba3f9f8fb51e0e3df2d18f039a391822f3bdbb/cli/lib/kontena/cli/stacks/yaml/reader.rb) cleaned up quite a bit.

cli/spec/fixtures/kontena-with-env-file.yml
```diff
@@ -5,7 +5,7 @@ wordpress:
   stateful: true
   env_file: .env
   environment:
-    WORDPRESS_DB_PASSWORD: ${project}_secret
+    WORDPRESS_DB_PASSWORD: %{project}_secret
```
jakolehm: why `%` ?

kke: This is only used by the `app` specs, for stacks there's stack-with-env-file.yml. 

I modified the wrong file and used % instead of $ when reverting like it was in another yml.

cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
```diff
@@ -1,5 +1,6 @@
 require_relative '../../../../spec_helper'
 require 'kontena/cli/stacks/yaml/reader'
+require 'liquid'
```
jakolehm: why reader does not require liquid?

cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
```diff
@@ -194,21 +195,28 @@ def absolute_yaml_path(file = 'kontena_v3.yml')
           .and_return(fixture('stack-with-variables.yml'))
         allow(ENV).to receive(:[])
           .with('MYSQL_IMAGE')
-          .and_return(nil)
+          .and_return('')
+        allow(ENV).to receive(:[])
+          .with('TAG')
+          .and_return('4.1')
 
+        #STDERR.puts(subject.variables.to_h.inspect)
```
jakolehm: I think we can remove this?

cli/lib/kontena/cli/stacks/validate_command.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'common'
+
+module Kontena::Cli::Stacks
```
jakolehm: We should add this command to docs.

cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
kke: it does, in the `def initialize`

cli/lib/kontena/cli/stacks/validate_command.rb
kke: done

PR #1562, opened 2016-12-15
Update Weave Net to 1.8.2

jakolehm:
https://github.com/weaveworks/weave/releases/v1.8.2


SpComb:
Need weave 1.8.3 to fix #1582

jakolehm:
1.8.3 is not released yet.. imho we should take this in and do a separate PR for new updates.

jnummelin:
Oh, somehow just by reading the related issue I had the wrong impression that 1.8.3 was already released. :/

Quickly tested locally and everything worked as expected.

So, LGTM

SpComb:
Sure, not say we shouldn't update to 1.8.2 :) Haven't tested it yet, though

PR #1564, opened 2016-12-15
Grid default affinity

### LINKED ISSUE #1032: Add a default affinity rule for a grid
it’d be nice if there was, say a special label that Kontena interpreted as the node being ineligible for automatic assignment.  It would take an affinity rule to get services on the node in question.  Or another way to view this, as the title suggests, is for the grid to have a default affinity rule which one might set to something like `label!=reserved=true`  Assuming one put a reserved=true label on certain nodes.  Perhaps this particular suggested affinity rule might even be the default, so I wouldn't need to set my grid default affinity rule once I know I want to reserve some nodes.  Otherwise, services created beforehand wouldn't know they shouldn't migrate to these reserved nodes.

jakolehm:
Implements #1032


kke:
@jakolehm needs rebase

jnummelin:
LGTM

PR #1566, opened 2016-12-16
HTTP GET /?health will not get logged

kke:
If you deploy the HA master stack or run some other kind of health monitoring, the checks will generate a lot of `GET /` lines in the master log. This PR skips logging for /?health requests and leaves  everything else as is.

The root path is excluded from token processing, so this should not enable brute forcing the tokens in the shadows.


jnummelin:
This needs to be "synced" with #1580

PR #1568, opened 2016-12-19
Use YAML.safe_load to avoid possible code execution

kke:
It's better to use safe_load to avoid a situation where a crafted yaml could execute code when parsing.

### DIFF
--- a/cli/kontena-cli.gemspec
+++ b/cli/kontena-cli.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler", "~> 1.7"
   spec.add_development_dependency "rake", "~> 10.0"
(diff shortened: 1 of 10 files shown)


nevalla:
OS X comes with ruby 2.0 and psych 2.0. Installing newer version of the gem builds native extension:
```
$ sudo gem install psych
Password:
Fetching: psych-2.2.2.gem (100%)
Building native extensions.  This could take a while...
Successfully installed psych-2.2.2
Parsing documentation for psych-2.2.2
unable to convert "\xCA" from ASCII-8BIT to UTF-8 for lib/psych.bundle, skipping
Installing ri documentation for psych-2.2.2
1 gem installed
```

I think we should also search for alternative solutions.

kke:
There's also this: https://github.com/dtao/safe_yaml

kke:
#1573 is better.

Issue #1526, opened 2016-12-09
Agent WeaveExecutor fails with dead weaveexec containers on CoreOS beta, alpha


SpComb:
This started to happen on my CoreOS Beta 1235.2.0 vagrant node recently. Seen something similar before with Ubuntu xenial, and switching to the overlayfs2 volume driver fixed it.

```
kontena-agent | E, [2016-12-09T11:12:50.978477 #1] ERROR -- : Actor crashed!
kontena-agent | Docker::Error::ServerError: Driver overlay failed to remove root filesystem 9835a3da45f02f0d0f000ffbb0221612a36d4f233c950e4a49aba52b1797995e: remove /var/lib/docker/overlay/b492d218b9b7ab546b11b24dc7f46aef107c264e425e15b5ef7a0ff364bc20f8/merged: device or resource busy
kontena-agent | 
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:50:in `rescue in request'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:38:in `request'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/container.rb:234:in `remove'
kontena-agent | 	/app/lib/kontena/network_adapters/weave_executor.rb:79:in `execute'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
[...]

PR #1569, opened 2016-12-19
Change default cadvisor image to official

### LINKED ISSUE #1567: Cadvisor / Weave confuses Docker in CoreOS Beta
See: https://github.com/weaveworks/weave/issues/2559#issuecomment-267553417

jnummelin:
Our own cAdvisor image uses `nsenter` to move the cAdvisor process to host side. This seems to cause some instabilities with docker and weave. See https://github.com/weaveworks/weave/issues/2559#issuecomment-267553417

The official cAdvisor image does not seem to suffer from this and hence this change back to the original image. cAdvisor has also fixed the original `du -sh` hogging all CPU in https://github.com/google/cadvisor/releases/tag/v0.23.3 which was the main reason to switch to custom image using nsenter.

Tested this with Ubuntu/aufs which AFAIK had the worst symptoms with `du -sh` with 100+ container and did not see any issues. Also testing with CoreOS beta (1235.2.0, docker 1.12.3) works without any issues.

Fixes #1567

### DIFF
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -6,8 +6,8 @@ module Kontena::Launchers
     include Celluloid::Notifications
     include Kontena::Logging
 
-    CADVISOR_VERSION = ENV['CADVISOR_VERSION'] || '0.23.2'
-    CADVISOR_IMAGE = ENV['CADVISOR_IMAGE'] || 'kontena/cadvisor'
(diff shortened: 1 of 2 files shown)


SpComb:
Also fixes #1526

jnummelin:
@SpComb Yes, the mounts are weird since the container mounts `/var/lib/docker`. Mounting that is NOT really recommended by Docker folks. (see https://docs.docker.com/engine/admin/troubleshooting_volume_errors/). That basically means that we have to come up with some other solution for longer term to get rid of cAdvisor.

PR #1579, opened 2016-12-20
Container exec/logs/inspect tries to guess full container name

### LINKED ISSUE #1563: kontena container exec is too complicated to use
Back in the days of 0.1x.x you could do things like:

    #assuming that there's an instance -1 of your service..
    kontena container exec rails-app-1 rake db:migrate

Now:
```
kontena version
cli: 1.0.2
master: 1.0.2 
````

...you need to do something like this:

    kontena container ls | grep rails-app | head -n1 | cut -d " " -f1 | xargs -I_ kontena container exec _ rake db:migrate

### PR DESCRIPTION
**Pull request by kke**
Fixes #1563

Makes `kontena container exec`, `kontena container logs` and `kontena container inspect` to try finding `{any_node_name}/container_id` if the user supplied container_id was not found.

Returns an error if multiple matches are found.

### DIFF
--- a/cli/lib/kontena/cli/containers/container_id_param.rb
+++ b/cli/lib/kontena/cli/containers/container_id_param.rb
@@ -13,11 +13,7 @@ module Kontena::Cli::Containers
               ENV["DEBUG"] && STDERR.puts("Invalid response from master: #{containers.inspect}")
               exit_with_error('Invalid response from master')
             end
-            targets = containers['containers'].select do |c|
-              c['name'].end_with?(container_id) ||
-              c['name'].end_with?(container_id + "-1") ||
-              c['name'] =~ /\A\w+\-#{container_id}(?:\-\d+)/
-            end.map { |c| "#{c['node']['name']}/#{c['name']}" }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by jakolehm** [comment]
Imho too much magic is just too much. If user does not know (or does not want to know) the full path to the container it should be possible to grep/pipe that information without too much effort.

Maybe something like this:

```
$ kontena container exec $(kontena container ls -q | grep redis-1) redis-cli ping
PONG
```

/cc @timolehto @SpComb @jnummelin

timolehto:
Supporting something like `-q` like docker in general would be nice as now it often time is the case that we need to do things like `kontena command ls | tail -n +2 | cut -d " " -f1` possibly combined with `xargs` or something depending on what your doing and it gets quite messy. That might be enough here too.

timolehto:
That's maybe even better. Supporting the `-q` option in general would be great, but this usecase `kontena service exec` would really hit the sweet spot, without being too magical.

timolehto:
Historically with heroku we've been doing it like `heroku run --app foo rake db:migrate` and `kontena service exec` certainly feels like the same thing and I would think many other heroku converts would appreciate this.

SpComb:
See  #1726

kke:
Too magical

Issue #1460, opened 2016-12-01
Timing issue with initial admin code


matti:
When a new master is created there is a small time window that will invalidate the initial admin code for good. I think we've seen this on the slack channel as well?

repro (following the updated DEVELOPMENT.md) #1459 :

```
docker-compose up --service-ports api
```

another terminal (pseudobash):

```
while true; do
  timeout(1.5); do
    kontena master login --name devmaster --code initialadmincode http://localhost:9292
  end
end
```

This will result in:
```
 [error] invalid_grant
```

matti:
and to clarify: grant will always be invalid, so you need to recreate the master.

SpComb:
I've also seen this when deploying a new master, and having the initial `kontena master login` error for whatever reason - unable to re-login anymore.

Looks like the master invalidates the initial admin code in mongo after login.

#### `docker logs kontena-server-api`
```
I, [2016-12-01T09:28:56.638811 #6]  INFO -- Mongodb::Migrator: migrating CreateInitialAdmin
*********************************************
Initial Admin Code: rf...eq3
*********************************************
```

#### `docker exec -it kontena-server-mongo mongo kontena_development`
```
> db.access_tokens.find()
{ "_id" : ObjectId("583fed93b0783c000600000e"), "token_type" : "bearer", "internal" : true, "scopes" : [ "user", "owner" ], "code" : "rf...eq3", "user_id" : ObjectId("583fed93b0783c000600000d"), "expires_at" : null, "token" : "$2a$10$B6...dUS", "token_last_four" : "6350", "updated_at" : ISODate("2016-12-01T09:29:55.494Z"), "created_at" : ISODate("2016-12-01T09:29:55.494Z") }
```

The initial admin code is there

#### `kontena master login --code rf...eq3 http://.../`

#### `docker exec -it kontena-server-mongo mongo kontena_development`

```
> db.access_tokens.find()
{ "_id" : ObjectId("583fedb3b0783c0006000013"), "token_type" : "bearer", "internal" : true, "expires_at" : null, "scopes" : [ "user", "owner" ], "user_id" : ObjectId("583fed93b0783c000600000d"), "token" : "$2a$10$B6O...VDa", "token_last_four" : "617f", "updated_at" : ISODate("2016-12-01T09:30:27.407Z"), "created_at" : ISODate("2016-12-01T09:30:27.407Z") }
```

The intial admin code is gone

#### `DEBUG=true kontena master login --code rf...eq3 http://.../`
```
[...]

PR #1580, opened 2016-12-20
Quick-check db when requesting / from master api

kke:
In case the DB connection is dead, requesting `/` will return an error. Handy for healthcheck. Also a slim chance of helping with the timing issue in #1460

### DIFF
--- a/server/server.rb
+++ b/server/server.rb
@@ -41,6 +41,7 @@ class Server < Roda
   route do |r|
 
     r.root do
+      Mongoid.default_session.collection_names # crashes if db is not connected
       {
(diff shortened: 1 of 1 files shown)


server/server.rb
```diff
@@ -41,6 +41,7 @@ def self.name
   route do |r|
 
     r.root do
+      Mongoid.default_session.collection_names # crashes if db is not connected
```
jakolehm: Imho we should do this in a `/health` or similar route.

jnummelin: I agree, separate health endpoint plays nicer also with #1566

PR #1587, opened 2016-12-23
docs (deploy) Container affinity.

### LINKED ISSUE #1402: container affinity
Currently there is `service` affinity like:

```
affinity:
  - service==mysql
```

But there are some (raw docker) cases when one would like to schedule a service in the same machine as where a container is running (see https://github.com/jeroenpeeters/docker-ssh)

Affinity setting `container==mysql-1` would ensure this.

matti:
Fixes #1402

### DIFF
--- a/docs/using-kontena/deploy.md
+++ b/docs/using-kontena/deploy.md
@@ -74,11 +74,16 @@ When creating services, you can direct the Node(s) to which the containers shoul
 
 An affinity condition happens when Kontena is trying to find a field that matches (`==`) a given value. An anti-affinity condition happens when Kontena is trying to find a field that does not match (`!=`) a given value.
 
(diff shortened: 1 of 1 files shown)


jakolehm:
Imho container affinity is a power-user feature and we should note that in the docs.. I'm a bit against the whole idea of container affinity because it exposes too much internals.

matti:
Not really sure. I want to deploy a service that uses the same docker daemon (https://github.com/jeroenpeeters/docker-ssh) and this is very handy. Not sure how this differs from any other affinities.

jakolehm:
Because we reserve right to change container naming ;) (reason why it should be used with care). I think it would be make more sense to have "service instance" affinity.

matti:
So still no docs on container affinity.

PR #1589, opened 2016-12-27
Add container hours telemetry data

jakolehm:


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -22,8 +22,10 @@ class NodeInfoWorker
     def initialize(queue, autostart = true)
       @queue = queue
       @statsd = nil
+      @stats_since = Time.now
```
kke: Maybe use `Time.now.utc` everywhere to avoid any DST issues

PR #1594, opened 2016-12-29
[WIP] Support canary deployment with lb

jnummelin:
This PR implements support for canary deployments behind the loadbalancer. See the docs change for example.

Not sure if the check for existing upstreams is the best way to determine if the whole service config should be removed or not. Other option is to keep track of the aliased services in etcd and only remove the config when last service gets removed.

### DIFF
--- a/agent/lib/kontena/load_balancers/configurer.rb
+++ b/agent/lib/kontena/load_balancers/configurer.rb
@@ -29,10 +29,10 @@ module Kontena::LoadBalancers
     # @param [Docker::Container] container
     def ensure_config(container)
       name = container.labels['io.kontena.load_balancer.name']
-      service_name = container.service_name_for_lb
+      env_hash = container.env_hash
+      service_name = env_hash['KONTENA_LB_SERVICE_ALIAS'] || container.service_name_for_lb
       check_uri = container.labels['io.kontena.health_check.uri']
       etcd_path = "#{ETCD_PREFIX}/#{name}"
-      env_hash = container.env_hash
 
       mode = env_hash['KONTENA_LB_MODE'] || 'http'
       balance = env_hash['KONTENA_LB_BALANCE'] || 'roundrobin'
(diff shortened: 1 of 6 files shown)


agent/lib/kontena/load_balancers/configurer.rb
```diff
@@ -77,8 +77,11 @@ def ensure_config(container)
     # @param [Docker::Container] container
     def remove_config(container)
       name = container.labels['io.kontena.load_balancer.name']
-      service_name = container.service_name_for_lb
+      service_name = container.env_hash['KONTENA_LB_SERVICE_ALIAS'] || container.service_name_for_lb
+
       mode = container.env_hash['KONTENA_LB_MODE'] || 'http'
+      return if count_upstreams(name, service_name, mode) > 0
```
SpComb: The `lb:remove_config` -> `remove_config` only happens when terminating the `instance_number == 1` container. Are all the other upstreams guaranteed to be gone at that point?

i.e. do we terminate service instances in reverse-order, and wait for each service instance to be terminated before terminating the first instance?

docs/using-kontena/loadbalancer.md
```diff
@@ -183,6 +183,40 @@ loadbalancer:
...
+Now when you are ready to deploy a new version of your app you can deploy that using following configuration:
+```
+app_version2:
+  image: my-app:2.0
+  environment:
+    - KONTENA_LB_MODE=http
+    - KONTENA_LB_BALANCE=roundrobin
+    - KONTENA_LB_INTERNAL_PORT=80
+    - KONTENA_LB_VIRTUAL_HOSTS=my-app.example.com
+    - KONTENA_LB_SERVICE_ALIAS=my-app
```
SpComb: This alias is global across all stacks, maybe it would be a good idea to use `KONTENA_LB_SERVICE_ALIAS=$STACK/my-app` as the example?

tecnobrat:
Hey all!

We've got some requirements for handling canary deploys, specifically around splitting the traffic based on some additional logic besides just random.

We use JWTs to authenticate our users, the JWT is sent either via an `Authorization: Bearer` header, or via a cookie.  We would love to be able to have the load balancer parse this header/cookie and make decisions based on values encoded in the JWT.

Consider the JWT has the following info encoded in it:
`{ user_id: "000-00000-000", channel: "beta" }`

We would like to be able to send all users who are in the "beta" channel to the canary build, while everyone else stays on the normal build.  But we also want to be able to support % traffic flows as well.

jnummelin:
This has fallen too far behind to make it work. Also there's lot of problems/difficulties in this approach both internally and also for confusing the users. As we've been "speculating" bit different approach to whole LB config, these kind of things would be lot easier to implement using the (still to be decided) "ingress" config model.

PR #1597, opened 2016-12-29
Send labels with the initial ws connection headers

### LINKED ISSUE #1136: Master does not take account node engine labels on initial agent connection
Master sends node information back to agent before agent  sends docker engine labels to master.

This is troublesome in situations where docker engine labels have a custom `region` label because the master will resolve overlay (weave) peer addresses using this label information. Basically this means that node tries to connect to other nodes via public ip which in most cases is not a valid assumption.

jnummelin:
fixes #1136

### DIFF
--- a/agent/lib/kontena/websocket_client.rb
+++ b/agent/lib/kontena/websocket_client.rb
@@ -58,7 +58,8 @@ module Kontena
       headers = {
           'Kontena-Grid-Token' => self.api_token.to_s,
           'Kontena-Node-Id' => host_id.to_s,
-          'Kontena-Version' => Kontena::Agent::VERSION
+          'Kontena-Version' => Kontena::Agent::VERSION,
+          'Kontena-Node-Labels' => labels
       }
       @ws = Faye::WebSocket::Client.new(self.api_uri, nil, {headers: headers})
 
@@ -159,6 +160,10 @@ module Kontena
     # @return [String]
     def host_id
       Docker.info['ID']
+    end
+
+    def labels
+      Docker.info['Labels'].join(',')
     end
 
     def verify_connection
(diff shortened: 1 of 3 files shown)


jakolehm:
Not 100% sure this is the right way... should we send node information on websocket connect event (now it's `websocket:connected` which happens too late)?

jakolehm:
As discussed, sending info on websocket connect does not help here because it's also too late.

jnummelin:
If we don't want to use the headers to send the info, I can think of two alternative approaches. Both need some refactoring on the node plugging sequence.

sequencediagram1

In this case server would request the info from agent and after receiving it would execute the node plugging functionality. This would require new RPC api on agent side and refactoring of the node_plugger sequence.

sequencediagram2

In this case agent would send the full node info data after the ws connection is opened. Server executes the node plugger functionality after receiving this initial node info. This would require some mechanism to keep track of whether or not initial node plugging functionality is executed or not as agent periodically send the full info to the server.


Some other options, all which might be bad ideas :) :
- Send the node_info data on the initial connect as base64 encoded query param
- Send the node_info data on the initial connect as base64 encoded header field


As for now, the only critical extra info is the labels and to send those the custom header field is pretty simple solution.

agent/lib/kontena/websocket_client.rb
```diff
@@ -161,6 +162,10 @@ def host_id
       Docker.info['ID']
     end
 
+    def labels
+      Docker.info['Labels'].join(',')
```
jakolehm: Is this always an array? Maybe it would be safer to cast `.to_a` first?

jakolehm:
Ok, maybe using headers is fine for bug fixing this race condition. We can refactor this logic later to match your diagram.

agent/lib/kontena/websocket_client.rb
jnummelin: Good point, will change.

PR #1599, opened 2016-12-30
Remove exact docker patch version from Ubuntu install docs

jakolehm:
Instructions now install latest 1.12.x Docker engine version. Also the apt repo is written to a separate sources.list file.


docs/getting-started/installing/ubuntu.md
```diff
@@ -123,7 +123,7 @@ Now you can go ahead and install the `kontena-agent` Ubuntu package using the ap
 
 ```
 $ wget -qO - https://bintray.com/user/downloadSubjectPublicKey?username=bintray | sudo apt-key add -
-$ echo "deb http://dl.bintray.com/kontena/ubuntu xenial main" | sudo tee -a /etc/apt/sources.list
+$ echo "deb http://dl.bintray.com/kontena/ubuntu xenial main" | sudo tee -a /etc/apt/sources.list.d/kontena.list
```
SpComb: Probably better to drop the `-a` from `tee` here, to replace any existing `kontena.list` rather than appending

jakolehm: Dropped.

PR #1605, opened 2017-01-02
Allow to set custom image for VPN

jnummelin:
Allows easier experimentation with VPN

### DIFF
--- a/cli/lib/kontena/cli/vpn/create_command.rb
+++ b/cli/lib/kontena/cli/vpn/create_command.rb
@@ -8,6 +8,7 @@ module Kontena::Cli::Vpn
 
     option '--node', 'NODE', 'Node name where VPN is deployed'
     option '--ip', 'IP', 'Node ip-address to use in VPN service configuration'
+    option '--image', 'IMAGE', 'Custom Docker image to use', hidden: true
 
     def execute
       require_api_url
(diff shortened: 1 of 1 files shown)


SpComb:
It would be great if we could replace `kontena vpn create` with a stack file... it would need some new custom resolvers, like

```yaml
variables:
  image:
    default: kontena/openvpn:ethwe
  node:
    from:
      - host_node:
          public_ip: true
services:
  vpn:
    image: $image
    affinity:
      - node==$node
    environment:
      - OVPN_SERVER_URL=udp://{{node.public_ip}:1194
```

SpComb:
Nothing against this PR specifically, it just feels slightly meh in general to start adding a million `kontena vpn create` / `kontena registry create` flags when we could be extending and dogfooding our stacks instead.

cli/lib/kontena/cli/vpn/create_command.rb
```diff
@@ -8,6 +8,7 @@ class CreateCommand < Kontena::Command
 
     option '--node', 'NODE', 'Node name where VPN is deployed'
     option '--ip', 'IP', 'Node ip-address to use in VPN service configuration'
+    option '--image', 'IMAGE', 'Custom Docker image to use', hidden: true
```
SpComb: default: 'kontena/openvpn:ethwe'

jnummelin:
I do agree that creating custom stack files is much better to try out things like this. Closing.

PR #1620, opened 2017-01-04
Fix ubuntu build file permissions

jakolehm:
Copies sources to ubuntu build image (no bind mounting anymore).

### DIFF
--- a/build/Dockerfile.ubuntu
+++ b/build/Dockerfile.ubuntu
@@ -1,5 +1,5 @@
 FROM ubuntu:trusty
-MAINTAINER jari@kontena.io
+MAINTAINER Kontena, Inc. <info@kontena.io>
 
 RUN echo 'deb http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu trusty main' >> /etc/apt/sources.list && \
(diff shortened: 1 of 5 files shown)


build/Dockerfile.ubuntu
```diff
@@ -9,12 +9,15 @@ RUN echo 'deb http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu trusty main' >>
     rm -rf /var/lib/apt/lists/* && \
     gem install bundler
 
-ADD server/Gemfile /build/server/Gemfile
-ADD server/Gemfile.lock /build/server/Gemfile.lock
+COPY server/Gemfile /build/server/Gemfile
+COPY server/Gemfile.lock /build/server/Gemfile.lock
```
SpComb: BTW in general, it might be slightly better to use `COPY server/Gemfile server/Gemfile.lock /build/server/` to reduce the number of image layers?

SpComb:
I don't think this works either, the `docker run --rm ... rake release:build_ubuntu` drops the resulting .deb package into the Docker container rootfs, and it goes away when the build container is removed...

This still needs some bind mount to put the resulting .deb file in.

build/Dockerfile.ubuntu
jakolehm: Fixed

jakolehm:
@SpComb @jnummelin ping

PR #1628, opened 2017-01-05
Install plugins without shelling out and put them under .kontena/gems

kke:
This will allow installing plugins without sudo and it plays nicer with funny installations where `gem` command does not point to the same ruby as the kontena-cli installation.

### DIFF
--- a/cli/lib/kontena/plugin_manager.rb
+++ b/cli/lib/kontena/plugin_manager.rb
@@ -2,27 +2,45 @@ require 'singleton'
 
 module Kontena
   class PluginManager
+
+    module SayOverride
+      def say(*args)
+      end
+    end
+
     include Singleton
 
     CLI_GEM = 'kontena-cli'.freeze
     MIN_CLI_VERSION = '0.15.99'.freeze
 
-    attr_reader :plugins
+    def init
+      plugins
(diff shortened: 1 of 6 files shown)


cli/lib/kontena/plugin_manager.rb
```diff
@@ -2,27 +2,174 @@
...
     include Singleton
 
     CLI_GEM = 'kontena-cli'.freeze
     MIN_CLI_VERSION = '0.15.99'.freeze
 
-    attr_reader :plugins
+    # Initialize plugin manager
+    def init
+      ENV["GEM_HOME"] = install_dir
+      Gem.paths = ENV
```
jakolehm: What does this actually do?

kke: GEM_HOME is the directory where gems will be installed to.
Gem.paths for some reason contains a dump of ENV. This reinitializes it after setting GEM_HOME.

```ruby
def self.paths=(env)
  clear_paths
  @paths = Gem::PathSupport.new env
  Gem::Specification.dirs = @paths.path # FIX: home is at end
end
```

kke:
Merging, revert if there's problems during RC evaluation.

PR #1629, opened 2017-01-08
Disable cAdvisor disk metrics & give lower cpu priority

jakolehm:
1.0.3 release (#1569) moved cAdvisor process back to container and this seems to cause resource usage explosion on some nodes (memory usage creeps over 1GB etc). This PR tries to fix this situation by disabling disk stats and giving cAdvisor lower cpu priority. Problematic nodes show ~40MB memory usage after this patch and also cpu usage is much lower (few percentages).

PR also removes some leftovers from old behaviour.

### DIFF
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -80,7 +80,7 @@ module Kontena::Launchers
           'Binds' => volume_binds,
           'NetworkMode' => 'host',
           'CpuShares' => 128,
-          'Memory' => (256 * 1024 * 1024)
+          'Memory' => (256 * 1024 * 1024),
           'RestartPolicy' => {'Name' => 'always'}
         }
       )
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -80,6 +80,7 @@ module Kontena::Launchers
           'Binds' => volume_binds,
           'NetworkMode' => 'host',
           'CpuShares' => 128,
+          'Memory' => (256 * 1024 * 1024)
           'RestartPolicy' => {'Name' => 'always'}
         }
       )
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -51,7 +51,9 @@ module Kontena::Launchers
     # @param [String] image
     def create_container(image)
       container = Docker::Container.get('kontena-cadvisor') rescue nil
-      if container && container.info['Config']['Image'] != image
+      if container && config_changed?(container)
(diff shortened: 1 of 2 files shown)


agent/lib/kontena/launchers/cadvisor.rb
```diff
@@ -67,14 +69,17 @@ def create_container(image)
           '--listen_ip=127.0.0.1',
           '--port=8989',
           '--storage_duration=2m',
-          '--housekeeping_interval=30s'
+          '--housekeeping_interval=10s',
+          '--disable_metrics=tcp,disk'
```
jnummelin: Why do we disable tcp metrics?

agent/lib/kontena/launchers/cadvisor.rb
```diff
@@ -67,14 +69,17 @@ def create_container(image)
...
+        'Labels' => {
+          'io.kontena.agent.version' => Kontena::Agent::VERSION
+        },
         'Volumes' => volume_mappings,
         'HostConfig' => {
           'Binds' => volume_binds,
           'NetworkMode' => 'host',
-          'PidMode' => 'host',
-          'Privileged' => true,
+          'CpuShares' => 128,
```
jnummelin: Should we also set mem limits?

agent/lib/kontena/launchers/cadvisor.rb
jakolehm: They are disabled by default

agent/lib/kontena/launchers/cadvisor.rb
```diff
@@ -89,37 +94,31 @@ def log_error(exc)
 
     # @return [Hash]
     def volume_mappings
-      if kontena_image?
```
SpComb: Also clean out `#kontena_image?` if we're no longer using that?

SpComb:
I suspect that disabling the `du` metrics collection would also ease all of the dead Docker container / remove errors that we've been having.

agent/lib/kontena/launchers/cadvisor.rb
jakolehm: Maybe, any idea what is proper limit for this? 128MB?

jnummelin: I'd say 256M should allow some temporary mem hogging also and not get immediate OOM shot-in-the-head.

jakolehm: Set to 256M

SpComb:
> I suspect that disabling the du metrics collection would also ease all of the dead Docker container / remove errors that we've been having.

This probably isn't the case - we still have the problematic `/:/rootfs:ro` and `/var/lib/docker:/var/lib/docker:ro` bind mounts. Just having those bind mounts with mount propagation may be enough to cause issues..

Issue #1633, opened 2017-01-09
Export/Import secrets from vault to/from yaml


counterbeing:
In the even that I ever wanted to spin up a new grid, be it for testing, or to replace production, it would be super useful to be able to grab all of my secrets. Additionally, there are other tools out there like Helm that use yaml config to describe variables. I think it might be really useful to have a yaml import and export to update all variables at once. 

It would also offer the piece of mind that if something happened to my grid, I could restore it quickly with a few commands.

counterbeing:
I just noticed there's an issue #357 that resolves #354. #354 mentions the ability to export all secrets. Maybe this should stand as a feature request?

kke:
Kind of duplicates #1625

counterbeing:
I agree, it is partially, however that issue is requesting service specific secrets. I could see the functionality being listed under the same command. 

I just pushed a branch/PR. I tested it with my own kontena cluster, and everything looks good. If you guys like it, I can implement the `import` command as well, and update the docs.

PR #1634, opened 2017-01-09
Add export command to vault

counterbeing:
This is half of the fix for #1633

### DIFF
--- a/cli/lib/kontena/cli/vault/import_command.rb
+++ b/cli/lib/kontena/cli/vault/import_command.rb
@@ -1,0 +1,58 @@
+module Kontena::Cli::Vault
+  class ImportCommand < Kontena::Command
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+
+    option ['-u', '--upsert'], :flag, 'Create secret unless already exists', default: false
+    parameter "FILENAME", "Secret yaml file"
+
+    def execute
+      require_api_url
+      require_current_grid
+      token = require_token
+
+      # Not sure of a better way to do this for now. Skips the first two ARGF
+      # reads in order to get to the third without raising an error.
+      2.times do
+        begin
+          ARGF.read
+        rescue
+        end
+      end
+
+      raw = ARGF.read
+      exit_with_error('No data recieved from yaml file') if raw.to_s == ''
+      begin
+        secrets = YAML.load(raw)
+      rescue
+        exit_with_error('STDIN did not contain valid yaml')
+      end
+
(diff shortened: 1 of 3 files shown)


counterbeing:
I've just added a bit to do imports as well. I could use some feedback on my use of `ARGF`, but it works.

cli/lib/kontena/cli/vault/import_command.rb
```diff
@@ -0,0 +1,58 @@
...
+    parameter "FILENAME", "Secret yaml file"
+
+    def execute
+      require_api_url
+      require_current_grid
+      token = require_token
+
+      # Not sure of a better way to do this for now. Skips the first two ARGF
+      # reads in order to get to the third without raising an error.
+      2.times do
```
counterbeing: This is the piece I really don't like. If anyone has any thoughts on how to improve it, I'm all ears.

kke: You can just `File.read(filename)`

cli/lib/kontena/cli/vault/import_command.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      raw = ARGF.read
+      exit_with_error('No data recieved from yaml file') if raw.to_s == ''
+      begin
+        secrets = YAML.load(raw)
+      rescue
+        exit_with_error('STDIN did not contain valid yaml')
+      end
+
+      current_secret_keys =
```
kke: You could just run the vault update command like this:
```ruby
result = Kontena.run(['vault', 'update', '--upsert', name, secret])
count += 1 if result == 0
```

instead of reimplementing the vault reading + writing here

kke:
(i merged this to a feature branch to make some finetuning)

counterbeing:
Woohoo! Thanks @kke. Please let me know what you end up doing around the `ARGF` statement.

cli/lib/kontena/cli/vault/import_command.rb
counterbeing: Ah, I just found this. That would be cool too. I was hoping to accept pipes as well. Like `cat xyz.yml | kontena vault import`.  It's not required, but it would be nice to keep things as automate-able as possible.

cli/lib/kontena/cli/vault/import_command.rb
counterbeing: Ah, great. I was sort of working off of what I had seen, but that sounds like an improvement for sure.

PR #1635, opened 2017-01-10
Fix possible nil stack case in grid services API

jnummelin:
There seems to be possibility in the API to request services of a non existing stack in which case the API will produce internal errors.
See: https://kontenacommunity.slack.com/archives/general/p1484031542001669

I'm not entirely sure what could cause the CLI to actually even request services of a missing stack, but it should not cause internal error anyway.

This PR fixes this by halting the request in controlled manner if the given stack is not found.


jnummelin:
duplicate for #1613

PR #1636, opened 2017-01-10
Feature/master cli version check

### LINKED ISSUE #1632: CLI should print warning if using older version than master
Currently, I had a case where a breaking change between versions simply caused CLI to crash during a deploy script. It took a bit of figuring out to realize I needed to update my Kontena CLI version. Would be good to do a check if current CLI is old and show a warning.

kke:
Fixes #1632

Problems with this solution:
Does nothing if the server is not sending a version header (which it won't before this PR). 

Alternatives:
1. Query the server before talking to it.
2. Only check & complain when logging in. 
3. Restore the min_cli_version interceptor to master.

### DIFF
--- a/cli/lib/kontena/client.rb
+++ b/cli/lib/kontena/client.rb
@@ -484,14 +484,18 @@ module Kontena
     end
 
     def check_version_and_warn(server_version)
-      return nil if server_version.nil?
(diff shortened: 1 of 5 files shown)


jakolehm:
Travis says no?

PR #1655, opened 2017-01-13
Vault import/export

### LINKED ISSUE #1633: Export/Import secrets from vault to/from yaml

kke:
```
$ kontena vault export | kontena vault import --force
```

Completes #1634 
Fixes #1633 

ping @counterbeing

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@ docs
 */vendor
 */.bundle
 */coverage
+cli/omnibus/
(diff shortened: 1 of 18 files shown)


counterbeing:
Looks great! You even added specs. Thanks for the ping 👍

PR #1657, opened 2017-01-13
Agent: fix cAdvisor stats to ignore systemd Docker container mount cgroups

### LINKED ISSUE #1656: Kontena service CPU & Memory stats may be zero-valued due to cAdvisor cgroup confusion
## Symptom

Some `kontena service stats` will show zero CPU/memory usage for random Docker containers.

## Background
cAdvisor works by collecting the Docker container stats directly from the `/sys/fs/cgroup` files, such as `/sys/fs/cgroup/cpu/*/cpu.cfs_period_us`: https://github.com/google/cadvisor/blob/v0.24.1/container/common/helpers.go#L77

It locates the Docker container cgroup by recursively scanning through all `/sys/fs/cgroup/*` directories, and matching any cgroup that contains an existing Docker container ID in the basename: https://github.com/google/cadvisor/blob/v0.24.1/container/docker/factory.go#L148

This means that with CoreOS 1235.6.0 + systemd 231 + Docker 1.12.3 + Linux 4.7.3-coreos-r2, the cAdvisor Docker driver will pick up two cgroups for each Docker container:

```
I0113 12:22:45.621838       1 factory.go:111] Using factory "docker" for container "/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount"
I0113 12:22:45.626448       1 manager.go:874] Added container: "/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount" (aliases: [null-stackless-whoami-1 acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536], namespace: "docker")

I0113 12:22:45.755197       1 factory.go:111] Using factory "docker" for container "/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536"
I0113 12:22:45.756418       1 manager.go:874] Added container: "/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536" (aliases: [null-stackless-whoami-1 acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536], namespace: "docker")
```

In this case, `/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536` is the correct cgroup, which contains the processes within the Docker container. The `/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount` cgroup is empty, and associated with the `/var/lib/docker/containers/*/shm` mountpoint.

Both of these container entries will associate themselves with the information returned by the Docker API for that container ID, and will thus have an identical set of cAdvisor aliases. Both container entries will add identical `namespacedContainerName{"docker", "acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536"}` entries in the `github.com/google/cadvisor/manager:manager.containers` map, which override each other depending on the order that the cgroups are listed in, and may change as events come in?

The `/api/v1.2/docker/ID` will then semi-randomly return one of these two cgroup entries:

```
core@core-01 ~ $ curl -v http://localhost:8989/api/v1.2/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536
{"/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536":{"id":"acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536","name":"/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536","aliases":["null-stackless-whoami-1","acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536"],"namespace":"docker", ...
core@core-01 ~ $ docker restart kontena-cadvisor
kontena-cadvisor
core@core-01 ~ $ curl http://localhost:8989/api/v1.2/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536
{"/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount":{"id":"acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536","name":"/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount","aliases":["null-stackless-whoami-1","acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536"],"namespace":"docker", ...
```

See https://github.com/google/cadvisor/issues/1572

## Workaround
The `/api/v1.2/containers/CGROUP` API will always return the actual container cgroup stats, assuming that Docker places it's container cgroups directly under `/docker`.

Alternative is to somehow hide the systemd mount cgroups from cAdvisor.

SpComb:
Fixes #1656

This replaces the `/api/v1.2/docker/...` queries with a single `/api/v1.2/subcontainers` query, which returns all containers that cAdvisor knows about. This response includes both `/docker/*` and `/system.slice/var-lib-docker-containers-*-shm.mount` entries for each Docker container.

The agent ignores the `*.mount` entries that have zero-valued stats.

### DIFF
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -49,9 +49,12 @@ module Kontena::Workers
       info 'starting collection'
 
       if response = get("/api/v1.2/subcontainers")
-        response.each do |stat|
-          next unless stat[:namespace] == 'docker'
-          next if stat['name'].ends_with? '.mount'
+        response.each do |data|
+          next unless data[:namespace] == 'docker'
+
+          # Skip systemd mount units that confuse cadvisor
+          # @see https://github.com/kontena/kontena/issues/1656
+          next if data[:name].end_with? '.mount'
 
           send_container_stats(data)
         end
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -47,25 +47,26 @@ module Kontena::Workers
 
     def collect_stats
       info 'starting collection'
-      begin
-        Docker::Container.all.each do |container|
-          if container.running?
-            data = self.collect_container_stats(container.id)
-            send_container_stats(data) if data
-            sleep 0.5
-          end
+
+      if response = get("/api/v1.2/subcontainers")
+        response.each do |stat|
+          next unless stat[:namespace] == 'docker'
+          next if stat['name'].ends_with? '.mount'
+
+          send_container_stats(data)
         end
-      rescue => exc
-        error "error on stats fetching: #{exc.message}"
-        error exc.backtrace.join("\n")
       end
+
+    rescue => exc
+      error "error on stats fetching: #{exc.message}"
+      error exc.backtrace.join("\n")
     end
 
 
-    def collect_container_stats(container_id)
+    def get(path)
       retries = 3
       begin
-        response = client.get(:path => "/api/v1.2/containers/docker/#{container_id}")
+        response = client.get(:path => path)
         if response.status == 200
           JSON.parse(response.body, symbolize_names: true) rescue nil
         else
@@ -77,7 +78,7 @@ module Kontena::Workers
         if retries > 0
           retry
         end
-        error "error getting container(#{container_id}) stats. #{exc.class.name}: #{exc.message}"
+        error "get #{path}: #{exc.class.name}: #{exc.message}"
         nil
       end
     end
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -85,13 +85,10 @@ module Kontena::Workers
     ##
     # @param [Hash] container
     def send_container_stats(container)
-      # when single container stats are used cadvisor "prefixes" all the data with some stupid system slice sutff
-      slice_key = container.keys[0]
-
-      prev_stat = container.dig(slice_key, :stats)[-2] if container
+      prev_stat = container.dig(:stats)[-2] if container
       return if prev_stat.nil?
 
-      current_stat = container.dig(slice_key, :stats, -1)
+      current_stat = container.dig(:stats, -1)
       # Need to default to something usable in calculations
       cpu_usages = current_stat.dig(:cpu, :usage, :per_cpu_usage)
       num_cores = cpu_usages ? cpu_usages.count : 1
@@ -101,8 +98,8 @@ module Kontena::Workers
       event = {
         event: 'container:stats'.freeze,
         data: {
-          id: container.dig(slice_key, :aliases, 1),
-          spec: container.dig(slice_key, :spec),
+          id: container.dig(:aliases, 1),
+          spec: container.dig(:spec),
           cpu: {
             usage: raw_cpu_usage,
             usage_pct: (((raw_cpu_usage / interval_in_ns ) / num_cores ) * 100).round(2)
@@ -117,7 +114,7 @@ module Kontena::Workers
         }
       }
       self.queue << event
-      send_statsd_metrics(container.dig(slice_key, :aliases, 0), event[:data])
+      send_statsd_metrics(container.dig(:aliases, 0), event[:data])
     end
 
     def client
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -65,7 +65,7 @@ module Kontena::Workers
     def collect_container_stats(container_id)
       retries = 3
       begin
-        response = client.get(:path => "/api/v1.2/docker/#{container_id}")
+        response = client.get(:path => "/api/v1.2/containers/docker/#{container_id}")
         if response.status == 200
           JSON.parse(response.body, symbolize_names: true) rescue nil
         else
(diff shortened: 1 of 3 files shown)


agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -47,25 +47,26 @@ def start
...
       info 'starting collection'
-      begin
-        Docker::Container.all.each do |container|
-          if container.running?
-            data = self.collect_container_stats(container.id)
-            send_container_stats(data) if data
-            sleep 0.5
-          end
+
+      if response = get("/api/v1.2/subcontainers")
```
jakolehm: Interesting change.. what is the reason behind this?

SpComb: See the PR description, you were too quick :)

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -117,7 +118,7 @@ def send_container_stats(container)
         }
       }
       self.queue << event
-      send_statsd_metrics(container.dig(slice_key, :aliases, 0), event[:data])
+      send_statsd_metrics(container.dig(:aliases, 0), event[:data])
```
SpComb: Is this assumption of `"aliases": [name, id]` sensible?

It looks like the current cAdvisor implementation always returns these in this order: https://github.com/google/cadvisor/blob/v0.24.1/container/docker/handler.go#L222

But it's certainly not part of the API...

jakolehm: Good point, do we have any better ways?

agent/lib/kontena/workers/stats_worker.rb
jakolehm: Ok, makes sense :)

SpComb:
It looks like upstream cAdvisor is fixing the original issue in https://github.com/google/cadvisor/pull/1573, but maybe this change makes sense anyways? I would expect the single `/subcontainers` API approach to be cheaper than the current approach of multiple API queries, and generate mostly the same results?

This is probably 1.1 if there's an fixed upstream release we can use instead.

jakolehm:
Actually there has been reason for doing multiple calls. The `/api/v1.2/docker` did error out if one of the containers were in a bad state.. not sure if `subcontainers` handles this better.

SpComb:
> Actually there has been reason for doing multiple calls. The /api/v1.2/docker did error out if one of the containers were in a bad state.. not sure if subcontainers handles this better.

Any example of what kind of errors it returned in that case? It may have been fixed since.

jakolehm:
Api just throwed 500 error so agent did not get any metrics (#769).

jnummelin:
Yes, the whole point of doing multiple cAdvisor calls was to avoid one container erroring out all the containers. I'm afraid that this PR brings those issues back as this gets all the stats in single call. Although the cAdvisor version was 0.23.0 at that point, so some of the things might be fixed:
https://github.com/google/cadvisor/releases/tag/v0.23.2
> Better handling of partial failures when fetching subcontainers

https://github.com/google/cadvisor/releases/tag/v0.24.0
> Improved robustness to partial failures

SpComb:
> Api just throwed 500 error so agent did not get any metrics (#769).

Looking at the fixed issue (#678), it's `undefined method 'count' for nil:NilClass` at `num_cores = current_stat[:cpu][:usage][:per_cpu_usage].count` -  it can't have been a HTTP 500 response, because it wouldn't have actually parsed the JSON response and gotten even that far.

> Although the cAdvisor version was 0.23.0 at that point, so some of the things might be fixed

Yes, that's why it would be good to know what the actual reason behind the workaround was, we may not need that workaround anymore?

agent/lib/kontena/workers/stats_worker.rb
SpComb: Changed to use `name = container[:aliases].select{|a| a != container[:id]}.first`

jnummelin:
The cAdvisor 500 errors was fixed at the same PR, so the descriptions are not 100% accurate. :/

But in some cases where cAdvisor failed to collect some container stats (maybe dead containers, maybe containers gone missing during the collection) really gave 500 for the request. So one failing container made all others fail too and hence I did the "workaround" to get the stats one-by-one. I tried to find chat archives for more detailed error descriptions but failed. :/

jnummelin:
Tried to break this by running continuous scale up&down which should cause containers to disappear during collection and now cAdvisor seems to be able to handle that. From agent point-of-view all the collections were ok, on cAdvisor logs one error can be seen:
```
W0116 09:29:50.499561       1 container.go:365] Failed to get RecentStats("/docker/763a22cb225d6e083951c36b4210f879dae3c97373e00e383c46f89a3f722516") while determining the next housekeeping: unable to find data for container /docker/763a22cb225d6e083951c36b4210f879dae3c97373e00e383c46f89a3f722516
```
But this did not cause the cAdvisor API to "misbehave".

PR #1662, opened 2017-01-16
Save the variable values with stack and return them in response

kke:
The answers / values of variables were not being saved in stack data on master.


server/app/mutations/stacks/update.rb
```diff
@@ -44,6 +44,7 @@ def execute
         stack_name: self.stack,
         expose: self.expose,
         source: self.source,
+        variables: self.variables,
```
jakolehm: Does cli actually send these variables?

kke: Yes, master just didn't save them.

PR #1672, opened 2017-01-18
Add .netrc file to CLI Docker image

nevalla:
When using Drone CI, Drone tries to write data to `/root/.netrc` file. That returns error because CLI container is run by `kontena` user and permission is denied. This PR adds `.netrc` file into `/root` directory and gives access to kontena user to it.

Related discussion on Drone forum: http://discourse.drone.io/t/netrc-permission-denied/171


jakolehm:
Should we just switch to root? /cc @jnummelin

jnummelin:
I think using root makes life much easier, there might be other similar uses for Kontena CLI in the future. I think the only "real" issue using root is that when user mounts something to the container. But mounting your private stuff into random container is always potential user-error.

PR #1673, opened 2017-01-18
Optionally use sudo when running docker build/push

### LINKED ISSUE #1548: Add optional (or not) sudo to docker commands
By default `kontena stack build` runs `docker` without sudo, I suggest to add sudo to every docker command, that way user running kontena doesn't needs to be in docker group (security risk)

kke:
Fixes #1548

### DIFF
--- a/cli/lib/kontena/cli/stacks/build_command.rb
+++ b/cli/lib/kontena/cli/stacks/build_command.rb
@@ -1,4 +1,5 @@
 require_relative 'common'
+require 'shellwords'
 
 module Kontena::Cli::Stacks
   class BuildCommand < Kontena::Command
@@ -11,6 +12,8 @@ module Kontena::Cli::Stacks
     option ['--no-cache'], :flag, 'Do not use cache when building the image', default: false
     option ['--no-push'], :flag, 'Do not push images to registry', default: false
     option ['--no-pull'], :flag, 'Do not attempt to pull a newer version of the image', default: false
+    option ['--[no-]sudo'], :flag, 'Run docker using sudo', hidden: Kontena.on_windows?, environment_variable: 'KONTENA_SUDO', default: false
+
     parameter "[SERVICE] ...", "Services to build"
 
     def execute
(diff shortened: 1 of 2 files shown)


kke:
Is it actually needed for build? Or just push?

nevalla:
I think both. Both commands are executing `docker` command which might require sudo

jalberto:
@kke it's also needed for docker login adn pretty much any docker command (if some docker  subcommands use sudo and other doesn't it will create 2 different contexts)

In other hand, will be handy if --sudo is aglobal option, that way it can be used in an alias

    alias kontena="kontena --sudo"

kke:
@jalberto The build and push here are the only docker commands that the cli is running.

jakolehm:
Why we need an option for sudo.. why not just use `sudo kontena ...`?

jalberto:
@jakolehm because run a ruby gem as sudo is a bad practice for security reasons, also can cause problems with RVM/rbenv and similar tools

cli/lib/kontena/cli/stacks/build_command.rb
```diff
@@ -11,6 +12,8 @@ class BuildCommand < Kontena::Command
     option ['--no-cache'], :flag, 'Do not use cache when building the image', default: false
     option ['--no-push'], :flag, 'Do not push images to registry', default: false
     option ['--no-pull'], :flag, 'Do not attempt to pull a newer version of the image', default: false
+    option ['--[no-]sudo'], :flag, 'Run docker using sudo', hidden: Kontena.on_windows?, environment_variable: 'KONTENA_SUDO', default: false
```
jnummelin: Why do we need the `no-sudo` option when the default is false?

kke: I was hoping someone would ask :)

If you have `KONTENA_SUDO=true` but want to run the command without sudo, you would use `--no-sudo` which overrides the environment variable.

jnummelin: That's a bit weird combo to use I think. :) But it should not harm to have that option.

PR #1675, opened 2017-01-18
If master deploy returns :provider set it to config server.provider

kke:
Sets master config `server.provider` to the :provider returned by provisioning plugins.

This is needed by the `kontena master ssh` in #1205 to figure out if the provider is vagrant.

image

-->

image

### DIFF
--- a/cli/lib/kontena/callbacks/master/deploy/56_set_server_provider_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/56_set_server_provider_after_deploy.rb
@@ -16,9 +16,7 @@ module Kontena
 
         cmd = ['master', 'config', 'set', "server.provider=#{command.result[:provider]}"]
         spinner "Setting Master configuration server.provider to '#{command.result[:provider]}'" do
-          Retriable.retriable do
-            Kontena.run(cmd.shelljoin)
-          end
+          Kontena.run(cmd.shelljoin)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/callbacks/master/deploy/56_set_server_provider_after_deploy.rb
```diff
@@ -0,0 +1,27 @@
...
+        return unless config.current_master
+        return unless config.current_master.name == command.result[:name]
+        return unless command.result[:provider]
+
+        require 'shellwords'
+
+        cmd = ['master', 'config', 'set', "server.provider=#{command.result[:provider]}"]
+        spinner "Setting Master configuration server.provider to '#{command.result[:provider]}'" do
+          Retriable.retriable do
+            Kontena.run(cmd.shelljoin)
```
jakolehm: I'm starting to consider this `Kontena.run` harmful.. Why don't we just call method from the command object?

kke: `Kontena.run` would run callbacks if there were any. Also instance variables inside the method would work. If you directly call a method on a command, then you need to write the method so that they don't use any instance variables (options/flags/params).


The ssh improvement PR #1205 makes it possible to pass an array instead of a string to `Kontena.run` so you don't have to use shellescape/shelljoin.

kke: It would be nice though to have an initializer for commands, so that you could do something like: `c = Kontena::Cli::FooCommand.new(grid: 'foo', json: true, service: 'bar')` and then something like `c.get_foos` instead of `c.get_foos('foo', 'bar', true)`

kke: The retry here is pointless, exceptions won't be raised.

PR #1677, opened 2017-01-18
Show etcd health status

### LINKED ISSUE #1470: kontena grid health for etcd
With the Kontena IPAM based on etcd, the health of the grid's initial nodes and the kontena-etcd cluster is more critical than before; service deploys will not succeed without a functional kontena-etcd and kontena-ipam. Before Kontena 1.0 with the kontena-ipam, this kind of unhealthy grid would still let you deploy services, but the LB linking/registrations would be broken, which you might not even notice.

These kind of etcd cluster failures can be caused by configuration issues, doing e.g. `kontena grid create --initial-size=3` but only creating one node. #1468 addresses the configuration issues by having `kontena node ls` show warnings/errors when trying to run a grid with fewer nodes than required.

However, the `/v1/grid/:grid/nodes` configuration itself is not really enough to verify that the grid's etcd cluster is actually healthy. This would need to actually query the etcd cluster-health on each of the host node agents.

SpComb:
Fixes  #1470

Uses the etcd `/health` [API](https://coreos.com/etcd/docs/latest/v2/admin_guide.html#health-monitoring)

## Agent

Add agent `/etcd/health` RPC

```json
{"health": true}
```

```json
{"health": false}
```
```json
{"error": "proxy: zero endpoints currently available"}
```

## Server

Add server `/v1/nodes/:grid/:node/health` API. This always returns a HTTP 200 with a JSON response of the form:

#### Online healthy node

```json
{
   "name" : "core-01",
   "id" : "XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5",
   "connected" : true,
   "etcd_health" : {
      "error" : null,
      "health" : true
   },
   "node_number" : 1,
   "initial_member" : true
}
```

#### Online broken node

```json
{
   "id" : "XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5",
   "connected" : true,
   "initial_member" : true,
   "node_number" : 1,
   "name" : "core-01",
   "etcd_health" : {
      "health" : null,
      "error" : "Failed to open TCP connection to 172.17.0.1:2379 (Connection refused - connect(2) for \"172.17.0.1\" port 2379)"
   }
}
```

#### Online unhealthy node

```json
{
   "name" : "core-02",
   "initial_member" : false,
   "connected" : true,
   "etcd_health" : {
      "health" : null,
      "error" : "proxy: unable to get response from 1 endpoint(s)"
   },
   "node_number" : 2,
   "id" : "CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q"
}
```

#### Offline node

```json
{
   "node_number" : 2,
   "initial_member" : false,
   "name" : "core-02",
   "connected" : false,
   "etcd_health" : {
      "health" : null,
      "error" : null
   },
   "id" : "CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q"
}
```
* Add `kontena etcd health [NODE]` CLI

    ```
    Node core-01 is healthy
    Node core-02 is unhealthy: proxy: unable to get response from 1 endpoint(s)
    ``` 

## Testing

* `sudo ip link set weave down`
* `docker stop kontena-etcd` 
* `docker pause kontena-etcd`


SpComb:
Once #1468 is in, I can extend `kontena node health` to also use this API.

The `/v1/nodes/:grid/:node/health` API could then also be extended to include other non-etcd health checks.

I don't think adding this to `kontena node ls` is necessarily a good idea though, because it's O(N) RPCs.

jnummelin:
```
$ docker pause kontena-etcd
kontena-etcd

$ k etcd health
 [error] Internal Server Error
```

I'd guess this is due to some timeout but the error should be something more understandable.

SpComb:
> On agent side the timeout is seen in the logs:

The agent's etcd client timeout is something on the order of 90s, so the server RPC client times out. Fixed that to return HTTP 503 and handle that in the CLI.

SpComb:
Rebased, improved error handling, health icons; added agent/server/cli specs.

SpComb:
BTW the `Etcd::Health` monkey-patching here should eventually be replaced with https://github.com/SpComb/kontena-etcd/issues/7

server/app/views/v1/host_nodes/health.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.id @node.node_id
+json.name @node.name
+json.etcd @etcd_health
```
jakolehm: Is this an object or what?

agent/lib/kontena/rpc/etcd_api.rb
```diff
@@ -55,6 +55,13 @@ def delete(key, opts = {})
         {error: exc.message}
       end
 
+      # @return [Hash] {'health' => Boolean} or {'error' => String}
+      def health
+        {health: etcd.health}
+      rescue Etcd::Health::Error => error
+        {error: error.message}
```
jakolehm: Should error case have `health: false` ?

server/app/views/v1/host_nodes/health.json.jbuilder
SpComb: This is what the agent `/etcd/health` RPC returns, so either `{health: etcd.health}` or `{error: error.message}`.

agent/lib/kontena/rpc/etcd_api.rb
SpComb: Not sure, I think it depends on if it's possible for the `etcd` API `GET /health` to also return `{"health": false}`?

server/app/views/v1/host_nodes/health.json.jbuilder
SpComb: The reason this is a separate JSON key is that this response would be extended with other health signals in the future, like `{"weave": ..., "etcd": ..., "ipam": ...}` etc.

agent/lib/kontena/rpc/etcd_api.rb
SpComb: Looking at the etcd server implementation, yes, it's also possible to get a HTTP 503 `{"health": "false"}` response.

Never seen that when testing, not sure how the `Etcd::Health#health` code handles it ATM.

server/app/views/v1/host_nodes/health.json.jbuilder
jakolehm: Yes, etcd key is good but the content of that object is somewhat confusing to me.

SpComb:
Fixed the etcd health client, unified the server API responses, fixed the CLI.

PR #1678, opened 2017-01-19
Stack service_instances resolver

### LINKED ISSUE #1671: Deploying stack with daemon strategy and scale=0
I'm migrating my galera cluster to stacks and I wanted to launch the stack initially with the "node" service scaled to 0 although the .yml file specifies 1. I was going to then bootstrap it with the "seed" service and finally return the "node" service back to "scaling: 1". Steps:

    $ kontena stack install kontena-galera-cluster.yml
    $ kontena stack build -f kontena-galera-cluster.yml
    $ kontena service scale galera-cluster/node 0
    $ kontena service show galera-cluster/node
    # showed scaling: 0
    $ kontena stack deploy galera-cluster

Expected result: the "node" service would be deployed to 0 nodes.
Actual result: the "node" service was deployed to each node as if "scaling: 1".

After it completed deploying I noticed that "scaling:" had been reverted to 1. So is this simply because deploying a stack overrides the services?

I'm looking for a way to scale down a service that is part of a stack before the stack is deployed, or perhaps deploy only a subset of services from a stack.

Thanks!

jakolehm:
This resolver is handy if you don't want to overwrite service instances (scaling) when doing a stack upgrade.

Example:

```yaml
stack: jakolehm/mariadb-galera-cluster
variables:
  seed_instances:
    type: integer
    default: 1
    from: 
      service_instances: seed
services:
  seed:
    image: colinmollenhour/mariadb-galera-swarm:latest
    stateful: true
    instances: $seed_instances
```

Fixes #1671

### DIFF
--- a/cli/lib/kontena/cli/stacks/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/show_command.rb
@@ -17,8 +17,12 @@ module Kontena::Cli::Stacks
       show_stack(name)
     end
 
+    def fetch_stack(name)
+      client.get("stacks/#{current_grid}/#{name}")
+    end
+
     def show_stack(name)
-      stack = client.get("stacks/#{current_grid}/#{name}")
+      stack = fetch_stack(name)
 
       puts "#{stack['name']}:"
       puts "  state: #{stack['state']}"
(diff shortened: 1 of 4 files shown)


kke:
Hmm, nice. This can be used to generate a string like `mongodb://mongodb-1,mongodb-2,mongodb-3:27017/foo` when using liquid.

SpComb:
> This can be used to generate a string like mongodb://mongodb-1,mongodb-2,mongodb-3:27017/foo when using liquid

Only when upgrading a stack that has already been installed to the server, though.

jakolehm:
> Only when upgrading a stack that has already been installed to the server, though.

But you can combine prompt (or default instance count)... then it works also on initial install/deploy.

colinmollenhour:
Interesting solution! Thanks!

PR #1689, opened 2017-01-20
Improve upgrade docs

jakolehm:
+ fix minor doc bugs


nevalla:
Maybe not part of this PR but maybe we should add also instruction how to update CLI.

PR #1692, opened 2017-01-21
docker-compose restart to unless stopped

matti:

PR #1694, opened 2017-01-21
Docs/compose agent to version 2

matti:
Agent is defined in old compose format (master is as version 2 couple of lines up) -- also see #1692


jakolehm:
Thanks!

PR #1696, opened 2017-01-21
Add Summary of Using Kontena

nevalla:


docs/using-kontena/summary.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Summary
+toc_order: 1
+---
```
jakolehm: Do we need this anymore with gitbook?

nevalla: Nope. Removed

PR #1699, opened 2017-01-22
Stop container health check also on kill event

nevalla:
When container is stopped (`docker stop <container>`), Docker will send container event with status `kill`. Currently agent will stop container health check worker only with `die` event.

### DIFF
--- a/agent/lib/kontena/workers/health_check_worker.rb
+++ b/agent/lib/kontena/workers/health_check_worker.rb
@@ -11,7 +11,7 @@ module Kontena::Workers
     finalizer :terminate_workers
 
     START_EVENTS = ['start']
-    STOP_EVENTS = ['die']
+    STOP_EVENTS = ['die', 'kill']
     ETCD_PREFIX = '/kontena/log_worker/containers'
 
     ##
--- a/agent/spec/lib/kontena/workers/health_check_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/health_check_worker_spec.rb
@@ -62,6 +62,12 @@ describe Kontena::Workers::HealthCheckWorker do
   end
 
   describe '#on_container_event' do
+    it 'stops check on kill' do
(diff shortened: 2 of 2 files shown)


nevalla:
@jakolehm @jnummelin Do you see any side effects because of this?

nevalla:
Hmm... or should we only track `kill` event?

jnummelin:
Hmm, stopping health check on both `die` and `kill` does not really make sense to me. At least according to Docker event state here https://github.com/docker/docker/issues/12164#issuecomment-94940333, all containers should get `die` event when they are being stopped. I.e. when you stop a container, you should get events `kill`-> `die` -> `stop`.

SpComb:
It definitely makes more sense to stop health-checking on `kill`, because the process may stop responding to health checks while in the process of shutting down, and we don't want to restart it while it's shutting down...

Even with the `kill` coming before the `die`, the `Kontena::Workers::ContainerHealthCheckWorker` is still racy with the `Kontena::ServicePods::Stopper/Terminator`, though... the agent should not be health-checking once it has stopped the container, but this process currently relies on the `Docker::Container.stop` -> Docker event -> `Celluloid::Notifications` `container:event`  -> `Celluloid::Actor.kill` chain completing before the `ContainerHealthCheckWorker` periodic interval kicks in, and it's just a question of timing.

agent/lib/kontena/workers/health_check_worker.rb
```diff
@@ -11,7 +11,7 @@ class HealthCheckWorker
     finalizer :terminate_workers
 
     START_EVENTS = ['start']
-    STOP_EVENTS = ['die']
+    STOP_EVENTS = ['die', 'kill']
     ETCD_PREFIX = '/kontena/log_worker/containers'
```
SpComb: Mmm stale copy-pasta :)

PR #1702, opened 2017-01-23
Exit with error if deploy fails during stack upgrade

### LINKED ISSUE #1637: Stack upgrade --deploy return code should not be zero
When the command `kontena stack upgrade --deploy stack` has one or more errors, the return code should be non-zero. It's currently zero.

```
* Deploying stack stack. 
        halting deploy of grid/stack/service, one or more instances failed
```

The return code can be utilized in a build server to fail the build if the stack upgrade & deploy command fails.

Relevant code:
https://github.com/kontena/kontena/blob/master/cli/lib/kontena/cli/stacks/upgrade_command.rb#L24

Also, I would be interested to know if that error message means that every other service in the stack deployed successfully.

kke:
Fixes #1637

### DIFF
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -31,7 +31,7 @@ module Kontena::Cli::Stacks
 
       if deploy?
         result = Kontena.run(['stack', 'deploy', name])
-        unless result.zero?
+        unless result.to_i.zero?
           msg = "Deploy failed"
           msg << ": #{$!.message}" if $!
           exit_with_error msg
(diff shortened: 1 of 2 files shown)


SpComb:
:-1: 

This fixes this specific case, but there's a lot of similar cases, pretty much everywhere that `Kontena.run(...)` is used...

kke:
It could have the option `raise: true` that would let the cmd crash normally.

jnummelin:
I agree with @SpComb, there should be a common way to make this happen instead of patching each command separately.

panuhorsmalahti:
Why would this not be needed anymore?

SpComb:
Closing in favor of #2209 that fixes all cases of this issue in general.

PR #1706, opened 2017-01-23
Prettier initializer method for commands

kke:
Instead of:

```ruby
>> cmd = Kontena::Cli::Nodes::ListCommand.new('kontena') # (.new takes the command name for using in help output, not arguments.)
>> cmd.run(['--all'])
```

or more specifically:

```ruby
# def get_nodes(all = false)
#  nodes = client.get("foofoo")['nodes']
#   unless all
#     nodes.reject! { |n| !n['connected']}
#   end
#   nodes
# end

>>> cmd = Kontena::Cli::Nodes::ListCommand.new('kontena')
>>> cmd.get_nodes(false)
```

You can now do:

```ruby
# def get_nodes
#   nodes = client.get("foofoo")['nodes']
#   unless all?
#     nodes.reject! { |n| !n['connected']}
#   end
#   nodes
# end

>> cmd = Kontena::Cli::Nodes::ListCommand.new(all: true)
>> nodes = cmd.get_nodes
```

Because something like this does not make sense:

```ruby
class FooCommand
  attr_accessor :name

  def initialize(name)
    @name = name
  end

  def execute
    do_foo(name)
  end

  # code smell: method does nothing with the host object
  def do_foo(name)
    foo = Foo.find(name)
    foo.do_foo
  end
end
```

Instance methods should use instance variables:

```ruby
class FooCommand
  attr_accessor :name

  def initialize(name)
    @name = name
  end

  def execute
    do_foo
  end

  def do_foo
    foo = Foo.find(name)
    foo.do_foo
  end
end
```

### DIFF
--- a/cli/lib/kontena/command.rb
+++ b/cli/lib/kontena/command.rb
@@ -23,7 +23,7 @@ class Kontena::Command < Clamp::Command
       return if self.has_subcommands?
       return if self.callback_matcher
 
-      name_parts = self.name.split('::')[-2, 2]
+      name_parts = self.name.split('::')[-2, 2] || []
 
       unless name_parts.compact.empty?
         # 1: Remove trailing 'Command' from for example AuthCommand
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/command.rb
```diff
@@ -23,7 +23,7 @@ def finalize
       return if self.has_subcommands?
       return if self.callback_matcher
 
-      name_parts = self.name.split('::')[-2, 2]
+      name_parts = self.name.split('::')[-2, 2] || []
```
kke: (specs have commands that are not nested Module::Module::Class but just Class)

cli/spec/kontena/command_spec.rb
```diff
@@ -0,0 +1,55 @@
+require_relative '../spec_helper'
+require 'kontena/command'
+
+describe Kontena do
+  let(:subject) do
+    class TestCommand < Kontena::Command
```
SpComb: Use `Class.new(Kontena::Command) do ... ` for spec classes, because the `TestCommand` pollutes the global ruby namespace.

IIRC this will fail if you have two specs that both try and use `class TestCommand`.

kke: that's why it removes them in the after hook.

SpComb: Uh, ok, but maybe the `Class.new(...) do ... end` is still a better idea :)

SpComb: OTOH I guess this test case is sensitive to the `class.name`...? I'm not sure how that goes, I've seen it have some really weird magic where it picks up the variable name from `foo_class = Class.new ...`..

SpComb:
This PR on its own seems a little odd, but I vaugely see the point of using the CLI commands as some kind of pseudo-mutations.

Perhaps also refactor the existing `Kontena.run(...)` things to use this as part of the same PR, or whatever the intention was?

jnummelin:
@kke Still valid?

kke:
Perhaps but people didn't like it.

PR #1710, opened 2017-01-23
Improve stack deploy progress output

jakolehm:
```
$ kontena stack install --deploy foo.yml
 [done] Creating stack foo      
 [done] Triggering deployment of stack foo     
 [done] Deploying service foo     
 [done] Deploying service foo1     
 [done] Deploying service foo2     
 [done] Deploying service foo3     
```


jakolehm:
@kke open a separate issue for that. I like the idea because you really don't want to wait always... and hitting ctrl+c will show "fail" which is a bit misleading.

PR #1718, opened 2017-01-25
Add docker-compose support table

nevalla:


docs/references/docker-compose-support.md
```diff
@@ -0,0 +1,62 @@
+---
+title: docker-compose.yml comparison
+toc_order: 4
```
jakolehm: We don't need these anymore?

nevalla: No

PR #1726, opened 2017-01-26
cli: kontena service exec

### LINKED ISSUE #1563: kontena container exec is too complicated to use

SpComb:
Fixes #1563

This adds a new `kontena service exec SERVICE CMD...` command to complement `kontena container exec NODE/CONTAINER`.

The `kontena container` commands are difficult to use with the unpredictable node name prefix, and the changing container naming schemas in Kontena 0.16 -> 1.0 -> #1494.

This PR also removes the undocumented and broken `POST /v1/services/:grid_name/:stack_name/:service_name/exec` API.

## Examples

For a `redis` service with four instances:

```
⊝ redis                                                      4 / 4      no       running                                                      
```

#### `kontena service exec redis hostname`
```
redis-1.development.kontena.local
```

#### `kontena service exec --instance=3 redis hostname`
```
redis-3.development.kontena.local
```

#### `kontena service exec --all redis hostname`
```
 [done] Executing command on null-redis-1     
redis-1.development.kontena.local
 [done] Executing command on null-redis-2     
redis-2.development.kontena.local
 [done] Executing command on null-redis-3     
redis-3.development.kontena.local
 [done] Executing command on null-redis-4     
redis-4.development.kontena.local
```

#### `kontena service exec --silent --all redis hostname`
```
redis-1.development.kontena.local
redis-2.development.kontena.local
redis-3.development.kontena.local
redis-4.development.kontena.local
```

### Exit code

The exit code is propagated by the `kontena-cli`:

#### `kontena service exec redis ls /nonexist || echo fail`

```
ls: 
cannot access /nonexist: No such file or directory
fail
```

#### `kontena service exec --all --shell redis '[ $KONTENA_SERVICE_INSTANCE_NUMBER != 5 ]' && echo yay`
```
 [done] Executing command on null-redis-1     
 [done] Executing command on null-redis-2     
 [done] Executing command on null-redis-3     
 [done] Executing command on null-redis-4     
yay
```

#### `kontena service exec --all --shell redis '[ $KONTENA_SERVICE_INSTANCE_NUMBER != 2 ]' && echo yay`
```
 [done] Executing command on null-redis-1     
 [fail] Executing command on null-redis-2     
```

#### `kontena service exec --all --skip --shell redis '[ $KONTENA_SERVICE_INSTANCE_NUMBER != 2 ]' && echo yay`
```
 [done] Executing command on null-redis-1     
 [fail] Executing command on null-redis-2     
 [done] Executing command on null-redis-3     
 [done] Executing command on null-redis-4     
```

#### `kontena service exec --instance=2 redis hostname`
```
 [error] Service redis container null-redis-2 is not running, it is paused
```

####  `kontena service exec --all redis hostname`
```
 [done] Executing command on null-redis-1     
redis-1.development.kontena.local
 [warn] Service redis container null-redis-2 is paused, skipping
 [done] Executing command on null-redis-3     
redis-3.development.kontena.local
 [done] Executing command on null-redis-4     
redis-4.development.kontena.local
```

#### `kontena service stop redis && kontena service exec redis hostname`
```
 [error] Service redis does not have any running containers
```

### Optional shell command

This does not execute the command using a shell by default, unlike `kontena container exec`

####  `kontena service exec redis touch 'foo bar' && kontena service exec redis ls -l`
```
total 12
-rw-r--r--. 1 redis redis 76 Jan 26 14:51 dump.rdb
-rw-r--r--. 1 root  root   0 Jan 26 14:55 foo bar
```

#### `kontena service exec --shell redis 'touch foo bar' && kontena service exec redis ls -l`
```
-rw-r--r--. 1 root  root   0 Jan 26 15:04 bar
-rw-r--r--. 1 redis redis 76 Jan 26 14:51 dump.rdb
-rw-r--r--. 1 root  root   0 Jan 26 15:04 foo
```

#### `kontena service exec --shell redis 'for name in *; do echo "have file: $name"; done'`
```
have file: bar
have file: dump.rdb
have file: foo
```

### DIFF
--- a/cli/lib/kontena/cli/services/exec_command.rb
+++ b/cli/lib/kontena/cli/services/exec_command.rb
@@ -12,8 +12,9 @@ module Kontena::Cli::Services
     option ["-i", "--instance"], "INSTANCE", "Exec on given numbered instance, default first running" do |value| Integer(value) end
     option ["-a", "--all"], :flag, "Exec on all running instances"
     option ["--shell"], :flag, "Execute as a shell command"
-    option ["--[no-]spinner"], :flag, "Execute with interactive output", attribute_name: :with_spinner, default: nil
     option ["--skip"], :flag, "Skip failed instances when executing --all"
+    option ["--silent"], :flag, "Do not show exec status"
+    option ["--verbose"], :flag, "Show exec status"
 
     requires_current_master
     requires_current_grid
@@ -28,7 +29,7 @@ module Kontena::Cli::Services
 
       stdout = stderr = exit_status = nil
 
-      if with_spinner? || (@with_spinner.nil? && all?)
+      if !silent? && (verbose? || all?)
         spinner "Executing command on #{container['name']}" do
           stdout, stderr, exit_status = client.post("containers/#{container['id']}/exec", {cmd: cmd})
 
--- a/cli/lib/kontena/cli/services/exec_command.rb
+++ b/cli/lib/kontena/cli/services/exec_command.rb
@@ -35,6 +35,7 @@ module Kontena::Cli::Services
 
     def execute
       service_containers = client.get("services/#{parse_service_id(name)}/containers")['containers']
+      service_containers.sort_by! { |container| container['instance_number'] }
       running_containers = service_containers.select{|container| container['status'] == 'running' }
 
       if running_containers.empty?
--- a/cli/lib/kontena/cli/services/exec_command.rb
+++ b/cli/lib/kontena/cli/services/exec_command.rb
@@ -13,6 +13,7 @@ module Kontena::Cli::Services
     option ["-a", "--all"], :flag, "Exec on all running instances"
     option ["--shell"], :flag, "Execute as a shell command"
     option ["--[no-]spinner"], :flag, "Execute with interactive output", attribute_name: :with_spinner, default: nil
+    option ["--skip"], :flag, "Skip failed instances when executing --all"
 
     requires_current_master
     requires_current_grid
@@ -40,9 +41,9 @@ module Kontena::Cli::Services
       stdout.each do |chunk| $stdout.write chunk end
       stderr.each do |chunk| $stderr.write chunk end
 
-      exit exit_status if exit_status != 0
+      exit exit_status if exit_status != 0 && !skip?
 
-      return true
+      return exit_status == 0
     end
 
     def execute
@@ -55,13 +56,17 @@ module Kontena::Cli::Services
       end
 
       if all?
+        ret = true
         service_containers.each do |container|
           if container['status'] == 'running'
-            exec_container(container)
+            if !exec_container(container)
+              ret = false
+            end
           else
             warning "Service #{name} container #{container['name']} is #{container['status']}, skipping"
           end
         end
+        return ret
       elsif instance
         if !(container = service_containers.find{|container| container['instance_number'] == instance})
           exit_with_error "Service #{name} does not have container instance #{instance}"
--- a/cli/lib/kontena/cli/services/exec_command.rb
+++ b/cli/lib/kontena/cli/services/exec_command.rb
@@ -12,6 +12,7 @@ module Kontena::Cli::Services
     option ["-i", "--instance"], "INSTANCE", "Exec on given numbered instance, default first running" do |value| Integer(value) end
     option ["-a", "--all"], :flag, "Exec on all running instances"
     option ["--shell"], :flag, "Execute as a shell command"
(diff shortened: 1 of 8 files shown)


SpComb:
Not really sure how the `cmd_list` should be handled when using `--shell`. The existing `kontena container exec` command uses `cmd: ["sh", "-c", Shellwords.join(cmd_list)]}`, but I feel like that's a bit weird...

```
$ kontena container exec core-02/null-redis-1 'for name in *; do echo "have file: $name"; done'
sh: 1:  for name in *; do echo "have file: $name"; done: not found
```

It quotes the entire given string as a single shell token, and thus tries to run it as a command... not sure what kind of syntax you would use to get `kontena container exec` to actually run some kind of shell command?

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -0,0 +1,63 @@
...
+    option ["-a", "--all"], :flag, "Exec on all running instances"
+    option ["--shell"], :flag, "Execute as a shell command"
+
+    requires_current_master
+    requires_current_grid
+
+    # Exits if exec returns with non-zero
+    def exec_container(container)
+      if shell?
+        cmd = ['sh', '-c', cmd_list.join(' ')]
```
SpComb: Not entirely what the right way to do this shell-command joining is.

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -0,0 +1,63 @@
...
+      else
+        cmd = cmd_list
+      end
+
+      stdout, stderr, exit_status = client.post("containers/#{container['id']}/exec", {cmd: cmd})
+
+      STDOUT.puts stdout unless stdout.empty?
+      STDERR.puts stderr unless stderr.empty?
```
SpComb: There's some weirdness with the output in the `ls /nonexist` example.

The existing `kontena container exec` code does `stdout.join(" ")`, but I couldn't immediately figure out what the the format of the exec response actually is. Looks like it comes directly from the `docker-api` gem's `Docker::Container.exec`.

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -0,0 +1,63 @@
...
+      exit exit_status if exit_status != 0
```
SpComb: Thinking this could output which container failed, if we're running with `--all`?

jnummelin: Sound like good idea to show which of the containers failed.

SpComb: Currently this stops execing on the first container that fails.

jnummelin: That's reasonable.

SpComb: Added a default `--spinner`that shows which container failed to exec.

SpComb:
Added shell spinner that's enabled by default for `--all`, with support for `--skip`.

cli/lib/kontena/cli/services/exec_command.rb
SpComb: Fixed

kke:
`--spinner` arg seems a bit hacky.

For some commands `--silent` will not show spinners and for some commands they're shown when `--verbose`. (the vspinner method automatically picks these up if the option has been declared)

SpComb:
> the vspinner method automatically picks these up if the option has been declared

Combining the spinner with the command output needs some trickery, like actually exiting the block before dumping the command output, but still showing a `[fail]`... I think I'll skip the `vspinner` for now.

SpComb:
Happaned to also notice a `POST /v1/services/:grid_name/:stack_name/:service_name/exec` API on the server, but I suspect that API is broken...

```ruby
@grid_service.containers.where(name: "#{@grid_service.name}-#{data['instance']}")
```

is missing the stack prefix from the container name.

cli/lib/kontena/cli/services/exec_command.rb
SpComb: But this seems to work fine. Each argument is a shell token, separated by anything in the shells's `$IFS` (internal field separator).

PR #1734, opened 2017-01-30
Allow usage of local envs that have been set through to: env

kke:
Only variable names were accepted. Now it also accepts env variables that are set by a `to:` setter.

Now this works:

```
variables:
  foo:
    type: string
    from:
      env: BAR
    to:
      env: BAR
services:
  mysql:
    image: $BAR
```


SpComb:
Please also fix the usage of `to: env` in the docs:

* https://github.com/kontena/kontena/blob/master/docs/references/kontena-yml.md#variables
* https://github.com/kontena/kontena/blob/master/docs/references/kontena-yml-variables.md

PR #1735, opened 2017-01-30
Add user as master_admin after init-cloud

### LINKED ISSUE #1733: Init cloud command should add current user to master as master_admin role

kke:
Fixes #1733


jakolehm:
Can we have a test for this?

kke:
Now it will add to current grid unless it's nil.

jnummelin:
Seems to work:
```
$ k whoami
Master: mac
URL: http://localhost
Grid: subnet-test
User: admin

$ k master init-cloud
...

$ k whoami
Master: mac
URL: http://localhost
Grid: subnet-test
User: jussi@kontena.io
```
Some specs for the callback would be nice. :)

PR #1737, opened 2017-01-30
Deploy stack by default on install/upgrade

jakolehm:


SpComb:
:+1: although this could also be linked to some issue on implementing a `kontena stack deploy --noop`, that shows what changed would be deployed on upgrade.

kke:
(--noop = --dry-run)

SpComb:
Docs need updating:

* https://github.com/kontena/kontena/blob/master/docs/getting-started/quick-start.md#step-4-deploy-your-first-application-stack

    > `$ kontena stack install --deploy kontena.yml`

* https://github.com/kontena/kontena/blob/master/docs/using-kontena/stacks.md#deploying-stacks

    > Use the --deploy flag to simultaneously deploy the stack services to the grid [...] If you omit the kontena stack install --deploy flag, then you must run kontena stack deploy wordpress-red separately.

jakolehm:
Something is off here.. let's see if I can cleanup this with a new rebase.

PR #1742, opened 2017-01-30
Fix service volume update

jakolehm:
Allow to update service volume definitions, except when service is stateful and added volume is not a bind-mounted or a named volume.

### DIFF
--- a/server/app/mutations/grid_services/update.rb
+++ b/server/app/mutations/grid_services/update.rb
@@ -18,7 +18,6 @@ module GridServices
       if self.links
         validate_links(self.grid_service.grid, self.grid_service.stack, self.links)
       end
-
       if self.strategy && !self.strategies[self.strategy]
         add_error(:strategy, :invalid_strategy, 'Strategy not supported')
       end
@@ -27,6 +26,12 @@ module GridServices
       end
       if self.secrets
         validate_secrets_exist(self.grid_service.grid, self.secrets)
+      end
+      if self.volumes && self.grid_service.stateful?
+        changed_volumes = self.volumes.select { |v| !self.grid_service.volumes.include?(v) }
+        if changed_volumes.any? { |v| !v.include?(':') }
+          add_error(:volumes, :invalid, 'Adding a non-named volume is not supported to a stateful service')
+        end
       end
     end
 
@@ -53,6 +58,7 @@ module GridServices
       attributes[:devices] = self.devices if self.devices
       attributes[:deploy_opts] = self.deploy_opts if self.deploy_opts
       attributes[:health_check] = self.health_check if self.health_check
+      attributes[:volumes] = self.volumes if self.volumes
 
       if self.links
         attributes[:grid_service_links] = build_grid_service_links(
--- a/server/app/mutations/grid_services/update.rb
+++ b/server/app/mutations/grid_services/update.rb
@@ -27,10 +27,15 @@ module GridServices
       if self.secrets
         validate_secrets_exist(self.grid_service.grid, self.secrets)
       end
-      if self.volumes && self.grid_service.stateful?
-        changed_volumes = self.volumes.select { |v| !self.grid_service.volumes.include?(v) }
-        if changed_volumes.any? { |v| !v.include?(':') }
-          add_error(:volumes, :invalid, 'Adding a non-named volume is not supported to a stateful service')
+      if self.grid_service.stateful?
+        if self.volumes_from && self.volumes_from.size > 0
+          add_error(:volumes_from, :invalid, 'Cannot combine stateful & volumes_from')
+        end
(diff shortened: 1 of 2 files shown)


jnummelin:
Why not allow to update anonymous volumes for stateful service?

jakolehm:
Because anonymous volume cannot be updated to a data-volume container. Bind & named volume are "outside" so they can be added later.

jnummelin:
Technically nothing stops to add anonymous volumes to the service container. The data-volume container is immutable. But as we discussed, it might be too confusing for the users as they will not see what data is persisted on the data-volume and what are only on the local volumes. I think this should be explained in the documentation.

server/app/mutations/grid_services/update.rb
```diff
@@ -28,6 +27,17 @@ def validate
...
+        if self.volumes_from && self.volumes_from.size > 0
+          add_error(:volumes_from, :invalid, 'Cannot combine stateful & volumes_from')
+        end
+        if self.volumes
+          changed_volumes = self.volumes.select { |v| !self.grid_service.volumes.include?(v) }
+          if changed_volumes.any? { |v| !v.include?(':') }
+            add_error(:volumes, :invalid, 'Adding a non-named volume is not supported to a stateful service')
+          end
+        end
+      end
```
jnummelin: Shouldn't these be also updated in the create mutation to have same "rules" in both create and update? This basically means that it should be separate method in `common`.

jakolehm: Only `volumes_from` is the same in both cases.. but yes, that can be in `common`.

SpComb:
:+1: on fixing the `stack upgrade --deploy` after editing the service `volumes_from`:
```
core@core-01 ~ $ docker inspect -f '{{.HostConfig.VolumesFrom}}' wordpress.volumes-test-1
[weavewait-1.8.2:ro]
core@core-01 ~ $ docker inspect -f '{{.HostConfig.VolumesFrom}}' wordpress.volumes-test-1
[wordpress.wordpress-1 weavewait-1.8.2:ro]
```

server/app/mutations/grid_services/update.rb
SpComb: Looks like the existing `GridServices::Create#validate` already enforces this:

```
 [fail] Creating stack wordpress      
 [error] {"services"=>"Service create failed for service 'wordpress': Cannot combine stateful & volumes_from"}
```

server/app/mutations/grid_services/update.rb
```diff
@@ -28,6 +27,17 @@ def validate
...
         validate_secrets_exist(self.grid_service.grid, self.secrets)
       end
+      if self.grid_service.stateful?
+        if self.volumes_from && self.volumes_from.size > 0
+          add_error(:volumes_from, :invalid, 'Cannot combine stateful & volumes_from')
+        end
+        if self.volumes
+          changed_volumes = self.volumes.select { |v| !self.grid_service.volumes.include?(v) }
+          if changed_volumes.any? { |v| !v.include?(':') }
+            add_error(:volumes, :invalid, 'Adding a non-named volume is not supported to a stateful service')
```
SpComb: minor wording fix: "Adding a non-named volume to a stateful service is not supported"

server/spec/mutations/grid_services/update_spec.rb
```diff
@@ -93,6 +93,111 @@
...
+        it 'allows to add bind mounted volume' do
+          outcome = described_class.new(
+            grid_service: stateful_service,
+            volumes: ['/data', '/foo:/foo']
+          ).run
+          expect(outcome.success?).to be_truthy
+          expect(outcome.result.volumes).to eq(['/data', '/foo:/foo'])
+        end
+
+        it 'allows to remove a volume' do
```
SpComb: Removing an anonymous `volumes: /var/lib/test` from a stateful service doesn't actually remove volumes present when it was initially deployed, since those anonymous volumes will still be configured on the `*-volumes` container, and will still be visible to the re-deployed service container.

```
core@core-01 ~ $ docker inspect -f '{{.Config.Volumes}}' wordpress.wordpress-1
map[/var/www/html:{}]
core@core-01 ~ $ docker inspect -f '{{range .Mounts}}{{.Destination}} {{end}}' wordpress.wordpress-1-volumes
/var/lib/test /var/www/html 
core@core-01 ~ $ docker exec wordpress.wordpress-1 ls -la /var/lib/test/
total 20
drwxr-xr-x. 2 root root 4096 Feb  1 13:29 .
drwxr-xr-x. 1 root root 4096 Feb  1 13:29 ..
-rw-r--r--. 1 root root    0 Feb  1 13:29 asdf
```

Probably not much of an issue, but that's how it behaves :)

PR #1745, opened 2017-01-31
Vault ssl cert resolver for stacks

jakolehm:
Can be used to multiselect certificates from Kontena Vault.

Example:
```
variables:
  lb_certs:
    type: array
    required: false
    from:
      vault_cert_prompt: Select SSL cert secrets
services:
  lb:
    image: kontena/lb:latest
    ports:
      - 80:80
      - 443:443
    secrets:
      # {% for cert in lb_certs %}
      - secret: {{ cert }}
        name: SSL_CERTS
        type: env
      # {% endfor %}
```

screen shot 2017-01-30 at 8 55 59


SpComb:
> secrets: # {% for cert in lb_certs %} - secret: {{ cert }} name: SSL_CERTS type: env # {% endfor %}

I would be curious to know how this kind of configuration with multiple identically named envs is implemented.

kke:
Docs?

jakolehm:
Added some docs about this resolver.

PR #1747, opened 2017-01-31
Fix cli specs to use an explicit client instance_double

SpComb:
Using `let :client do spy() end` means that specs could mock out the wrong `allow(client).to receive(...)` expectations and the specs would either fail with weird errors or pass through using the resulting `spy` mock return value. The specs would also pass for code that performed additional requests (like `client.delete(...)`) not described in the specs.

With the `instance_double`, any mis-expectations in the specs will be immediately obvious in the form of `#<InstanceDouble(Kontena::Client) :client> received unexpected message ... with ...` failures. This makes the CLI specs more robust and easier to debug while developing.

* Removes unnecessary specs on "requires api url" and "requires token"

* Fixes some missing API requests in some of the specs, but nothing that would have hidden any bugs in the CLI code that I could spot.

* Minor improvements to the `kontena version` and `kontena grid trusted-subnet list` specs to also check output.


kke:
There's RequirementsHelper that has:

`expect_to_require_current_master`
`expect_to_require_current_master_token`
`mock_current_master`

SpComb:
> There's RequirementsHelper that has:
> 
> `expect_to_require_current_master`
> `expect_to_require_current_master_token`

Hmm... those are for the new CLI classes using the new class-level `requires_*` annotations. Using those would be a refactoring for all of the old-style CLI classes that do the instance-level `require_*`, so not part of this PR.

I'm assuming that the separate test cases for `expect(subject).to receive(:require_api_url)` are kind of pointless, and very difficult to test separately, since they're part of the completel `execute` function.

> mock_current_master

That overlaps with the existing `ClientHelpers`... that looks weird for those specs that incldue both, like the `Kontena::Cli::Stacks::DeployCommand` spec...

PR #1751, opened 2017-01-31
Read variable defaults from master when running stack upgrade

### LINKED ISSUE #1551: Read stack variable defaults from master or a file
When installing or upgrading read defaults for variable answers from the previous answers stored to stack info on master or a settings file.

Also add the possibility to answer questions and write the answers to such a file.

Also maybe something like `stack reconfigure` (which does almost the same as upgrade, re-answer questions and create a new revision)

kke:
Fixes #1551

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -59,8 +59,8 @@ module Kontena::Cli::Stacks
       @stack_name ||= self.name || stack_name_from_yaml(filename)
     end
 
-    def reader_from_yaml(filename, from_registry: false, name: nil, values: nil)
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -59,17 +59,17 @@ def stack_name
       @stack_name ||= self.name || stack_name_from_yaml(filename)
     end
 
-    def reader_from_yaml(filename, from_registry: false, name: nil, values: nil)
-      reader = Kontena::Cli::Stacks::YAML::Reader.new(filename, from_registry: from_registry, values: values)
+    def reader_from_yaml(filename, from_registry: false, name: nil, values: nil, defaults: nil)
```
jakolehm: Where is this used?

kke: `stack validate` needs the reader.

PR #1754, opened 2017-01-31
Fix Ubuntu xenial packaging to support dpkg-reconfigure

### LINKED ISSUE #1750: Ubuntu xenial dpkg-reconfigure is broken
The Ubuntu xenial `kontena-server` and `kontena-agent` dpkg `postinstall` scripts only work for an initial install, where they configure the commented defaults:

```sh
# Fetching configuration from debconf
db_get kontena-agent/server_uri
URI="KONTENA_URI=${RET}"
sed -i -r "s#\#KONTENA_URI=.*#${URI}#" /etc/kontena-agent.env

db_get kontena-agent/grid_token
TOKEN="KONTENA_TOKEN=${RET}"
sed -i -r "s#\#KONTENA_TOKEN=.*#${TOKEN}#" /etc/kontena-agent.env
```

Using `dpkg-reconfigure kontena-*` to re-run the postinst script will prompt with the previous value stored in debconf (?), but using the prompt to change the value will not have any effect on the actual `/etc/kontena-*.env` configuration or running systemd services.

Does this debconf prompting also happens on package upgrades?

Best option may be to just skip the debconf prompts for anything except the initial install. Alternatively, we have to write out the shell-gunk to load the existing `.env` values, and the enhance the sed-magic to update the values.

SpComb:
Fixes #1750

This reads any existing `/etc/kontena-*.env` configuration during the package `config` step, and uses that to set the debconf values that will be shown in the `dpkg-reconfigure` prompt.

The `postinst` step can then also replace any existing `KONTENA_FOO=...` values.

This means that `dpkg-reconfigure kontena-agent` can be used to e.g. fix the `KONTENA_URI` and reload the service.

This complexity also introduces new failure modes for the the maintscripts. For example, invalid syntax in the `/etc/kontena-agent.env` file will break any package operations, like:

```
/var/lib/dpkg/info/kontena-agent.config: 11: /etc/kontena-agent.env: nope: not found
```

However, the error messages seem reasonable, and those cases would also break the systemd units, so perhaps it's robust enough?

### DIFF
--- a/server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
+++ b/server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
@@ -12,7 +12,7 @@ sed -i -r "s#\#VAULT_IV=.*#VAULT_IV=${VAULT_IV}#" /etc/kontena-server.env
 
 # Fetching configuration from debconf
 db_get kontena-server/initial_admin_code
-sed -i -r "s#\#INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
+sed -i -r "s#\#?INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
 
 # reduce permissions on config file
 chmod 0600 /etc/kontena-server.env
--- a/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
+++ b/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
@@ -4,6 +4,13 @@ set -e
 # Source debconf library.
 . /usr/share/debconf/confmodule
 
+if [ -e /etc/kontena-agent.env ]; then
+  . /etc/kontena-agent.env
+
+  db_set kontena-agent/server_uri "${KONTENA_URI:-}"
+  db_set kontena-agent/grid_token "${KONTENA_TOKEN:-}"
+fi
+
 db_input high kontena-agent/server_uri || true
 db_input high kontena-agent/grid_token || true
 db_go || true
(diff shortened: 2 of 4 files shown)


server/packaging/ubuntu_xenial/kontena-server/DEBIAN/postinst
```diff
@@ -12,7 +12,7 @@ sed -i -r "s#\#VAULT_IV=.*#VAULT_IV=${VAULT_IV}#" /etc/kontena-server.env
 
 # Fetching configuration from debconf
 db_get kontena-server/initial_admin_code
-sed -i -r "s#\#INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
+sed -i -r "s#\#?INITIAL_ADMIN_CODE=.*#INITIAL_ADMIN_CODE=${RET}#" /etc/kontena-server.env
```
SpComb: Is updating this in the `/etc/kontena-server.env` enough to allow a new `kontena master login --code ...`, or should the postinst perhaps perform some kind of master admin reset thing if this changes?

jakolehm: Imho it's not enough to change this `.env`. /cc @kke

kke: Nope, need to run `rake kontena:reset_admin` inside the container and that generates a new random code.

SpComb: Fixed by only prompting for the initial admin code if not already set.

SpComb:
Ready for review, waiting for approval.

This should be ok to handle the simple `dpkg-reconfigure` cases. If we want fancier `dpkg-reconfigure` functionality for things like resetting the initial admin code, then those are separate issues/PRs.

PR #1760, opened 2017-02-01
Return healthy status when health check's response status is < 400

nevalla:


SpComb:
See https://github.com/kontena/kontena/issues/1790#issuecomment-283935164, IMO :-1: because the set of health check statuses should be configurable.

I might imagine a situation where accepting HTTP 3xx responses might turn all health checks into false positives, e.g. if you have some fancy `GET /status/` handler that checks that the database is alive, but you configure a health-check for `GET /status`, which just returns a `HTTP 302 /status/`...

SpComb:
Closing as obsolete. Feel free to re-open if hardcoding the agent to accept `< 400` as healthy turns out to be a valid fix for  #1790, but my understanding is that it would need a bit more finesse.

> Yes, but we have other cases where 403 is to be considered healthy (http basic auth means healthy)

PR #1761, opened 2017-02-01
Move osx builds to jenkins

jakolehm:
Travis has been too slow lately...


SpComb:
re https://www.traviscistatus.com/incidents/k79mjcv403c4

> MacOS queue backup & emergency maintenance

Issue #1769, opened 2017-02-02
Stack upgrade can remove linked-to services, breaking linking services


SpComb:
Having a service that links to a service in a different stack, and then using `kontena stack upgrade` to remove that linked-to service, results in a nil [`grid_service.grid_service_links[...].linked_grid_service`. This is normally impossible, as the `kontena service rm` -> server `DELETE /v1/services/:grid_name/:stack_name/:service_name` -> `GridServices::Delete` validation refuses to remove a service that is linked to.

This results in `kontena service ls` HTTP 500 errors:  #1741

The server's `StackDeployWorker#remove_services`  should not remove services that are still linked. The `Stacks::Update` mutation should also validate these.

## Example
```
$ kontena stack install --deploy links-test.yaml
$ kontena stack install --deploy links-test2.yaml
# edit the links-test.yaml to remove the test1 service
$ kontena stack upgrade --deploy links links-test.yaml
$ kontena service ls
 [error] Internal Server Error
```

The server error is on the `links2/test2` service, and running `kontena service rm links2/test2` fixes this situation.

#### `links-test.yaml`
```yaml
stack: test/links
services:
    test1:
        image: redis
    test2:
        image: redis
```

#### `links-test2.yaml`
```yaml
stack: test/links2
services:
    test2:
        image: redis
        links:
          - links/test1
```

### Server error
```

NoMethodError (undefined method `to_path' for nil:NilClass
Did you mean?  to_a):
  /home/kontena/kontena/kontena/server/app/views/v1/grid_services/_grid_service.json.jbuilder:37:in `block (3 levels) in evaluate'
  /home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/proxy.rb:150:in `map'
  /home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/proxy.rb:150:in `method_missing'
  /home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/embedded/many.rb:404:in `method_missing'
  /home/kontena/kontena/kontena/server/app/views/v1/grid_services/_grid_service.json.jbuilder:36:in `block (2 levels) in evaluate'
[...]

PR #1770, opened 2017-02-02
Render only working service links

### LINKED ISSUE #1766: Cannot remove grid XYZ
Hello -
I want to delete my kontena grid using **"kontena grid remove --force testing"**. In order to do it properly I removed each node in this grid and stopped the kontena-agent service on them.  **"kontena node ls"** shows zero nodes, so I can start to delete my grid.

```sh
foo@bar:~$ kontena node ls
Name                                                                   Status     Initial    Labels                                  
foo@bar:~$ kontena service ls
 [error] Internal Server Error
foo@bar:~$ kontena grid remove --force testing
 [fail] removing testing grid      
 [error] {"grid"=>"Grid has services"}
foo@bar:~$ 
``` 

Is there any config file/ folder of kontena which define the services? Or is there another way to solve it.

Best regards,

vinhmmkh

jakolehm:
Fixes #1766

### DIFF
--- a/server/app/views/v1/grid_services/_grid_service.json.jbuilder
+++ b/server/app/views/v1/grid_services/_grid_service.json.jbuilder
@@ -33,7 +33,7 @@ json.stack do
   json.id grid_service.stack.to_path
   json.name grid_service.stack.name
 end
-json.links grid_service.grid_service_links.map{|s|
+json.links grid_service.grid_service_links.select{ |s| !s.linked_grid_service.nil? }.map { |s|
   { id: s.linked_grid_service.to_path, alias: s.alias, name: s.linked_grid_service.name }
 }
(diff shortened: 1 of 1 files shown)


SpComb:
Not sure it makes sense to add code cruft to every possible place that might be nil in the case of what's essentially database corruption?

We should fix the code to not leave dangling references... like a `before_destroy` hook on the `GridService` or such?

jakolehm:
Yes, we could probably use mongo `pull` to remove item from links when service is destroyed.

SpComb:
:-1: on this in favor of a fix for #1769 

That might be a mongoid hook that raises an error if anything tries to removed a linked-to service, in addition to a `Stacks::Update` mutation validation for the same?

SpComb:
Superseded by #1873

PR #1779, opened 2017-02-03
Docs: note about osx installer

jakolehm:


kke:
We could have an automatic redirect to latest release (and latest beta) installer on the web page.

PR #1792, opened 2017-02-05
Initial e2e test suite

jakolehm:
Currently has setup/teardown only for vagrant.

```
$ rake vagrant:setup
$ bundle exec rspec -f d spec/
$ rake vagrant:teardown
```


matti:
What's the preferred method to use install CLI with this?

matti:
works.

test/spec/features/app/deploy_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+
+describe 'app deploy' do
+  it 'deploys a simple stack' do
```
jnummelin: Seems to really deploy an app... :)

test/spec/features/app/ps_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'app ps' do
```
jnummelin: There's no guarantee that there is anything deployed, does it make sense to check the `app ps` without anything deployed? (IMO no)

test/spec/features/app/remove_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+
+describe 'app remove' do
+  it 'removes a deployed stack' do
```
jnummelin: Removes an app... :)

test/spec/features/plugin/uninstall_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe 'plugin uninstall' do
+  before(:each) do
+    run('kontena plugin install aws')
+  end
+
+  it 'removes installed plugin' do
+    k = run('kontena plugin uninstall --force aws')
+    expect(k.code).to eq(0)
```
jnummelin: Should we also verfiy that we get at least some of the known plugins? (DO, AWS, vagrant, ...)

test/spec/features/service/deploy_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+describe 'service deploy' do
+  it 'deploys a service' do
+    run("kontena service create test-1 redis:3.0")
```
jnummelin: Why service is created with `run(...)` and other command with `kommando`? What's the difference?

test/spec/features/service/scale_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'service scale' do
+  it 'scales a service' do
+    run("kontena service create test-1 redis:3.0")
+    k = kommando("kontena service deploy test-1")
+    expect(k.run).to be_truthy
+    k = kommando("kontena service scale test-1 3")
+    expect(k.run).to be_truthy
+
```
jnummelin: Should we verify that three instances are actually created within some decent time?

test/spec/features/app/ps_spec.rb
jakolehm: we should test both.. with and without deploy ;)

test/spec/features/service/scale_spec.rb
jakolehm: Probably yes, although scale blocks until deployment has succeeded.

test/spec/features/plugin/uninstall_spec.rb
jakolehm: I don't get this? Can you explain a bit more.

test/spec/features/service/deploy_spec.rb
jakolehm: `support/shell.rb` should explain.

test/spec/features/plugin/uninstall_spec.rb
jnummelin: Ignore, was reading badly :)

PR #1796, opened 2017-02-06
Prompt echo can be disabled

### LINKED ISSUE #1794: Stack variable type "password"
```
variables:
  registry_htpasswd_password:
    type: string
    from:
      prompt:
```

something that would not echo to the screen would be nice

kke:
Fixes #1794

```yaml
stack: kke/wutwut
version: 0.0.1-pre1
variables:
  pass:
    type: string
    echo: false
    from:
      prompt: Enter secret password

  if_foo:
    only_if:
      pass: "foo"
    type: string
    from:
      prompt: you win a prize
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
@@ -50,10 +50,16 @@ module Kontena::Cli::Stacks
         prompt.yes?(question_text, default: option.default == false ? false : true)
       end
 
-      def ask
-        prompt.ask(question_text, default: option.default, echo: option.handler.options[:echo].nil? ? true : option.handler.options[:echo])
+      def echo?
+        return true if option.handler.nil?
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
```diff
@@ -51,7 +51,7 @@ def bool
       end
 
       def ask
-        prompt.ask(question_text, default: option.default)
+        prompt.ask(question_text, default: option.default, echo: option.handler.options[:echo].nil? ? true : option.handler.options[:echo])
```
jakolehm: This seems a bit too magical (and long) line to me. Is it possible that `option.handler.options` is nil?

kke: No.

https://github.com/kontena/opto/blob/master/lib/opto/type.rb#L93

PR #1797, opened 2017-02-06
Services can now have post install hooks that display messages after installing

### LINKED ISSUE #1788: stack post install messages
Once the stack has been installed, it would be great to give some info to the user "now go run this, then open this URL..."

kke:
Fixes #1788

The post install messages are not stack but service specific.

```
services:
  wordpress:
    hooks:
      post_install:
        - message: You have just installed wordpress, let's hope someone reads your blog.
```


SpComb:
Perhaps some example of how this behaves with interpolated values?

The liked issue suggests "then open this URL...", which either implies using the `$STACK.$GRID.kontena.local` names with a VPN connection, or then something more complicated to generate a URL for the linked load balancer or the exposed host port. The later would need some interesting new support for interpolating the actual deployed service state to get at things like the host node's public IP address.

kke:
Does @jnummelin or @jakolehm approve the design?

SpComb:
I think this is an okay start, although most of the usecases that I see involve getting at the deployed `kontena service show` state in the `post_install` hook. No idea how to get that interpolated in, though.

docs/references/kontena-yml.md
```diff
@@ -178,6 +178,16 @@ hooks:
       cmd: grunt dist
 ```
 
+**post_install**
+
+Display a message in the CLI after the service has been installed.
```
SpComb: Please have some explicit reference to `kontena stack install`, because that's the only CLI command that uses these.

jnummelin:
Just thinking should `kontena stack upgrade` also show these messages? If I now install a stack and later on add one service to it (with post install msg) I never see the message?

jakolehm:
Why only `message`? There is already `pre_build` client side hook that executes `cmd`, should we follow that pattern?

kke:
Hmm yes maybe it should be post_create or something.

SpComb:
> Why only message? There is already pre_build client side hook that executes cmd, should we follow that pattern?

:-1: on anything that executes local commands from stack registry YAML files.

I think the `kontena stack build` -> `hooks: pre_build: cmd: ...` is okay, because that requires a local stack file, and does not pull from the registry.

SpComb:
Needs a design review to figure out what the `post_install` hook should look like.

jnummelin:
IMO `post_install` hooks should be run in both `install` and `upgrade`. And as said, the hook needs to be consistent with other hook yamls.

I'd leave running commands out of it, as @SpComb said it will open up big holes to run random commands on the client host when installing stacks from registry.

SpComb:
Didn't make it for 1.2

kke:
Outdated & forgotten.

matti:
😢

PR #1802, opened 2017-02-07
Simple menus were not enabled on windows by default

### LINKED ISSUE #1783: Windows detector is not actually used for selecting simple menus
Simple menus are only enabled when env `KONTENA_SIMPLE_TERM` is set. Not by default on windows as was intended.

```ruby
 35   def self.on_windows?
 36     ENV['OS'] == 'Windows_NT' && RUBY_PLATFORM !~ /cygwin/
 37   end
 38
 43   def self.simple_terminal?
 44     ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
 45   end
 46
 51   def self.prompt
 52     return @prompt if @prompt
 53     if simple_terminal?
 ```

Should be

```ruby
 44     on_windows? || ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
```

kke:
Fixes #1783

The "simple menus" were not used on Windows by default as  was intended. Only when you set env KONTENA_SIMPLE_TERM.

This PR makes windows choose the simple terminal always.

### DIFF
--- a/cli/lib/kontena_cli.rb
+++ b/cli/lib/kontena_cli.rb
@@ -41,7 +41,7 @@ module Kontena
   end
 
   def self.simple_terminal?
-    ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
+    on_windows? || ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
   end
 
   def self.pastel
--- a/cli/spec/kontena/kontena_cli_spec.rb
+++ b/cli/spec/kontena/kontena_cli_spec.rb
@@ -4,8 +4,7 @@ require 'kontena_cli'
 describe Kontena do
   context 'prompt' do
     it 'uses light prompt on windows' do
-      allow(ENV).to receive(:[]).and_call_original
-      expect(ENV).to receive(:[]).with('OS').and_return('Windows_NT')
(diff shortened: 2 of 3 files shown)


cli/lib/kontena_cli.rb
```diff
@@ -41,7 +41,7 @@ def self.browserless?
   end
 
   def self.simple_terminal?
-    ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
+    on_windows? || ENV['KONTENA_SIMPLE_TERM'] || !$stdout.tty?
```
jakolehm: Test would be nice

cli/spec/kontena/kontena_cli_spec.rb
```diff
@@ -2,27 +2,35 @@
 require 'kontena_cli'
 
 describe Kontena do
-  let(:subject) { described_class }
+  context 'prompt' do
+    it 'uses light prompt on windows' do
+      expect(ENV).to receive(:[]).with('OS').and_return('Windows_NT')
```
jnummelin: `ENV` will get also other reads than `OS`, maybe use `allow` instead

kke: using allow + expect.. funny it worked on ruby 2.3+ without the allow

kke:
It still doesn't work.

PR #1805, opened 2017-02-07
Change node->containers removal to use callback

### LINKED ISSUE #1785: Node remove does not remove volume containers properly
Although there is a rule for that. Maybe this is a bug in Mongoid?

jnummelin:
Mongoid removes dependant models using the `default_scope` only if one is specified. In this case `Container` model has default scope to find only non-volume containers. This caused volume containers to be left behind when a node was removed. For some reason the removal works as was expected during specs, see https://github.com/mongoid-rspec/mongoid-rspec/issues/182.

fixes #1785

### DIFF
--- a/server/app/models/host_node.rb
+++ b/server/app/models/host_node.rb
@@ -29,7 +29,7 @@ class HostNode
   attr_accessor :schedule_counter
 
   belongs_to :grid
-  has_many :containers, dependent: :destroy
+  has_many :containers
   has_many :host_node_stats
   has_and_belongs_to_many :images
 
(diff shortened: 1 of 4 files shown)


server/app/models/host_node.rb
```diff
@@ -42,6 +42,10 @@ class Error < StandardError
 
   scope :connected, -> { where(connected: true) }
 
+  after_destroy do |node|
+    node.containers.unscoped.delete_all
```
jakolehm: `delete_all` skips callbacks.. is it safe to use?

jnummelin: Currently there's no callback on the container model, so it is safe. But to be more future-proof, I'll change it to use destroy so that we don't have to worry when adding possible callbacks on the container model.

PR #1830, opened 2017-02-09
More e2e specs

jakolehm:


test/spec/features/stack/install_spec.rb
```diff
@@ -41,4 +42,8 @@
       end
     end
   end
+
+  context 'service_link' do
+
```
jnummelin: Placeholder for test cases?

SpComb:
Conflict on #1876

SpComb:
The conflict was such trivial that I went ahead and resolved it. The github web editor isn't too-bad, it seems.

PR #1839, opened 2017-02-10
Display server version on master container startup

### LINKED ISSUE #1835: output master version in the logs when starting
When you manually update master, it gives a nice visual confirmation that the desired version is running.

kontena-agent says `INFO -- Kontena::Agent: initializing agent (version 1.1.1)`

kke:
Show: `** Starting Kontena Master version x.y.z **` on master container startup

Fixes #1835

### DIFF
--- a/server/run.sh
+++ b/server/run.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+echo "** Starting Kontena Master version `cat VERSION` **"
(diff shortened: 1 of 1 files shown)


matti:
:beer:

kke:
I think it's supposed to go to puma's before_fork normally, but this way it gets echoed before all the "Starting puma...." stuff.

PR #1840, opened 2017-02-10
Change year 2016 to 2017 in CLI logo

kke:


matti:
PRO-TIP: make it ERB and use Time.now

kke:
Then someone changes the year on their machine to 1820 and says the copyright expired already and steals it!

I don't think it serves any real purpose there anyway so it could just be taken out..

matti:
it should sign that year and validate it against year.kontena.io

Issue #1803, opened 2017-02-07
Support for managed volumes


jnummelin:
Kontena should support named volumes as top level concept in stack yaml. This naturally also means that volumes are then "schedulable" objects.

Below is crude list of things needed:
- Yaml parser needs to support volumes
- Master has volumes API (CRUD)
- Migration picks up named volumes from existing services
- Agent sends available volume drivers to master
    - Master can use this info when scheduling services using volumes

Volumes definition follows docker-compose [syntax](https://docs.docker.com/compose/compose-file/#/volume-configuration-reference):

```
volumes:
  data:
    scope: instance-shared
    driver: rexray
    driver_opts:
      foo: bar
      baz: 1
```

Volumes can be defined in three modes: `instance-shared`, `instance-private` or `node`

### instance-shared

```
# Volume shared between web and sidekick
# web-1, sidekick-1 shares vol-1
# web-2, sidekick-2 shares vol-2
services:
  web:
    stateful: true
    volumes:
      - vol


  sidekick:
    volumes:
      - vol

volumes:
  vol:
    scope: instance-shared
    driver: ebs
```

So there will be a volume created per instance of the service that uses the volume. Also different service instances will be able to share the same volumes. This implies also service affinity between the services sharing the volumes in this mode.

### instance-private
```
# All instances get their own volume
# web-1, web-vol-1
# sidekick-1 sidekick-vol-1
# web-2, web-vol-2
# sidekick-2, sidekick-vol-2
# Deployment fails if trying to share "vol"
services:
  web:
    stateful: true
    volumes:
      - vol

  sidekick:
    volumes:
      - vol

volumes:
  vol:
    scope: instance-private
```

All service instances will get their own volume created during deploy.

### node

```
# Only one volume per node
# web-1 -> vol
# sidekick-1 -> vol
# web-2 -> vol
# sidekick-2 -> vol
services:
  web:
    stateful: true
    volumes:
      - vol

  sidekick:
    volumes:
      - vol

volumes:
  vol:
    scope: node
```

In this mode only one volume will be created per node and many services can use the same volume. A volume is created only on the nodes where the referencing services are deployed to.

matti:
in kontena `instance` is used as "container instance", not as "vm instance"

jnummelin:
My first gut feeling was that volumes should be global. So if one stack defines volume `xyz`, any other stack can use that automatically.

[...]

Issue #1442, opened 2016-11-29
Server: current_user.grid_ids.include? vs current_user.accessible_grids.find_by


SpComb:
Different server API methods use a different grid access check, which behaves differently for users with the master_admin role:

```ruby
  def accessible_grids
    if self.master_admin?
      Grid.all
    else
      self.grids
    end
  end
```

### `current_user.grid_ids.include?(...)`
```
server/app/routes/v1/etcd_api.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/external_registries.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/containers_api.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/containers_api.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/nodes_api.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/secrets.rb:        unless current_user.grid_ids.include?(grid.id)
server/app/routes/v1/services_api.rb:        unless current_user.grid_ids.include?(grid_service.grid_id)
```

### `current_user.accessible_grids.find_by(...)`
```
server/app/routes/v1/certificates_api.rb:        @grid = current_user.accessible_grids.find_by(name: name)
server/app/routes/v1/grids_api.rb:        @grid = current_user.accessible_grids.find_by(name: name)
server/app/routes/v1/stacks_api.rb:        @grid = current_user.accessible_grids.find_by(name: name)
```

kke:
There's also the `grid_admin?`

PR #1849, opened 2017-02-13
Volumes api

jnummelin:
Implementation for volume management API and functionality.

Volumes themselves are always grid scoped but can be fully defined also within a stack. If stack refers to externally created volume it has to be still defined in stack using the external attribute.
```
stack: foo
services:
  web:
    image: nginx
    volumes:
      - my-vol:/web

volumes:
  my-vol:
    external: someOtherVolume
```

CLI has command to create grid scoped volumes and is able to list and show both types of volumes.

Master parses together the scope of the volume thus agent just ensures the named volume is present when creating containers.


agent/lib/kontena/models/service_pod.rb
```diff
@@ -37,7 +37,8 @@ class ServicePod
                   :pid,
                   :hooks,
                   :secrets,
-                  :networks
+                  :networks,
+                  :volume_specs
```
jakolehm: Why not just `volumes`?

jnummelin: To distinguish between the volumes for the container (bind mounts, anon volumes, etc.) and the volumes which existence needs to be ensured when creating the container

server/app/models/stack.rb
```diff
@@ -11,6 +11,8 @@ class Stack
   has_many :stack_revisions, dependent: :destroy
   has_many :stack_deploys, dependent: :destroy
   has_many :grid_services
+  has_many :volumes
+  embeds_many :external_volumes
```
jakolehm: Why we need this `external_volumes` mapping?

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,113 @@
+module V1
+  class VolumesApi < Roda
```
jakolehm: Needs to be added to REST API docs.

server/app/models/stack.rb
jnummelin: To make it more flexible and robust to define external volumes in stacks. They are kind-of mappings to grid level volumes where there might be extra attributes.

server/app/models/volume.rb
```diff
@@ -0,0 +1,54 @@
...
+  field :scope, type: String # TODO: Should be enum once scope names are fixed
+  field :driver, type: String, default: 'local'
+  field :driver_opts, type: Hash, default: {}
+  field :external, type: Boolean
+
+  belongs_to :grid
+  belongs_to :stack, inverse_of: :volumes
+  belongs_to :external_volume
+
+  index({ grid_id: 1 })
```
kke: needs a migration with `Volume.create_indexes`

server/app/models/grid.rb
```diff
@@ -29,6 +29,7 @@ class Grid
   has_many :stacks, dependent: :destroy
   has_many :grid_domain_authorizations, dependent: :delete
   has_many :networks, dependent: :delete
+  has_many :volumes, dependent: :delete
```
kke: :delete will not run any callbacks, intentional?

server/app/models/stack.rb
jakolehm: Ok, but somehow this smells a bit .. could we map volumes always the same way (and just have extra attributes set whenever needed).

jnummelin: I did try it that way but it IMO ended up even smellier. :D The `external` volume is a kind-of reference to grid level volume, so it's not really doable with a single `Volume` model. If there would be only `:volumes` relation, I don't see a clear way to define an external volume on stack level that references a grid level volume. Yes, we can add extra attribute `external: true` for the volume, but how to convey the reference to a grid level volume? Having a real relation to a volume within volume sounds really bad IMO.

I do agree that this creates bit of repetition, but couldn't come up with any cleaner alternative. I'm perfectly fine to refactor this if any better ideas? :)

server/app/mutations/stacks/create.rb
```diff
@@ -47,6 +61,26 @@ def execute
         return
       end
 
+      if self.volumes
+        # TODO Separate into own method
+        attributes.delete(:volumes)
+        self.volumes.each do |volume|
+          if volume[:external]
+            vol = Volume.where(name: volume['name'], grid: grid, stack: nil).first
+            if vol
```
SpComb: I find the `if` a bit odd here... if we've validated it, then can't we just assume it's not nil? It feels scary to have code that just leaves out a volume.

I guess I might setup some kind of of `@external_volumes[name] = vol` in the `validate_volumes` and then re-use those models here...

server/app/mutations/stacks/create.rb
```diff
@@ -47,6 +61,26 @@ def execute
...
+              stack.external_volumes.create!(volume: vol)
+            end
+          else
+            outcome = Volumes::Create.run(grid: grid, stack: stack, **volume.symbolize_keys)
+            unless outcome.success?
+              outcome.errors.message.each do |key, msg|
+                add_error(:volumes, :key, "Volume create failed for '#{volume[:name]}': #{msg}")
```
SpComb: This needs to also `return`, or the execute will just continue to create the stack and services.

server/app/mutations/volumes/create.rb
```diff
@@ -0,0 +1,42 @@
+module Volumes
+  class Create < Mutations::Command
+
+    required do
+      model :grid, class: Grid
+      string :scope # TODO Match the predefined scopes
```
SpComb: `in: %w(....)`

server/app/mutations/volumes/create.rb
```diff
@@ -0,0 +1,42 @@
...
+      string :name  # TODO Force same rules for naming as docker
```
SpComb: `matches: /.../`

SpComb: Make sure this does not allow `.` in the name.

server/app/mutations/volumes/delete.rb
```diff
@@ -0,0 +1,29 @@
+module Volumes
+  class Delete < Mutations::Command
+
+    required do
+      model :volume, class: Volume
+    end
+
+    def validate
+      self.volume.grid.stacks.each do |stack|
+        if stack.external_volumes.find {|e| e == self.volume}
```
SpComb: `e.volume == self.volume`? I feel this is missing specs... :)

server/app/mutations/volumes/delete.rb
```diff
@@ -0,0 +1,29 @@
...
+          add_error(service.name, :volume_in_use, "Volume still in use in service #{service.name}")
```
SpComb: `stack.name`

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,113 @@
...
+        grid = find_grid(grid_name)
+        if stack_name
+          stack = find_stack(stack_name, grid)
+          volume = stack.volumes.find_by(name: volume_name)
+        else
+          volume = grid.volumes.find_by(name: volume_name)
+        end
+        halt_request(404, {error: "Volume #{volume_name} not found"}) if !volume
+
+        unless current_user.grid_ids.include?(volume.grid_id)
```
SpComb: Shouldn't these checks all be fixed to use `current_user.accessible_grids` per #1442 ?

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,113 @@
...
+          halt_request(403, {error: 'Access denied'})
+        end
+
+        volume
+      end
+
+      # Parse stack/vol out of path
+      # @param [String] volume path
+      # @return [Array] [stack, volume]
+      def parse_params(path)
```
SpComb: This is weird... alternative would be the same `:grid/null/:volume` vs `:grid/:stack/:volume` as used for the services API. Keep the `[nil, volume]` return value, though :)

Mainly this kind of schema prevents ever adding new `POST :grid/:volume/snapshot` API endpoints.

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,113 @@
...
+            @volume = load_volume(grid_name, stack, volume)
+            render('volumes/show')
+          end
+        end
+        r.post do
+          r.is do
+            grid = find_grid(grid_name)
+            data = parse_json_body rescue {}
+            data[:grid] = grid
+            outcome = Volumes::Create.run(data)
```
SpComb: This is missing access control checks?

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,113 @@
...
+          [nil, path]
+        end
+      end
+
+      r.on ':grid_name' do |grid_name|
+        @grid = find_grid(grid_name)
+        r.get do
+          r.is do
+            @volumes = @grid.volumes
+            render('volumes/index')
```
SpComb: This is missing access control checks?

server/app/routes/v1/volumes_api.rb
SpComb: This check should probably be moved to the `find_grid` helper..

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +175,46 @@ def build_networks
...
+      end
+
+      volume_specs
+    end
+
+    VOLUME_FLAGS = [':ro', ':rw', ':z', ':Z', ':rshared', ':shared', ':rslave', ':slave', ':rprivate', ':private', ':nocopy']
+    # Removes known mount flags from the volume specification to make it parseable
+    # @param [String]
+    # @return [String]
+    def remove_volume_flags(volume)
```
SpComb: Ugh, enumerating all of the known volume flags seems like a really bad idea.

I suspect the `:ro` suffixes are not accepted for anonymous volumes? i.e. they will always be the third `:`-separated part?

Also for a `host-src:container-dest:ro` volume spec, the `container-dest` MUST always be an absolute path, i.e. starting with `/`. That should be enough to figure it out in case without enumerating the flags.

server/app/mutations/stacks/create.rb
```diff
@@ -47,6 +61,26 @@ def execute
...
       end
 
+      if self.volumes
+        # TODO Separate into own method
+        attributes.delete(:volumes)
+        self.volumes.each do |volume|
+          if volume[:external]
+            vol = Volume.where(name: volume['name'], grid: grid, stack: nil).first
+            if vol
+              stack.external_volumes.create!(volume: vol)
```
SpComb: Hmm... you don't set any `name: ...`? But the  `external_volume.name` is always the same as the `external_volume.volume.name`?

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +175,46 @@ def build_networks
       networks
     end
 
+    def build_volumes(instance_number)
+      volume_specs = []
+      grid_service.volumes.each do |vol|
+        vol_string = remove_volume_flags(vol.dup)
```
SpComb: I think all this complexity can be avoided by just matching what the agent does...

```ruby
next unless vol.include?(':') # anonymous volume
host_vol, container_path, flags = vol.split(':') # flags is optional
next if host_vol.start_with?('/') # bind mount from host
```

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +175,46 @@ def build_networks
...
+        elements = vol_string.split(':')
+        host_path_or_vol = elements[-2]
+        if host_path_or_vol.nil? || host_path_or_vol.start_with?('/')
+          next
+        end
+        container_path = elements[-2]
+        volume =
+          grid_service.stack.volumes.find_by(name: host_path_or_vol) ||
+          grid_service.stack.external_volumes.find_by(name: host_path_or_vol).volume
```
SpComb: AFAIK the `stack.external_volumes[...].name` is always just nil ATM?

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +175,46 @@ def build_networks
...
+          # TODO Need o handle nils? Should not be possible since the stack is invalid if no volumes found
+        if volume
+          volume_name = volume.name_for_service(grid_service, instance_number)
+          volume_specs << {
+              name: volume_name,
+              driver: volume.driver,
+              scope: volume.scope,
```
SpComb: Not used by the agent.

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +175,46 @@ def build_networks
...
+              driver_opts: volume.driver_opts
+          }
+          vol.sub!(host_path_or_vol, volume_name)
```
SpComb: Ew, string mutation, this is super scary, I don't like it. Please explicitly `map` the `spec[:volumes]`.

SpComb:
Missing `Stacks::Update` logic to add/remove volumes on `kontena stack upgrade`.

server/app/models/stack.rb
SpComb: Agreed that the separate processing of `external_volumes` seems odd, but I see the point.. this is a complex N:M relationship, where the same `Volume` will either belong to a single `Stack`, or it might belong to a `Grid` (i.e. `volume.stack == nil`), but be referenced by many `Stack`s.

With some creative interpretation of the [Mongoid Relations : Referenced N-N](https://mongoid.github.io/old/en/mongoid/docs/relations.html#has_and_belongs_to_many) docs... might this work somehow?

```ruby
class Grid
  has_many :volumes
end

class Stack
  has_and_belongs_to_many :volumes
end

class Volume
  belongs_to :grid
  belongs_to :stack
end
```

This would then have something like a stack S with `{ "volume_ids": [ 'A', 'B' ]}` with grid-level volume A `{ "grid_id": "G", "stack_id": null}` and stack-level volume B `{ "grid_id": "G", "stack_id": "S" }`.... needs testing, I don't know if I understood this correctly :)

SpComb:
I also find the design model of "external" vs "scope" a bit lacking... this is apparent in how the scope is completely ignored for external volumes. I'll try and give it a bit more thought, and post any design that I can come up with to #1803...

server/app/models/external_volume.rb
```diff
@@ -0,0 +1,8 @@
+class ExternalVolume
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :name, type: String
+  
+  has_one :volume
```
SpComb: Does this work like you expect it to? Per the [Mongoid relations doc](https://mongoid.github.io/old/en/mongoid/docs/relations.html#has_one), this means that each volume in the volumes` collection will have a `external_volume_id`. This means that each `Volume` can only be referenced by one `ExternalVolume`?

SpComb: This should be `belongs_to :volume` with a `has_many :external_volumes` on the `Volume` model.

server/app/models/stack.rb
SpComb: If that feels too complicated, and it probably is.... just make all of the volume references use a layer of indirection? Keep the `Grid`, `Stack` and `Volume` references as they are, but drop the `ExternalVolumes` completely, and then refactor the `GridService` to parse out the volumes as embedded `ServiceVolume` models:

```ruby
class GridService
  embeds_many :volumes, class: ServiceVolume
end

class ServiceVolume
  embedded_in :grid_service

  field :bind_mount, type: String # optional host bind mount
  field :path, type: String # mount path in the container
  field :flags, type: String # optional flags

  belongs_to :volume # optional
end
```

with associated `has_many :service_volumes` on the `Volume` model.

jakolehm: @SpComb I was actually going to propose similar solution.. I think this is the simplest way to go?

server/app/routes/v1/volumes_api.rb
jnummelin: But the difference with services is that the volumes are NOT part of the special `null` stack. :/ Hence having `null` in the path makes things confusing.

SpComb: Agreed that the `null` in the `/v1/volumes/:grid/null/:volume` URL is a bit odd, and I would also leave the special `null` stack out of this, the current `return [nil, volume]` is much better.

But if volumes are going to be first-class managed resources, then they might grow extra sub-resources at some point in the future, like say snapshots... if you look at how those work in e.g. the services `/v1/services/testgrid/null/fooservice/containers` API, then this kind of variable-length URL scheme would not work.

If we don't need to be consistent with the services API, then maybe even try `/v1/volumes/:grid/:stack.:volume` vs `/v1/volume/:grid/:volume`, like the Docker container names? But maybe it's better to just be consistent with what we already have...

SpComb: I tried something similar in #1373... it took the first `/v1/services/:name` component and tried to look it up as a stack. If you had a stack with that name, then it would expect the next part of the URL to be the service name, otherwise it would look it up as a stack: https://github.com/kontena/kontena/pull/1373/files#diff-55f0a23194a9889111f68dc75d7a1f25R24

Probably worse than just the `null` in the URL...

cli/lib/kontena/cli/volume_command.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'volumes/create_command'
+require_relative 'volumes/remove_command'
+require_relative 'volumes/list_command'
+
+class Kontena::Cli::VolumeCommand < Kontena::Command
+
+  subcommand "create", "Create a managed volume", Kontena::Cli::Volume::CreateCommand
```
kke: The lazy subcommand loader thing is in master now, so these should also be:

```
subcommand "create", "Create a managed volume", load_subcommand('volumes/create_command')
```

(also remove the requires above)

cli/lib/kontena/main_command.rb
```diff
@@ -55,6 +56,7 @@ class Kontena::MainCommand < Kontena::Command
   subcommand "whoami", "Shows current logged in user", Kontena::Cli::WhoamiCommand
   subcommand "plugin", "Plugin related commands", Kontena::Cli::PluginCommand
   subcommand "version", "Show version", Kontena::Cli::VersionCommand
+  subcommand "volume", "Volume specific commands", Kontena::Cli::VolumeCommand
```
kke: Same here, use the load_subcommand(...) thing

server/app/models/service_volume.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def to_s
+    elements = []
+    if self.volume
+      elements << self.volume.name
+    elsif self.bind_mount
+      elements << self.bind_mount
+    end
+    elements << self.path
+    elements << self.flags if self.flags && !self.flags.empty?
```
SpComb: I don't think you can use `path:flags` if there is no `volume:` or `/bind/mount:` prefix

server/app/models/grid.rb
jakolehm: I think this should be `:destroy`

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +84,26 @@ def build_grid_service_secrets(existing_secrets)
       service_secrets
     end
 
+    def build_service_volumes()
```
jakolehm: Really complex method.. maybe we should split this?

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +84,26 @@ def build_grid_service_secrets(existing_secrets)
...
+      service_volumes = []
+      self.volumes.each do |vol|
+        # TODO Check existing volumes, if needed
+        service_volume = nil
+        elements = vol.split(':')
+        if elements[0].start_with?('/') && elements[1] && elements[1].start_with?('/') # Bind mount
+          service_volume = ServiceVolume.new(bind_mount: elements[0], path: elements[1], flags: elements[2..-1].join(':'))
+        elsif !elements[0].start_with?('/') && elements[1].start_with?('/') # Real volume
+          volume = self.grid.volumes.find_by(name: elements[0])
+          # TODO What if the volume is not found?
```
SpComb: `raise` and then `add_error` in `def validate`?

server/app/models/volume.rb
```diff
@@ -0,0 +1,40 @@
...
+    when 'container'
+      "#{service.name_with_stack}.#{self.name}-#{instance_number}"
+    when 'service'
+      "#{service.name_with_stack}.#{self.name}"
+    end
+
+  end
+
+  # @return [Boolean]
+  def default_stack?
```
SpComb: mm... not used anywhere? Useful method though.

server/app/mutations/grid_services/update.rb
```diff
@@ -79,6 +86,9 @@ def execute
       if self.secrets
         attributes[:secrets] = self.build_grid_service_secrets(self.grid_service.secrets.to_a)
       end
+      if self.volumes
+        attributes[:service_volumes] = self.build_service_volumes
```
SpComb: Needs to handle validation errors.

Best to raise an error, and then have a def-level `except` that does `add_error`...

server/app/mutations/stacks/common.rb
```diff
@@ -52,6 +52,9 @@ def common_validations
           string :expose
           string :registry
           model :variables, class: Hash
+          array :volumes do
+            model :object, class: Hash
```
SpComb: Aren't these still strings at this point?

server/app/mutations/stacks/create.rb
```diff
@@ -47,6 +61,22 @@ def execute
         return
       end
 
+      if self.volumes
+        # TODO Separate into own method
+        attributes.delete(:volumes)
+        self.volumes.each do |volume|
+          unless volume[:external]
+            outcome = Volumes::Create.run(grid: grid, **volume.symbolize_keys)
```
SpComb: These volume names are not unique across stacks.

server/app/mutations/stacks/common.rb
jnummelin: No, this is stack mutation so they are hashes

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +84,26 @@ def build_grid_service_secrets(existing_secrets)
...
+          service_volume = ServiceVolume.new(volume: volume, path: elements[1], flags: elements[2..-1].join(':'))
+        elsif elements[0].start_with?('/') && (elements.size == 1 || !elements[1].start_with?('/')) # anon volume
+          service_volume = ServiceVolume.new(bind_mount: nil, path: elements[0], flags: nil) # anon vols do not support flags
```
jnummelin: I guess I was preparing for possibilitythat anon vol might have flags also. but that should not be possible afaik.

server/app/models/grid.rb
jnummelin: Yes, better to use `:destroy`, although there is no callbacks at the moment.

server/app/models/service_volume.rb
jnummelin: true, but the `VolumesHelpers#parse_volume` strips those out already when parsin the volume info.

server/app/models/volume.rb
```diff
@@ -0,0 +1,54 @@
...
+  belongs_to :external_volume
+
+  index({ grid_id: 1 })
+  index({ name: 1 })
+  index({ stack_id: 1 })
+
+  validates_presence_of :name, :scope, :grid_id
+  validates_uniqueness_of :name, scope: [:grid_id, :stack_id]
+
+  def to_path
```
jnummelin: Now this does too :)

server/app/mutations/grid_services/common.rb
jnummelin: Still too complex? Refactored heavily when implemented support for the external volume "mapping"

server/app/mutations/grid_services/update.rb
jnummelin: What do you mean with validation errors?

server/app/routes/v1/volumes_api.rb
jnummelin: It was there but the code was bit messy to see it. :) Cleaned it so now it's handled on the top level route when the grid is loaded.

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +86,30 @@ def build_grid_service_secrets(existing_secrets)
       service_secrets
     end
 
+    def build_service_volumes()
+      service_volumes = []
+      self.volumes.each do |vol|
+        vol_spec = parse_volume(vol)
+        if vol_spec[:volume]
```
jakolehm: What happens if `vol_spec[:volume].nil?`

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +86,30 @@ def build_grid_service_secrets(existing_secrets)
...
+          # Named volume, try to find the proper mapping for external volumes
+          # NULL stack services don't have revs, for those just use the "global" name
+          volume_name = vol_spec[:volume]
+          if self.stack.latest_rev
```
jakolehm: Does this work when stack is not yet created?

server/app/mutations/grid_services/common.rb
jakolehm: Much better but added couple notes.

server/app/mutations/grid_services/common.rb
jakolehm: Ok, because this is in GridService stack already exists.

server/app/mutations/grid_services/update.rb
```diff
@@ -32,7 +34,12 @@ def validate
           add_error(:volumes_from, :invalid, 'Cannot combine stateful & volumes_from')
         end
         if self.volumes
-          changed_volumes = self.volumes.select { |v| !self.grid_service.volumes.include?(v) }
+          changed_volumes = self.volumes.select { |v|
+            vols = self.grid_service.service_volumes.map {
+              |sv| sv.to_s
```
jakolehm: `self.grid_service.service_volumes.map { |sv| sv.to_s }`

server/app/mutations/stacks/create.rb
```diff
@@ -36,6 +37,19 @@ def validate_services
       end
     end
 
+    def validate_volumes
+      if self.volumes
```
jakolehm: Less nesting if `return unless self.volumes`

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,83 @@
...
+            render('volumes/show')
+          end
+        end
+        r.post do
+          r.is do
+            data = parse_json_body rescue {}
+            data[:grid] = @grid
+            outcome = Volumes::Create.run(data)
+            if outcome.success?
+              volume = outcome.result
```
jakolehm: duplicate `outcome.result` call

server/app/services/docker/service_creator.rb
```diff
@@ -173,6 +174,30 @@ def build_networks
...
+        if sv.volume
+          volume_name = sv.volume.name_for_service(grid_service, instance_number)
+          volume_specs << {
+              name: volume_name,
+              path: sv.path,
+              flags: sv.flags,
+              driver: sv.volume.driver,
+              driver_opts: sv.volume.driver_opts
+          }
+        elsif
```
jakolehm: `else`?

server/app/views/v1/grid_services/_grid_service.json.jbuilder
```diff
@@ -21,7 +21,7 @@ json.secrets grid_service.secrets.as_json(only: [:secret, :name, :type])
 json.memory grid_service.memory
 json.memory_swap grid_service.memory_swap
 json.cpu_shares grid_service.cpu_shares
-json.volumes grid_service.volumes
+json.volumes grid_service.service_volumes.map {|sv| sv.to_s}
```
jakolehm: Is this right?

test/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       - mongodb
 
   agent:
-    container_name: kontena-agent
+    container_name: e2e-kontena-agent
```
jakolehm: Not part of this pr, right? :)

server/app/mutations/grid_services/common.rb
jnummelin: yes, there were some specs that did not set stack at all for services, but I fixed those in this PR.

server/app/mutations/stacks/create.rb
jnummelin: 👍

server/app/mutations/grid_services/common.rb
jnummelin: then it just creates the `ServiceVolume` with the other args that has been parsed from the volume string. So this `if`block only "replaces" the parsed named volume string with the actual volume reference.

server/app/mutations/grid_services/update.rb
jnummelin: yes, I'll clean this up.

server/app/views/v1/grid_services/_grid_service.json.jbuilder
jnummelin: yes, service itself does not anymore have the "raw" volume definitions thus the `ServiceVolume`s are stringified to have the API backwards compatible. I'll add a test case to verify it :)

test/docker-compose.yml
jnummelin: oh, I've managed to put this same change to other PR also... :/ I'll revert this

server/app/mutations/grid_services/common.rb
jakolehm: I think we should avoid this "backlink" to `latest_rev` .. reason: we are already passing latest_rev as a hash to mutation.

jnummelin: The service mutations do not currently get `latest_rev` as input. (Or am I blind? :D ) Also the service mutations are actually called from multiple places (`Stack::Create`, `Stack::Update`, `Stack::Deploy`, ...) so there would be multiple places to "fix" to give that as input. In that light I see the "backlinking" actually quite clean solution.

jakolehm: They get service hash from latest_rev (see :https://github.com/kontena/kontena/blob/master/server/app/mutations/stacks/update.rb#L22-L38).

jnummelin: Yes, but the plain service hash is not enough as now we need to access the volume info also from `stack.latest_rev`.

SpComb: The other `GridServices::Common` helper methods take `grid` and `stack` as arguments.

This could take a `{ volume: Volume }` mapping as an argument that passed to the `Services` mutations by the `Stacks` mutations? For plain services where this is run directly from the grid services API without any such volume mappings, this would then interpret the named volumes as literal `Volume.name`s.

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +86,30 @@ def build_grid_service_secrets(existing_secrets)
...
+            stack_volume = self.stack.latest_rev.volumes.find {|v|
+              v['name'] == vol_spec[:volume]
+            }
+            # Use external volume definition if given
+            volume_name = stack_volume['external'] == true ? stack_volume['name'] : stack_volume.dig('external', 'name') || stack_volume['name']
```
SpComb: This line takes like more than a minute to read, and I'm sitll not quite sure what it does.

I think this line should be somewhere in the `Stacks` mutations, not the `Services` mutations?

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +86,30 @@ def build_grid_service_secrets(existing_secrets)
...
+    def build_service_volumes()
+      service_volumes = []
+      self.volumes.each do |vol|
+        vol_spec = parse_volume(vol)
+        if vol_spec[:volume]
+          # Named volume, try to find the proper mapping for external volumes
+          # NULL stack services don't have revs, for those just use the "global" name
+          volume_name = vol_spec[:volume]
+          if self.stack.latest_rev
+            stack_volume = self.stack.latest_rev.volumes.find {|v|
```
SpComb: This will be `nil` if there is no such volume defined in the stack, and then explodes without saying which volume was not found in the stack.

server/app/mutations/grid_services/common.rb
```diff
@@ -84,6 +86,30 @@ def build_grid_service_secrets(existing_secrets)
       service_secrets
     end
 
+    def build_service_volumes()
+      service_volumes = []
+      self.volumes.each do |vol|
+        vol_spec = parse_volume(vol)
```
SpComb: Needs to handle the error case of `vol` not being a valid volume spec, like `"foo"`. Currently `parse_volume` may return `nil`, and then this explodes.

server/db/migrations/19_volume_migration.rb
```diff
@@ -0,0 +1,25 @@
...
+class VolumeMigration < Mongodb::Migration
+
+  extend VolumesHelpers
+
+  def self.up
+    Volume.create_indexes
+
+    GridService.each do |service|
+      service.volumes.each do |v|
+        volume_spec = parse_volume(v)
```
SpComb: There might be services with invalid volume specs. Not much we can do about those, but maybe not explode the entire mutation?

server/app/helpers/volumes_helpers.rb
```diff
@@ -0,0 +1,14 @@
+module VolumesHelpers
+
+  def parse_volume(vol)
+    elements = vol.split(':')
+    if elements[0].start_with?('/') && elements[1] && elements[1].start_with?('/') # Bind mount
+      {bind_mount: elements[0], path: elements[1], flags: elements[2..-1].join(':')}
```
SpComb: > -v|--volume[=[[HOST-DIR:]CONTAINER-DIR[:OPTIONS]]]
> The OPTIONS are a comma delimited list and can be: ...

So this should be `.join(',')`?

SpComb:
Needs merge after #1873

agent/spec/lib/kontena/models/service_pod_spec.rb
```diff
@@ -345,8 +353,9 @@
     it 'returns correct array when volumes are defined' do
       data['volumes'] = volumes
       bind_vols = subject.build_bind_volumes
-      expect(bind_vols.size).to eq(1)
-      expect(bind_vols[0]).to eq('/proc:/host/proc:ro')
+      expect(bind_vols.size).to eq(2)
+      expect(bind_vols[0]).to eq('app.someVol:/data:')
```
jnummelin: yes, I actually tested it manually. :) But now a big issue to take out in code for clarity.

server/app/mutations/stacks/common.rb
```diff
@@ -14,9 +14,9 @@ def handle_service_outcome_errors(service_name, messages, type)
       end
     end
 
-    def handle_volume_outcome_errors(volume_name, messages)
-      messages.each do |key, msg|
-        add_error(:volumes, key.to_sym, "Volume create failed for volume '#{volume_name}': #{msg}")
+    def handle_volume_outcome_errors(volume_name, errors)
+      errors.each do |key, atom|
+         add_service_error(volume_name, key, atom.symbolic, atom.message)
```
SpComb: Hmm, #1976 isn't merged yet... and this should not be under the `services => ...` key?

```
add_error("volumes.#{volume_name}.#{key}", atom.symbolic, message)
```

server/app/routes/v1/volumes_api.rb
```diff
@@ -16,7 +16,7 @@ class VolumesApi < Roda
       def find_grid(grid_name)
```
SpComb: #1970 is merged, so this should be replaced with `load_grid`.

server/app/mutations/grid_services/update.rb
```diff
@@ -41,6 +46,8 @@ def validate
     end
 
     def execute
+      self.grid = self.grid_service.grid
```
jnummelin: sure, I'll refactor it so. clearer that way as it follows the same pattern with other helper methods.

server/app/mutations/grid_services/update.rb
jnummelin: I've added volume validation also now.

server/app/mutations/grid_services/common.rb
jnummelin: I can try to refactor it to be more clearer, but this essentially tries to use the given `external` volume name either with `external.name`or just the volume name when `external: true` is used.

The only option to take this functionality out from `Services` mutations is to pass some volume "mapping" hash into `Services` mutation. The issue I see there is that `Services` mutations are called from multiple sources, thus the mapping needs to be  generated in multiple places.

server/app/routes/v1/volumes_api.rb
jnummelin: And also uses now the common way of checking grid access.

server/app/routes/v1/volumes_api.rb
```diff
@@ -0,0 +1,82 @@
...
+    include LogsHelpers
+
+    plugin :backtracking_array
+
+    route do |r|
+
+      validate_access_token
+      require_current_user
+
+      def find_grid(grid_name)
```
jnummelin: fixed

jnummelin:
Commented stuff has been fixed and polished, PTAL @jakolehm @SpComb

server/app/mutations/stacks/create.rb
```diff
@@ -57,6 +61,20 @@ def execute
       stack
     end
 
+    # @param [Array<Hash>] volumes
+    def create_volumes(volumes)
+      return unless volumes
+      volumes.each do |volume|
+        unless volume[:external]
+          outcome = Volumes::Create.run(grid: self.grid, **volume.symbolize_keys)
```
SpComb: AFAIK this prevents installing multiple copies of the same stack, and if fix this naming later, it will lose any installed stack volumes.

server/app/routes/v1/volumes_api.rb
SpComb: I meant drop this entire method and just use the common `load_grid` instead :)

SpComb:
@jnummelin please mark `status/merge` unless you still want to fix some of the minor things.

server/app/mutations/stacks/common.rb
```diff
@@ -14,6 +14,12 @@ def handle_service_outcome_errors(service_name, messages, type)
       end
     end
 
+    def handle_volume_outcome_errors(volume_name, errors)
+      errors.each do |key, atom|
+         add_service_error(volume_name, key, atom.symbolic, atom.message)
```
SpComb: This uses the  #1976 `add_service_error` which is not yet in git master, and adds a `services.$volume.$key` error, which doesn't make very much sense. See previous comment: https://github.com/kontena/kontena/pull/1849#discussion_r107435002

```ruby
add_error("volumes.#{volume_name}.#{key}", atom.symbolic, message)
```

jnummelin: I'll fix this

PR #1607, opened 2017-01-02
Improve how agent rpc server handles requests

jakolehm:
This PR has following improvements:

- RpcServer is now pool of actors
- WebsocketClient is passed to RpcServer actor and actor is called with async
- ServicePod requests return only after they have completed
- server can wait request to finish (no polling)
- wait for port functionality has been moved to agent side

PR #1855, opened 2017-02-14
Refactor all agent communication to msgpack rpc

jakolehm:
Depends on #1607


SpComb:
I think you can edit the PR to temporarily set the merge-base to `feature/improved-agent-rpc-server`, which would make this easier to review.

agent/lib/kontena/workers/container_info_worker.rb
```diff
@@ -63,14 +63,11 @@ def publish_info(container)
...
-        data: {
-          node: self.node_info['ID'],
-          container: data
-        }
+      data = {
+        node: self.node_info['ID'],
+        container: data
       }
-      self.queue << event
+      rpc_client.async.notification('/containers/save', [data])
```
SpComb: `/containers/save` vs `/nodes/update`?

server/app/services/rpc/node_handler.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def initialize(grid)
+      @grid = grid
+    end
+
+    def get(id)
+      node = @grid.host_nodes.find_by(node_id: id)
+      if node
+        template = Tilt.new('app/views/v1/host_nodes/_host_node.json.jbuilder')
+        JSON.parse(template.render(nil, node: node))
```
SpComb: Oh... is this really the easiest way? :D

server/app/services/rpc_server.rb
```diff
@@ -13,30 +22,76 @@ def initialize(code, message, backtrace = nil)
...
-    if klass = HANDLERS[handler]
+    if actor = handling_actor(grid_id, handler)
       begin
-        result = klass.new(grid).send(method, *message[3])
-        return [1, msg_id, nil, result]
+        result = actor.send(method, *message[3])
+        send_message(ws_client, [1, msg_id, nil, result])
+      rescue Celluloid::DeadActorError
+        error "actor for handler #{handler} is dead, removing it from cache"
+        @handlers[grid_id].delete(handler)
```
SpComb: mm.... should we be removing the actor from `@handlers` whenever any exception is raised, because that is going to be a dead actor.

server/app/services/rpc_server.rb
```diff
@@ -13,30 +22,76 @@ def initialize(code, message, backtrace = nil)
...
       rescue RpcServer::Error => exc
```
SpComb: The handlers can't really raise `RpcServer::Error` anymore if they are separate actors.

server/app/services/rpc_server.rb
```diff
@@ -1,6 +1,15 @@
...
+require_relative 'rpc/node_service_handler'
+
 class RpcServer
+  include Celluloid
+  include Logging
 
   HANDLERS = {
+    'containers' => Rpc::ContainerHandler,
+    'nodes' => Rpc::NodeHandler,
+    'node_services' => Rpc::NodeServiceHandler
```
SpComb: This looks a bit odd here, part of a follow-up PR?

jakolehm: yes, it somehow got here from other pr while I was doing the split

server/app/services/rpc/node_handler.rb
jakolehm: Yes and no. This PR does not include enhanced serializers... but the follow up PR does.

jakolehm:
> I think you can edit the PR to temporarily set the merge-base to feature/improved-agent-rpc-server, which would make this easier to review.

Done

SpComb:
Needs to be rebased over the squashed #1607 based branch in master: 7c2b4654c3cd8c7a18a0814b5d4af67484a74729

SpComb:
The e2e tests pass (using the rebased branch).

server/app/services/rpc_server.rb
SpComb: I mean: if any RPC `handling_actor.send(...)` raises an exception, then that actor is dead. We should invalidate that actor immediately on any errors. The current implementation will unnecessarily also fail the following RPC with an avoidable `Celluloid::DeadActorError`.

agent/lib/kontena/rpc_client.rb
```diff
@@ -0,0 +1,68 @@
...
+      future
+    end
+
+    def handle_response(response)
+      type, msgid, error, result = response
+      @requests[msgid] = [result, error]
+    end
+
+    def request_id
+      rand(2_147_483_647)
```
jnummelin: Collisions might be extremely rare, but they could happen. :) Maybe use UUID? Although UUID is naturally slower to generate, I don't see it being any bottleneck.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -105,16 +101,13 @@ def run_hooks(service_container, type)
...
-                  data: chunk
-              }
+          data = {
+              id: id,
+              time: Time.now.utc.xmlschema,
+              type: type,
+              data: chunk
           }
-          Celluloid::Actor[:queue_worker].send_message(msg)
+          rpc_client.async.notification('/containers/log', [data])
```
jnummelin: Isn't notification always async from callers point of view? Could the `async` part be moved within the rcp_client somehow?

agent/lib/kontena/rpc_client.rb
SpComb: MessagePack-RPC specs this as an `uint32`. Since this is an Actor, this could just be `@request_id += 1` instead?

agent/lib/kontena/service_pods/creator.rb
SpComb: The `async` makes sense here, no need to block the `ServicePods::Creator` actor if the `RpcClient` actor is busy.

agent/lib/kontena/rpc_client.rb
jakolehm: This should be fixed now.. `#request_id` takes into account existing request id's and retries until it gets available number.

SpComb:
New comment because the existing comment thread somehow went read-only O_o

> sure @request_id += 1 is much simpler/faster but what happens if rpc client crashes after first request is sent (spoiler: next one gets the same id)?

Right... it would need to initialize `@request_id = rand(...)` and then also handle uint32 wrapping on `@request_id += 1`.

> Actually that could happen also with this implementation (if moon phase is right etc..)

Should be fine. Just need to verify that the `@requests` hash doesn't leak too badly...

agent/lib/kontena/rpc_client.rb
```diff
@@ -0,0 +1,75 @@
...
+    def request(method, params)
+      id = request_id
+      @requests[id] = nil
+      future = Celluloid::Future.new {
+        sleep 0.01 until @client.connected?
+        @client.send_request(id, method, params)
+        time = Time.now.utc
+        until !@requests[id].nil?
+          sleep 0.01
+          raise TimeoutError.new(500, 'Request timed out') if time < (Time.now.utc - 30)
```
SpComb: This leaks the `@requests[id]`... but it has to, if it wants to avoid re-using that `request_id` :)

server/app/services/rpc/node_handler.rb
SpComb: Indeed, this matches the existing server approach, so NVM.

server/app/services/rpc_server.rb
SpComb: I guess it's okay for the handling Actor to sacrifice itself if throwing some "invalid request" error... :/

The handler caching thing here is just a performance optimization, I guess.

server/app/services/rpc/container_handler.rb
```diff
@@ -0,0 +1,126 @@
...
+          type: data['type'],
+          data: data['data']
+        }
+        if @logs.size >= 5
+          flush_logs
+        end
+      end
+    end
+
+    def flush_logs
```
SpComb: Because the new server `RpcServer` shares a single `Rpc::ContainerHandler` actor for multiple RPC calls, there can be multiple `Rpc::ContainerHandler.send(:log, ...)` tasks running at the same time... is this code Celluloid task-safe?

server/app/services/rpc_server.rb
```diff
@@ -13,30 +22,81 @@ def initialize(code, message, backtrace = nil)
...
-    return [1, msg_id, {error: 'grid not found'}, nil] if grid.nil?
-
     handler = message[2].split('/')[1]
     method = message[2].split('/')[2]
-    if klass = HANDLERS[handler]
+    if actor = handling_actor(grid_id, handler)
       begin
-        result = klass.new(grid).send(method, *message[3])
-        return [1, msg_id, nil, result]
+        result = actor.send(method, *message[3])
```
SpComb: Thinking... these are run as multiple concurrent Celluloid actor `rpc_server.async.handle_notification` tasks, and so this sync `actor.send(...)` is an Celluloid task scheduling point: https://github.com/celluloid/celluloid/wiki/Pipelining%20and%20execution%20modes

 Does this risk any concurrency issues between the shared `@handlers` state before and after this line? I guess maybe not, since the only operation we do after this line is the `@handlers[...].delete(...)`, which probably won't explode..?

server/app/services/rpc_server.rb
```diff
@@ -13,30 +22,81 @@ def initialize(code, message, backtrace = nil)
...
   end
+
+  # @param [String] grid_id
+  # @param [Array] message msgpack-rpc notification array
+  def handle_notification(grid_id, message)
+    handler = message[1].split('/')[1]
+    method = message[1].split('/')[2]
+    if actor = handling_actor(grid_id, handler)
+      begin
+        debug "rpc notification: #{actor.class.name}##{method} #{message[2]}"
```
SpComb: This makes the server `LOG_LEVEL=0` significantly more verbose than before... but also more useful. It's just pretty spammy now with all the very large parameter hashes being logged.

server/app/services/rpc/container_handler.rb
SpComb: It probably is, because I don't spot any celluloid task scheduling points in this Actor. I assume the `@db_session` operations will block the actor and prevent any concurrent `logs(...)` tasks from appending to the `@logs`.

It just needs to be really careful... insert some random `sleep` call somewhere in here, and you might end up with duplicated `ContainerLogs` entries in mongodb?

SpComb: Confirmed that the following experiment results in duplicated logs:

```patch
--- a/server/app/services/rpc/container_handler.rb
+++ b/server/app/services/rpc/container_handler.rb
@@ -52,6 +52,7 @@ module Rpc
 
     def flush_logs
       @db_session[:container_logs].insert(@logs)
+      sleep 5
       @logs.clear
     end
 ```

Just something to watch out for and consider when working on these actors...

The results look amusing when restarting redis :)

2017-03-01-134656_665x2159_scrot

server/app/services/rpc_server.rb
SpComb: BTW: if we did want to fix this, then that would probably mean moving this request-handler-reply logic into the actual handler actors themselves... that is, include some `RpcHandler` mixin into each RPC handler class that contains the method dispatch, with local error handling within the actor.

server/app/services/rpc_server.rb
jakolehm: I agree that this is too verbose.. maybe we should remove it?

SpComb:
YOLO

PR #1757, opened 2017-01-31
Agent: Upgrade to faye-websocket 0.10.7 with connection error reasons, close timeouts

SpComb:
Fixes #1749

Bumps `faye-websocket` to [0.10.7](https://github.com/faye/faye-websocket-ruby/releases/tag/0.10.7) to emit `:error` reasons from EM: https://github.com/faye/faye-websocket-ruby/issues/102 https://github.com/faye/faye-websocket-ruby/commit/39e26115b39c83e923dc0656548fad52dd736956

This also replaces the close timeout implemented by the agent `Websocketclient` in #1602 with the new faye-websocket close timeout. This includes a workaround for https://github.com/faye/faye-websocket-ruby/issues/103

#### Connecting to a server that is stopped:
```
kontena-agent | I, [2017-02-01T12:22:55.926625 #1]  INFO -- Kontena::WebsocketClient: connecting to master at ws://192.168.66.1:9292
kontena-agent | E, [2017-02-01T12:22:56.001138 #1] ERROR -- Kontena::WebsocketClient: connection refused: ws://192.168.66.1:9292
kontena-agent | I, [2017-02-01T12:22:56.001186 #1]  INFO -- Kontena::WebsocketClient: connection closed with code 1006
```

#### Connecting to a `ws://` server using `wss://`
```
kontena-agent | I, [2017-02-01T12:23:43.189657 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://192.168.66.1:9292
kontena-agent | E, [2017-02-01T12:23:43.213118 #1] ERROR -- Kontena::WebsocketClient: protocol error, check ws/wss: wss://192.168.66.1:9292
kontena-agent | I, [2017-02-01T12:23:43.213176 #1]  INFO -- Kontena::WebsocketClient: connection closed with code 1006
```

PR #1860, opened 2017-02-15
Fluentd log shipping

### LINKED ISSUE #1719: Enable us to pipe service(/cluster?) logs to ELK Stack for example
Currently kontana stores the logs on the master, which is better than having them on the nodes, but searchability is limited and we can't for example build reports and dashboards around that using common tools like Kibana.
As you mentioned in Gutter it's possible to scrape the logs from the mongo DB on the master, but this has to be done outside of the cluster which is not ideal, because we have to maintain dedicated hosts to handle the logs and can't use docker resp. kontena to host elasticsearch, kibana etc.

It would be great if additionally(?) we could activate something like filebeat/logspout/logstash or something like that, that also pipes the data into elasticsearch so we can better search and monitor the logfiles. Especially when you have things like zipkin etc. to follow request through your system, you need some aggregated view on the logs.

I think it would be sufficient for kontena to provide a service that gathers the logs and is able to pipe them into logstash server or elasticsearch directly. This would help a lot!

https://hub.docker.com/r/sebp/elk/ is a nice combined docker image that I used with docker swarm.

jnummelin:
Added support for direct log shipping using fluentd.

Can be enabled from cli using:
`kontena grid update --log-driver fluentd --log-opt fluentd-address=localhost:24224 test`

Can be disabled using:
`kontena grid update --log-driver none test`

Still few todos:

- [x] add proper tagging based on service, stack, instance, node etc., https://github.com/kontena/kontena/pull/1860/commits/9451655ad1719aaebf779aba27ffc5ca20d8ab0e
- [x] try to direct fluentd logger logs to standard logger, https://github.com/kontena/kontena/pull/1860/commits/f814868759744afc7d79ff8507768d95a73d79e4
- [x] docs, https://github.com/kontena/kontena/pull/1860/commits/84016882ea240252a874b1c4a6dd9efd5d217d94
- [x] check if there are other relevant options to support, IMO no
- [x] should we validate fluentd opts on the server mutation

fixes #1719

### DIFF
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -118,6 +118,14 @@ module Docker
       self.labels['io.kontena.service.instance_number'].to_i
     end
 
+    def service_name
+      self.labels['io.kontena.service.name'].to_s
+    end
+
+    def stack_name
+      self.labels['io.kontena.stack.name'].to_s
+    end
+
     # @return [Hash]
     def env_hash
       if @env_hash.nil?
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -119,11 +119,11 @@ module Docker
     end
 
     def service_name
-      self.labels['io.kontena.service.name'].to_s
+      (self.labels['io.kontena.service.name'] || self.name).to_s
     end
 
     def stack_name
-      self.labels['io.kontena.stack.name'].to_s
+      (self.labels['io.kontena.stack.name'] || 'system').to_s
     end
 
     # @return [Hash]
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -63,13 +63,16 @@ module Kontena::Workers
           event: EVENT_NAME,
           data: {
               id: id,
+              service: @container.service_name,
+              stack: @container.stack_name,
+              instance: @container.instance_number,
               time: time.utc.xmlschema,
               type: stream,
               data: data
           }
       }
       @queue << msg
-      publish_log(msg)
+      publish_log(msg.dig('data', 'data'))
     end
 
     def publish_log(log)
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -60,7 +60,7 @@ module Kontena::Workers
       data = match[2]
       msg = {
         id: id,
-        service: @container.service_name,
+        service: @container.service_name || @container.name,
         stack: @container.stack_name,
         instance: @container.instance_number,
         time: time.utc.xmlschema,
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -4,7 +4,6 @@ module Kontena::Workers
     include Kontena::Logging
 
     CHUNK_REGEX = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)\s(.*)$/
-    EVENT_NAME = 'container:log'
 
     # @param [Docker::Container] container
     # @param [Queue] queue
@@ -60,7 +59,7 @@ module Kontena::Workers
       time = DateTime.parse(match[1])
       data = match[2]
       msg = {
-          event: EVENT_NAME,
+          event: 'container:log',
           data: {
               id: id,
               service: @container.service_name,
@@ -76,7 +75,7 @@ module Kontena::Workers
     end
 
     def publish_log(log)
-      Celluloid::Notifications.publish(EVENT_NAME, log)
+      Celluloid::Notifications.publish('container:log', log)
     end
   end
 end
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -59,19 +59,16 @@ module Kontena::Workers
       time = DateTime.parse(match[1])
       data = match[2]
       msg = {
-          event: 'container:log',
-          data: {
-              id: id,
-              service: @container.service_name,
-              stack: @container.stack_name,
-              instance: @container.instance_number,
-              time: time.utc.xmlschema,
-              type: stream,
-              data: data
-          }
+        id: id,
+        service: @container.service_name,
+        stack: @container.stack_name,
+        instance: @container.instance_number,
+        time: time.utc.xmlschema,
+        type: stream,
+        data: data
       }
       @queue << msg
-      publish_log(msg.dig('data', 'data'))
+      publish_log(msg)
     end
 
     def publish_log(log)
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -76,7 +76,7 @@ module Kontena::Workers
     end
 
     def publish_log(log)
-      Celluloid:: Notifications.publish(EVENT_NAME, log)
+      Celluloid::Notifications.publish(EVENT_NAME, log)
     end
   end
 end
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -60,7 +60,7 @@ module Kontena::Workers
       data = match[2]
       msg = {
         id: id,
-        service: @container.service_name || @container.name,
+        service: @container.service_name,
         stack: @container.stack_name,
         instance: @container.instance_number,
         time: time.utc.xmlschema,
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -38,7 +38,9 @@ module Kontena::Workers
       }
     rescue Excon::Errors::SocketError => exc
       since = Time.now.to_f
-      error "log socket error: #{@container.id}"
+      error "log socket error for container: #{@container.id}"
+      error "#{exc.class.name}: #{exc.message}"
+      error exc.backtrace.join("\n")
       retry
     rescue Docker::Error::TimeoutError
       since = Time.now.to_f
@@ -57,12 +59,21 @@ module Kontena::Workers
       return unless match
       time = DateTime.parse(match[1])
       data = match[2]
-      @queue << {
-          id: id,
-          time: time.utc.xmlschema,
-          type: stream,
-          data: data
+      msg = {
+          event: EVENT_NAME,
+          data: {
+              id: id,
+              time: time.utc.xmlschema,
+              type: stream,
+              data: data
+          }
       }
+      @queue << msg
+      publish_log(msg)
+    end
+
+    def publish_log(log)
+      Celluloid:: Notifications.publish(EVENT_NAME, log)
     end
   end
 end
--- a/agent/lib/kontena/workers/fluentd_worker.rb
+++ b/agent/lib/kontena/workers/fluentd_worker.rb
@@ -32,7 +32,8 @@ module Kontena::Workers
         @fluentd = Fluent::Logger::FluentLogger
           .new("#{node_name}.#{info.dig('grid' 'name')}",
               :host => host,
-              :port => port || 24224)
+              :port => port || 24224,
+              :logger => Kontena::Logging.logger)
         @forwarding = true
       else
         info "stopping fluentd log streaming"
--- a/agent/lib/kontena/workers/fluentd_worker.rb
+++ b/agent/lib/kontena/workers/fluentd_worker.rb
@@ -24,16 +24,15 @@ module Kontena::Workers
     # @param [Hash] info
     def on_node_info(topic, info)
       node_name = info['name']
-      driver = info.dig('grid', 'logs', 'driver')
+      driver = info.dig('grid', 'logs', 'forwarder')
       if driver == 'fluentd'
         fluentd_address = info.dig('grid', 'logs', 'opts', 'fluentd-address')
         info "starting fluentd log streaming to #{fluentd_address}"
         host, port = fluentd_address.split(':')
         @fluentd = Fluent::Logger::FluentLogger
-          .new("#{node_name}.#{info.dig('grid' 'name')}",
+          .new("#{node_name}.#{info.dig('grid', 'name')}",
               :host => host,
-              :port => port || 24224,
-              :logger => Kontena::Logging.logger)
+              :port => port || 24224)
         @forwarding = true
       elsif @forwarding
         info "stopping fluentd log streaming"
--- a/agent/lib/kontena/workers/fluentd_worker.rb
+++ b/agent/lib/kontena/workers/fluentd_worker.rb
@@ -27,7 +27,7 @@ module Kontena::Workers
       driver = info.dig('grid', 'logs', 'driver')
       if driver == 'fluentd'
         fluentd_address = info.dig('grid', 'logs', 'opts', 'fluentd-address')
-        puts "starting fluentd log streaming to #{fluentd_address}"
+        info "starting fluentd log streaming to #{fluentd_address}"
         host, port = fluentd_address.split(':')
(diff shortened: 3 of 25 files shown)


agent/lib/kontena/workers/container_log_worker.rb
```diff
@@ -61,12 +63,20 @@ def on_message(id, stream, chunk)
...
               type: stream,
               data: data
           }
       }
       @queue << msg
+      publish_log(msg.dig('data', 'data'))
+    end
+
+    def publish_log(log)
+      Celluloid:: Notifications.publish(EVENT_NAME, log)
```
SpComb: Odd space after `Celluloid::`

jnummelin:
I thought first of having a `LogShipperWorker` to listen the `agent:node_info` events which would then spin up the protocol specific actor to do the actual log shipping. It would make the actual log shipper implementation a bit simplified but maybe bit over engineering at this point when we support only fluentd?

jakolehm:
Just thinking if `fluentd` is a good choice as name here? Isn't it a bit confusing to have possibility to configure grid level fluentd log driver that is not the same as service (or container) level log driver?

jnummelin:
@jakolehm So you mean it might be confusing in sense that users would expect this to configure container log driver instead of shipping the logs from agent?

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -0,0 +1,60 @@
...
+    ##
+    # @param [Queue] queue
+    # @param [Boolean] autostart
+    def initialize(autostart = true)
+      @fluentd = nil
+      @queue = []
+      @forwarding = false
+      info 'initialized'
+      subscribe('agent:node_info', :on_node_info)
+      subscribe(Kontena::Workers::ContainerLogWorker::EVENT_NAME, :on_log_event)
```
SpComb: IMO it would be clearer to just use the literal `subscribe('container:log')` event name both here and in the publish. Using this kind of constant is harder to grep for.

jnummelin: 50/60 :)
Maybe I'm just old-school, but I find using "random strings" a bit smelly in sense. We do it quite a lot in our code base... 
But I can live with plain strings if others see that as better option.

SpComb: Finding all usages of a generic `EVENT_NAME` constant across the different module/class level scopes is harder than grepping for a string literal.

server/app/models/grid.rb
```diff
@@ -11,6 +11,7 @@ class Grid
   field :trusted_subnets, type: Array, default: []
   field :stats, type: Hash, default: {}
   field :default_affinity, type: Array, default: []
+  field :logs, type: Hash, default: {}
```
jnummelin: Sure, I think its good idea.

SpComb:
Needs merge with #1855

jnummelin:
rebased

agent/lib/kontena/workers/container_log_worker.rb
```diff
@@ -57,12 +58,24 @@ def on_message(id, stream, chunk)
...
+              id: id,
+              service: @container.service_name,
+              stack: @container.stack_name,
+              instance: @container.instance_number,
+              time: time.utc.xmlschema,
+              type: stream,
+              data: data
+          }
       }
+      @queue << msg
```
SpComb: This should just be `@queue << { id: ... }`

agent/lib/kontena/workers/container_log_worker.rb
```diff
@@ -57,12 +58,24 @@ def on_message(id, stream, chunk)
...
+      publish_log(msg.dig('data', 'data'))
```
SpComb: eh... just `publish_log(data)` :D

SpComb: ...except the `FluentdWorker` expects this to be the `{service: ...}` hash.

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -0,0 +1,60 @@
...
+        info "starting fluentd log streaming to #{fluentd_address}"
+        host, port = fluentd_address.split(':')
+        @fluentd = Fluent::Logger::FluentLogger
+          .new("#{node_name}.#{info.dig('grid' 'name')}",
+              :host => host,
+              :port => port || 24224,
+              :logger => Kontena::Logging.logger)
+        @forwarding = true
+      else
+        info "stopping fluentd log streaming"
```
SpComb: Doesn't this get logged spuriously every time the `node_info` updates, if fluentd logging is not configured?

SpComb: Maybe `elsif @forwarding`?

cli/lib/kontena/cli/grids/update_command.rb
```diff
@@ -8,10 +8,13 @@ class UpdateCommand < Kontena::Command
     parameter "NAME", "Grid name"
     option "--statsd-server", "STATSD_SERVER", "Statsd server address (host:port)"
     option "--default-affinity", "[AFFINITY]", "Default affinity rule for the grid", multivalued: true
+    option "--log-driver", "LOG_DRIVER", "Set grid wide log driver"
+    option "--log-opt", "[LOG_OPT]", "Set log options (key=value)", multivalued: true
```
SpComb: Ideally these same options would also exist on `kontena grid create`.... but it's not just the `--log` options that suffer from this.

docs/using-kontena/logs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Kontena streams all logs of the services deployed and stores them in Kontena Master for easy access.
+
+There are various methods to access this log information.
+
+- `kontena grid logs` for grid wide logs (all services and Kontena internal components)
+- `kontena service logs xyz` for individual service logs
+- `kontena stack logs xyz` for stacks service logs
+
+The logs are stored in capped collection which ensures that old events gets automatically removed and do not consume disk space on master database.
```
SpComb: The logs are stored in a capacity-bounded MongoDB collection, which limits the disk space consumed on the master by automatically removing old logs.

docs/using-kontena/logs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Often there is a need to further process the logs and gather some relevant statistics and insight what is happening in your services. To ship the logs to some other system there are three different alternatives outlined in below chapters.
+
+### Enable fluentd shipping
+
+Kontena supports fluentd log shipping that can be configured on each grid. When fluentd shipping is enabled, all container logs are automatically sent to fluentd for further processing **in addition** of storing them in Kontena Master.
+
+You can enable fluentd forwarding using:
+```
+kontena grid update --log-driver fluentd --log_opt fluentd-address=server:port my-grid
```
SpComb: `--log-opt`

test/spec/features/grid/update_spec.rb
```diff
@@ -0,0 +1,14 @@
+describe 'update grid' do
+  it 'validates log opts' do
+    k = run "kontena grid update --log-opt server=foo e2e"
+    expect(k.code).not_to eq(0)
+    expect(k.out).to match /Need to specify --log-driver when using --log-opt/
```
SpComb: These kind of CLI-only tests should be cli specs...?

SpComb:
Also document the new grid options in `docs/using-kontena/grids.md`

SpComb:
> Just thinking if fluentd is a good choice as name here? Isn't it a bit confusing to have possibility to configure grid level fluentd log driver that is not the same as service (or container) level log driver?

Agreed, `kontena grid update --log-driver fluentd` vs `kontena service create --log-driver fluentd` are confusingly similar... 

The fluentd logging structure also seems to be different than the example in the [fluentd docs](http://www.fluentd.org/guides/recipes/docker-logging)? `log` vs `event`?

```
2015-09-01 15:10:40 -0600 docker.3fd8678d487e: {"source":"stdout","log":"Hello Fluentd!","container_id":"3fd8678d487e540c7a303e1613101e746c5012f3317434eda93f24351c1928f7","container_name":"/angry_kalam"}
```

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -0,0 +1,60 @@
...
+        @fluentd = nil
+      end
+    end
+
+    def on_log_event(topic, log)
+      # TODO Get more tags
+      # maybe stack.service.instance
+      #puts "forwarding: #{@forwarding}, fluentd: #{@fluentd}"
+      if @forwarding && @fluentd
+        tag = [log[:stack], log[:service], log[:instance]].join('.')
```
SpComb: Missing the `grid` mentioned in the docs: 

> Each event sent to fluentd is tagged with following notation:`grid.stack.service.instance`

SpComb: Not familar with fluentd... should this tag also include some `kontena` prefix?

SpComb:
Leaving the `status/design-review` label because of the `fluentd` vs `fluentd` log driver confusion.

agent/lib/kontena/workers/fluentd_worker.rb
jnummelin: No, the grid part is actually configured when instantiating the fluentd forwarder:
```
@fluentd = Fluent::Logger::FluentLogger
          .new("#{node_name}.#{info.dig('grid' 'name')}",
              :host => host,
              :port => port || 24224,
              :logger => Kontena::Logging.logger)
```

jnummelin:
Any proposals how to handle the driver confusion? I was thinking whether it makes it more clear if we rename the `--log-driver` to `log-shipper`?

agent/lib/kontena/workers/container_log_worker.rb
jnummelin: Yes, the `FluentdWorker` expects to get the service related data to be able to tag the event properly.

test/spec/features/grid/update_spec.rb
jnummelin: added cli spec for this also. A bit redundant but didn't want to remove from e2e tests :D

jnummelin:
Fixed the log sending structure, dunno what brainfart I had... :D 

The only thing still to figure out is the driver confusion.

SpComb:
The `agent/Gemfile.lock` needs a merge with #1757

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -44,13 +44,10 @@ def on_node_info(topic, info)
     end
 
     def on_log_event(topic, log)
-      # TODO Get more tags
-      # maybe stack.service.instance
-      #puts "forwarding: #{@forwarding}, fluentd: #{@fluentd}"
```
SpComb: BTW: These agent log-on-log things are amusing... you get an infinite log amplification DoS :)

We should figure out some fix for that, there's some parts of the agent that you just can't do any debug logging in ^^

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -44,13 +44,10 @@ def on_node_info(topic, info)
...
       if @forwarding && @fluentd
         tag = [log[:stack], log[:service], log[:instance]].join('.')
         record = {
-          event: log[:data], # the actual log event
+          log: log[:data], # the actual log event
```
SpComb: Also needs update to the docs `{ event: <log data> }` bit

:+1: for conforming to the Docker fluent logging driver

docs/using-kontena/logs.md
```diff
@@ -0,0 +1,92 @@
...
+kontena grid update --log-driver fluentd --log-opt fluentd-address=server:port my-grid
+```
+
+Each event sent to fluentd is tagged with following notation:
+`grid.stack.service.instance`
+
+The record itself is a hash with following semantics:
+```
+{
+  event: <log data>,
```
SpComb: Commented on the wrong commit, I guess; this needs to be `log: `

jakolehm:
Imho `--log-driver` sounds like a grid wide setting for services. And because this PR implements a bit different thing we should really come up with a name that describes this feature better (or deprecate service level log-driver).

jnummelin:
Some options to "rename" the options:
- `--log-forwarder=fluentd ...`
- `--log-shipper=fluentd ...`

@jakolehm @SpComb WDYT?

jakolehm:
`--log-forwarder` will get my vote.

docs/using-kontena/grids.md
```diff
@@ -59,6 +59,13 @@ The `kontena grid update --statsd-server HOST:PORT` option configures each host
 
 See the [Statistics](stats.md#exporting-stats) documentation for further details.
 
+### Logging options
+
+The `kontena grid update --log-driver some-driver --log-opt xyz=foo` options configures each host node to send container logs to a remote log collection service.
```
SpComb: `--log-forwarder`

docs/using-kontena/logs.md
```diff
@@ -0,0 +1,92 @@
...
+The record itself is a hash with following semantics:
+```
+{
+  log: <log data>,
+  type: stdout / stderr
+}
+```
+
+To disable log forwarding, use:
+`kontena grid update --log-driver none my-grid`
```
SpComb: `--log-forwarder`

server/app/mutations/grids/update.rb
```diff
@@ -23,6 +23,14 @@ class Update < Mutations::Command
       array :default_affinity do
         string
       end
+      hash :logs do
+        required do
+          string :forwarder, matches: /^(fluentd|none)$/ # Only fluentd now supported, none removes log shipping
```
SpComb: Could also be `in: ['fluentd', 'none']`?

server/app/mutations/grids/update.rb
```diff
@@ -71,5 +93,14 @@ def notify_nodes
         GridScheduler.new(grid).reschedule
       }
     end
+
+    def validate_fluentd_opts(opts)
+      address = opts['fluentd-address']
+      add_error(:logs, :forwarder, 'fluentd-address option must be given') unless address
+      if address
+        host, port = address.split(':')
+        add_error(:logs, :forwarder, 'fluentd-address must be given in format address:port') unless host && port
```
SpComb: The `:port` is optional on the agent?

test/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       - mongodb
 
   agent:
-    container_name: kontena-agent
+    container_name: e2e-kontena-agent
```
SpComb: Odd, does this somehow affect the log forwarding?

server/app/mutations/grids/update.rb
SpComb: Sadly the resulting errors are a bit obtuse either way, although with `in`, the resulting error could include the valid options...

### `matches: ...`
```
$ kontena grid update --log-forwarder asdf development
 [error] {"logs"=>{"forwarder"=>"Forwarder isn't in the right format"}}
```

### `in: ...`
```
$ kontena grid update --log-forwarder asdf development
 [error] {"logs"=>{"forwarder"=>"Forwarder isn't an option"}}
```

agent/lib/docker/container.rb
```diff
@@ -118,6 +118,14 @@ def instance_number
       self.labels['io.kontena.service.instance_number'].to_i
     end
 
+    def service_name
+      (self.labels['io.kontena.service.name'] || self.name).to_s
```
jnummelin: Yes, only for "system" containers

server/app/mutations/grids/update.rb
jnummelin: Should I change this to be "manual" validation in the mutation to make it give better errors?

test/docker-compose.yml
jnummelin: Nope, just something I fiddled with and seemed to commit by accident, will revert

SpComb:
Tweaked the squash-commit title and message a bit.

PR #1863, opened 2017-02-17
Do not start health check if no protocol specified

jnummelin:
Due to #1837 server added health check labels with nil protocol and port for every service. With those agent started the health check defaulting to use tcp and port 0. As this naturally doesn't provide healthy statuses, master was on constant re-scheduling loop for every single service.

Fixed by adding health check labels only if valid protocol is set.

### DIFF
--- a/server/app/models/grid_service_health_check.rb
+++ b/server/app/models/grid_service_health_check.rb
@@ -10,4 +10,11 @@ class GridServiceHealthCheck
 
   embedded_in :grid_service
 
+  def is_valid?()
+    return false if protocol.nil?
+    return false if protocol.empty?
+    return false if port.nil?
+    true
+  end
+
 end
--- a/server/app/models/grid_service_health_check.rb
+++ b/server/app/models/grid_service_health_check.rb
@@ -10,7 +10,7 @@ class GridServiceHealthCheck
 
   embedded_in :grid_service
 
-  def is_valid?()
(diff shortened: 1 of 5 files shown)


server/app/services/docker/service_creator.rb
```diff
@@ -131,7 +131,7 @@ def build_labels
         labels['io.kontena.load_balancer.internal_port'] = internal_port
         labels['io.kontena.load_balancer.mode'] = mode
       end
-      if grid_service.health_check
+      if grid_service.health_check && grid_service.health_check.protocol
```
nevalla: Is only protocol only required? Should we check that all health check options are in place?

server/app/models/grid_service_health_check.rb
```diff
@@ -10,4 +10,11 @@ class GridServiceHealthCheck
 
   embedded_in :grid_service
 
-end
\ No newline at end of file
+  def is_valid?()
```
nevalla: Very small coding style issue here. I think `valid?` without brackets is more commonly used with ruby.

jnummelin: `valid?` clashes with Mongoid hence is used ìs_valid?` instead. But I'll remove the unnecessary brackets.

nevalla: okay :+1:

PR #1865, opened 2017-02-19
Return better error message if cloud user not found

nevalla:


server/app/services/cloud/rpc/server_api.rb
```diff
@@ -55,7 +55,7 @@ def put(user_id, path, options)
 
       def require_user(user_id)
         @user = User.find_by(external_id: user_id)
-        raise RpcServer::Error.new(403, 'Forbidden') unless @user
+        raise RpcServer::Error.new(403, 'Cloud user not found. Please try re-login to master with CLI: kontena master login --force') unless @user
```
jakolehm: Isn't this a wrong place for these kind of messages? Client that consumes this api should generate proper end-user message based on error code...

nevalla: Yes you are right. Good error message but in wrong place. I still want to tweak error message sent by master a bit, but I'll try to make it more general.

nevalla:
Closed this. Will handle on the Cloud Dashboard

PR #1870, opened 2017-02-20
The registry url was not stored properly in stack metadata

### LINKED ISSUE #1844: kontena stack upgrade stackname searches for kontena.yml
```
kontena stack upgrade lb-ingress

 [done] Reading stack lb-ingress metadata from Kontena Master
 [error] No such file or directory @ rb_sysopen - /Users/mpa/dev/kontena-compose/kontena.yml
```

Shouldn't it upgrade from the original stack instead?

kke:
Fixes #1844

The "registry" field always contained "file:///" no matter what.

Now it contains the url of the registry in case the stack came from a registry. Otherwise it's nil.

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -71,7 +71,7 @@ module Kontena::Cli::Stacks
         'expose' => outcome[:expose],
         'version' => outcome[:version],
         'source' => reader.raw_content,
-        'registry' => 'file://',
+        'registry' => outcome[:registry],
         'services' => kontena_services,
         'variables' => outcome[:variables]
       }
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -31,7 +31,7 @@ module Kontena::Cli::Stacks
           @registry = file
         else
           @raw_content = File.read(File.expand_path(file))
-          @registry = "file:///#{File.basename(file)}"
+          @registry = "file://#{File.basename(file)}"
         end
 
         @errors           = []
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -25,13 +25,13 @@ module Kontena::Cli::Stacks
         if from_registry?
           require 'shellwords'
           @raw_content = Kontena::StacksCache.pull(file)
-          @registry    = current_account.name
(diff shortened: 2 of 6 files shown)


jnummelin:
> It's original intention was to store the url of the registry being used
Do you mean like `stack-registry.myorg.com` ?

kke:
Yes. Because it's possible to use other registries. And if we store the full url to the stack on the registry, then it's indistinguishable from any other url.

Perhaps need to add a field to stack on master.

jnummelin:
Why couldn't it be stored with full url in the same field?

kke:
Stacks have `stack` and `version` fields which will tell the original stack name + version on the registry.

I now changed the registry field to say "kontena" when the stack is from the kontena registry.

kke:
> Why couldn't it be stored with full url in the same field?

Stack can come from 1) file 2) url 3) registry.

When it's from a registry, there could be some special stuff in the future like pulling the `˜> 2.0` version. 

Having both registry urls and regular urls in there would make it difficult to identify if the target is a registry  or not.

SpComb:
> Having both registry urls and regular urls in there would make it difficult to identify if the target is a registry or not.

It doesn't make sense to store a non-registry thing in the `registry` attr.

I understand it to be the base URL / DNS name of something implementing the `http://stacks.kontena.io/stack/...` API.

SpComb:
For the case of `kontena stack install test/foo`, that would result in something like `{ registry: 'stacks.kontena.io', stack: 'test/foo', version: '0.1.0' }` on the server. The CLI `kontena stack upgrade` command would then know to request `https://stacks.kontena.io/stack/matti/whoami` to get the newest version.

The case of `kontena stack install http://whatever/stack.yml` is more interesting. I guess you would point it at something like `https://raw.githubusercontent.com/SpComb/kontena-examples/whoami/whoami/kontena.yml`, which always returns the newest version. I suppose that could be stored in some generic `url` attribute, and the CLI would fetch that again on `kontena stack upgrade`... OTOH the current stack registry also works exactly the same way, so ¯\\_(ツ)_/¯ it's all the same difference.

kke:
Well now it's always the registry url. Otherwise it's nil.

Figure out a solution for url based stacks in the future.

SpComb:
The server's `Stacks::Common` mutation's registry param should probably be moved from `required` to `optional`.

kke:
Well, it was a bug. Maybe it should have been solved in a more backwards compatible manner.

matti:
I thought I had gone insane when my other laptop worked (1.1.1) and the other didn't (1.1.2).

PR #1873, opened 2017-02-20
Refactor agent to pull services desired state from the master

jakolehm:
This PR changes how scheduling decisions happen inside the master and how scheduling decisions are transferred to agents.

- scheduler writes `GridServiceInstance` objects (instead of sending service pod information directly to agents)
- agent fetches scheduled instances via rpc api from master (`ServicePodManager`)
- `ServicePodManagerWorker` creates an actor per service pod (`ServicePodWorker`)
- `ServicePodWorker` handles the desired state of actual containers within container engine
- load balancer config removal is refactored into `GridServices::Delete` mutation, removal is triggered to a node via rpc call

Depends on #1855


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -23,37 +24,59 @@ def initialize(autostart = true)
...
-    # @param [String] topic
-    # @param [Hash] info
-    def on_node_info(topic, info)
-      statsd_conf = info.dig('grid', 'stats', 'statsd')
-      if statsd_conf
+    # @return [Node]
+    def node
+      if @node.nil?
+        while @node.nil?
+          @node = fetch_node
```
SpComb: This never updates the `@node` info?

agent/spec/lib/kontena/launchers/etcd_launcher_spec.rb
```diff
@@ -94,14 +98,14 @@
       allow(container).to receive(:running?).and_return(true)
       allow(container).to receive(:info).and_return({'Config' => {'Image' => 'etcd'}})
       expect(subject.wrapped_object).to receive(:add_dns)
-      node_info = {
+      node = Node.new(
```
SpComb: I sense a great conflict with #1323 :(

It refactors this spec: https://github.com/kontena/kontena/pull/1323/files#diff-88ba28942ba5ccb7c85e35a737d4f14f

SpComb:
Would it be possible to split the `node_info` refactoring from the `service_pods` refactoring? AFAIK they're not strictly tied together.

It would also make it easier to deal with the impending merge conflicts in #1323

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -211,20 +204,21 @@ def ensure_default_pool()
       @default_pool = @ipam_client.reserve_pool(DEFAULT_NETWORK, '10.81.0.0/16', '10.81.128.0/17')
     end
 
-    # @param [Hash] info
-    def start(info)
-      wait { images_exist? && !starting? }
+    def start
```
SpComb: This seems to still lack support for reconfiguring on `node_info` changes?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,106 @@
...
+        'stopped'
+      end
+    end
+
+    # @param [String] current_state
+    def sync_state_to_master(current_state)
+      state = {
+        service_id: service_pod.service_id,
+        instance_number: service_pod.instance_number,
+        state: current_state
```
SpComb: This overlaps with what the Docker event worker does?

I don't think we should necessarily be storing the actual running container state in the `GridServiceInstance`. It should just be the current manager state.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,106 @@
...
+      @service_pod = service_pod if service_pod
+      ensure_desired_state
+    end
+
+    def destroy
+      @service_pod.mark_as_terminated
+      update
+    end
+
+    def ensure_desired_state
```
SpComb: Must also consider error handling of the various `ensure_*` operations.

SpComb: The design needs something to report the resulting error states back to the master as part of the service instance state.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -159,6 +159,7 @@ def ensure_image(name)
       # @param [Docker::Container] service_container
       # @return [Boolean]
       def service_uptodate?(service_container)
+        return false unless service_container.running?
```
SpComb: No. This will nuke a stopped container instead of just starting it?

agent/lib/kontena/workers/node_info_worker.rb
jakolehm: Yes, just wondering should we just fetch/update node info on every iteration or not.

agent/spec/lib/kontena/launchers/etcd_launcher_spec.rb
jakolehm: Let's make conflicts great again... I think we should merge #1323 first and then we can see how to proceed with this.

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: I disagree.. storing the actual state in `GridServiceInstance` makes scheduler much more simpler (it does not have to care about containers).

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Yes, reporting errors (or progress) is still missing.

agent/lib/kontena/service_pods/creator.rb
jakolehm: Does it really matter? We don't really protect containers so why not swap to fresh one here?

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+
+    def populate_workers_from_docker
+      info "populating service pod workers from docker"
+      fetch_containers.each do |c|
+        service_pod = Kontena::Models::ServicePod.new(
+          'id' => "#{c.service_id}/#{c.instance_number}",
+          'service_id' => c.service_id,
+          'instance_number' => c.instance_number,
+          'desired_state' => "unknown"
```
jnummelin: I do miss enums fron Java days... :)  Would it make sense to use some enum like solution for these states instead of semi-random strings? E.g. https://github.com/dblock/ruby-enum looks pretty ok.

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -0,0 +1,126 @@
...
+      begin
+        unless workers[service_pod.id]
+          worker = ServicePodWorker.new(node)
+          self.link worker
+          workers[service_pod.id] = worker
+        end
+        workers[service_pod.id].async.update(service_pod)
+      rescue Celluloid::DeadActorError => exc
+        workers.delete(service_pod.id)
+        if retries == 1
```
jnummelin: I don't see `retries` being increased, so now this is not retried at all. The condition also seems a bit odd, did you intend `retries <= 1` with incrementing retry counter?

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Then this needs to look at the Docker events, or the state will be out of date?

SpComb:
Note that the MessagePack `server/app/serializers/rpc` bits here somewhat overlap with the JSON serializers in #1906. The `/service_pods/notify_update` mechanism here is also vaugely similar to the more general `EventStream` implementation in #1906.

They are not exactly equivalent in their current form, but one `EventStream` style implementation of mongodb pub-sub based real-time updates to models seems more appealing than having a mess of three/four different mechanims to react to state updates across our architecture...

* CLI API polling
* Cloud EventStream `GridService#update` RPCs
* Agent `/node_service_pods/list` polling + `/service_pods/notify_update` notifications
* LB Docker API + etcd based state

agent/lib/kontena/workers/service_pod_manager.rb
SpComb: IMO we should be avoiding any kind of retry logic like this. Either use a retry helper function to avoid these kinds of bugs, or better yet, just leave out the retry logic entirely.

Unexpected errors should just propagate and crash the actor. It must be able to restart itself in any case, and if we can rely on the restart logic, then we can get rid of a lot of complex and badly tested retry logic like this.

agent/lib/kontena/workers/service_pod_manager.rb
SpComb: I think this `'desired_state' => "unknown"` is perhaps a sign of mixing up two different things into the same `Kontena::Models::ServicePod`: the identity of a container <-> grid service instance vs the configuration state from the server.

The `populate_workers_from_docker` does not know anything about the server configuration state. It should only initialize the `ServicePodWorker` with the `{:id, :service_id, :instance_number}` identity, and then leave the first `populate_workers_from_master` -> `update` to actuall provide the `ServcePod` configuration state?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,106 @@
...
+      exclusive {
+        debug "state of #{service_pod.name}: #{service_pod.desired_state}"
+        service_container = get_container(service_pod.service_id, service_pod.instance_number)
+        if service_pod.running? && service_container.nil?
+          info "creating #{service_pod.name}"
+          ensure_running
+        elsif service_container && service_pod.running? && !service_container.running?
+          info "starting #{service_pod.name}"
+          ensure_running
+        elsif service_pod.running? && (service_container && service_container.service_revision != service_pod.service_revision)
```
SpComb: Some design thoughts re the currently missing support for restarting the container here: The `kontena service restart ...` should set some `restart_at` timestamp on the `GridService` / `GridServiceInstance`s, and then this agent `ServicePodWorker` can trigger a restart if `service_container.state.started_at < service_pod.restart_at`.

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Which is not necessarily a bad approach, it just means that this `ServicePodManager` has to also handle the agent's Docker `EventWorker` and `ContainerHealthCheckWorker` events.

Moving the `ContainerHealthCheckWorker` actions into the `ServicePodManager` should definitely be done in any case, to effetively solve the race conditions in https://github.com/kontena/kontena/pull/1699#issuecomment-274474724

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Compared to the current approach of leaving the non-declarative `/service_pods/restart` API in place, this would have the advantage of being more robust against temporary RPC errors, which will become more common if we simplify the agent code per let-it-crash.. If the agent is busy restarting to recover from a crash while the user triggers a `kontena service restart`, then the declarative `service_pod.restart_at` would let the agent pick up the restart command after recovering.

agent/lib/kontena/models/node.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  attr_reader :id,
+              :created_at,
+              :updated_at,
+              :name,
+              :labels,
+              :peer_ips,
+              :node_number,
+              :initial_member,
+              :grid
```
jakolehm: Probably yes.

agent/lib/kontena/workers/service_pod_manager.rb
jakolehm: I think we can actually get that initial state from docker...

SpComb:
Needs rebase over  0d28788

agent/lib/kontena/service_pods/creator.rb
jakolehm: No, because then desired_state is stopped and creator is not called at all.

jakolehm:
Rebased against master.

agent/lib/kontena/service_pods/creator.rb
SpComb: Yes, the `kontena service stop` will mark the `desired_state: 'stopped'`, and the agent `ServicePodWorker` will execute `Kontena::ServicePods::Stopper`.

Then you run `kontena service start`, the service pod is updated to `desired_state: running`, and the agent `ServicePodWorker` will execute `Kontena::ServicePods::Creator`.... which now destroys the existing stopped container and creates a new one.

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -110,10 +110,13 @@ def check_tcp_status(ip, port, timeout)
         data['status'] = 'unhealthy'
       end
       data
+<<<<<<< 822939a96818996084bfc7b1fd436f4c2573444b
```
SpComb: Your rebase was close, but not quite :D

SpComb: NB: this is just in the commit history, not in the complete branch diff, so it just looks weird, but it'll be fine in the squash commit.

agent/lib/kontena/service_pods/creator.rb
SpComb: This kind of change to `kontena service stop+start` would cause things like the service container's weave overlay IP to change.

server/app/mutations/grid_services/delete.rb
```diff
@@ -15,14 +15,29 @@ def validate
...
         notify_node(node) if node
       end
     end
 
+    # @param [HostNode] node
     def notify_node(node)
       RpcClient.new(node.node_id).notify('/service_pods/notify_update', 'terminate')
     end
+
+    def notify_lb_remove
```
SpComb: This isn't ideal, because it's again one of those things where if the RPC happens to fail somehow, then the service will get left behind in etcd indefinitely. I suppose that itself shouldn't really cause any issues though, as long as the upstream backends get cleaned out.

SpComb: Actually, this could be fixed, if the agent `/load_balancers/remove_service` RPC fails with an error, then it will propagate here and cause the `self.grid_service.destroy` to be skipped... the service delete will fail, and it can be retried later by the client... :)

Unfortunately it looks like the agent `Celluloid::Actor[:lb_configurer].remove_config` will just swallow exceptions, so it will currently fail silently on things like etcd errors :(

agent/lib/kontena/load_balancers/configurer.rb
```diff
@@ -14,7 +14,6 @@ class Configurer
     def initialize
       @etcd = Etcd.client(host: self.class.gateway, port: 2379)
       subscribe('lb:ensure_config', :on_ensure_config)
-      subscribe('lb:remove_config', :on_remove_config)
```
SpComb: Also remove the `def on_remove_config`.

server/app/mutations/grid_services/delete.rb
jakolehm: Not optimal but I don't want to bloat this PR more than necessary.. old way of removing lb config just does not work with this new logic so some changes are necessary.

agent/lib/kontena/service_pods/creator.rb
jakolehm: Yes, but container is swapped also on any scheduler change (for example if image has been updated). Anyway, I'll change this to "start" so that we can move on.

agent/lib/kontena/workers/service_pod_manager.rb
jakolehm: Fetching state from Docker is more complicated than I first thought.. not going to implement here.

jakolehm:
Rebased

agent/lib/kontena/load_balancers/configurer.rb
SpComb: :1st_place_medal:

agent/lib/kontena/load_balancers/configurer.rb
```diff
@@ -14,7 +14,6 @@ class Configurer
...
       subscribe('lb:remove_service', :on_remove_service)
```
SpComb: Wait, now the `remove_service` should be cleaned out, and this needs to be `lb:remove_config` instead... The `Kontena::ServicePods::Creator` still sends `lb:remove_config`.

agent/lib/kontena/load_balancers/configurer.rb
```diff
@@ -79,26 +74,8 @@ def ensure_config(container)
...
-    rescue => exc
-      error "#{exc.class.name}: #{exc.message}"
-    end
-
-
-    # @param [Docker::Container] container
-    def remove_service(container)
-      service_name = container.service_name_for_lb
+    # @param [String] service_name
+    def remove_config(service_name)
```
SpComb: BTW this function will behave hilarously if `service_name = nil`... it will wipe out all LB services from etcd :P

agent/lib/kontena/rpc/lb_api.rb
```diff
@@ -0,0 +1,16 @@
+require_relative '../models/node'
```
SpComb: Unnecessary `require`?

agent/lib/kontena/rpc/agent_api.rb
```diff
@@ -1,3 +1,5 @@
+require_relative '../models/node'
```
SpComb: Should these `models/...` requires be in the `kontena-agent.rb`, where everything else is required?

agent/lib/kontena/rpc/lb_api.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Kontena
+  module Rpc
+    class LbApi
+
+      # @param [String] lb_name
+      # @param [Hash] service_pod
+      def remove_service(lb_name, service_pod)
```
SpComb: The `lb_name` argument is not used?

agent/lib/kontena/service_pods/creator.rb
SpComb: Can this `return false` be reverted now that the `ServicePodWorker#ensure_running` invokes the `Kontena::ServicePods::Starter`?

The `Kontena::ServicePods::Creator` should never be called anymore if the container already exists and is up to date...

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -23,7 +24,6 @@ def initialize(autostart = true)
       @statsd = nil
       @stats_since = Time.now
       @container_seconds = 0
-      subscribe('websocket:connected', :on_websocket_connected)
```
SpComb: This should probably be restored.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -38,22 +38,36 @@ def start
       end
     end
 
-    # @param [String] topic
-    # @param [Hash] data
-    def on_websocket_connected(topic, data)
-      self.publish_node_info
-      self.publish_node_stats
+    # @return [Node, NilClass]
+    def fetch_node
```
SpComb: Not used anywhere ATM?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    def ensure_terminated
+      Kontena::ServicePods::Terminator.new(
+        service_pod.service_id, service_pod.instance_number
+      ).perform
+    end
+
+    def service_container_outdated?(service_container, service_pod)
+      creator = Kontena::ServicePods::Creator.new(service_pod)
```
SpComb: Should all of this `container_outdated?` logic be moved into the `ServicePodWorker`? It doens't make sense to make this decision here, and then also have the `Kontena::ServicePods::Creator` check the same condition... it's a bug if the `Kontena::ServicePods::Creator` gets invoked but decides not to do anything?

agent/spec/lib/kontena/rpc/agent_api_spec.rb
```diff
@@ -12,13 +12,4 @@
       sleep 0.01
     end
   end
-
-  describe '#node_info' do
```
SpComb: The RPC method is still there.

server/app/serializers/rpc/host_node_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+    def created_at
+      object.created_at.to_s
+    end
+
+    def updated_at
+      object.updated_at.to_s
+    end
+
+    def peer_ips
+      object.grid.host_nodes.ne(id: object.id).map{|n|
```
SpComb: As in https://github.com/kontena/kontena/pull/1906#discussion_r104626917, this logic is now duplicated in many places :(

It should be a `HostNode#peer_ips` method.

server/app/serializers/rpc/host_node_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+      }.compact
+    end
+
+    def initial_member
+      object.initial_member?
+    end
+
+    def grid
+      if grid = object.grid
+        GridSerializer.new(grid).to_hash
```
SpComb: :+1: for de-duplicating this vs the JSON version.

agent/lib/kontena/load_balancers/configurer.rb
```diff
@@ -76,6 +76,8 @@ def ensure_config(container)
 
     # @param [String] service_name
     def remove_config(service_name)
+      return if service_name.to_s.empty?
```
SpComb: Would it hurt to `raise` instead?

jakolehm: Probably yes, this is now called via notifications.. raising error would only crash this actor.

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: I agree but did not want to go any deeper here.

agent/lib/kontena/workers/node_info_worker.rb
jakolehm: removed

server/app/serializers/rpc/host_node_serializer.rb
jakolehm: I agree, not going to fix here.

agent/spec/lib/kontena/rpc/agent_api_spec.rb
jakolehm: fixed

SpComb:
Should the agent `Kontena::ServicePods::Creator#ntoify_master` -> RPC `/containers/event { status: 'deployed', deploy_rev: ...}` -> server `Rpc::ContainerHandler` `container.set(:deploy_rev => ...)` stuff be removed? Is the container `deploy_rev` still used for anything?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,119 @@
...
+      creator.container_outdated?(service_container) ||
+        creator.labels_outdated?(service_pod.labels, service_container) ||
+          creator.recreate_service_container?(service_container)
```
SpComb: This skips the `service_container.config['Image'] != service_pod.image_name` and `image_outdated?` checks in the `Kontena::ServicePods::Creator#service_uptodate?`

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    # @param [String] current_state
+    def sync_state_to_master(current_state)
+      state = {
+        service_id: service_pod.service_id,
+        instance_number: service_pod.instance_number,
+        state: current_state
+      }
+      rpc_client.async.notification('/node_service_pods/set_state', [node.id, state])
```
SpComb: The `state` is confusingly named here, it's called `pod` in the server handler, which makes it look like this is sending just  the `current_state` string at first glance...

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Needs some TODO issue for the out-of-date  server `ServiceInstance` state if the container crashes or whatever.

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -0,0 +1,131 @@
...
+      rescue Celluloid::DeadActorError => exc
+        workers.delete(service_pod.id)
+      end
+    end
+
+    def on_worker_exit(worker, reason)
+      workers.delete_if { |k, w| w == worker }
+    end
+
+    def finalize
```
jakolehm: It's just simpler to terminate all workers if manager crashes. We can improve this later if needed.

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -0,0 +1,131 @@
...
+      loop do
+        populate_workers_from_master
+        sleep 30
+      end
+    end
+
+    # @return [Node]
+    def node
+      while @node.nil?
+        @node = Actor[:node_info_worker].node
```
jakolehm: I agree, maybe we should remove `Actor[:node_info_worker].node` call and just wait until `@node` is available (this logic was ok before I restored `:on_node_info`)

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,119 @@
...
+    attr_accessor :service_pod
+
+    def initialize(node, service_pod = nil)
+      @node = node
+      @service_pod = service_pod
+      @prev_state = 'unknown'
+    end
+
+    def update(service_pod = nil)
+      @service_pod = service_pod if service_pod
```
jakolehm: Leftovers from refactoring, will cleanup.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,119 @@
...
+      ensure_desired_state
+    end
+
+    def destroy
+      @service_pod.mark_as_terminated
+      update
```
jakolehm: yes

server/app/mutations/grid_services/delete.rb
```diff
@@ -7,18 +7,39 @@ class Delete < Mutations::Command
...
       linked_from_services = self.grid_service.linked_from_services
       if linked_from_services.count > 0
         add_error(:service, :invalid, "Cannot delete service that is linked to another service (#{linked_from_services.map{|s| s.name}.join(', ')})")
       end
     end
 
     def execute
-      worker(:grid_service_remove).async.perform(self.grid_service.id)
+      nodes = self.grid_service.grid_service_instances.map{ |i| i.host_node }
+      notify_lb_remove if self.grid_service.linked_to_load_balancer?
```
SpComb: Hmm... if the service is in a restart loop, and restarts after this removal, but before the agent kills it, then the service will be re-added to etcd?

SpComb:
Server DB migrations need bumping:

* `19_add_indexes_to_container_log.rb`
* `19_create_grid_service_instance.rb`

server/app/services/grid_scheduler.rb
```diff
@@ -19,49 +19,114 @@ def reschedule
...
+      if should_reschedule_service?(service)
+        reschedule_service(service)
       end
     end
   rescue => exc
     error exc.message
     debug exc.backtrace.join("\n") if exc.backtrace
   end
 
+  def check_service(service)
```
SpComb: Not used

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -27,68 +23,37 @@ def deploy(node, instance_number, deploy_rev, creds = nil)
...
 
-  # @param [HostNode] node
-  # @param [String] instance_number
-  # @param [String] deploy_rev
-  def wait_for_service_to_start(node, instance_number, deploy_rev)
-    Timeout.timeout(30) do
-      sleep 0.5 until self.deployed_service_container_exists?(instance_number, deploy_rev)
+  # @param [GridServiceInstance] service_instance
+  def wait_for_service_to_start(service_instance)
+    Timeout.timeout(300) do
```
SpComb: Must fix in  #1987

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -10,15 +10,11 @@ def initialize(grid_service)
...
-  def deploy(node, instance_number, deploy_rev, creds = nil)
-    if !self.service_exists_on_node?(node, instance_number)
-      self.terminate_service_instance(instance_number)
-    end
-
-    self.create_service_instance(node, instance_number, deploy_rev, creds)
-    self.wait_for_service_to_start(node, instance_number, deploy_rev)
+  def deploy(node, instance_number, deploy_rev)
+    service_instance = create_service_instance(node, instance_number, deploy_rev)
+    notify_node(node)
```
SpComb: Also notify any old node where the service instance is currently running?

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -0,0 +1,48 @@
...
+      )
+      service_instance.set(state: pod['state']) if service_instance
+    end
+
+    # @return [Boolean]
+    def migration_done?
+      if @migration_done.nil?
+        last = SchemaMigration.last
+        if last
+          @migration_done = last.version >= 19
```
SpComb: Probably not version 19 anymore.

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -0,0 +1,48 @@
...
+    def initialize(grid)
+      @grid = grid
+    end
+
+    # @param [String] id
+    # @return [Array<Hash>]
+    def list(id)
+      node = @grid.host_nodes.find_by(node_id: id)
+      return { error: 'Node not found' } unless node
+      return { error: 'Migration not done' } unless migration_done?
```
SpComb: This protects against wiping out all containers in... what kind of cases?

server/app/services/rpc_server.rb
```diff
@@ -40,6 +40,10 @@ def handle_request(ws_client, grid_id, message)
       begin
         result = actor.send(method, *message[3])
         send_message(ws_client, [1, msg_id, nil, result])
+        unless actor.alive?
+          error "actor for handler #{handler} did die, removing it from cache"
```
SpComb: Does this ever trigger? If we're the only actor sending it messages, then we should see any exceptions that kill it?

server/app/views/v1/grid_services/_grid_service.json.jbuilder
```diff
@@ -33,9 +33,9 @@ json.stack do
   json.id grid_service.stack.to_path
   json.name grid_service.stack.name
 end
-json.links grid_service.grid_service_links.map{|s|
-  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.linked_grid_service.name }
-}
+json.links grid_service.grid_service_links.map { |s|
+  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.linked_grid_service.name } if s.linked_grid_service
```
SpComb: Looks like #1770?

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -27,68 +23,37 @@ def deploy(node, instance_number, deploy_rev, creds = nil)
...
-  # @param [HostNode] node
-  # @param [String] instance_number
-  # @param [String] deploy_rev
-  def wait_for_service_to_start(node, instance_number, deploy_rev)
-    Timeout.timeout(30) do
-      sleep 0.5 until self.deployed_service_container_exists?(instance_number, deploy_rev)
+  # @param [GridServiceInstance] service_instance
+  def wait_for_service_to_start(service_instance)
+    Timeout.timeout(300) do
+      until service_instance.reload.state == 'running' do
```
SpComb: This needs to wait for the `service_instance.reload.deploy_rev >= deploy_rev`? The instance might already be running when we deploy it.

server/app/services/rpc/node_service_pod_handler.rb
jakolehm: Cases where migration fails -> no scheduled instances -> agents wipes every service instance... I just wanted to be sure that this cannot happen.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -0,0 +1,122 @@
...
+        elsif service_pod.stopped? && (service_container && service_container.running?)
+          info "stopping #{service_pod.name}"
+          ensure_stopped
+        elsif service_pod.terminated?
+          info "terminating #{service_pod.name}"
+          ensure_terminated if service_container
+        elsif service_pod.desired_state_unknown?
+          info "desired state is unknown for #{service_pod.name}, not doing anything"
+        else
+          warn "unknown state #{service_pod.desired_state} for #{service_pod.name}"
```
SpComb: Oops, looks like this now also needs a "service_pod and container are in sync" (`service_pod.running? && service_container.running?` etc.) case...

```
kontena-agent | W, [2017-03-17T14:56:37.575027 #1]  WARN -- Kontena::Workers::ServicePodWorker: unknown state running for redis-1
```

server/app/services/grid_service_instance_deployer.rb
jakolehm: Might be a good idea.

server/app/views/v1/grid_services/_grid_service.json.jbuilder
jakolehm: Yes, it just hits here harder because we are just removing services from db.. if client is polling rest api at the same time this starts to throw exceptions (no transactions).

SpComb: I don't understand what you mean... the `GridServices::Delete` mutation should still only be the only thing that destroys any `GridService` model, and that validates that the service is not linked to anymore? Apart from the bug of the `Stacks::Update` -> `StackDeployWorker` skipping the `GridService` validations... which still needs to be fixed...

server/app/services/grid_service_instance_deployer.rb
jakolehm: State is reseted before this check but it's not very good idea.. I'll improve this a bit.

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -27,68 +30,51 @@ def deploy(node, instance_number, deploy_rev, creds = nil)
...
+    instance = current_service_instance(instance_number)
+    unless instance
+      instance = GridServiceInstance.create!(
+        host_node: node, grid_service: self.grid_service, instance_number: instance_number
+      )
     end
+    instance.set(host_node_id: node.id, deploy_rev: deploy_rev, desired_state: 'running')
+
+    instance
+  rescue => exc
```
SpComb: This will end badly, just let this propagate up to `deploy`?

jakolehm: Yes.

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    def initialize(autostart = true)
+      @workers = {}
+      subscribe('agent:node_info', :on_node_info)
+      subscribe('service_pod:update', :on_update_notify)
+      async.start if autostart
+    end
+
+    def start
+      wait!(interval: 0.1, message: 'waiting for node info') { self.node }
```
jakolehm: We have a problem here.. if actor crashes then node information is lost and we are not getting a new notification. Actor is basically in a restart loop after that. /cc @SpComb

server/app/mutations/grid_services/delete.rb
jakolehm: Any ideas how to fix?

SpComb: Not as part of this PR. Needs an issue to track potential LB registration race conditions / inconsistencies.

agent/lib/kontena/workers/service_pod_manager.rb
SpComb: Re-reading the logic that I commented on earlier, I see the difference now... what about the brute-force:

```ruby
def node
  wait! { Actor[:node_info_worker].node }
end
```

SpComb: BTW this same issue also affects pretty much every other Actor in the agent... it's just better to fix this here than add in `rescue`'s everywhere to make the Actor not crash.

SpComb: > `wait! { ... }`

Scratch that specific suggestion, the Agent's `WaitHelper#wait!` is broken in that it doesn't return the value like it's IMO supposed to.

PR #1874, opened 2017-02-21
Display service logs correctly when giving instance option

nevalla:
This PR will change the logic how service logs are queried when giving instance option.
Previously logs were matched based on container name, but there are variation on container names that made it problematic.

This PR will add `instance_number` field to ContainerLog class and use it to query logs for instance.


server/app/routes/v1/services/service_container_logs.rb
```diff
@@ -5,7 +5,7 @@
     r.is do
       scope = @grid_service.container_logs
 
-      scope = scope.where(name: r['container']) unless r['container'].nil?
```
SpComb: Our 1.0 API probably means that we must keep the existing `?container=` param. Add the `?instance=` as a new parameter?

nevalla: Just thinking what is the correct behaviour when both instance and container are present?

SpComb: Dunno, just include both `where`'s in the scope... I can't think of any usecase for that. The `?container` would just be for backwards-compatibility.

nevalla: Added `instance_number` field to ContainerLog model. That field is used when querying logs for instance.

SpComb: Hmm... won't work for existing logs, but why not, I guess.

SpComb:
The e2e specs are not reliable. They must wait for the services to actually start up before scanning the logs:

```
Failures:

  1) service logs displays service logs
     Failure/Error: expect(k.out.scan('The server is now ready to accept connections on port 6379').size).to eq(2)
     
       expected: 2
            got: 1
     
       (compared using ==)
     # ./spec/features/service/logs_spec.rb:16:in `block (2 levels) in <top (required)>'

Finished in 25.45 seconds (files took 0.21632 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/features/service/logs_spec.rb:13 # service logs displays service logs
```

I suppose the quick fix is a bit of `sleep 1`... after that it falls under the general topic of #1904

SpComb:
The `server/db/migrations/18_*` needs to be renumbered after #1323

nevalla:
Merged Master and resolved conflicts. Also bumped migration number.

SpComb:
Hmm... still test spec failing on the `displays service logs`.... it's expecting to see the final "The server is now ready to accept connections" line, so could this be consistently failing on the server's 5-line log buffering?

Maybe scan for `/Running in standalone mode/` instead.... it's earlier in the startup logging, and guaranteed to get flushed out of the server logs buffer.

SpComb:
Indeed, using `kontena grid logs --tail` during the test, it looks like the `test-1-2` "The server is now ready to accept connections" is always within 5 lines of the `DELETE /v1/services/e2e/null/test-1`line logged by the api container, so that line may indeed end up getting buffered, and only show up once the spec's `kontena service rm` flushes the server log buffer...

test/spec/features/service/logs_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  after(:each) do
+    run("kontena service rm --force test-1")
+  end
+
+  it 'displays service logs' do
+    k = kommando("kontena service logs test-1")
+    expect(k.run).to be_truthy
+    expect(k.out.scan('The server is now ready to accept connections on port 6379').size).to eq(2)
```
SpComb: Scanning for the last line of logs output from the server startup is broken, because it will get stuck in the server logs buffer until some other action flushes it out:

```
test-1-2 | 1:M 07 Mar 09:00:13.961 * The server is now ready to accept connections on port 6379
weave | INFO: 2017/03/07 09:00:14.043612 [nameserver be:69:a5:ac:cb:73] adding entry for 9ac49592502d1503762f936957382438d372ca84d867e27bf5367279662b9f51: test-1-2.e2e.kontena.local. -> 10.81.128.89
test_api_1 | 172.19.0.1 - - [07/Mar/2017:09:00:18 +0000] "GET /v1/services/e2e/null/test-1/container_logs?limit=100 HTTP/1.1" 200 17818 0.5099
test_api_1 | 172.19.0.1 - - [07/Mar/2017:09:00:19 +0000] "DELETE /v1/services/e2e/null/test-1 HTTP/1.1" 200 2 0.0879
```

In this case, the output line is within 5 lines of the server `DELETE ...`, so no amount of sleep is really going to help.

SpComb:
@nevalla I suggest this kind of change to the tests:

```patch
diff --git a/test/spec/features/service/logs_spec.rb b/test/spec/features/service/logs_spec.rb
index d88e347..afa0410 100644
--- a/test/spec/features/service/logs_spec.rb
+++ b/test/spec/features/service/logs_spec.rb
@@ -11,17 +11,17 @@ describe 'service logs' do
     run("kontena service rm --force test-1")
   end
 
-  it 'displays service logs' do
+  it 'displays all service instance logs' do
     k = kommando("kontena service logs test-1")
     expect(k.run).to be_truthy
-    expect(k.out.scan('The server is now ready to accept connections on port 6379').size).to eq(2)
+    expect(k.out.scan(/PID: 1/).size).to eq(2)
   end
 
   context 'when passing instance number' do
     it 'diplays logs only from related service instance' do
       k = kommando("kontena service logs -i 1 test-1")
       expect(k.run).to be_truthy
-      expect(k.out.scan('The server is now ready to accept connections on port 6379').size).to eq(1)
+      expect(k.out.scan(/PID: 1/).size).to eq(1)
     end
   end
 end
```

That line is earlier in the output, so it is guaranteed to get flushed through the server logs buffer.

PR #1884, opened 2017-02-22
CLI: Fix stacks YAML reader handling of undefined variables 

### LINKED ISSUE #1846: Referencing a non-existing stack variable error is hard to understand
```
 [error] undefined method `match' for nil:NilClass
Did you mean?  catch
```

SpComb:
Fixes #1846, #1890

The current CLI stacks YAML reader uses `Liquid::Template.render(..., strict_variables: true)`. This means that that any references to undefined variables will omit the tag from the output and push an error message into `template.errors`. The CLI ignores those errors, which cascades into other difficult to debug issues when parts are missing from the resulting YAML:

* #1846
* #1890

A Liquid bug also means that optional `required: false` stack variables also behave as undefined variables: https://github.com/Shopify/liquid/issues/749

This PR fixes the CLI stack YAML reader to:

* Workaround Liquid bugs to treat optional  `required: false` stack variables as falsey values for Liquid conditional blocks

* Fail `kontena stack ...` fast with a  `Liquid error: undefined variable ... ` message if the YAML contains invalid variable references.

    These errors would be nicer if they also contained the source YAML `file:line`, but that is missing from the `Liquid::UndefinedVariable` error...

## Examples

```yaml
stack: test/variables
variables:
  asdf:
    type: string
    default: asdf
services:
  test:
    image: redis
    affinity:
      - {{ asdflol }}
```

#### `kontena stack install`

```
 [error] Liquid error: undefined variable asdflol
```

#### `DEBUG=stack` without `render!` errors
```yaml
stack: test/variables
variables:
  asdf:
    type: string
    default: asdf
services:
  test:
    image: redis
    affinity:
      - 
```

which results in random nil errors.

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -147,10 +147,11 @@ module Kontena::Cli::Stacks
         raise RuntimeError, "Variable validation failed: #{variables.errors.inspect}" unless variables.valid?
       end
 
+      # @raise [Liquid::Error]
       def interpolate_liquid(content, vars)
         Liquid::Template.error_mode = :strict
         template = Liquid::Template.parse(content)
-        template.render(vars, strict_variables: true, strict_filters: true)
+        template.render!(vars, strict_variables: true, strict_filters: true)
       end
 
       def stack_name
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -7,6 +7,12 @@ module Kontena::Cli::Stacks
       module Setters; end
     end
 
+    class LiquidNull
+      def to_liquid
+        nil
+      end
+    end
+
     class Reader
       include Kontena::Util
       include Kontena::Cli::Common
@@ -151,6 +157,14 @@ module Kontena::Cli::Stacks
       def interpolate_liquid(content, vars)
         Liquid::Template.error_mode = :strict
         template = Liquid::Template.parse(content)
+
+        # Liquid skips any values given as nil
+        # Use an explicit value that gets translated to nil for optional variables
+        vars = vars.clone
+        vars.each do |key, value|
+          vars[key] = LiquidNull.new if value.nil?
+        end
+
         template.render!(vars, strict_variables: true, strict_filters: true)
       end
 
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -108,7 +108,7 @@ module Kontena::Cli::Stacks
         )
         if defaults
           defaults.each do |key, val|
-            var = variables.option(key)
+            var = @variables.option(key)
             var.default = val if var
           end
         end
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -163,10 +163,7 @@ module Kontena::Cli::Stacks
         template = Liquid::Template.parse(content)
 
         # Wrap nil values in LiquidNull to not have Liquid consider them as undefined
-        vars = vars.clone
-        vars.each do |key, value|
-          vars[key] = LiquidNull.new if value.nil?
-        end
+        vars = Hash[vars.map {|key, value| [key, value.nil? ? LiquidNull.new : value]}]
 
         template.render!(vars, strict_variables: true, strict_filters: true)
       end
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -7,6 +7,10 @@ module Kontena::Cli::Stacks
       module Setters; end
     end
 
+    # Workaround for nil-valued variables in Liquid templates:
+    #   https://github.com/Shopify/liquid/issues/749
+    # This is something that we can pass in to `Liquid::Template.render` that gets evaluated as nil.
+    # If we pass in a nil value directly, then Liquid ignores it and considers the variable to be undefined.
     class LiquidNull
       def to_liquid
         nil
@@ -158,8 +162,7 @@ module Kontena::Cli::Stacks
         Liquid::Template.error_mode = :strict
         template = Liquid::Template.parse(content)
 
-        # Liquid skips any values given as nil
-        # Use an explicit value that gets translated to nil for optional variables
+        # Wrap nil values in LiquidNull to not have Liquid consider them as undefined
         vars = vars.clone
         vars.each do |key, value|
           vars[key] = LiquidNull.new if value.nil?
--- a/cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
+++ b/cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
@@ -474,7 +474,7 @@ describe Kontena::Cli::Stacks::YAML::Reader do
     end
 
     before do
-      ENV.delete('asdf')
+      allow(ENV).to receive(:[]).with('asdf').and_return(nil)
     end
 
     it "omits the env" do
@@ -491,7 +491,7 @@ describe Kontena::Cli::Stacks::YAML::Reader do
     end
 
     before do
-      ENV['asdf'] = 'test'
+      allow(ENV).to receive(:[]).with('asdf').and_return('test')
     end
 
     it "defines the env" do
--- a/cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
+++ b/cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
@@ -5,14 +5,6 @@ require 'liquid'
 describe Kontena::Cli::Stacks::YAML::Reader do
   include FixturesHelpers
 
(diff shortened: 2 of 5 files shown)


kke:
There's `kontena stack validate` that can dump the interpolated yaml to stdout.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -147,10 +151,11 @@ def process_variables
         raise RuntimeError, "Variable validation failed: #{variables.errors.inspect}" unless variables.valid?
       end
 
+      # @raise [Liquid::Error]
       def interpolate_liquid(content, vars)
         Liquid::Template.error_mode = :strict
         template = Liquid::Template.parse(content)
-        template.render(vars, strict_variables: true, strict_filters: true)
+        template.render!(vars, strict_variables: true, strict_filters: true)
```
kke: I think this exclamation sign is pretty much what this PR should contain :)

SpComb:
> There's `kontena stack validate` that can dump the interpolated yaml to stdout.
Right.

Pulled the debug yaml thing, it was horribly broken by not actually returning the data anyways.

cli/lib/kontena/cli/stacks/yaml/reader.rb
SpComb: Now it is :)

SpComb:
This fails with `required: false` variables - they are left undefined, and then any `{% if asdflol %}` explodes with a `Liquid::Error`...

SpComb:
It looks like liquid does not like nil values. We call `Liquid::Template.render!` with `vars = {"image"=>"influxdb:1.2-alpine", "affinity"=>"label!=no-influxdb", "lb_service"=>nil, "lb_domain"=>nil, "lb_https_redirect"=>nil, "STACK"=>"influxdb", "GRID"=>"development"}`, but the `Liquid::Context#find_variable` -> `lookup_and_evaluate` loop over the `@environments` will skip any values that evaluate to `nil`:

```
Liquid::Context#find_variable: key=lb_service
Liquid::Context#lookup_and_evaluate: key=lb_service in obj={"image"=>"influxdb:1.2-alpine", "affinity"=>"label!=no-influxdb", "lb_service"=>nil, "lb_domain"=>nil, "lb_https_redirect"=>nil, "STACK"=>"influxdb", "GRID"=>"development"}
Liquid::Context#lookup_and_evaluate: key=lb_service in obj={}
```

This explodes on the final empty `Liquid::Template#assigns` hash in the context environment...

kke:
We could use `strict_variables: false`

SpComb:
It looks like Liquid will call `#to_liquid` on any variable values. Hack in some `LiquidNull` for those optional variables.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -7,6 +7,12 @@ module Resolvers; end
       module Setters; end
     end
 
+    class LiquidNull
```
kke: Simple (using extend):
```
module LiquidNull
  def to_liquid
    nil
  end
end

... 

value.extend LiquidNull
vars[key] = value
```

Advanced (using refinements):
```
module Kontena
  module LiquidNull
    refine NilClass do
       def to_liquid
         nil
       end
     end
  end
end

class Reader
  using Kontena::LiquidNull  # all nulls get to_liquid in this class
   ...
end
```

SpComb:
> `value.extend LiquidNull`
> `using Kontena::LiquidNull # all nulls get to_liquid in this class`

Can't just extend the `nil` values, because the lookup results would still get rejected by the `Liquid::Context#find_variable` loop's `variable.nil?`.

kke:
https://github.com/Shopify/liquid/issues/749 looks like this has been reported but not fixed.

SpComb:
Added specs, please take another look.

cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
```diff
@@ -5,14 +5,6 @@
 describe Kontena::Cli::Stacks::YAML::Reader do
   include FixturesHelpers
 
-  def absolute_yaml_path(file = 'kontena_v3.yml')
-    "#{Dir.pwd}/#{file}"
-  end
-
-  let(:subject) do
```
SpComb: @kke do these spec cleanups look reasonable?

cli/spec/kontena/cli/stacks/yaml/reader_spec.rb
```diff
@@ -457,4 +457,48 @@
...
+    end
+  end
+
+  context "for an optional variable that is defined" do
+    subject do
+      described_class.new(fixture_path('stack-with-liquid-optional.yml'))
+    end
+
+    before do
+      ENV['asdf'] = 'test'
```
kke: `allow(ENV).to receive(:[]).with('asdf').and_return('test')`

SpComb: This is here because the YAML `Reader` will also set `ENV['asdf'] = 'lol'` in the (earlier) `stack-with-liquid-undefined.yaml` case... Leaking those `ENV` between spec examples will cause weird behavior if any two specs use the same variable names :(

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -147,10 +157,18 @@ def process_variables
...
       end
 
+      # @raise [Liquid::Error]
       def interpolate_liquid(content, vars)
         Liquid::Template.error_mode = :strict
         template = Liquid::Template.parse(content)
-        template.render(vars, strict_variables: true, strict_filters: true)
+
+        # Wrap nil values in LiquidNull to not have Liquid consider them as undefined
+        vars = vars.clone
```
kke: I'm not sure if it matters, but Hash#clone makes a shallow copy, meaning any hash/array values inside the top level hash are still references.

SpComb: I assume the stack variables are always going to be a flat hash... I suppose Liquid does have some support for indexing into hash values, but I'm not sure how it then deals with nil values in there..

SpComb: Point being to not mutate the given `vars` hash argument.

kke: There can be arrays as values. Maybe LiquidNull should be a subclass of string or something.

Hmm, I wonder how these look in the reader.execute[:variables] that gets sent to master. Maybe there needs to be some "deliquidnullify" after the interpolation?

SpComb: > There can be arrays as values. Maybe LiquidNull should be a subclass of string or something.
Are those arrays going to contain nil values? How does liquid behave when those nil values get used for something?

> Hmm, I wonder how these look in the reader.execute[:variables] that gets sent to master. Maybe there needs to be some "deliquidnullify" after the interpolation?

They shouldn't end up there, the `#execute[:variables]` are from a separate `variables.to_h(...)` call without this mapping of `LiquidNull`s.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -163,10 +163,7 @@ def interpolate_liquid(content, vars)
         template = Liquid::Template.parse(content)
 
         # Wrap nil values in LiquidNull to not have Liquid consider them as undefined
-        vars = vars.clone
-        vars.each do |key, value|
-          vars[key] = LiquidNull.new if value.nil?
-        end
+        vars = Hash[vars.map {|key, value| [key, value.nil? ? LiquidNull.new : value]}]
```
kke: I guess this is still a reference unless you `value.clone`. 

That may make it possible to change the values of options in liquid so that master gets different values that were set to variables in the variable parsing.

SpComb: I would be surprised if Liquid let you mutate template values.

kke: Right, it probably doesn't.

SpComb:
> Are those arrays going to contain nil values? How does liquid behave when those nil values get used for something?

Liquid seems to deal with nils in array-valued variables:

```yaml
stack: test/variables
variables:
  asdf:
    type: array
    value:
      - "foo"
      - "bar"
      - null
services:
  test:
    image: redis
    environment:
    # {% for item in asdf %}
      - ASDF={{ item }}
    # {% endfor %}
```

```
Liquid::Template.render! {"asdf"=>["foo", "bar", nil], "STACK"=>"variables", "GRID"=>"development"}
```

Results in

```yaml
variables:
  asdf:
    label: asdf
    type: array
    from:
      env: asdf
    to:
      env: asdf
    errors: {}
    value:
    - foo
    - bar
    - 
services:
  test:
    image: redis
    environment:
    - ASDF=foo
    - ASDF=bar
    - ASDF=
```

kke:
```yaml
    value:
    - foo
    - bar
    - 
```

And that generates `["foo", "bar", nil]` when yaml is parsed.

SpComb:
> And that generates ["foo", "bar", nil] when yaml is parsed.

Yes, it seems to behave correctly.

SpComb:
@kke do I assume that this is approved now?

PR #1887, opened 2017-02-22
Fix server stack remove service reverse-dependency sorting on links

### LINKED ISSUE #1836: kontena stack rm --force hangs
with: 1.1.1 & 1.1.0
```
kontena stack install matti/lb-ingress
kontena stack install matti/sentry
# .. wait until booted up
kontena stack rm --force sentry
# always hangs, rerun works almost immediately
```

Once I also got this on the second try
```
 [fail] Removing stack sentry
 [error] invalid_token
```

SpComb:
Fixes #1836

#1358 used `Stacks::SortHelper.sort_services` to sort stack services into reverse-dependency order for removal, to avoid `GridServices::Delete` errors when removing a service that still has links to it. The implementation of `Stacks::SortHelper.sort_services` was broken, and used the `GridServiceLink.grid_service` instead of the `GridServiceLink.linked_grid_service`, causing it to consider each service as only linking to itself:

```
a=foo links=[{:name=>"foo"}]
b=bar links=[{:name=>"bar"}]
a=redis links=[{:name=>"redis"}]
b=foo links=[{:name=>"foo"}]
.a=redis links=[{:name=>"redis"}]
b=bar links=[{:name=>"bar"}]
a=foo links=[{:name=>"foo"}]
b=redis links=[{:name=>"redis"}]
```

This means that it effectively only sorted on the number of links. This worked for the simple case of services in a stack linking to one service, where that the linked-to service would thus have fewer links (zero). However, if that service then also linked to e.g. an lb service outside of the stack, the services could get sorted into the wrong order.

### DIFF
--- a/server/app/mutations/stacks/sort_helper.rb
+++ b/server/app/mutations/stacks/sort_helper.rb
@@ -21,7 +21,7 @@ module Stacks
     # @return [Array<Hash>]
     def __links_for_service(service)
       if service.is_a?(GridService)
-        service.grid_service_links.map{ |l| { name: l.grid_service.name } }
+        service.grid_service_links.map{ |l| { name: l.linked_grid_service.name } }
       else
         service[:links] || []
       end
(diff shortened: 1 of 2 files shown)


SpComb:
It looks like the `StackDeployWorker` also uses the same `Stacks::SortHelper.sort_services`, which means that the same scenarios would also apply to stack deployments, where linked services could get deployed in the wrong order?

PR #1889, opened 2017-02-22
Modifications to simplify kontena-cli homebrew formula

kke:
Without this PR the homebrew formula for kontena-cli needs to be too complex for them to accept it.

With this PR the formula is practically : `gem build kontena-cli.gemspec && gem install kontena-cli-*.gem && bin.install "out/kontena"`.

- Simplify bin/kontena exec wrapper
- Make bin/kontena detect if it's being loaded from a homebrew installation
- Add `+homebrew` tag to `kontena --version` build tags when homebrew detected
- Add `.pre.head_abc012` to `Kontena::Cli::VERSION` when a homebrew HEAD build detected
- Make bin/kontena load the completer instead of MainCommand when running `kontena complete`, so we only have one exec script and one less shebang to worry about
- Modified the gemspec to avoid warnings
- The completion init script doesn't need the wizardry to figure out a path for the completer script, as `kontena complete` should work


cli/lib/kontena/scripts/completer
```diff
@@ -3,238 +3,7 @@
 
 # resolve bin path, ignoring symlinks
 require 'pathname'
-require 'yaml'
 bin_file = Pathname.new(__FILE__).realpath
+$:.unshift File.expand_path('../../..', bin_file)
```
SpComb: Should this be `.../lib`?

kke: Then we would be going to `lib/../lib`which is kind of redundant.

SpComb: Looking at the context, this is replacing `$:.unshift File.expand_path('../../lib', bin_file)` -> `$:.unshift File.expand_path('../../..', bin_file)`, that supposed to be different?

SpComb:
Trying to think about this upstream vs homebrew distro packaging split, I'm not sure it's a good idea to embed homebrew-specific stuff into the `kontena/cli` upstream either:

* `cli/bin/kontena` reference to `../../libexec`
* `cli/lib/kontena/cli/version.rb` reference to `ENV["_KONTENA_HOMEBREW"]`
* `cli/lib/kontena/main_command.rb` reference to `ENV["_KONTENA_HOMEBREW"]`

Instead, I think we should add some generic/non-homebrew-specific mechanisms that would allow the use of a trivial shell wrapper for distro packaging, like:

```sh
export GEM_HOME=...
export KONTENA_BUILD_TAGS=homebrew

exec $INSTALL_PATH/bin/kontena "$@"
```

I'm not ruby packaging expert, but that could be something along the lines of...

### Fix the `cli/bin/kontena` `$LOAD_PATH` manipulation

Ideally, we wouldn't even need this at all. The minimal `cli/bin/kontena` would just be

```ruby
#!/usr/bin/env ruby
# encoding: UTF-8

STDOUT.sync = true

require 'kontena_cli'

Kontena::PluginManager.instance.init
Kontena::MainCommand.run
```

This works with `bundle exec bin/kontena` during development, and then uses the `gem install --bindir` generated wrappers when deployed?

### Add `Kontena::Cli::VERSION` support for external build tags

Use something like `ENV["KONTENA_BUILD_TAGS]` to pick up the `+homebrew` stuff?

cli/lib/kontena/scripts/completer
kke: I think it was broken previously but still worked because the correct libpath came from the fact that kontena-cli gem was installed to system gems.

```
irb(main):001:0> File.expand_path('../../lib', '/Users/kimmo/Projects/kontena/cli/lib/kontena/scripts/completer')
=> "/Users/kimmo/Projects/kontena/cli/lib/kontena/lib"
```

kke:
Yep, that's how it is now.

```
require 'kontena_cli'
```


This in a perfect world would be:
```ruby
require 'kontena/cli'
```

And instead of `lib/kontena_cli.rb` there would be `lib/kontena/cli.rb`.

cli/lib/kontena/scripts/completer
SpComb: Ah yes, I see now. The fixed version expands to `cli/lib/kontena/scripts/completer/../../..` = `cli/lib`.

kke:
### regular gem install - works like before

End result is same as when installing from rubygems: `gem install kontena-cli`
```
$ gem uninstall -a -x --force kontena-cli
$ gem build kontena-cli.gemspec
$ gem install kontena-cli-*.gem
...
$ kontena --version
kontena-cli 1.1.1 [ruby2.4.0+x86_64-darwin16]
$ source `kontena whoami --bash-completion-path`
$ kontena m<tab> --> master
```

### brew install - works like before

```
$ brew tap kke/kcli
$ brew install kontena-cli # the formula points to this branch now
$ kontena --version
kontena-cli 1.1.1 [ruby2.4.0+x86_64-darwin16+homebrew]
$ kontena m<tab> --> master
```

### from local dir - works like before
```
$ bin/kontena --version
kontena-cli 1.1.1 [ruby2.2.3+x86_64-darwin16]
$ KONTENA_EXTRA_BUILDTAGS=foo,head bin/kontena --version
kontena-cli 1.1.1-head [ruby2.2.3+x86_64-darwin16+foo+head]
```

cli/lib/kontena/scripts/completer
```diff
@@ -3,238 +3,7 @@
 
 # resolve bin path, ignoring symlinks
 require 'pathname'
-require 'yaml'
```
nevalla: Is `yaml` available on `kontena/scripts/completer.rb`? Or is this totally redundant require?

kke: It does come through some of the requires, but added it to completer.rb to be sure.

cli/kontena-cli.gemspec
```diff
@@ -22,15 +21,16 @@ Gem::Specification.new do |spec|
...
-  spec.add_runtime_dependency "clamp", "~> 1.1.0"
-  spec.add_runtime_dependency "ruby_dig", "~> 0.0.2"
-  spec.add_runtime_dependency "launchy", "~> 2.4.3"
-  spec.add_runtime_dependency "hash_validator", "~> 0.7.0"
-  spec.add_runtime_dependency "retriable", "~> 2.1.0"
+  spec.add_runtime_dependency "clamp", ">= 1.1", "< 1.2"
+  spec.add_runtime_dependency "ruby_dig", "< 0.1"
+  spec.add_runtime_dependency "launchy", ">= 2.4.3", "< 2.5"
+  spec.add_runtime_dependency "hash_validator", ">= 0.7", "< 0.8"
+  spec.add_runtime_dependency "retriable", ">= 2.1", "< 2.2"
```
SpComb: Some particular reason for breaking out the `~>` syntax?

cli/kontena-cli.gemspec
```diff
@@ -22,15 +21,16 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler", "~> 1.7"
   spec.add_development_dependency "rake", "~> 10.0"
-  spec.add_runtime_dependency "excon", "~> 0.49.0"
+  spec.add_development_dependency "rake-compiler", "~> 1.0"
```
SpComb: Hmm... why does the CLI need `rake compile`?

cli/kontena-cli.gemspec
kke:
```
WARNING:  description and summary are identical
WARNING:  pessimistic dependency on clamp (~> 1.1.0) may be overly strict
  if clamp is semantically versioned, use:
    add_runtime_dependency 'clamp', '~> 1.1', '>= 1.1.0'
WARNING:  pessimistic dependency on launchy (~> 2.4.3) may be overly strict
  if launchy is semantically versioned, use:
    add_runtime_dependency 'launchy', '~> 2.4', '>= 2.4.3'
WARNING:  pessimistic dependency on retriable (~> 2.1.0) may be overly strict
  if retriable is semantically versioned, use:
    add_runtime_dependency 'retriable', '~> 2.1', '>= 2.1.0'
WARNING:  pessimistic dependency on liquid (~> 4.0.0) may be overly strict
  if liquid is semantically versioned, use:
    add_runtime_dependency 'liquid', '~> 4.0', '>= 4.0.0'
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
```

cli/kontena-cli.gemspec
kke: It doesn't.

cli/kontena-cli.gemspec
kke: ~> 2.3 means "equal to 2.3 or greater than 2.3, but less than 3.0", while ~> 2.3.0 would mean "equal to 2.3.0 or greater than 2.3.0, but less than 2.4.0

So while these `~> 2.3.0` are working, and getting 2.3.* like intended, gem build warns about them.

SpComb: That feels like a bad warning, because the suggested variant is totally not equivalent. I guess this is the best way to workaround it, unless there's something to supress it... what about `gem build --force`, is that a bad idea?

kke: Well `gem build --silent` mutes the warnings.

cli/kontena-cli.gemspec
```diff
@@ -23,9 +22,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.7"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_runtime_dependency "excon", "~> 0.49.0"
-  spec.add_runtime_dependency "tty-prompt", "~> 0.10"
-  spec.add_runtime_dependency "clamp", "~> 1.1.0"
-  spec.add_runtime_dependency "ruby_dig", "~> 0.0.2"
+  spec.add_runtime_dependency "tty-prompt", "0.11.0"
```
SpComb: Looks like #1901

cli/kontena-cli.gemspec
```diff
@@ -23,9 +22,9 @@ Gem::Specification.new do |spec|
...
+  spec.add_runtime_dependency "clamp", "~> 1.1"
```
SpComb: Keep `~> 1.1.0`?

PR #1896, opened 2017-02-24
Avoid leaking auth codes to cloud in referer header 

kke:
As realized by @SpComb - the CLI authentication mechanism can leak the auth code in referrer header when doing the final redirect to https://cloud.kontena.io/terminal-success during the oauth dance.

This PR adds `Referrer-Policy: no-referrer` header to the request which should eliminate the referrer on most browsers.

### DIFF
--- a/cli/lib/kontena/cli/localhost_web_server.rb
+++ b/cli/lib/kontena/cli/localhost_web_server.rb
@@ -17,7 +17,7 @@ module Kontena
     attr_accessor :server, :success_response, :error_response, :port
 
     DEFAULT_ERROR_MESSAGE   = "Bad request"
-    SUCCESS_URL             = "http://cloud.kontena.io/terminal-success"
+    SUCCESS_URL             = "https://cloud.kontena.io/terminal-success"
 
     # Get new server instance
     #
@@ -73,6 +73,7 @@ module Kontena
           socket.print [
             'HTTP/1.1 301 Found',
             "Location: #{SUCCESS_URL}",
+            "Referrer-Policy: no-referrer"
             "Connection: close",
             ''
           ].join("\r\n")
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/localhost_web_server.rb
```diff
@@ -73,6 +73,7 @@ def serve_one
           socket.print [
             'HTTP/1.1 301 Found',
             "Location: #{SUCCESS_URL}",
+            "Referrer-Policy: no-referrer"
```
SpComb:
```
SyntaxError: /home/kontena/kontena/kontena/cli/lib/kontena/cli/localhost_web_server.rb:77: syntax error, unexpected tSTRING_BEG, expecting ']'
            "Connection: close",
             ^
```

kke:
List of browsers that support this: http://caniuse.com/#feat=referrer-policy

Alternative solution: do a double redirect, problem is that this is a bit cumbersome with the current localhost_web_server implementation.

SpComb:
I'm not seeing any `Referer: ...` headers on the `https://cloud.kontena.io/terminal-success` requests, either with `http://`, `https://` or the `Referrer-Policy: no-referrer` with either Firefox or Chromium, so I'm not entirely sure the browsers are behaving.

SpComb:
:+1: on the HTTP 302 change though

kke:
Yeah this is actually mostly redundant.

Browsers tend to drop referrer headers when redirecting between https (cloud) and http (localhost webserver). Any browser that doesn't, probably will not respect the referrer-policy header either.

kke:
The most safe method for the paranoid kind would be to do a double redirect.

1) cloud -> localhost/cb?code=abcd
2) localhost/cb -> localhost/redirect
3) localhost -> cloud.kontena.io/terminal-success

kke:
I guess with HTTP 301 the browser would remember it has been permantently moved if a duplicate auth code would be generated and open the success page directly, making the code to never be received by the cli.

PR #1898, opened 2017-02-24
Bump version to 1.2.0.dev

SpComb:
This begins the `1.2` release cycle in preparation for merging 1.2 PRs.

Any future `1.1.x` releases will come from a separate `v1.1` branch.


SpComb:
It's not strictly a semver, it's a rubygems version:

`Gem::Version.new('1.2.dev') < Gem::Version.new('1.2.0') => true`

But yes, let's stick to the `1.x.y.ZZZ` structure. No point trying to do something fancy with "branch" versions, git master is going to hit `1.2.x` point versions anyways.

PR #1906, opened 2017-02-28
Send events to Kontena Cloud in real time

nevalla:
This PR tracks changes in `Grid`, `GridService`, `HostNode`, `Container` and `User` models and sends fresh data to Kontena Cloud when the model changes.

json-serializer gem is used to serialize models to json (later on all the views can be replaced with the related serializers).

Currently `Container` models are not serialized but instead when container changes, update event for the related `GridService` will be sent. Not sure should we send also container data in real time to Cloud.

Also MongoId has a [bug(?)](https://jira.mongodb.org/browse/MONGOID-3552) on relation callbacks: When the relation object is removed, model still has it on callback method execution. Workaround for this is not to execute callbacks on remove event but call explicitly update method when relation is removed from the model.

WebSocket sends events as [MessagePack RPC notification](https://github.com/msgpack-rpc/msgpack-rpc/blob/master/spec.md#notification-message) message: `[2, method, [params]]`. Method is combination of resource#action, for example `GridService#create`. In params the first parameter is grid_id, the second is the array of users' external ids that can receive the message and the third is the object itself as json string.


server/Gemfile.lock
```diff
@@ -150,6 +151,7 @@ DEPENDENCIES
   faye-websocket (~> 0.10.0)
   httpclient (~> 2.3)
   json-jwt (= 1.5.2)
```
kke: (I have a hunch that this json-jwt thing is a remnant from some auth experiments)

SpComb:
Note that #1873 also adds server model serializers.

server/app/services/cloud/websocket_client.rb
```diff
@@ -26,13 +26,22 @@ def self.api_uri
 
     include Logging
     KEEPALIVE_TIME = 30
+
+    if defined? Faye::Websocket::Client.CLOSE_TIMEOUT
```
SpComb: We should update to faye-websocket 0.10.7, so that we don't need this monkey-patching.

server/app/services/cloud/websocket_client.rb
```diff
@@ -195,5 +263,22 @@ def verify_connection
     rescue => exc
       error exc.message
     end
+
+    # Abort the connection, closing the websocket, with a timeout
+    def close
```
SpComb: This copy-pasting of the websocket-rpc related code between the server and agent in the same git repo is really starting to bug me.

Couldn't we somehow refactor these into shared code between the two components?

SpComb:
This event stream concept is also relevant for both any grid metadata API implementation designed to replace the current Docker API + etcd approach, as well as the agent refactoring stuff currently going on.

The `EventStream` -> `MongoPubsub` implementation here would also work for that, it would just be a question of subscribing to those mongo pubsubs in the agent `RpcServer` and any possible metadata API implementation.

server/app/services/agent/node_plugger.rb
```diff
@@ -18,6 +18,7 @@ def plugin!
       begin
         prev_seen_at = node.last_seen_at
         self.update_node
+        self.publish_update_event
```
SpComb: Hmm, why does this need an explicit publish, doesn't the `node.set` trigger the `after_update`?

nevalla: "When executing atomic operations via these methods, no callbacks will ever get run, nor will any validations."
https://mongoid.github.io/old/en/mongoid/docs/persistence.html#atomic

server/app/mutations/grids/unassign_user.rb
```diff
@@ -24,7 +24,8 @@ def validate
 
     def execute
       user.grids.delete(grid)
+      user.save # relation callback has bug so we have to explicitly call update here
```
SpComb: Does this need to be `save`, vs an explicit `publish_update_event` like elsewhere in the code?

nevalla: Not actually sure... have to test

server/app/services/cloud/websocket_client.rb
```diff
@@ -124,55 +135,112 @@ def send_message(msg)
...
       error exc.message
     end
 
+    # @param [Hash] msg
+    def send_notification_message(msg)
+      invalidate_users_cache if msg[:type] == 'User' # clear cache if users are modified
+      grid_id = resolve_grid_id(msg)
+      users = resolve_users(grid_id)      
+      params = [grid_id, users, msg[:object]]
+      message = [2, "#{msg[:type]}##{msg[:event]}", params]
```
SpComb: Could these use similar `/grid_service/update` style methods as the Agent RPC does? Ideally they would be something vaugely consistent with the HTTP API URL structure...

server/app/models/event_stream.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.channel
+    CHANNEL
+  end
+
+  def publish_create_event    
+    event = {
+      event: 'create',
+      type: self.class.name,
+      object: find_serializer.to_json
```
SpComb: Hmm, comparing this to the RPC serializers in #1873, perhaps it would be better to just use MessagePack serialization for everything on the RPC wire, i.e. this should just be `.to_hash`, not `.to_json`?

Current format is something like hash in json-string in messagepack-string in messagepack-hash?

server/app/models/grid_service.rb
```diff
@@ -92,7 +94,10 @@ def default_stack?
 
   # @param [String] state
   def set_state(state)
-    self.set(:state => state)
+    if self.state != state
```
SpComb: Is this necessarily safe with multiple master processes doing work at the same time?

server/app/serializers/_kontena_json_serializer.rb
```diff
@@ -0,0 +1,12 @@
+class KontenaJsonSerializer < JsonSerializer
+
+  def decorate
+    return nil unless object
+
+    if object.is_a?(Array)
+      to_arry
```
SpComb: Should this be `to_array`?

nevalla: `to_arry` is right method call: https://github.com/frodsan/json-serializer/blob/master/lib/json_serializer.rb#L43-L51

The purpose of overriding this method is that Mongoid model has `#to_a` method that parent class uses to recognise arrays.

server/app/services/agent/node_plugger.rb
SpComb: So if I understand this correctly, changing the `update_node` method to use `node.update_attributes(...)` would also trigger the `publish_update_event`?

nevalla: yes, but I think here is some good reason to use `#set`.

server/app/services/cloud/websocket_client.rb
nevalla: yes. If we have a consistent way, let's use that.

SpComb: I don't think we really do have any consistent way as such, yet. I'm not sure where the current Agent RPC `/subject/verb` pseudo-URL methods come from exactly.

IMO I would love to see some kind of REST-style push-event stream along the lines of `UPDATE /services/:grid/:stack/:service {'image': ...}` instead of the current MessagePack-RPC thing of `/services/update {'grid': ..., 'stack': ..., 'image': ...}`, but the difference isn't really that dramatic, I suppose.

server/app/services/agent/node_plugger.rb
SpComb: If I'm interpreting the mongoid docs correctly, then `node.update_attributes(...)` should result in exactly the same atomic mongo `collections[...].find(...).update($set => ...)` operation, with the difference of also triggering the mongoid hooks?

It's not really that significant, I'm mostly just curious as to if I'm understanding the Mongoid behavior here correctly.

SpComb:
Needs merge of #1855

server/app/services/cloud/websocket_client.rb
SpComb: See #1757 for the latest agent `WebsocketClient` changes

server/app/serializers/host_node_serializer.rb
```diff
@@ -0,0 +1,83 @@
...
+    object.last_seen_at.try(:iso8601)
+  end
+
+  def engine_root_dir
+    object.docker_root_dir
+  end
+
+  def peer_ips
+    if object.grid
+      object.grid.host_nodes.ne(id: object.id).map{|n|
```
SpComb: This logic is now duplicated here and in the `server/app/views/v1/host_nodes/_host_node.json.jbuilder`... refactor to a `HostNode` model method?

server/app/services/mongo_pubsub.rb
```diff
@@ -100,24 +100,28 @@ def publish(channel, data)
...
   def self.publish_async(channel, data)
-    @supervisor.actors.first.async.publish(channel, data)
+    @supervisor.actors.first.async.publish(channel, data) if self.started?
   end
 
   # @param [String] channel
   # @return [Subscription]
   def self.subscribe(channel, &block)
-    @supervisor.actors.first.subscribe(channel, block)
+    @supervisor.actors.first.subscribe(channel, block) if self.started?
```
SpComb: This seems a little odd... do the initializers/migrations also trigger `MongoPubsub.subscribe`?

nevalla: No it does not, but if some component is calling that before `MongoPubsub.start!` it will explode.

server/app/serializers/host_node_serializer.rb
nevalla: I think eventually we will replace views with serializers, so I did not touch views on purpose. But yes that can be refactored.

server/app/services/mongo_pubsub.rb
SpComb: Now instead of exploding, it just silently does not subscribe, and gets stuck instead?

nevalla:
@SpComb I changed this now so that only EventStream module checks that MongoPubsub has been started before publishing events.

server/app/services/cloud/websocket_client.rb
```diff
@@ -124,55 +138,109 @@ def send_message(msg)
...
+      object = msg[:object]
+      if msg[:type] == "Grid"
+        object['id']
+      else
+        object.dig('grid', 'id')
+      end
+    end
+
+    def resolve_users(grid_id)
+      if(grid_id)
```
SpComb: `if grid_id`

server/app/services/cloud/websocket_client.rb
```diff
@@ -111,6 +124,7 @@ def connect
     def send_message(msg)
       EM.next_tick {
         begin
+          debug "Sending websocket message"
```
SpComb: This debug message isn't very useful.

```
D, [2017-03-14T11:28:16.523215 #18391] DEBUG -- Cloud::WebsocketClient: Sending notification message: [2, "HostNode#update", ["development", [], {"id"=>"XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5", ...}}}]]
D, [2017-03-14T11:28:16.530596 #18391] DEBUG -- Cloud::WebsocketClient: Sending websocket message
```

nevalla: I think it's some kind of leftover. I'll remove it.

nevalla:
Just removed token from serializer classes. Will do separate PR to remove it from responses returned by Rest API.

SpComb:
Needs merge after #1873

server/app/services/cloud/websocket_client.rb
SpComb: Bump

server/app/services/cloud/websocket_client.rb
```diff
@@ -124,55 +138,109 @@ def send_message(msg)
...
 
+    # @param [Hash] msg
+    def send_notification_message(msg)
+      invalidate_users_cache if msg[:type] == 'User' # clear cache if users are modified
+      grid_id = resolve_grid_id(msg)
+      users = resolve_users(grid_id)
+      params = [grid_id, users, msg[:object]]
+      message = [2, "#{msg[:type]}##{msg[:event]}", params]
+      debug "Sending notification message: #{message}"
+      EM.defer {
```
SpComb: Bump

SpComb:
Waiting for `status/merge` once this is smoke-tested to work.

Issue #1905, opened 2017-02-27
HostNode statistics are missing timestamps


nevalla:
`HostNodeStat` model contains `created_at` and `updated_at` fields (via `Mongoid::Timestamps`). However, when saving stats, `HostNodeStat` model is not used and `created_at` and `updated_at` fields are not set:

```
      stat = {
        grid_id: grid.id,
        host_node_id: node.id,
        memory: data['memory'],
        load: data['load'],
        filesystem: data['filesystem'],
        usage: data['usage']
      }
      db_session[:host_node_stats].insert(stat)
```

https://github.com/kontena/kontena/blob/master/server/app/services/agent/message_handler.rb#L141-L163

nevalla:
The same issue with ContainerStat

PR #1908, opened 2017-02-28
HostNode statistics are missing timestamps

daniel-bytes:
Fixes [1905](https://github.com/kontena/kontena/issues/1905)

In MessageHandler#on_container_stat and MessageHandler#on_node_stat, Mongoid timestamp values are now manually set, by either parsing incoming data['time'] values or defaulting to Time.now.utc.

Refactored MessageHandler#on_container_stat buffering to use a configurable buffer size.

Additional unit tests added to verify timestamp fields, buffer size refactor code, and on_container_stat calls (previously on_container_stat did not have any unit tests).


server/app/services/agent/message_handler.rb
```diff
@@ -188,13 +194,17 @@ def on_node_stat(grid, data)
       node = grid.host_nodes.find_by(node_id: data['id'])
       return unless node
       data = fixnums_to_float(data)
+      time = data['time'] ? Time.parse(data['time']) : Time.now.utc
```
jakolehm: we don't actually send `time` from the agent.. should we? /cc @nevalla

SpComb: Agreed, it looks like both the agent host and container stats are missing any `time` field:

* https://github.com/kontena/kontena/blob/fix/hostnode_statistics_missing_timestamp/agent/lib/kontena/workers/node_info_worker.rb#L126
* https://github.com/kontena/kontena/blob/fix/hostnode_statistics_missing_timestamp/agent/lib/kontena/workers/stats_worker.rb#L109

The agent events may be queued up for long periods of time if there's e.g. websocket connection issues.

IMO either also fix the agent side as part of this PR, or file a separate issue for that.

nevalla: I agree, it would be more reliable if agent sends the time. Could be also a separate PR.

daniel-bytes:
Had to fix several merge conflicts as a result of refactoring done in [PR 1855](https://github.com/kontena/kontena/pull/1855)

daniel-bytes:
@jakolehm @nevalla @SpComb  I have a question:  Since these data points are immutable facts, would it make more sense to only store the created_at timestamp in the database? [It looks like mongoid provides this](https://mongoid.github.io/old/en/mongoid/docs/extras.html).  Considering we will probably store a lot of these statistics, it could end up saving a decent amount of duplicated data.

jakolehm:
@daniel-bytes yes, that makes sense.

daniel-bytes:
Updated PR to handle timestamp values being sent from the agent.  Added unit tests on agent side.

agent/spec/lib/kontena/workers/node_info_worker_spec.rb
```diff
@@ -214,4 +214,22 @@
       expect(subject.private_ip).to eq('192.168.2.10')
     end
   end
+
+  describe '#publish_node_stats' do
+    it 'sends stats via rpc' do
+      expect(rpc_client).to receive(:notification).once do |key, msg|
```
SpComb: You should be able to compose rspec matchers to do this, without having to use a block:

```ruby
expect(rpc_client).to receive(:notification).once.with('/nodes/stats', [Hash])
```

Or even better for this spec:

```ruby
expect(rpc_client).to receive(:notification).once.with('/nodes/stats', [hash_including(time: String)])
```

server/spec/services/rpc/container_handler_spec.rb
```diff
@@ -107,23 +107,148 @@
       #puts bm
       total_time = Time.now.to_f - start_time
       expect(grid.container_logs.count).to eq(1_000)
-      expect(total_time < 0.5).to be_truthy
+      expect(total_time).to be < 0.5
```
SpComb: :+1: for much more useful rspec errors with this kind of expectation

server/spec/services/rpc/container_handler_spec.rb
```diff
@@ -107,23 +107,148 @@
...
     end
   end
 
-  describe '#on_container_health' do
-    it 'updates health status' do
+  describe '#save' do
```
SpComb: Did this group of `#save` specs get duplicated during the merge?

server/app/services/rpc/node_handler.rb
```diff
@@ -7,6 +7,11 @@ class NodeHandler
 
     def initialize(grid)
       @grid = grid
+      @db_session = ContainerLog.collection.session.with(
```
SpComb: The `ContainerLog` looks a bit odd here, but I suppose it doesn't really make any difference. Maybe use `HostNode` instead, though?

agent/spec/lib/kontena/workers/node_info_worker_spec.rb
daniel-bytes: Awesome, let me take a look.  I was having issues with hash_including but I probably was doing it wrong.

server/spec/services/rpc/container_handler_spec.rb
daniel-bytes: Yes good call I think there is some duplication going on in this spec file.  The merge got pretty nasty....

server/app/services/rpc/node_handler.rb
daniel-bytes: Yes, that makes sense.  Let me fix this.

SpComb:
Has this been tested to check that the agent timestamps in the stats entries make it all the way to the server MongoDB intact?

daniel-bytes:
E2E tests on my local environment looked good.

PR #1909, opened 2017-02-28
Updating development.md guide to include step to delete master nodes from local cli config file

daniel-bytes:
Small fix to developer setup documentation.  

At the end of document, when describing how to clean up the local dev environment, I added some notes and script explaining how the ```.kontena_client.json``` file may need to be cleaned up as well to prevent issues with previous CLI state being maintained.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -96,4 +96,10 @@ docker ps -aq -f name=weave | xargs docker rm -f
 dontena cloud master rm
 ```
 
+You may also want to remove the cli config file to prevent previous state from causing issues:
+
+```bash
+rm $HOME/.kontena_client.json 
+```
+
 and start from the beginning.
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -96,4 +96,10 @@ docker ps -aq -f name=weave | xargs docker rm -f
 dontena cloud master rm
 ```
 
+You may also want to remove the cli config file to prevent previous state from causing issues:
+
+```bash
+rm $HOME/.kontena_client.json
```
jakolehm: Is `kontena master rm devmaster` not enough? Does it still leave some traces to config file?

SpComb:
Please update the PR title to match the actual change. It's what shows up as the git master commit summary, so it's misleading ATM.

PR #1911, opened 2017-03-01
CLI global --master option for selecting a master from config by name

kke:
Mostly just an idea.

```console
$ kontena --master snowy-summer-34 service ls
```

Of course you can already use the ENV variables `KONTENA_MASTER`, `KONTENA_URL` and `KONTENA_TOKEN`.


jnummelin:
What's the benefit of these compared to envs? The drawback I see is that using these explicit options these are very easily leaked into build logs or similar.

pcornelissen:
I would suggest to also have a parameter that just takes the local label for a configured master. So you don't have to pass the secret master token around and just have it stored in the default kontena config file in the users home dir.

jakolehm:
Imho we only need envs and `--master <name>`.

kke:
Yep maybe the url+token envs are enough for scripting.

SpComb:
So this PR effectively just adds new `--master` flags that do exactly the same thing as the existing `KONTENA_MASTER=` etc envs?

cli/lib/kontena/cli/common.rb
```diff
@@ -177,7 +177,7 @@ def client(token = nil, api_url = nil)
 
         @client ||= Kontena::Client.new(
           api_url || require_current_master.url,
-          token || require_current_master.token
+          token || ENV['KONTENA_TOKEN'] || require_current_master.token
```
SpComb: Note to self: Currently the `KONTENA_TOKEN` seems to be ignored unless also using `KONTENA_URL`.

SpComb: AKA with this change, just setting `KONTENA_TOKEN=asdf` will actually cause the CLI to fail with HTTP 403. Probably not a bug, seems intended.

SpComb:
Looks like using `KONTENA_URL=...` and `KONTENA_TOKEN=...` also requires `KONTENA_GRID=....`, but trying to use ` kontena --master-url=... node ls --grid ...` doesn't work:

`DEBUG=true bundle exec bin/kontena --master-url=http://localhost:9291 node ls --grid development`
```
DEBUG Kontena CLI 1.4.0.dev (ruby-2.3.1+x86_64-linux-gnu)
DEBUG Loading plugin kontena-plugin-hello
DEBUG Running Kontena::MainCommand with ["--master-url=http://localhost:9291", "node", "ls", "--grid", "development"] -- callback matcher = 'nil'
DEBUG Running Kontena::Cli::NodeCommand with ["ls", "--grid", "development"] -- callback matcher = 'nil'
DEBUG Loading configuration from ENV
DEBUG Configuration loaded with 1 servers.
DEBUG Current master: default
DEBUG Current grid: (not selected)
DEBUG Running Kontena::Cli::Nodes::ListCommand with ["--grid", "development"] -- callback matcher = 'nodes list'
ArgumentError: You have not selected a grid. Use: kontena grid
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/config.rb:378:in `require_current_grid'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:157:in `verify_current_grid'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:211:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:216:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:216:in `run'
  bin/kontena:18:in `<top (required)>'
DEBUG Execution took 0.428 seconds
DEBUG SystemExit status 1
```

(with a `KONTENA_TOKEN` env exported earlier)

kke:
Replaced by #2923

PR #1912, opened 2017-03-01
Link env variables reference to summary

jakolehm:


SpComb:
Shouldn't the `references/environment-variables` show up in the https://kontena.io/docs/ ToC anways? Or is the docs site just trailing behind git master?

SpComb:
Oh... I suppose the new docs site uses the `SUMMARY.md` as the ToC? Are the old ToC headers not used anymore?

PR #1921, opened 2017-03-02
Fix master auth config race condition issues

### LINKED ISSUE #1919: AuthProvider configuration is invalid with multiple master processes or nodes
Kontena server master nodes store oAuth related configuration in the `AuthProvider` class (`server/app/services/auth_provider.rb`).  The configuration is provided by a PATCH or PUT HTTP request to the config API (`server/app/routes/v1/config_api.rb`), generally sent from the cli `kontena master config import` command.

The issue is that this configuration is stored in memory as part of the `AuthProvider` class OpenStruct fields, and as a result will only be available for a single master node process after being set.  This is an issue for cli commands like `kontena master init-cloud`, which first set configuration and then initiate an oAuth authorization.

To recreate this issue, follow the steps in `development.md` on a system with more than one available CPU core (I was able to reproduce easily with 4 cores).  When you get to the `dontena master init-cloud --force` command you should see a failure.

kke:
Fixes #1919

This PR makes AuthProvider stop being a singleton, removing any thread issues with the configuration when running multiple master web processes.

### DIFF
--- a/server/app/services/auth_provider.rb
+++ b/server/app/services/auth_provider.rb
@@ -79,23 +79,13 @@ class AuthProvider < OpenStruct
     @table[:root_url] = cfg['server.root_url']
     @table[:cloud_api_url] = cfg['cloud.api_url'] || 'https://cloud-api.kontena.io'
     @table[:ignore_invalid_ssl] = cfg['cloud.ignore_invalid_ssl'].to_s == 'true'
-
-    if @table['cloud.provider_is_kontena'].to_s == "true"
-      @is_kontena = true
-    elsif @table[:authorize_endpoint]
-      uri = URI.parse(@table[:authorize_endpoint]) rescue nil
-      if uri && uri.host.end_with?('kontena.io')
-        @is_kontena = true
-      else
-        @is_kontena = false
-      end
-    else
-      @is_kontena = false
-    end
+    @table[:provider_is_kontena] = cfg['cloud.provider_is_kontena'].to_s == "true"
   end
 
   def is_kontena?
-    @is_kontena
+    return true if self.provider_is_kontena
+    uri = URI.parse(self.authorize_endpoint) rescue nil
+    uri && uri.host.end_with?('kontena.io')
   end
 
   def update_kontena
--- a/server/app/services/auth_provider.rb
+++ b/server/app/services/auth_provider.rb
@@ -36,9 +36,12 @@ class AuthProvider < OpenStruct
   ]
 
   def self.reset_instance
-    @last_update = Configuration.last_update
     Singleton.send :__init__, self
     self
+  end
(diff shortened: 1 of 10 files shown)


server/app/services/auth_provider.rb
```diff
@@ -40,43 +40,66 @@ def self.reset_instance
...
 
+  def self.set_updated
+    @last_update = Configuration.last_update
+  end
+
+  def self.last_update
+    @last_update
+  end
+
+  def self.valid?
```
jakolehm: Missing specs...

SpComb:
This branch seems to include all the `1.2-dev` master commits...

For these 1.1.x fixes, just have the PR merge into master. Once it's fixed in master, you cherry-pick the commit from master into a (new) separate `git checkout -b v1.1 1.1.2` branch.

server/app/services/auth_provider.rb
```diff
@@ -35,48 +30,65 @@ class AuthProvider < OpenStruct
...
+  attr_accessor :userinfo_username_jsonpath
+  attr_accessor :userinfo_email_jsonpath
+  attr_accessor :userinfo_user_id_jsonpath
+  attr_accessor :root_url
+  attr_accessor :cloud_api_url
+  attr_accessor :ignore_invalid_ssl
+  attr_accessor :provider_is_kontena
+
+  def self.instance
+    new(Configuration.decrypt_all)
```
jakolehm: Does this dump every value in Configuration to AuthProvider?

jakolehm:
@daniel-bytes pls test if this solves multi-core problems that you had.

server/app/services/auth_provider.rb
kke: Yes. Otherwise there would be 14 separate queries.

SpComb:
Is this `status/merge`, or waiting for pre-merge testing?

server/app/services/auth_provider.rb
```diff
@@ -35,48 +30,65 @@ class AuthProvider < OpenStruct
...
+  attr_accessor :userinfo_endpoint
+  attr_accessor :userinfo_username_jsonpath
+  attr_accessor :userinfo_email_jsonpath
+  attr_accessor :userinfo_user_id_jsonpath
+  attr_accessor :root_url
+  attr_accessor :cloud_api_url
+  attr_accessor :ignore_invalid_ssl
+  attr_accessor :provider_is_kontena
+
+  def self.instance
```
daniel-bytes: A static method named `instance` makes me think this is holding on to a lazy-loaded singleton, not creating a new instance each call.

daniel-bytes:
I tested this branch against a 4 core Docker for Mac VM and it looks good.

PR #1926, opened 2017-03-02
[WIP] Statistics and Metrics APIs

daniel-bytes:
This is a work in progress for the APIs required in Kontena server to expose key metrics for consumption by Kontena cloud.

Experimenting with ways of aggregating a stream of records pulled from the database to ensure data is aggregated for all records in the stream and then in specific time slices (ex: average per second).


SpComb:
Needs rebase over #1908

`git rebase --onto master 371d6ef`

PR #1929, opened 2017-03-03
Metrics API > Host Nodes and Grids

daniel-bytes:
API endpoints to expose metrics for nodes and grids.


server/db/migrations/19_create_host_node_stats_created_at_index.rb
```diff
@@ -0,0 +1,5 @@
+class CreateHostNodeStatsCreatedAtIndex < Mongodb::Migration
+  def self.up
+    HostNodeStat.create_indexes
```
SpComb: Is this for #1908?

daniel-bytes: Kind of... I added the created_at index to the model definition for #1908, but it wasn't necessary at that point as we weren't using it yet.  I still need to verify if that index is enough for the aggregations being done in this PR, or if it needs to be modified (ie making it a compound index).

nevalla: Migration number 19 is already used in #1874, so prepare to change (depending on which one got merged first)

server/app/routes/v1/nodes_api.rb
```diff
@@ -52,6 +52,17 @@ def load_grid_node(grid_name, node_id)
 
             render('host_nodes/health')
           end
+
+          # GET /v1/nodes/:grid/:node/stats
+          r.on 'stats' do
+            default_seconds = 60 * 60
```
nevalla: You can use also `1.hour.to_i` which is more human readable (ActiveSupport provides that functionality). Actually you don't even need to cast it to integer `to - 1.hour` should work.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -236,6 +253,54 @@ def calculate_container_time(container)
...
+        idle_ticks = current.idle - prev.idle
+
+        total_ticks = system_ticks + user_ticks + idle_ticks
+
+        [system_ticks, user_ticks, idle_ticks].map do |ticks|
+          (ticks / total_ticks.to_f) * 100.0
+        end
+      end
+
+      averages = all.transpose.map do |stats|
```
SpComb: If I understood this correctly, these would be total percentages over all CPUs?

```ruby
total_system_percentage, total_user_percentage, total_idle_percentage = all.transpose.map do ...
```

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -236,6 +253,54 @@ def calculate_container_time(container)
       0
     end
 
+    # @param [Array<Vmstat::Cpu>] prev_cpu_stats
+    # @param [Array<Vmstat::Cpu>] current_cpu_stats
+    # @return [Hash]
+    def calculate_average_cpu(prev_cpu_stats, current_cpu_stats)
+      all = prev_cpu_stats.zip(current_cpu_stats).map do |prev, current|
```
SpComb: I think this needs some comments to understand :) I guess the `all` will be shaped like:

```
[
  [cpu0_system_percentage, cpu0_user_percentage, cpu0_idle_percentage],
  ...
  [cpuN_system_percentage, cpuN_user_percentage, cpuN_idle_percentage],
]
```

SpComb:
:+1: for the CPU averages, the agent is the correct place to calculate those

agent/lib/kontena/workers/node_info_worker.rb
SpComb: And then the `all.transpose`will look like

```ruby
[
  [cpu0_system_percentage, ..., cpuN_system_percentage],
  [cpu0_user_percentage, ..., cpuN_user_percentage],
  [cpu0_idle_percentage, ..., cpuN_idle_percentage],
]
```

server/db/migrations/19_create_host_node_stats_created_at_index.rb
kke: maybe migration numbering should be changed to something like "201703081345_foofoo.rb"

agent/lib/kontena/workers/node_info_worker.rb
SpComb: BTW re summing up CPU utilization %, there's two schools of thought.. This kind of averaging is like Windows, where two busy-looping processes on a four-core machine will be 50% usage. Generally, in Linux (like what `top` etc do), two busy-looping processes/threads on an N-core machine will show as 200% usage.

I personally prefer the Linux style... You don't need to know how many CPUs your machine has to see if something is bottlenecking on a single CPU .

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -236,6 +253,54 @@ def calculate_container_time(container)
...
+          (ticks / total_ticks.to_f) * 100.0
+        end
+      end
+
+      averages = all.transpose.map do |stats|
+        total = stats.reduce :+
+        total / stats.size.to_f
+      end
+
+      return {
```
daniel-bytes: I know... I actually only used it because it looked so ugly to me starting the line with just the open parens

agent/lib/kontena/workers/node_info_worker.rb
daniel-bytes: Yes thats correct.  I actually found someone online doing something similar w/ VmStat data, that's how I came to this.  Pretty cool once you wrap your head around it.

agent/lib/kontena/workers/node_info_worker.rb
daniel-bytes: Haha maybe too many years of Windows has corrupted my mind.  :)

I find it confusing because then 25% means something very different on my 8 core laptop than it does on a 1 core micro cloud VM.  Maybe we should return # of cores back to the cloud UI?

@nevalla what do you think?

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -236,6 +253,54 @@ def calculate_container_time(container)
...
+      in_bytes_per_second = (current_iface.in_bytes - previous_iface.in_bytes) / interval
+      out_bytes_per_second = (current_iface.out_bytes - previous_iface.out_bytes) / interval
+
+      return {
+        in_bytes_per_second: in_bytes_per_second,
+        out_bytes_per_second: out_bytes_per_second
+      }
+    end
+
+    def get_network_interface()
```
daniel-bytes: @SpComb @nevalla : Any thoughts on this algorithm for getting the network interface to monitor?  I'm just grabbing the ethernet adapter with the most outgoing traffic.  Wasn't sure if I should instead get a sum of all ethernet (or non-loopback) adapters I/O.

daniel-bytes:
Branch updated with fixes requested from code review.  Also adding stats API endpoint for grids (aggregate of node stats for all nodes on a grid).  Removed WIP status.

daniel-bytes:
Closing this WIP, going to open a new PR

PR #1930, opened 2017-03-06
Send file:// as registry url to allow backwards compatibility with pre v1.1.2 masters

kke:
The "registry" field in stack metadata was mandatory before v1.1.2. 

This PR sends 'file://' as registry url for stacks installed from local files or direct urls like pre v1.1.2 CLI's did.

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -38,10 +38,10 @@ module Kontena::Cli::Stacks
           @registry    = current_account.stacks_url
         elsif from_url?
           @raw_content = load_from_url(file)
-          @registry = nil
+          @registry = 'file://'
         else
           @raw_content = File.read(File.expand_path(file))
-          @registry = nil
+          @registry = 'file://'
         end
 
         @errors           = []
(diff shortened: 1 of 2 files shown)


SpComb:
The `kontena vpn create` and `kontena registry create` commands also use `file://`:

* https://github.com/kontena/kontena/blob/v1.1.2/cli/lib/kontena/cli/vpn/create_command.rb#L28
* https://github.com/kontena/kontena/blob/v1.1.2/cli/lib/kontena/cli/registry/create_command.rb#L95

kke:
And that is ok because the vpn and registry are not in the registry.

SpComb:
> And that is ok because the vpn and registry are not in the registry.

Yes, just leaving a comment on those two, because we forgot to "fix" them in #1870

SpComb:
Do we want to stick to the old `{"registry": "file://"}` for both 1.1.3 and 1.2.0?

Alternatively, do we want to fix this for 1.1.3, and then fix 1.2.0 to always use `{"registry": null}`?

jakolehm:
Is there some benefit for using `null`?

SpComb:
> Is there some benefit for using null?

Semantics... once we implement some `kontena stack upgrade --auto foo`, which looks at the `registry` and `stack` of the existing stack to figure out if there's some new version of it available (possibly compatible with some potential `~ 0.2` semver matcher...), then it will need to ignore the `file://` stuff... which basically means that they're just placeholders for `null`.

SpComb:
I guess we'll just stick with `"file://"` everywhere, both for 1.1.3 and 1.2.

PR #1932, opened 2017-03-06
Remove unnecessary spec_helper requires in tests

kke:
Rspec adds spec/ to load path. You don't need any `require_relative "../../../spec_helper"` sorcery, 'require "spec_helper"` works fine.

### DIFF
--- a/agent/.gitignore
+++ b/agent/.gitignore
@@ -8,3 +8,4 @@ log/*.log
 /build/
 /release/
 /vendor/
+.rspec-local
--- a/agent/.gitignore
+++ b/agent/.gitignore
@@ -8,4 +8,3 @@ log/*.log
 /build/
 /release/
 /vendor/
-.rspec-local
--- a/cli/.gitignore
+++ b/cli/.gitignore
@@ -15,3 +15,4 @@ mkmf.log
 .idea
 *.gem
 /vendor/
+/.rspec-local
--- a/cli/.gitignore
+++ b/cli/.gitignore
@@ -15,4 +15,3 @@ mkmf.log
 .idea
 *.gem
 /vendor/
-/.rspec-local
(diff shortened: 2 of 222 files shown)


SpComb:
The `agent` includes a `.rspec` with `--require spec_helper`, so the agent specs don't need this require at all.

kke:
I don't like including `.rspec` in git, because I you may want to make local changes to it.

SpComb:
> I don't like including .rspec in git, because you may want to make local changes to it.

Hmm... well, the `server` and `agent` already include their `.rspec`... but the server doesn't have `--require spec_helper`, so it also has `require_relative '../../spec_helper'` in each spec :confounded:

Could we fix all three of the components to work the same way? My vote is for `.rspec` with `--require spec_helper`, but I'm also okay with `require 'spec_helper'`.

kke:
I like to have `--profile --color --format d` in my cli `.rspec` :(

SpComb:
sez the [rspec docs](http://www.rubydoc.info/github/rspec/rspec-core/RSpec/Core/Configuration)

> Configuration options are loaded from ~/.rspec, .rspec, .rspec-local, command line switches, and the SPEC_OPTS environment variable (listed in lowest to highest precedence; for example, an option in ~/.rspec can be overridden by an option in .rspec-local).

So we can stick `.rspec` into git, and you can use `.rspec-local`?

kke:
Nice.

cli/.gitignore
```diff
@@ -15,3 +15,4 @@ mkmf.log
 .idea
 *.gem
 /vendor/
+/.rspec-local
```
SpComb: This could be in the top-level `.gitignore`

agent/.gitignore
```diff
@@ -8,3 +8,4 @@ log/*.log
 /build/
 /release/
 /vendor/
+.rspec-local
```
SpComb: Doesn't git pick it up from `../.gitignore`? I don't think this is needed here or in the cli/server.

PR #1934, opened 2017-03-06
Fixing paths for nested sub commands in calls to load_subcommand.

daniel-bytes:
Fixes an issue in [PR 1093](https://github.com/kontena/kontena/pull/1093) where paths are incorrect for some nested sub commands.


SpComb:
BTW I noticed that the `kontena stack registry` command isn't using the `subcommand_loader` either... want to fix that too at the same time?

daniel-bytes:
@SpComb: I updated stack registry

SpComb:
BTW you don't need to write out the `[PR xyz](...)` links, you can just write `#1093`.

PR #1936, opened 2017-03-07
Inform user that input from STDIN is expected when doing vault write without value as param

### LINKED ISSUE #1375: kontena vault write silently hangs and waits for input
If you try:

```bash
$ kontena vault write secret_name
```

You will end up with the cursor on the next line and nothing ever happening. It's actually waiting for input from STDIN.

To fix, replace:

```ruby
      if secret.to_s == ''
        secret = STDIN.read
      end
```

with

```ruby
      if secret.to_s == ''
        puts "Enter content and end with ctrl-d" if $stdout.tty?
        secret = STDIN.read # maybe a good idea to chomp/strip
      end
```

kke:
Fixes #1375

From:

```
$ kontena vault write foo
<nothing until you ctrl-c or ctrl-d>
```

To:

```
$ kontena vault write foo
Enter a value for foo, press ctrl-d when finished:
```


kke:
Replaced by #2045

PR #1937, opened 2017-03-07
Only warn about server/cli version mismatch when minor version differs by 2+

kke:
Super annoying.

Now it only warns if:
- trying to talk with 0.x or 2.x servers using 1.x cli
- trying to talk with 1.6 server using 1.4 cli. 

1.2 CLI does not warn when talking to 1.3 or 1.1 server.

### DIFF
--- a/cli/lib/kontena/client.rb
+++ b/cli/lib/kontena/client.rb
@@ -491,7 +491,13 @@ module Kontena
       return nil if $VERSION_WARNING_ADDED
       return nil unless server_version.to_s =~ /^\d+\.\d+\.\d+/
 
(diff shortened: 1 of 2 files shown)


SpComb:
But I would expect a warning if using a `1.2` CLI to talk to a `1.1.x` master?

kke:
1.2 CLI should talk with 1.1.x master just fine. So should 1.3. Maybe it should just nag about major versions.

jakolehm:
Imho this is wrong... where do we define that cli/rest api is compatible like this? Probably it's a good idea to annoy for now so that users will upgrade (or downgrade) to match version.

PR #1940, opened 2017-03-07
Require --force or confirmation when upgrading to a different stack

### LINKED ISSUE #1938: stack upgrade can be done with wrong kontena.yml
Something like this: will "work"
```
git clone git@github.com:matti/kontena-stacks
cd kontena-stacks/lb-ingress
kontena stack install
cd ../weavescope
kontena stack upgrade lb-ingress
```

I managed to craft a very specific kontena.yml that broke the master for good (http 500 for all requests)
```
kontena_master_1 | 192.168.81.1 - - [07/Mar/2017:08:56:13 +0000] "GET /v1/grids/vagrant/stacks HTTP/1.1" 500 35 0.0975
kontena_master_1 |
kontena_master_1 | NoMethodError (undefined method `to_path' for nil:NilClass
kontena_master_1 | Did you mean?  to_a):
```

kke:
Fixes #1938

Require confirmation or `--force` when user/stackname is different on master than in the source you are trying to upgrade from.

### DIFF
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -15,6 +15,8 @@ module Kontena::Cli::Stacks
 
     option '--[no-]deploy', :flag, 'Trigger deploy after upgrade', default: true
 
+    option '--force', :flag, 'Force upgrade'
+
     requires_current_master
     requires_current_master_token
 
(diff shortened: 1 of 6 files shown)


cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -25,6 +27,10 @@ def execute
 
       stack = stack_from_yaml(filename, name: name, values: values, defaults: master_data['variables'])
 
+      unless force? || master_data['stack'] == stack['stack']
+        confirm "Going to replace stack #{Kontena.pastel.cyan(master_data['stack'])} on master with #{Kontena.pastel.cyan(stack['stack'])} from #{Kontena.pastel.yellow(filename)}. Are you sure?"
```
SpComb: Just "Replacing stack ..."

PR #1944, opened 2017-03-07
Fixing node_id issue in server node_handler_spec.

daniel-bytes:
Fixing missing `node_id` field for `HostNode` object in `server/spec/services/rpc/node_handler_spec.rb`.

### DIFF
--- a/server/spec/services/rpc/node_handler_spec.rb
+++ b/server/spec/services/rpc/node_handler_spec.rb
@@ -3,14 +3,15 @@ require_relative '../../spec_helper'
 describe Rpc::NodeHandler, celluloid: true do
   let(:grid) { Grid.create! }
   let(:subject) { described_class.new(grid) }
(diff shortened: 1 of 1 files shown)


server/spec/services/rpc/node_handler_spec.rb
```diff
@@ -3,14 +3,15 @@
 describe Rpc::NodeHandler, celluloid: true do
   let(:grid) { Grid.create! }
   let(:subject) { described_class.new(grid) }
-  let(:node) { HostNode.create!(grid: grid, name: 'test-node') }
+  let(:node) { HostNode.create!(node_id: 'a', grid: grid, name: 'test-node') }
 
   describe '#stats' do
     it 'saves host_node_stat item' do
-      node
+      puts 'save:'
```
SpComb: `puts`

daniel-bytes: argghh!  fixed.

PR #1952, opened 2017-03-09
Read stack name from ./kontena.yml unless given in stack commands

### LINKED ISSUE #1920: kontena stack commands should use kontena.yml ?
Currently `install` and `upgrade` default to kontena.yml so that you can say `kontena stack install` in a directory that has a `kontena.yml` (stack version of that).

Wouldn't it make sense that `logs` would also "read" the stack name from the kontena.yml (use the same logic what install uses to generate stack name..?) -- currently you need to `kontena stack logs -t stackname`

kke:
The stack name in all stack commands is now optional in case you are in a directory that has `kontena.yml`, in which case the stack name is read from it.

Fixes #1920

However, there are several stack commands that have multiple parameters and that brings a problem.

for example:
```
$ kontena stack monitor [STACK_NAME] [SERVICE_NAMES..]
```

Now if stack name is optional and left out, the first service name would be used as the stack name.
In this PR, using `.` as stack_name will read it from the kontena.yml.

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -7,6 +7,28 @@ module Kontena::Cli::Stacks
   module Common
     include Kontena::Cli::Services::ServicesHelper
 
+    module StackNameParamWithKontenaYmlFallback
+      def self.included(where)
+        where.class_eval do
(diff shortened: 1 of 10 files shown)


SpComb:
Feels a bit magical to me, I don't really like it, but I also always type out the `kontena stack install kontena.yml` explicitly... Opinions @jakolehm @jnummelin?

PR #1956, opened 2017-03-09
Metrics API > Host Nodes and Grids

daniel-bytes:
APIs for fetching HostNodeStat metrics data, by node and by grid.

API will default to the last hour of stats, 1 stat per minute, with network stats from network adapter `eth0`.  The client can pass optional query parameters `from` and `to` to modify the timespan, and `iface` to return network stats for an alternate network interface (such as `docker0`).


server/app/views/v1/grid_stats/stats.json.jbuilder
```diff
@@ -0,0 +1,15 @@
+json.from @node_stats[:from_time]
+json.to @node_stats[:to_time]
+json.stats @node_stats[:stats] do |stat|
+  json.timestamp stat[:timestamp]
+  json.data_points stat[:data_points]
+  json.cpu_usage_percent stat[:cpu_usage_percent]
+  json.memory_used_bytes stat[:memory_used_bytes]
```
jakolehm: Should these be objects instead of flat keys? Example:
```json
{
  "memory": {
    "used_bytes": 1111
  }
}
```

@nevalla what do you think?

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -236,6 +253,58 @@ def calculate_container_time(container)
...
+
+      {
+        interface_name: current_iface.name,
+        in_bytes_per_second: in_bytes_per_second,
+        out_bytes_per_second: out_bytes_per_second
+      }
+    end
+
+    def get_network_interface
+      Vmstat.network_interfaces.select { |x| x.ethernet? and x.in_bytes > 0 and x.out_bytes > 0 }
```
SpComb: Selecting the network interface with the most traffic isn't a good idea, each node will have several network interfaces with different levels of traffic. Enumerating the interfaces isn't that straightforward either, since they will vary across different platforms....

On virtual machines, it seems like `ethernet?` may exclude the bridge and veth interfaces, leaving only the virtualized ethernet interfaces: `Vmstat.network_interfaces.select{|x|x.ethernet?}.map{|x| x.name} => [:eth0, :eth1]` Using Kontena terms, `eth1` would be the `private_interface` (`ENV['KONTENA_PEER_INTERFACE']`), and then the remaining `eth0` could be assumed to be the `public_interface`. However, on some platforms like Packet, the interfaces will be entirely different... 

Other interesting interfaces will well-known names would be the `docker0` bridge interface, and the `weave` bridge interface. The `docker0` bridge interface stats are the total of all the container `eth0` network stats, whereas the `weave` bridge interface stats would be the total of all container `ethwe` network stats.

jakolehm: @SpComb what if we just collect stats from `docker0` and `weave`? `docker0` traffic can be thought as "public" and `weave` as "internal".

SpComb: I suppose the difference will be `network_mode: host` containers.

server/app/views/v1/grid_stats/stats.json.jbuilder
```diff
@@ -0,0 +1,15 @@
+json.from @node_stats[:from_time]
+json.to @node_stats[:to_time]
+json.stats @node_stats[:stats] do |stat|
+  json.timestamp stat[:timestamp]
+  json.data_points stat[:data_points]
```
nevalla: what `data_points` indicates? Just thinking the use case where you might need that.

server/db/migrations/19_create_host_node_stats_created_at_index.rb
```diff
@@ -0,0 +1,5 @@
+class CreateHostNodeStatsCreatedAtIndex < Mongodb::Migration
```
nevalla: Prepare to change migration number

server/app/views/v1/grid_stats/stats.json.jbuilder
daniel-bytes: Got it.  I agree this is much nicer.

server/app/views/v1/grid_stats/stats.json.jbuilder
daniel-bytes: Just the number of data points in this stat.  It's probably not helpful for the front end, I mostly had it in there for the sake of unit tests.  I'll take it out.

server/db/migrations/19_create_host_node_stats_created_at_index.rb
daniel-bytes: I'll remove the percent calculations.

agent/lib/kontena/workers/node_info_worker.rb
daniel-bytes: So, any consensus on this?  Even if we start storing multiple interfaces (like `docker0` and `weave` in the mongo collection we would still have to aggregate them somehow).

jakolehm: @SpComb @daniel-bytes if `x.ethernet?` returns only virtualized ethernet interfaces then we should probably gather stats from those and store them separately in the master database.

daniel-bytes: We could store all of them but we still need a plan for aggregating them... either sum or choose a single adapter, etc

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -158,17 +164,28 @@ def send_statsd_metrics(event)
...
       event[:filesystem].each do |fs|
-        name = fs[:name].split("/")[1..-1].join(".")
+        name = fs[:name] ? fs[:name].split("/")[1..-1].join(".") : ""
         statsd.gauge("#{key_base}.filesystem.#{name}.free", fs[:free])
         statsd.gauge("#{key_base}.filesystem.#{name}.available", fs[:available])
         statsd.gauge("#{key_base}.filesystem.#{name}.used", fs[:used])
         statsd.gauge("#{key_base}.filesystem.#{name}.total", fs[:total])
       end
+      event[:network].each do |network|
+        name = network[:name] ? network[:name] : ""
```
SpComb: The network interface names might also include `.`, for VLAN interfaces names like `eth0.8`. Escape those to something else to keep the statsd schema at a fixed number of levels?

server/app/models/host_node_stat.rb
```diff
@@ -6,10 +6,270 @@ class HostNodeStat
...
+        },
+        filesystem: {
+          used: '$filesystem_used',
+          total: '$filesystem_total'
+        }
+      }
+    }])
+  end
+
+  def self.get_aggregate_stats_for_grid(grid_id, from_time, to_time, network_iface)
```
SpComb: Exposing a single "host node network interface name" parameter for grid-level aggregation isn't that great... different host nodes running on different platforms might use different network interface names.

It would be better to have each agent figure out what network interface to monitor?

server/app/views/v1/grid_metrics/metrics.json.jbuilder
```diff
@@ -0,0 +1,22 @@
+json.from @from
+json.to @to
+json.stats @grid_stats do |stat|
```
SpComb: Is this identical to the `host_nodes/metrics.json`?

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -245,6 +262,51 @@ def calculate_container_time(container)
...
+        }
+      }.inject(result) { |memo, cpu_core|
+        memo[:system] += cpu_core[:system]
+        memo[:user] += cpu_core[:user]
+        memo[:idle] += cpu_core[:idle]
+        memo
+      }
+    end
+
+    def get_network_traffic
```
SpComb: It looks like the rate calculations for the host-level network stats are in #1995. Do we want to calculate aggregated rates from absolute counters on the server, or should the agent just report rates?

server/app/models/host_node_stat.rb
SpComb: I guess this makes sense for `?iface=docker0` and `?iface=weave`, but not for the default `eth0` iface...

server/app/views/v1/grid_metrics/metrics.json.jbuilder
daniel-bytes: Yes, we can technically share the same view file if we wanted to.

server/app/models/host_node_stat.rb
daniel-bytes: "It would be better to have each agent figure out what network interface to monitor?"

I wasn't sure which one would make sense to collect at this level so I just collected them all (the same as cAdvisor does for container stats), and push the decision of which to show down to the API consumer.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -245,6 +262,51 @@ def calculate_container_time(container)
...
+        user: 0.0,
+        idle: 0.0
+      }
+
+      prev_cpus.zip(current_cpus).map { |prev_cpu, current_cpu|
+        system_ticks = current_cpu.system - prev_cpu.system
+        user_ticks = current_cpu.user - prev_cpu.user
+        idle_ticks = current_cpu.idle - prev_cpu.idle
+
+        total_ticks = (system_ticks + user_ticks + idle_ticks).to_f
```
daniel-bytes: Do you think we should include the `nice` ticks?  Since `Vmstat` doesn't expose the others we will have to live with less-than-perfect potential values I suppose (or look into a different library, but that seems like overkill right now).

SpComb:
Closing in favor of #1995, which contains updates to this PR. Easier to just review as a whole then.

PR #1957, opened 2017-03-10
Stack yaml volume mapping parser and validation support

kke:
API support in #1849


cli/spec/fixtures/stack-with-volumes.yml
```diff
@@ -0,0 +1,33 @@
...
+      - wp-data-ext-2:/var/lib
+
+volumes:
+  wp-data:
+    driver: local
+    scope: instance
+    driver_opts:
+      foo: bar
+  wp-data-ext:
+    external: true
```
SpComb: Does `external: true` make sense? The server needs to know what volume to map, it should come from something like a CLI stack variables resolver?

kke: `external: true` is identical to compose.

jnummelin: In #1849 `external: true` tries to map the volume to a grid level volume with the given name.

jnummelin:
If the volume is specified to be `external`, the scope attribute should not be used. Now the validation fails with following stack:
```
stack: jussi/redis-ext-vol
description: Just a simple Redis stack with volume
version: 0.0.1
services:
  redis:
    image: redis:3.2-alpine
    stateful: true
    command: redis-server --appendonly yes
    volumes:
      - redis-data:/data
      - instanceVol:/foo

volumes:
  redis-data:
    external:
      name: testVol

  instanceVol:
    external: true
```
with error:
```
YAML validation failed! Aborting.

- "/Users/jussi/code/stacks/volumes/ext-vols.yml":
  - volumes:
      redis-data:
        scope: value from list required
  - volumes:
      instanceVol:
        scope: value from list required
```

Scope is always attribute of the actual volume as the usable scopes depends highly on the volume driver type.

SpComb:
We had some discussion about validating that the `external: xxx` `scope: ...` matches the scope that the volume was actaully created with... If the stack expects a `scope: instance` volume, and the user gives it a `scope: stack` volume, the service is not going to behave correctly.

OTOH I guess we figured we would implement that as a `variables: ext_vol: from: volume_resolver: scope: ...` resolver parameter...

cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
```diff
@@ -27,24 +39,91 @@ def validate(yaml)
...
+              if options['volumes']
+                mount_points = options['volumes'].inject(Hash.new(0)) { |hsh, vol| hsh[vol.split(':').last] += 1; hsh }
+                mount_points.each do |mount_point, occurences|
+                  next unless occurences > 1
+                  result[:errors] << { 'services' => { service => { 'volumes' => { mount_point => "mount point defined #{occurences} times" } } } }
+                end
+
+                options['volumes'].each do |volume|
+                  if volume.include?(':')
+                    volume_name, mount_point = volume.split(':', 2)
```
jnummelin: This will not work on "anonymous" volume definitions like `/data`

kke: And that's why there's this `if volume.include?(':')`?

jnummelin:
Not sure if the volume validations cover all the possible cases:
`/path/on/host:/data:ro,rprivate` - bind mount with optional flags
`a-volume:/data:ro,rprivate` - named volume with optional flags
'/data` - anonymous volume, no flags allowed

Also IMO the validations have too deep if/else branching, makes it really hard to read/understand.

SpComb:
Conflict on #1985

SpComb:
@kke can you merge this from master ASAP? We need to get this reviewed and merged today.

PR #1963, opened 2017-03-13
Client errors were not going to STDERR

### LINKED ISSUE #1959: Make `kontena vault read` scripting friendly
In some cases it would be helpful to be able to script reading values from the vault but currently `kontena vault read KEY` outputs multiple lines and in a format that is not easy to parse for scripting. So, two suggested improvements:

1. Add a '--raw' option so that only the value is output to stdout
2. Any errors such as "Not Found" should go to stderr instead of stdout (this really goes for all commands I think)

kke:
Fixes #1959

For example 404 errors were printed to STDOUT. This PR changes the output to STDERR.

Before:

```
$ kontena vault read foofoofoo > foo.txt
$ cat foo.txt
[error] Not found
```

After:

```
$ kontena vault read foofoofoo > foo.txt
[error] Not found
$ cat foo.txt
$
```

### DIFF
--- a/cli/lib/kontena/command.rb
+++ b/cli/lib/kontena/command.rb
@@ -69,7 +69,6 @@ class Kontena::Command < Clamp::Command
 
   def self.inherited(where)
     where.extend Finalizer
-    super
   end
 
   def self.callback_matcher(cmd_class = nil, cmd_type = nil)
--- a/cli/lib/kontena/command.rb
+++ b/cli/lib/kontena/command.rb
@@ -69,6 +69,7 @@ class Kontena::Command < Clamp::Command
 
   def self.inherited(where)
     where.extend Finalizer
+    super
   end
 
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/command.rb
```diff
@@ -69,6 +69,7 @@ def self.load_subcommand(path)
 
   def self.inherited(where)
     where.extend Finalizer
+    super
```
SpComb: What happens here?

SpComb:
Checked that the e2e tests still pass :+1:

cli/lib/kontena/command.rb
kke: Some unrelated local modifications to the file, removed.

kke: I was experimenting with something like:
```
class ListCommand < StackCommand
end
```
to replace
```
class Kontena::Cli::Stacks::ListCommand < Kontena::Command
end
```

PR #1970, opened 2017-03-14
Harmonize grid access checks

### LINKED ISSUE #1442: Server: current_user.grid_ids.include? vs current_user.accessible_grids.find_by

jnummelin:
fixes #1442 

There were couple of APIs that returned 404 in case no grid access, but I'd consider those a bugs as majority of APIs returned 403.

### DIFF
--- a/server/app/helpers/request_helpers.rb
+++ b/server/app/helpers/request_helpers.rb
@@ -52,6 +52,18 @@ module RequestHelpers
     request.halt
   end
 
+  ##
+  # @param [String] name
+  # @return [Grid]
+  def load_grid(name)
+    @grid = Grid.find_by(name: name)
+    halt_request(404, {error: 'Not found'}) unless @grid
+    unless current_user.has_access?(@grid)
+      halt_request(403, {error: 'Access denied'})
+    end
+    @grid
+  end
+
   def test_env?
     ENV['RACK_ENV'].to_s == 'test'
   end
--- a/server/app/models/user.rb
+++ b/server/app/models/user.rb
@@ -64,6 +64,12 @@ class User
   end
 
   ##
+  # @param [Grid] grid
+  def has_access?(grid)
+    self.accessible_grids.include?(grid)
+  end
+
+  ##
   # @param [String] role
   def in_role?(role)
     self.roles.where(name: role).exists?
--- a/server/app/models/user.rb
+++ b/server/app/models/user.rb
@@ -66,7 +66,11 @@ class User
   ##
   # @param [Grid] grid
   def has_access?(grid)
-    self.accessible_grids.include?(grid)
+    if self.master_admin?
+      true
+    else
+      self.grid_ids.include?(grid.id)
+    end
   end
 
   ##
--- a/server/app/routes/v1/certificates_api.rb
+++ b/server/app/routes/v1/certificates_api.rb
@@ -11,14 +11,6 @@ module V1
       validate_access_token
       require_current_user
 
-      ##
-      # @param [String] name
-      # @return [Grid]
-      def load_grid(name)
-        @grid = current_user.accessible_grids.find_by(name: name)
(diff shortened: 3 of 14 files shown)


server/app/helpers/request_helpers.rb
```diff
@@ -52,6 +52,18 @@ def halt_request(status, body = {})
     request.halt
   end
 
+  ##
```
kke: I think this load_grid thing could be done like the "require_current_user" and "validate_access_token" helpers, then you would always have a `current_grid` like you have `current_user`.

LGTM, just a thought.

SpComb: The grid name comes from the URL parameters, not the request headers.

jnummelin: Yes, that would make it troublesome to be parsed out genericly

server/app/models/user.rb
```diff
@@ -64,6 +64,12 @@ def accessible_grids
   end
 
   ##
+  # @param [Grid] grid
+  def has_access?(grid)
+    self.accessible_grids.include?(grid)
```
SpComb: Considering how hot this method is (called on all API requests?), this might be worth optimizing to just query for the specific grid, instead of querying for all grids and then doing a linear search?

Something equivalent to the existing `user.grid_ids.include?(grid.id)` or `accessible_grids.find_by(name: name)` methods, which AFAIK avoid loading all the grids?

```ruby
if master_admin?
  true
else
  self.grids.find(grid.id) # XXX: raises error?
end
```

Needs proper tests, though, because I'm not sure how those relation lookups behave :)

jnummelin:
```
Benchmark.bm(7) do |x|
        x.report("include?:")   { n.times do ; user.grids.include?(grids[rand(0..99)]); end }
        x.report("find by id:") { n.times do ; user.grids.find(grids[rand(0..99)].id); end }
      end
```
```
include?:  7.970000   0.490000   8.460000 ( 16.329947)
find by id: 14.170000   0.530000  14.700000 ( 21.880940)
```

jnummelin: So I'll refactor this to use `self.gird_ids.include?(grid.id)`

SpComb: Nice benchmark :1st_place_medal: 

Surprising that `user.grids.find(...)` is even slower than `user.grids.include?(...)`, it would be interesting to see what queries Mongoid does under the hood. OFC the `user.grids.find(...)` will also return the Grid object...

kke:
Fixes #2054

PR #1975, opened 2017-03-14
Fix Ubuntu xenial package install to not override prompted debconf values with empty values from the default config file

### LINKED ISSUE #1974: Ubuntu xenial install writes empty /etc/kontena-*.env variables from debconf
On a brand new system, without any `/etc/kontena-agent.env`, running `apt install kontena-agent` will prompt for the debconf `kontena-agent/server_uri` and `kontena-agent/grid_token` values during `config`, but fails to write them to the `/etc/kontena-agent.env` file during `postinst`:

```sh
# Set URL of your kontena server
KONTENA_URI=

# Set kontena token
KONTENA_TOKEN=

# kontena peer interface (overlay network interface)
KONTENA_PEER_INTERFACE=eth1
```

A later `dpkg-reconfigure kontena-agent` and re-entering the values will write them out.

SpComb:
Fixes #1974

Kontena 1.1.2 included support for `dpkg-reconfigure`(#1754), which involves having the package `config` scripts read any existing configuration values from the `/etc/kontena-*.env` conffile, and setting them to debconf. If the `config` script is only run once, this behaves correctly, as debconf will always prompt for the values when run the first time.

However, during normal `apt install`, the config script will be run twice: once during `dpkg-preconfigure` without any `/etc/kontena-*.env` conffiles in place, and then a second time after unpacking the `/etc/kontena-*.env` conffile. The first `config` run would prompt for the values, but the second `config` run would override the prompted values with the empty `KONTENA_*` values from the default `/etc/kontena-*.env` conffile. These empty debconf values would then get used by the `postinst` script and written into the conffile :disappointed: 

This fixes the `kontena-server` and `kontena-agent` `config` maintscripts to skip the debconf set if the config variables read from the default `/etc/kontena-*.env` conffile are empty. This preserves any existing debconf values from earlier prompts.

Testing this properly requires a 1.1.5 pre-release build, installed using `apt install`, ideally directly from the standard bintray source, as this issue does not happen when installing with `dpkg -i`: https://github.com/kontena/kontena/issues/1974#issuecomment-286403674

### DIFF
--- a/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
+++ b/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
@@ -8,7 +8,7 @@ if [ -e /etc/kontena-agent.env ]; then
   . /etc/kontena-agent.env
 
   [ -n "$KONTENA_URI" ] && db_set kontena-agent/server_uri "$KONTENA_URI"
-  [ -n "$KONTENA_URI" ] && db_set kontena-agent/grid_token "$KONTENA_TOKEN"
+  [ -n "$KONTENA_TOKEN" ] && db_set kontena-agent/grid_token "$KONTENA_TOKEN"
 fi
 
 db_input high kontena-agent/server_uri || true
--- a/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
+++ b/agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
@@ -7,8 +7,8 @@ set -e
 if [ -e /etc/kontena-agent.env ]; then
   . /etc/kontena-agent.env
 
-  db_set kontena-agent/server_uri "${KONTENA_URI:-}"
-  db_set kontena-agent/grid_token "${KONTENA_TOKEN:-}"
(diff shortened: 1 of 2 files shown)


agent/packaging/ubuntu_xenial/kontena-agent/DEBIAN/config
```diff
@@ -7,8 +7,8 @@ set -e
 if [ -e /etc/kontena-agent.env ]; then
   . /etc/kontena-agent.env
 
-  db_set kontena-agent/server_uri "${KONTENA_URI:-}"
-  db_set kontena-agent/grid_token "${KONTENA_TOKEN:-}"
+  [ -n "$KONTENA_URI" ] && db_set kontena-agent/server_uri "$KONTENA_URI"
+  [ -n "$KONTENA_URI" ] && db_set kontena-agent/grid_token "$KONTENA_TOKEN"
```
SpComb: Copy-pasta

jakolehm: Uh, how did I miss that.

PR #1976, opened 2017-03-14
Improve server stack mutatations to return errors for multiple services

SpComb:
Improve the server Stack mutations validation and error handling to use structured `service.$name.key` error keys. This allows the stack mutations to return multiple errors for different services, compared to the current `handle_service_outcome_errors` which only returns the last error. 

The current `handle_service_outcome_errors` uses the same `:services` key for each error, overrwriting any earlier service errors, and thus only returning the last error. When handling errors for multiple stack services in a loop, any cascading errors caused by the failure of one service would overwrite the initial root cause error.

## API
This changes the format of the stack API `{"errors": {"services": ....}}` error responses to also contain further nested errors for specific service fields.

## CLI
Change the generic API error handling to output the errors using YAML formatting, instead of the ruby hash syntax.

# Examples

For a stack containing multiple incorrect services, with incorrect links:

```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - bar

  asdf:
    image: redis
    depends_on:
      - quux
```

### Before

Only one of the service validation errors is returned:

```
 [fail] Creating stack test      
 [error] {"services"=>"Service validate failed for service 'asdf': Linked service 'quux' does not exist"}
```

### After

Validation errors are returned for both services, using the new YAML formatting on the CLI
```
 [fail] Creating stack test      
 [error] Unprocessable Entity:
	services:
	  foo:
	    links: Linked service 'bar' does not exist
	  asdf:
	    links: Linked service 'quux' does not exist
```


jnummelin:
Looking at the REST API documentation, we haven't documented the error responses that deeply so IMO it's ok to change the error responses.

The design look ok

SpComb:
TODO to improve the CLI `errors => { ... }` rendering to output something YAMLish

SpComb:
Updated for 1.2.dev, and improved the CLI API errors messages to benefit from the better API errors.

SpComb:
The TODO with handling multiple invalid `links` for a single service got complex enough that I made it a separate follow-up PR: #2058

cli/lib/kontena/errors.rb
```diff
@@ -4,10 +4,33 @@ class StandardError < ::StandardError
...
+      # @param status [Fixnum] HTTP response status
+      # @param message [String] short error message
       def initialize(status, message)
         @status = status
         super(message)
       end
     end
+
+    # The normal {error: {foo: "invalid foo"}} error response format used by the API
+    class StandardErrors < StandardError
```
jakolehm: `StandardErrors` vs `StandardError` .. is that name a bit too close (makes spotting usage hard)?

SpComb: What about `StandardErrorHash`?

PR #1980, opened 2017-03-15
Fix server stack services sorting

### LINKED ISSUE #1958: stack deploy fails due to missing service that is not missing
I recreated my staging grid and now I get an error when I try to install my stack:

```
# kontena stack install -n staging
 [fail] Creating stack staging
 [error] {"services"=>"Service create failed for service 'restaurants-host': Service staging/contactme-host does not exist"}
```

The kontena yaml contains the requested service of course:

```
services:
....
  contactme-host:
    extends:
      file: common-kontena.yaml
      service: java-service
    image: "nexus-dockerkulinariweb.orchit-dev.de/kulinariweb/contactme:${STAGE}_LATEST"
    depends_on:
      - eureka
      - confighub
      - rabbitmq
      - redis
      - dbinit
    links:
      - mariadb/node:mariadb
...
  restaurants-host:
    extends:
      file: common-kontena.yaml
      service: java-service
    image: "nexus-dockerkulinariweb.orchit-dev.de/kulinariweb/restaurants:${STAGE}_LATEST"
    depends_on:
      - eureka
      - rabbitmq
      - confighub
      - contactme-host
    links:
      - mariadb/node:mariadb
....
```
The services are created and in the initialized state. A kontena stack deploy afterwards seems to work. (still running)

SpComb:
Fixes #1958

The server must create and remove linked stack services in the correct order: a service can only be created when all of the services that it links to have been created, and it can only be removed once any linking services have been removed.

The current implementation of stack services sorting uses `Array#sort` with a comparison function based on the presence of shallow links, but this does not behave correctly for indirectly linked services, where the comparison function gives the wrong results. The following stack YAML results in a `sort_services` ordering of `["asdf1", "asdf2", "asdf3", "bar", "asdf", "foo"]`, which fails when creating the `bar` service:

```yaml
stack: test/test
services:
  bar:
    image: redis
    depends_on:
      - foo

  foo:
    image: redis
    depends_on:
      - asdf

  asdf:
    image: redis
    depends_on:
      - asdf1
      - asdf2
      - asdf3
  asdf1:
    image: redis
  asdf2:
    image: redis
  asdf3:
    image: redis
```

This sorting and validation also fails to detect loops in the services graph, including the trivial case of a service linking to itself.

### Complex indirect linking case
```
$ kontena stack install test/links-sorting.yaml 
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  bar:
	    links: Service test/foo does not exist
```
```
$ kontena stack show test
test:
  created: 2017-05-15T14:18:39.713Z
  updated: 2017-05-15T14:18:39.716Z
  state: initialized
```

### Direct recursive links
```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - foo
```
```
$ kontena stack install test/links-recursive.yaml 
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  foo:
	    links: Service test/foo does not exist
```
```
$ kontena stack ls
NAME                           STACK                                    SERVICES   STATE      EXPOSED PORTS                                     
⊝ test                         test/test:0.0.1                          0          initialized                                                   
```

### Indirect recursive links
```yaml
stack: test/test
services:
  bar:
    image: redis
    depends_on:
      - foo

  foo:
    image: redis
    depends_on:
      - asdf

  asdf:
    image: redis
    depends_on:
      - bar
```
```
$ kontena stack install test/links-recursive.yaml 
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  bar:
	    links: Service test/foo does not exist
	  asdf:
	    links: Service test/bar does not exist
	  foo:
	    links: Service test/asdf does not exist
```
```
$ kontena stack ls
NAME                           STACK                                    SERVICES   STATE      EXPOSED PORTS                                     
⊝ test                         test/test:0.0.1                          0          initialized                                                   
```

### Fix sorting to only consider services in the same stack

The `sort_services` was sorting services with external links incorrectly. For the API `{links: [ {name: 'foo/bar' } ]}` case, it would ignore the `foo/bar` links, but for the `GridService.GridServiceLinks` case, it would just take the name of the linked service in the external stack. If that matched the name of a service in the local stack, the sorting order could be incorrect.

Fixed by having the `sort_services` only select internal links within the same stack.

### Fix sorting of complex service linking structures

Change `sort_services` to use a deep list of linked services, including the recursive links of linked services, for the sort comparison. This should provide a stable comparison function that works with `Array#sort`.

### Use `sort_services` `raise LinkError` for service link validation

The mutations previously used a `validate_service_links` that both mutated the `service[:links]` to only leave external links, as well as checked for links to internal missing services. It would not validate other errors such as a service linking to itself, or indirect recursive links between services. These kinds of recursive links would lead to post-validation `GridServices::Create.run` errors, leaving behind an incompletely created stack. Without the fixes in #1976, these errors would also overwrite eachother, obscuring the actual service with broken links.

The new `sort_services` uses the deep `linked_services` array to validate both links to missing services, as well as recursive links. Both of these raise a `LinkError` with a `service` attribute. The mutations now rescue this to add the validation error. This no longer leaves an incomplete `test` stack behind, as it now fails early during validation.

The detection of recursive links relies on the use of nested `Array` references, and the ability of `Array#flatten!` to detect recursive references:

```ruby
irb(main):001:0> a = ['foo']
=> ["foo"]
irb(main):002:0> b = ['bar', a]
=> ["bar", ["foo"]]
irb(main):003:0> a << b
=> ["foo", ["bar", [...]]]
irb(main):004:0> a.flatten!
ArgumentError: tried to flatten recursive array
	from (irb):4:in `flatten!'
	from (irb):4
	from /usr/bin/irb:11:in `<main>'
```

### Complex indirect linking case
```
$ kontena stack install test/links-sorting.yaml 
 [done] Creating stack test      
 [done] Triggering deployment of stack test     
 [done] Deploying service asdf1     
 [done] Deploying service asdf2     
 [done] Deploying service asdf3     
 [done] Deploying service asdf     
 [done] Deploying service foo     
 [done] Deploying service bar     

```

### Missing link
```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - bar
```
```
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  foo:
	    links: 'service foo has missing links: bar'
```

The `test` stack does not get created.

### Direct recursive link
```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - foo
```
```
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  foo:
	    links: 'service foo has recursive links: ["foo", [...]]'
```

The `test` stack does not get created.

### Indirect recursive links
```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - bar
  bar:
    image: redis
    depends_on:
      - asdf
  asdf:
    image: redis
    depends_on:
      - foo
```
```
 [fail] Creating stack test      
 [error] Kontena::Errors::StandardErrorHash : Unprocessable Entity:
	services:
	  foo:
	    links: 'service foo has recursive links: ["bar", ["asdf", ["foo", [...]]]]'
```

The `test` stack does not get created.


SpComb:
Spotted another bug in the `Stacks::SortHelper`: when sorting `GridService`s, it only looks at the linked service name, ignoring the stack. Links to some service in a different stack would confuse it to think that service was linking to an identically named service in the same stack.

server/app/mutations/stacks/sort_helper.rb
```diff
@@ -1,29 +1,79 @@
...
+      end
+
+      # Flatten the nested references to arrays
+      # In case of recursive references, the Array#flatten! will fail with ArgumentError: tried to flatten recursive array
+      # {service => [linked_service, linked_service_link, ...]}
+      service_links.each do |service, links|
+        begin
+          service_links[service].flatten!
+        rescue ArgumentError
+          raise LinkError.new(service), "service #{service} has recursive links: #{links}"
```
SpComb: This is pretty random in terms of which service it gets raised for exactly, and what the links look like. It's not the service with the minimal link cycle, it could be any leaf service that links into a cycle of linked services.

server/app/mutations/stacks/common.rb
```diff
@@ -35,16 +35,11 @@ def validate_expose
...
+    def select_external_service_links(service)
       links = service[:links] || []
-      internal_links = links.select{ |l| !l['name'].include?('/') }
-      links = links - internal_links
-      internal_links.each do |l|
-        unless self.services.any?{|s| s[:name] == l['name']}
-          add_service_error(service[:name], :links, :exist,  "Linked service '#{l['name']}' does not exist")
-        end
-      end
-      service[:links] = links
```
SpComb: This kind of hidden mutation in a `validate` method also happens in the `GridServices::Common#validate_links`, which mutates the `link[:name]` to add the `stack/` prefix if missing. I think that ought to be fixed too.

server/app/mutations/stacks/sort_helper.rb
```diff
@@ -1,29 +1,105 @@
...
+          end
+        end
+      end
+
+      # Flatten the deep array references to a flat array
+      # In case of recursive references, the Array#flatten! will fail with ArgumentError: tried to flatten recursive array
+      # {service => [linked_service, linked_service_link, ...]}
+      service_links.each do |service, links|
+        begin
+          service_links[service] = service_links[service].flatten
```
SpComb: There was a bug here with `service_links[service].flatten!`, which the specs did not catch, but I do not wish to try writing a spec for that.

SpComb:
The recursive graph algorithm based on recursive Ruby arrays here turns out to be very similar to what's mentioned as an answer to [What are recursive arrays good for?](http://stackoverflow.com/questions/10606734/what-are-recursive-arrays-good-for/10608755#10608755). This is a nice example of a use-case for recursive arrays, also with the addition of the `Array#flatten => ArgumentError: tried to flatten recursive array` usecase.

PR #1987, opened 2017-03-15
Replace server timeout { sleep until ... } loops with non-interrupting wait_until { ... } loops

SpComb:
Fixes #1986 to not interrupt moped operations.

This differs from the [agent's `WaitHelper`](https://github.com/kontena/kontena/blob/master/agent/lib/kontena/helpers/wait_helper.rb) in some ways:

* Renamed to `wait_until { condition }` to more closely match the `sleep interval until condition` that this is replacing
* The `wait_until!` method correctly returns the result
* The `wait_until` method only yields once if the interval is longer than the timeout
* The `wait_until` method does not yield if the timeout is zero
* The `wait_until` method takes the message as an optional first positional argument
* The debug and exception messages contains timing information
* More debug messages

This differs from the existing `Timeout::timeout` in some ways:

* Does not impose any timeout on the actual block operation - you need to use e.g. moped's own timeout mechanisms instead

PR #1988, opened 2017-03-16
Backport timeout fixes to 1.1

jakolehm:
Backports #1987 to v1.1

### DIFF
--- a/server/app/helpers/wait_helper.rb
+++ b/server/app/helpers/wait_helper.rb
@@ -4,6 +4,10 @@ module WaitHelper
   WAIT_TIMEOUT = 300
   WAIT_INTERVAL = 0.5
 
+  def _wait_now
+    Time.now.to_f
+  end
+
   # Wait until given block returns truthy value, returning nil on timeout
   #
   # @param timeout [Fixnum] How long to wait
@@ -14,10 +18,10 @@ module WaitHelper
   def wait(timeout: WAIT_TIMEOUT, interval: WAIT_INTERVAL, message: nil, &block)
     raise ArgumentError, 'no block given' unless block_given?
 
-    wait_until = Time.now.to_f + timeout
+    wait_until = _wait_now + timeout
 
     loop do
-      return nil if Time.now.to_f > wait_until
+      return nil if _wait_now > wait_until
 
       value = yield
 
--- a/server/app/helpers/wait_helper.rb
+++ b/server/app/helpers/wait_helper.rb
@@ -1,0 +1,47 @@
+module WaitHelper
+  include Logging
+
+  WAIT_TIMEOUT = 300
+  WAIT_INTERVAL = 0.5
+
+  # Wait until given block returns truthy value, returning nil on timeout
+  #
+  # @param timeout [Fixnum] How long to wait
+  # @param interval [Fixnum] At what interval is the block yielded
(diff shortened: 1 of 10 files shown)


server/app/helpers/wait_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    loop do
+      return nil if _wait_now > wait_until
+
+      value = yield
+
+      if value
+        return value
+      else
+        debug "wait... #{message}"
```
jnummelin: Maybe guard this with `if message`? Seeing random `wait...` messages without context in log would be slightly confusing

jakolehm: done

server/app/helpers/wait_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+  # @raise [Timeout::Error] If block does not return truthy value within given timeout
+  def wait!(**opts, &block)
+    if value = wait(**opts, &block)
+      return value
+    else
+      raise Timeout::Error
+    end
+  end
+
+  # XXX: also used in class methods
```
SpComb: `git cherry-pick 36bb0c8`

server/app/models/distributed_lock.rb
```diff
@@ -13,17 +15,16 @@ def self.with_lock(name, timeout = 60)
     lock_id = nil
     begin
       if timeout.to_f > 0.0
-        Timeout.timeout(timeout) do
-          sleep 0.01 until lock_id = self.obtain_lock(name)
-        end
+        lock_id = WaitHelper.wait(timeout: timeout) { self.obtain_lock(name) }
```
SpComb: Should this use `interval: 0.01`? Dunno how much of a difference the default 0.5 interval makes.

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -32,7 +33,7 @@ def deploy(node, instance_number, deploy_rev, creds = nil)
   # @param [String] deploy_rev
   def wait_for_service_to_start(node, instance_number, deploy_rev)
     # node/agent has 5 minutes to do it's job
-    Timeout.timeout(300) do
+    wait!(timeout: 300) do
       sleep 0.5 until self.deployed_service_container_exists?(instance_number, deploy_rev)
```
SpComb: This looks broken... it's `sleep until ...` within the wait-block, and does this actually return `true`?

This does not ever time out anymore if the service fails to deploy... which is a problem, because the 1.1 RPC's don't report errors either.

server/app/mutations/grid_certificates/get_certificate.rb
```diff
@@ -54,7 +55,7 @@ def execute
           end
         end
 
-        Timeout::timeout(30) {
+        wait!(timeout: 30) {
           info 'waiting for DNS validation...'
           sleep 1 until challenge.verify_status == 'valid'
```
jnummelin: See: https://github.com/kontena/kontena/pull/1988#discussion_r106352492

SpComb: Yes, this should also be like

```
info "waiting for ..."
wait!(timeout: 30) { challenge.verify_status == 'valid' }
info "..."
```

PR #1989, opened 2017-03-16
Revert "Backport timeout fixes to 1.1"

kke:
Reverts kontena/kontena#1988


SpComb:
GG, this revert of a force-unpushed commit like un-reverted it or something.

PR #1991, opened 2017-03-16
Fix server v1.1: allow wait with zero timeout

SpComb:
The new server `WaitHelper#wait_until` had an edge case when called with a zero timeout. This happens when removing a service with zero containers, which uses a timeout proportional to the number of containers. Without the fix, this will immediately fail with a timeout, even if the wait condition is true. This breaks removal of a service that has failed to deploy.

The previous `Timeout::timeout(0)` would not impose any timeout on the block. The new `WaitHelper.wait_until(timeout: 0) { ... }` will always yield the block exactly once, and fail with a timeout if the block is not immediately true.

This is sufficient to fix the `GridServiceRemoveWorker` case:

```ruby
      wait_instance_removal(grid_service, grid_service.containers.scoped.count * 30)

  def wait_instance_removal(grid_service, timeout)
    wait_until!(timeout: timeout) { grid_service.reload.containers.scoped.count == 0 }
  end
```

### DIFF
--- a/server/app/helpers/wait_helper.rb
+++ b/server/app/helpers/wait_helper.rb
@@ -8,7 +8,9 @@ module WaitHelper
     Time.now.to_f
   end
 
-  # Wait until given block returns truthy value, returning nil on timeout
+  # Wait until given block returns truthy value, returning nil on timeout.
+  #
+  # For a zero timeout, only yields exactly once.
   #
   # @param timeout [Fixnum] How long to wait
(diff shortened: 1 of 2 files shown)


SpComb:
Confirmed zero-container service removal to work in 1.1.5.rc3:

```
$ kontena service scale redis 0
 [done] Scaling redis to 0 instances      
$ kontena service rm --force redis
 [done] Removing service redis      
```

PR #1993, opened 2017-03-16
Stop log streaming if queue grows too big

jakolehm:
We don't really want to exhaust master with logs. This PR implements a fix that stops log streaming if queue grows too big. Log streaming is re-enabled after queue has been processed.

### DIFF
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -4,7 +4,6 @@ module Kontena::Workers
     include Kontena::Logging
 
     CHUNK_REGEX = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)\s(.*)$/
-    QUEUE_LIMIT = 5000
 
     # @param [Docker::Container] container
     # @param [Queue] queue
@@ -68,7 +67,7 @@ module Kontena::Workers
         type: stream,
         data: data
       }
-      @queue << msg if @queue.size < QUEUE_LIMIT
+      @queue << msg
       publish_log(msg)
     end
 
--- a/agent/lib/kontena/workers/container_log_worker.rb
+++ b/agent/lib/kontena/workers/container_log_worker.rb
@@ -4,6 +4,8 @@ module Kontena::Workers
     include Kontena::Logging
 
     CHUNK_REGEX = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)\s(.*)$/
+    QUEUE_LIMIT = 5000
+    EVENT_NAME = 'container:log'
 
     # @param [Docker::Container] container
     # @param [Queue] queue
@@ -67,7 +69,7 @@ module Kontena::Workers
         type: stream,
         data: data
       }
-      @queue << msg
+      @queue << msg if @queue.size < QUEUE_LIMIT
       publish_log(msg)
     end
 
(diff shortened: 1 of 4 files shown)


agent/lib/kontena/workers/container_log_worker.rb
```diff
@@ -67,7 +68,7 @@ def on_message(id, stream, chunk)
         type: stream,
         data: data
       }
-      @queue << msg
+      @queue << msg if @queue.size < QUEUE_LIMIT
```
jnummelin: As a user I'd really like to see this action taking place in logs. Maybe a warning?

jakolehm: Do we really want to log every time we discard a log item? It will exhaust agent logs, yes?

jnummelin: Definitely not. :)

jakolehm: Maybe we should log only when agent hits this limit.. and then log again when queue can be used? Actually this might be better logic anyway..  we probably should not start to fill queue again if it's almost full.

jnummelin: yeah, sound like a good idea to start processing again when the queue is like 50% full or something.

jnummelin: Would it make sense to move the limit check to beginning of the method to avoid any parsing and matching if the queue is full. The agent has hands full at that point anyway so no need to do unnecessary parsing.

jakolehm: Not sure if that is a good idea.. then it will also skip fluentd shipping. 

I did some refactoring and now LogWorker is responsible of throttling.. it will terminate actual container log workers if queue grows too big, and start them again when queue is processed.

agent/lib/kontena/workers/log_worker.rb
```diff
@@ -51,31 +65,53 @@ def queue_processing?
...
         rescue Docker::Error::NotFoundError
           # Could be thrown since container.skip_logs? actually loads the container details
         end
       end
-
-      async.process_queue
     end
 
     def process_queue
+      current_actor = Actor.current
```
SpComb: This is going to be several (~5?) sync thread-to-actor calls for **each** log line, with all of the associated overhead. The `defer` does not smell good...

agent/lib/kontena/workers/log_worker.rb
```diff
@@ -51,31 +65,53 @@ def queue_processing?
...
+      elsif worker.throttling?
+        info "queue has been almost processed, releasing throttle"
+        worker.release_throttle!
+        worker.start_container_workers
+      else
+        sleep 0.05
+      end
+    end
+
+    def stop_container_workers
```
SpComb: The way this marks the current `Time.now` timestamp into etcd instead of the actual last processed log item timestamp means that we're going to lose any logs that are buffered in the Docker API queues.

agent/lib/kontena/workers/log_worker.rb
```diff
@@ -51,31 +65,53 @@ def queue_processing?
 
     def start
       sleep 1 until Actor[:etcd_launcher].running?
+      start_container_workers
+      async.process_queue
```
SpComb: This still gets called twice:

* `initialize` `@queue_processing = nil` -> `async.start` -> `async.process_queue` -> `defer { while queue_processing? ... }`
* `on_connect` `@queue_processing = true`  -> `async.start` -> `async.process_queue` -> `defer { while queue_processing? ... }`

What's stopping these two `process_queue` `defer { ... }` tasks from racing and having both of them run concurrently?

jakolehm:
Not sure if we need this after #2111 is merged.

SpComb:
Closing in favor of the #2111 `ContainerLogWorker`-based fix, the `LogWorker` needs some other refactoring.

agent/lib/kontena/workers/log_worker.rb
SpComb: Seems to really happen in current master:  #2232

PR #1994, opened 2017-03-16
Backport #1993 to v1.1

jakolehm:
Backports #1993 to v1.1


agent/lib/kontena/workers/container_log_worker.rb
```diff
@@ -66,7 +67,7 @@ def on_message(id, stream, chunk)
               data: data
           }
       }
-      @queue << msg
+      @queue << msg if @queue.size < QUEUE_LIMIT
```
jnummelin: I'd move the `if @queue.size < QUEUE_LIMIT` limit check the first step in the method, no need to parse and match if we're anyway going to ignore the event. So maybe something like:
```
return if @queue.size > QUEUE_LIMIT
```

jakolehm: done

PR #1995, opened 2017-03-16
[WIP] Service Level Metrics

daniel-bytes:
Service level metrics API endpoint.  Returns CPU/Memory/Network data for containers belonging to a service.

### DIFF
--- a/server/app/models/container_stat.rb
+++ b/server/app/models/container_stat.rb
@@ -20,6 +20,7 @@ class ContainerStat
   index({ grid_service_id: 1, created_at: 1 })
 
   def self.get_aggregate_stats_for_service(service_id, from_time, to_time, network_iface)
+
     self.collection.aggregate([
     {
       '$match': {
@@ -113,10 +114,11 @@ class ContainerStat
           tx_errors: '$network_tx_errors'
         }
       }
-    }]).map do |stat|
+    }
+    ]).map do |stat|
       # convert CPU mask to num_cores
-      stat["cpu"]["num_cores"] = (stat["cpu"]["mask"].split('-').last.to_i + 1)
-      stat["cpu"].delete("mask")
+      #stat["cpu"]["num_cores"] = (stat["cpu"]["mask"].split('-').last.to_i + 1)
+      #stat["cpu"].delete("mask")
       stat
     end
   end
--- a/server/app/models/container_stat.rb
+++ b/server/app/models/container_stat.rb
@@ -36,7 +36,7 @@ class ContainerStat
     },
     {
       '$match': {
-        'network.interfaces.name': "eth0"
+        'network.interfaces.name': network_iface
       }
     },
     {
@@ -117,8 +117,8 @@ class ContainerStat
     }
     ]).map do |stat|
       # convert CPU mask to num_cores
-      #stat["cpu"]["num_cores"] = (stat["cpu"]["mask"].split('-').last.to_i + 1)
-      #stat["cpu"].delete("mask")
+      stat["cpu"]["num_cores"] = (stat["cpu"]["mask"].split('-').last.to_i + 1)
+      stat["cpu"].delete("mask")
       stat
     end
   end
--- a/server/app/models/container_stat.rb
+++ b/server/app/models/container_stat.rb
@@ -17,4 +17,107 @@ class ContainerStat
   index({ grid_service_id: 1 })
   index({ container_id: 1 })
   index({ created_at: 1 })
+  index({ grid_service_id: 1, created_at: 1 })
+
+  def self.get_aggregate_stats_for_service(service_id, from_time, to_time, network_iface)
+    self.collection.aggregate([
+    {
+      '$match': {
+        grid_service_id: service_id,
+        created_at: {
+          '$gte': from_time,
+          '$lte': to_time
+        }
+      }
+    },
+    {
+      '$unwind': '$network.interfaces'
+    },
+    {
+      '$match': {
+        'network.interfaces.name': "eth0"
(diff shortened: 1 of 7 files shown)


server/app/services/metrics.rb
```diff
@@ -0,0 +1,35 @@
+class Metrics
+
+  # @param [Container Mongo Collection] containers_collection
```
nevalla: I think the correct class is `Moped::Collection`

server/app/views/v1/grid_stats/stats.json.jbuilder
```diff
@@ -0,0 +1,23 @@
...
+      json.usage stat[:stats].memory['usage']
+      json.limit stat[:stats].spec['memory']['limit']
+    end
+    json.network do
+      json.name stat[:stats].network['name']
+      json.rx_bytes stat[:stats].network['rx_bytes']
+      json.tx_bytes stat[:stats].network['tx_bytes']
+    end
+  else
+    json.cpu nil
```
nevalla: I think we should return empty Hash instead of nil if stats are not available

server/app/views/v1/host_nodes/stats.json.jbuilder
```diff
@@ -0,0 +1,23 @@
...
+      json.usage stat[:stats].memory['usage']
+      json.limit stat[:stats].spec['memory']['limit']
+    end
+    json.network do
+      json.name stat[:stats].network['name']
+      json.rx_bytes stat[:stats].network['rx_bytes']
+      json.tx_bytes stat[:stats].network['tx_bytes']
+    end
+  else
+    json.cpu nil
```
nevalla: I think we should return empty Hash instead of nil if stats are not available

server/app/views/v1/grid_stats/stats.json.jbuilder
daniel-bytes: @nevalla : The CLI actually expects this to be null/missing when calling the `service stats` command.  In the latest version of this PR I am sharing the same view for all stats APIs so I would either have to change this in the CLI or leave it alone.

daniel-bytes:
@SpComb : The changes to network stats gathering on the agent side as mentioned in #1956 are implemented in this PR.  Also adding the `nice` CPU counter.

server/app/views/v1/grid_stats/stats.json.jbuilder
nevalla: Maybe we can go with nil values. @jakolehm any preferences?

jakolehm: I agree, we can go with null/missing.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -245,6 +264,93 @@ def calculate_container_time(container)
...
+        internal: calculate_interface_traffic(prev_interfaces, internal_interfaces, interval_seconds),
+        external: calculate_interface_traffic(prev_interfaces, external_interfaces, interval_seconds)
+      }
+    end
+
+    def calculate_interface_traffic(prev_interfaces, current_interfaces, interval_seconds)
+      results = {
+        interfaces: [],
+        rx_bytes: 0,
+        prev_rx_bytes: 0,
```
SpComb: It's very possible (immediately after a reboot) for e.g. the `weave` interfaces to be missing during `#initialize`, but appear before `publish_node_stats` first runs. OTOH, those newly created interfaces are going to start with their counters at zero, so I guess this all works out.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -245,6 +264,93 @@ def calculate_container_time(container)
...
+          result[:prev_tx_bytes] += prev_iface[0].out_bytes
+        end
+
+        result
+      }
+
+      results[:rx_bytes_per_second] = ((results[:rx_bytes] - results[:prev_rx_bytes]).to_f / interval_seconds).round
+      results[:tx_bytes_per_second] = ((results[:tx_bytes] - results[:prev_tx_bytes]).to_f / interval_seconds).round
+
+      results.delete(:prev_rx_bytes)
```
SpComb: Instead of having a hash that changes form within the same function, just use separate variables and then return a new hash?

```ruby
{
  interfaces: interfaces,
  rx_bytes: total_rx_bytes,
  tx_bytes: total_tx_bytes,
  rx_bytes_per_second: ....
  tx_bytes_per_second: ....
}
```

IMO easier to figure out what's actually getting returned.

agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -155,15 +156,88 @@ def send_statsd_metrics(name, event)
...
+      }
+
+      results[:internal] = calculate_interface_traffic(internal_interfaces, prev_interfaces, interval_seconds)
+      results[:external] = calculate_interface_traffic(external_interfaces, prev_interfaces, interval_seconds)
+
+      results
+    end
+
+
+    def calculate_interface_traffic(current_interfaces, prev_interfaces, interval_seconds)
```
SpComb: This needs to go in some helper module, shared with the ~~`NodeInfoWorker`~~ `NodeStatsWorker`.

daniel-bytes: Agreed.  This and the other refactor fix have been pushed to a new StatsHelper module.

SpComb:
The `server/db/migrations` need renumbering after https://github.com/kontena/kontena/blob/master/server/db/migrations/21_volume_migration.rb

PR #2005, opened 2017-03-20
Mount cAdvisor volumes with rshared

### LINKED ISSUE #2004: Docker netns mounts are leaked to kontena-cadvisor
Repro steps for https://github.com/weaveworks/weave/issues/2842 with Kontena 1.1 and cAdvisor running on CoreOS with Linux 4.9.9, resulting in something very similar to https://github.com/weaveworks/weave/issues/1455

* `kontena service create redis redis`
* `kontena service scale redis 2` # create the redis-1 redis-2 containers
* `kontena node ssh ... docker restart kontena-cadvisor` # this leaks the netns mounts for the two redis containers
* `ping redis-2` # start pinging the weave overlay network IP address assigned to redis-2...
* `kontena service scale redis 1` # destroy the redis-2 container

At this point, the `redis-2` container will be gone from `docker ps -a`, but it will continue to respond to pings on the weave overlay network IP address for up to ~60s. What's happening:

* The `dockerd` runs in some mount namespace, starts some Docker containers, and mounts their network namespaces at `/run/docker/netns/*`
* The `kontena-cadvisor` container process starts up a new mount namespace, and recursively bind-mounts the `/run/docker/netns/*` netns mounts using the default `rprivate` mount propagation mode
* A Docker container gets stopped: all processes in the container exit, and Docker then [umounts](https://github.com/docker/docker/blob/1.12.x/vendor/src/github.com/docker/libnetwork/osl/namespace_linux.go#L401) the `/run/docker/netns/*` mountpoint
* The `/run/docker/netns/*` remains mounted in the `kontena-cadvisor` container's mount namespace, as the bind-mounts are using the `private` propagation mode
* Docker considers the container to be destroyed, but the container's netns remains alive with zero processes running, the `vethwepl*` interfaces remain linked to the `weave` bridge, and the `vethwe` inside the netns still has the overlay network address configured
  * At this point the container becomes a [weave zombie](https://gist.github.com/SpComb/9cc2a7ed46ff708547d09854b78c197f): overlay network address will continue to respond to ping, but there are no more processes listening for TCP connections
* The Docker [libnetwork GC](https://github.com/docker/libnetwork/pull/223) kicks in up to 60s later, and [unlinks](https://github.com/docker/docker/blob/1.12.x/vendor/src/github.com/docker/libnetwork/osl/namespace_linux.go#L103) the `/run/docker/netns/*` file
* On the current CoreOS Linux 4.9 kernel, unlinking the umounted mountpoint within the source namespace causes the Linux kernel to also umount the `/run/docker/net/*` mount within the `kontena-cadvisor` container's mount namespace
* Now that all netns mounts are gone, the kernel cleans up the network namespace, the `vethwepl` is destroyed and unlinked, and the containers stop responding to pings

Per the [cAdvisor README](https://github.com/google/cadvisor#quick-start-running-cadvisor-in-a-docker-container), the Kontena agent [launches](https://github.com/kontena/kontena/blob/v1.1/agent/lib/kontena/launchers/cadvisor.rb#L109-L112) the `kontena-cadvisor` container with  `/` and `/var/run` bind-mounts using Docker's default `rprivate` propagation mode. This means that any Docker containers that are running when the `kontena-cadvisor` container is started will have their `/run/docker/netns/*` mounts leaked into the `kontena-cadvisor` container. When those container are stopped, and Docker umounts `/run/docker/netns/*`, the netns will remain mounted within the `kontena-cdavisor` containers. This does not affect containers that were started after `kontena-cadvisor`, because the `private` mount propagation mode means that those new mounts will not be propagated to the `kontena-cadvisor` container namespace. However, after a node reboot, the `kontena-cadvisor` and any running service containers can be restarted by Docker in any random order.

One workaround for this would be to use `rshared` mount propagation for the `kontena-cadvisor` bind mounts: this means that when the Docker engine umounts the `/run/docker/netns/*` nsfs mount, this umount will immediately be propagated to the `kontena-advisor` container. Testing on CoreOS shows that in this case, the container netns gets destroyed immediately when Docker destroys the container, and there is no GC delay during which the old weave overlay network IP address continues to respond until it gets unlinked.

However, it is unclear how this change to use `rshared` mount propagation for cAdvisor would behave on older Linux kernels before 3.18, such as the RHEL/CentOS 7 Linux 3.10 kernels. It is also unclear why the resulting weave zombies will persist in some cases (https://github.com/weaveworks/weave/issues/2433 https://github.com/weaveworks/weave/issues/2842), whereas these repro steps only result in short-lived (under 1 minute) zombies on CoreOS Linux 4.9.

jakolehm:
Might fix #2004

### DIFF
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -32,8 +32,12 @@ module Kontena::Launchers
     end
 
     def start_cadvisor
-      pull_image(image)
-      create_container(image)
+      if cadvisor_enabled?
+        pull_image(image)
+        create_container(image)
+      else
+        warn "cadvisor is disabled"
+      end
     end
 
     def image
@@ -90,7 +94,7 @@ module Kontena::Launchers
     # @param [Exception] exc
     def log_error(exc)
       error "#{exc.class.name}: #{exc.message}"
-      error exc.backtrace.join("\n")
+      error exc.backtrace.join("\n") if exc.backtrace
     end
 
     # @return [Hash]
@@ -120,6 +124,12 @@ module Kontena::Launchers
       return true if cadvisor.labels['io.kontena.agent.version'].to_s != Kontena::Agent::VERSION
 
       false
+    end
+
+    # @return [Boolean]
+    def cadvisor_enabled?
+      return true unless ENV['CADVISOR_DISABLED']
+      ENV['CADVISOR_DISABLED'] != 'true'
     end
   end
 end
--- a/agent/lib/kontena/launchers/cadvisor.rb
+++ b/agent/lib/kontena/launchers/cadvisor.rb
@@ -106,10 +106,10 @@ module Kontena::Launchers
     # @return [Array<String>]
     def volume_binds
       [
-        '/:/rootfs:ro',
-        '/var/run:/var/run',
-        '/sys:/sys:ro',
-        '/var/lib/docker:/var/lib/docker:ro'
+        '/:/rootfs:ro,rshared',
+        '/var/run:/var/run:rshared',
+        '/sys:/sys:ro,rshared',
+        '/var/lib/docker:/var/lib/docker:ro,rshared'
       ]
     end
 
--- a/agent/spec/lib/kontena/launchers/cadvisor_launcher.rb
+++ b/agent/spec/lib/kontena/launchers/cadvisor_launcher.rb
@@ -29,6 +29,14 @@ describe Kontena::Launchers::Cadvisor do
       subject.start
       sleep 0.01
     end
+
+    it 'does not start if cadvisor is disabled' do
+      allow(ENV).to receive(:[]).with('CADVISOR_DISABLED').and_return('true')
+      expect(subject.wrapped_object).not_to receive(:pull_image)
+      expect(subject.wrapped_object).not_to receive(:create_container)
+      subject.start
+      sleep 0.01
+    end
   end
 
   describe '#volume_mappings' do
@@ -61,6 +69,17 @@ describe Kontena::Launchers::Cadvisor do
         labels: { 'io.kontena.agent.version' => "#{Kontena::Agent::VERSION}-special-edition" }
       )
       expect(subject.config_changed?(container)).to be_truthy
+    end
+  end
+
+  describe '#cadvisor_enabled?' do
+    it 'returns true by default' do
+      expect(subject.cadvisor_enabled?).to be_truthy
+    end
+
+    it 'returns false if disabled by env variable' do
+      allow(ENV).to receive(:[]).with('CADVISOR_DISABLED').and_return('true')
+      expect(subject.cadvisor_enabled?).to be_falsey
     end
   end
 end
--- a/docs/references/environment-variables.md
+++ b/docs/references/environment-variables.md
@@ -28,6 +28,7 @@ toc_order: 2
 - `ETCD_VERSION`: etcd image version
 - `CADVISOR_IMAGE`: cadvisor image (default: kontena/cadvisor)
 - `CADVISOR_VERSION`: cadvisor image version
+- `CADVISOR_DISABLED`: if set to `true` agent won't start cadvisor at all
 - `WEAVE_IMAGE`: weave net image (default: weaveworks/weave)
 - `WEAVEEXEC_IMAGE`: weave exec image (default: weaveworks/weaveexec)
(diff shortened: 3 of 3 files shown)


SpComb:
Breaks on CentOS:

```
[root@centos-1gb-ams2-01 ~]# docker start kontena-cadvisor
Error response from daemon: linux mounts: Path / is mounted on / but it is not a shared mount.
Error: failed to start containers: kontena-cadvisor
```

SpComb:
Using `rshared` for the cadvisor bind-mounts does not seem to behave as expected either. It works in that `/proc/$(docker inspect -f "{{.State.Pid}}" kontena-cadvisor)/mountinfo` shows the `/rootfs` and `/var/run` mounts as being `master:X` of the same peer group as the dockerd's `master:X` and the host's `shared:X` for those mounts.

However, the new docker `/run/docker/netns/*` mounts do not propagate to the cadvisor container, and neither do the umounts after a `kontena-cadvisor` start, so this behaves as the default `rprivate`.

SpComb:
Confirm that the cadvisor `rshared` bind-mounts also appear the fix the zombie netns issue on CentOS 7 with Linux 3.10:

```
[pid 26232] 09:12:47 umount("/var/run/docker/netns/e5cbd6596f4d", MNT_DETACH) = 0
[pid 28035] 09:13:21 unlinkat(AT_FDCWD, "/var/run/docker/netns/e5cbd6596f4d", 0) = 0
```

Both with and without a `kontena-cadvisor` restart.

SpComb:
Re the inability to use `rshared` bind mounts if the Docker engine runs as a `MountFlags=slave` systemd service, it seems like:

* Current RHEL/CentOS Docker packages use `MountFlags=slave`: https://access.redhat.com/articles/2938171
* Current upstream Docker RPM packages do not use `MountFlags`, but they are re-introducing it to resolve devicemapper issues: https://github.com/docker/docker/pull/31490

However, I'm still seeing plenty of devicemapper errors and dead containers when starting a cadvisor container with the default `rprivate` sharing, even with the default distro `MountFlags=slave` configuration. Which makes sense, because the cadvisor container is picking up those mounts from docker's mount namespace, regardless of the systemd `MountFlags=` configuration.

jakolehm:
@SpComb maybe we could introduce env for cadvisor mount flags (and default to rshared)?

SpComb:
Well... applying this PR would improve the total user experience on CentOS by breaking the `kontena-cadvisor` container, and preventing it from running... which would avoid a whole lot of `dead ` Docker container related errors that cause a whole lot more problems than missing stats :smile:

jakolehm:
@SpComb not sure if we need to change anything in here?

SpComb:
I think the design is okay, but perhaps it should behave better when the `rshared` fails:

* Fix the agent to log the cadvisor create errors when Docker is running with `MountFlags=slave`
* Add an env flag to disable cAdvisor on CentOS?
* Docs?

jakolehm:
@SpComb `Cadvisor` launcher already logs failed start attempts.

jakolehm:
Added `CADVISOR_DISABLED` env variable.

agent/lib/kontena/launchers/cadvisor.rb
```diff
@@ -32,8 +32,12 @@ def start
     end
 
     def start_cadvisor
-      pull_image(image)
-      create_container(image)
+      if cadvisor_enabled?
+        pull_image(image)
+        create_container(image)
+      else
+        warn "cadvisor is disabled"
```
SpComb: Ideally this would also stop it if it is running.

SpComb:
> @SpComb `Kontena::Launchers::Cadvisor` already logs failed start attempts.

BTW, it does not log any `Docker::Container#start` errors, because of #2134

#### `docker logs kontena-agent | grep -i cadvisor`
```
I, [2017-04-18T14:29:54.315908 #1]  INFO -- Kontena::Launchers::Cadvisor: initialized
I, [2017-04-18T14:29:54.379262 #1]  INFO -- Kontena::Launchers::Cadvisor: pulling image google/cadvisor:v0.24.1
I, [2017-04-18T14:30:02.352467 #1]  INFO -- Kontena::Launchers::Cadvisor: starting cadvisor service
I, [2017-04-18T14:30:10.593325 #1]  INFO -- Kontena::Workers::StatsWorker: waiting for cadvisor
```

#### `docker inspect -f '{{.State.Error}}' kontena-cadvisor`
```
linux mounts: Path / is mounted on / but it is not a shared mount.
```

PR #2008, opened 2017-03-21
Improve agent RPC request error handling

SpComb:
* Fix the agent RPC client actor to not crash on RPC errors, avoiding unexpected caller task terminations (#2063)

  * Sync requests use `RpcHelper#rpc_request -> RpcClient#request_with_error`, which returns an exception that gets raised locally within the calling actor

  * Any `rpc_client.async.request(...)` errors are now logged as warnings

  * RPC requests now use the `WaitHelper` and log pending requests at the DEBUG level

* Change the server RPC handlers to raise errors instead of returning succesful `{error: ...}` responses

  * Agents can use normal RPC semantics to handle the errors

  * The agent manager actors now rescue all RPC errors (including timeouts), log a warning, and retry later as part of their normal refresh loop

  * Keep some of the response validation as a sanity check to guard against accidential termination of containers/volumes

### DIFF
--- a/agent/lib/kontena/rpc_client.rb
+++ b/agent/lib/kontena/rpc_client.rb
@@ -66,6 +66,23 @@ module Kontena
       end
     end
 
+    # Async request wrapper.
+    #
+    # Logs a warning and returns nil on errors.
+    # Use Kontena::Helpers::RpcError.rpc_request to get a raised error instead.
+    #
+    # @return [Object, nil]
+    def request(method, params, **opts)
+      result, error = request_with_error(method, params, **opts)
+
+      if error
+        warn "RPC request #{method} failed: #{error}"
+        return nil
+      else
+        return result
+      end
+    end
+
     # Called from Kontena::WebsocketClient in the EM thread
     def handle_response(response)
       type, msgid, error, result = response
--- a/agent/lib/kontena/rpc_client.rb
+++ b/agent/lib/kontena/rpc_client.rb
@@ -1,9 +1,12 @@
 require_relative 'logging'
+require_relative 'helpers/wait_helper'
+
 
 module Kontena
   class RpcClient
     include Celluloid
     include Kontena::Logging
+    include Kontena::Helpers::WaitHelper
 
     class Error < StandardError
       attr_accessor :code, :message, :backtrace
@@ -34,26 +37,36 @@ module Kontena
       logger.error exc.message
     end
 
+    # This method should not raise, or the Actor will crash, and terminate any other pending requests.
+    #
     # @param [String] method
     # @param [Array] params
-    # @return [Object]
-    def request(method, params)
+    # @param [Fixnum] timeout seconds
+    # @return [Object, Exception]
+    def request_with_error(method, params, timeout: 30)
       id = request_id
       @requests[id] = nil
-      sleep 0.01 until @client.connected?
+
+      if !wait_until("websocket client is connected", timeout: timeout, interval: 0.01) { @client.connected? }
+        return nil, TimeoutError.new(500, 'WebsocketClient is not connected')
+      end
+
       @client.send_request(id, method, params)
-      time = Time.now.utc
-      until !@requests[id].nil?
-        sleep 0.01
-        raise TimeoutError.new(500, 'Request timed out') if time < (Time.now.utc - 30)
+
+      if !wait_until("request #{id} has response", timeout: timeout, interval: 0.01) { @requests[id] }
+        return nil, TimeoutError.new(500, 'Request timed out')
       end
+
       result, error = @requests.delete(id)
+
       if error
-        raise Error.new(error['code'], error['message'])
+        return result, Error.new(error['code'], error['message'])
+      else
+        return result, nil
       end
-      result
     end
 
+    # Called from Kontena::WebsocketClient in the EM thread
     def handle_response(response)
       type, msgid, error, result = response
       @requests[msgid] = [result, error]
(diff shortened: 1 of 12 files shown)


agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -44,10 +44,7 @@ def populate_workers_from_master
       exclusive {
         request = rpc_client.request("/node_service_pods/list", [node.id])
         response = request.value
-        unless response['service_pods'].is_a?(Array)
-          warn "failed to get list of service pods from master: #{response['error']}"
-          return
-        end
+        raise "Invalid response" unless response['service_pods'].is_a?(Array)
```
jakolehm: Why are we crashing here? This is not a rpc handler so no one is listening for this failure?

server/spec/services/rpc/node_service_pod_handler_spec.rb
```diff
@@ -10,15 +10,14 @@
...
 
-    it 'returns hash with error if id does not exist' do
-      list = subject.list('foo')
-      expect(list[:error]).not_to be_nil
+    it 'fails if node does not exist' do
+      expect{subject.list('foo')}.to raise_error RuntimeError, 'Node not found'
     end
 
-    it 'returns hash with error if migration is not done' do
-      allow(subject).to receive(:migration_done?).and_return(false)
```
SpComb: This spec was broken, it was failing on the `node_id: foo` and never checking `migration_done?`.

SpComb:
@jakolehm if you think the agents managers shouldn't crash on server RPC errors, then I can restore the warn-and-retry-later behavior.

agent/lib/kontena/workers/service_pod_manager.rb
SpComb: It lets the manager actor crash and restart. Are there expected to be any RPC errors that we can recover from in any more useful way?

jakolehm:
@SpComb it just sounds a bit weird to crash if client receives known error from the server?

SpComb:
Which of those errors are known or expected?

Perhaps the manager should still keep the workers alive even if it's is unable to get the list from the server. Current implementation has the workers terminated if the manager crashes.

jakolehm:
Correct if I'm wrong but those responses have `error` field. If we think that this kind of information is unknown or unexpected then it's fine to crash. It just throws actor in a restart loop until it gets valid array.

SpComb:
Current behavior is a bit of a mix... things like RPC timeouts will crash the agent actors, but then things like server internal errors or the host node going missing will have the actor warn, and try again later on the 30s loop.

This PR has the managers crash in all such situations... I'm thinking it might still be a better idea to just rescue and warn on all `RpcServer::Error` cases, including timeouts? The manager's 30s loop is probably more robust ATM than the actor-crash-restart loop?

SpComb:
This PR does have the side effect of making it slightly more likely for the agent `RpcClient` actor to crash on `request` errors, and I'm not entirely sure how safe that is... of course, it's already possible for it to crash on timeout errors.

OTOH it doesn't seem like the existing `rpc_client.future(...).value` makes any difference, the `RpcClient` actor still crashes even if you wrap the call in a future.

SpComb:
See #2063 for an analysis of the `RpcClient` actor crash side effects... it's not safe.

SpComb:
I'll update this to also fix the `RpcClient` actor to be crash-safe.

SpComb:
Changed the `RpcClient` actor to not crash on RPC errors. Any `rpc_client.async.request(...)` errors now get logged as warnings, and then the manager actors use a `rpc_request(...)` helper to raise exceptions locally within the calling actor. Using `WaitHelper` means that pending RPCs are now also logged at DEBUG level.

agent/lib/kontena/rpc_client.rb
```diff
@@ -34,26 +37,53 @@ def notification(method, params)
...
+    # Logs a warning and returns nil on errors.
+    # Use Kontena::Helpers::RpcError.rpc_request to get a raised error instead.
+    #
+    # @return [Object, nil]
+    def request(method, params, **opts)
+      result, error = request_with_error(method, params, **opts)
+
       if error
-        raise Error.new(error['code'], error['message'])
+        warn "RPC request #{method} failed: #{error}"
```
SpComb: This is a bit awkward, but it's still an improvement over how `rpc_client.async.request(...)` behaved before... it would not log any errors.

SpComb:
The `RpcClient` `wait_until(internval: 0.01)` is enough to effectively break agent `LOG_LEVEL=0` debug logging without `io.kontena.container.skip_logs=1`, because it's 100/s DEBUG container-log notifications per pending RPC request..

jakolehm:
Tested. Works really well (although debug logging is borked).

SpComb:
WaitHelper DEBUG logging fixes in #2072

agent/lib/kontena/rpc_client.rb
```diff
@@ -33,37 +38,67 @@ def notification(method, params)
...
+    # Called from Kontena::WebsocketClient in the EM thread
     def handle_response(response)
       type, msgid, error, result = response
       @requests[msgid] = [result, error]
     end
 
     # @return [Fixnum]
     def request_id
-      id = -1
-      until id != -1 && !@requests[id]
```
SpComb: BTW, I think this was broken... pending requests would have `@requests[id] = nil`, which would pass this check.

PR #2013, opened 2017-03-22
Remove grid token from Grid and HostNode JSON

nevalla:
This PR will remove grid token from Grid and HostNode JSON responses. Instead there will be new `/v1/grids/:grid/token` endpoint from where token can be requested (only for users with `master_admin` and `grid_admin` role). This way grid token is not leaked among other grid information.

### DIFF
--- a/server/app/routes/v1/grids_api.rb
+++ b/server/app/routes/v1/grids_api.rb
@@ -104,6 +104,11 @@ module V1
             render('grids/show')
           end
 
+          r.on 'token' do
+            halt_request(403, {error: 'Forbidden'}) unless current_user.can_update?(@grid)
+            render('grids/token')
+          end
+
           r.on 'container_logs' do
             scope = @grid.container_logs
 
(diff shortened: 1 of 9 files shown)


jnummelin:
Just a reminder that this will break the provisioning plugins :)

nevalla:
That's a valid point. When updating plugins we have to remember that there might be case when user is not allowed to fetch token, and on those cases node provision should fail with the proper message.

Also `kontena grid cloud-config` need also to be updated.

server/app/routes/v1/grids_api.rb
```diff
@@ -104,6 +104,11 @@ def load_grid(name)
             render('grids/show')
           end
 
+          r.on 'token' do
+            halt_request(403, {error: 'Forbidden'}) unless current_user.can_update?(@grid)
```
SpComb: :+1: for this access restriction, doing the same in #2504

server/app/views/v1/grids/_grid.json.jbuilder
```diff
@@ -1,6 +1,5 @@
 json.id grid.to_path
 json.name grid.name
-json.token grid.token
```
SpComb: :-1: can't break the existing API in the same PR, this needs to be done much later, once all the plugins have been updated.

PR #2014, opened 2017-03-22
Fix missing server Rpc::GridSerializer fields

### LINKED ISSUE #2012: #1873 server Rpc::GridSerializer is missing stats, logs fields
These are needed by the agent `StatsWorker` and `FluentdWorker`

SpComb:
Fixes #2012

### DIFF
--- a/server/spec/services/agent/node_plugger_spec.rb
+++ b/server/spec/services/agent/node_plugger_spec.rb
@@ -34,41 +34,13 @@ describe Agent::NodePlugger do
   describe '#send_node_info' do
     it "sends node info" do
       expect(rpc_client).to receive(:notify).with('/agent/node_info', hash_including(
-        id: 'xyz',
         name: 'test-node',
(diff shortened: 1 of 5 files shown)


server/spec/services/agent/node_plugger_spec.rb
```diff
@@ -1,31 +1,74 @@
...
+          id: 'test-grid',
+          name: 'test-grid',
+          stats: { statsd: nil },
+          logs: nil,
+        ),
+      ))
+      subject.send_node_info
+    end
+
+    context "for a grid with stats and logs" do
```
jakolehm: I think it would be better to move these tests to Rpc::GridSerializer spec, like this: https://gist.github.com/jakolehm/7516eab27dfcf69de2545770135cf39c

SpComb: Improved the RPC serializer specs.

PR #2018, opened 2017-03-24
Move management of container health checks into service pod worker

jakolehm:
Because now `ServicePodWorker` manages the desired state of containers it makes sense to move management of health checks to the same place.


agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -37,6 +39,10 @@ def ensure_desired_state
       exclusive {
         debug "state of #{service_pod.name}: #{service_pod.desired_state}"
         service_container = get_container(service_pod.service_id, service_pod.instance_number)
+        if service_container && !service_pod.running?
```
jnummelin: These conditions only apply to healthcheck, so maybe it would make sense to do them in `stop_health_check`?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -59,6 +65,10 @@ def ensure_desired_state
         else
           warn "unknown state #{service_pod.desired_state} for #{service_pod.name}"
         end
+
+        if service_pod.running? && service_container
```
jnummelin: These conditions only apply to healthcheck, so maybe it would make sense to do them in `ensure_health_check`?

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Maybe, what does @SpComb think about this?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -137,5 +147,30 @@ def sync_state_to_master(current_state)
...
+
+    # @param [Docker::Container] service_container
+    def ensure_health_check(service_container)
+      return if service_container.nil? || service_container.labels['io.kontena.health_check.uri'].nil?
+      return if @health_check && @health_check.alive?
+
+      @health_check = ContainerHealthCheckWorker.new(service_container)
+      @health_check.async.start
+    rescue => exc
+      error "#{exc.class.name}: #{exc.message}"
```
SpComb: Looking at the `ContainerHealthCheckWorker#initialize` code... it seems like it has a 0.001% probability of crashing.

Unnecessary rescue-and-log-without-giving-any-context-or-backtrace that hurts more than it helps?

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: The `HealthCheckWorker` could theoretically be running even if the `service_container` is nil? Unlikely, but this would be more robust if you leave off the `service_container` chckec...

SpComb: I think it makes sense to keep the `stop_health_check if !service_pod.running?` and `ensure_health_check if service_pod.running?` logic here.

Reason being: if we are stopping a container, then the health check worker must be terminated before we do that.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -137,5 +147,30 @@ def sync_state_to_master(current_state)
...
+
+    def stop_health_check
+      if @health_check
+        @health_check.terminate if @health_check.alive?
+      end
+    end
+
+    private
+
+    def finalize
```
SpComb: Oh right, this is a Celluloid finalizer, not a Ruby finalizer: https://github.com/celluloid/celluloid/wiki/Finalizers

Stil... linking instead?

SpComb:
Still :+1: for this refactoring, but this PR is AFAIK very broken, and definitely not something for 1.3.X

SpComb:
I assume this has been replaced by #2691

PR #2019, opened 2017-03-24
Validate hook names in kontena.yml

nevalla:


cli/lib/kontena/cli/stacks/yaml/custom_validators/hooks_validator.rb
```diff
@@ -6,10 +6,16 @@ def initialize
 
     def validate(key, value, validations, errors)
       unless value.is_a?(Hash)
-        errors[key] = 'hooks must be array'
+        errors[key] = 'hooks must be hash'
```
kke: I think the yaml terminology is "mapping" = hash, "list" = array

nevalla: Just looked what is used elsewhere. But agree on that.

nevalla: I suggest that we will replace all in once in separate PR.

kke: The volume PR has several uses of "mapping". 

```ruby
lib/kontena/cli/stacks/yaml/custom_validators/build_validator.rb
9:        errors[key] = 'build must be string or hash'

lib/kontena/cli/stacks/yaml/custom_validators/extends_validator.rb
9:        errors[key] = 'extends must be string or hash'
```

I think these are the only ones talking about "hash". I don't know what HashValidator says on a type error, probably not mapping.

kke: also most of the errors seem to say something like `must be string` , should be `must be a string`.

nevalla: I have updated error message to use 'mapping' instead of 'hash'

PR #2020, opened 2017-03-24
Volume instance scheduling

jnummelin:
follows #1849


server/app/services/volume_instance_schduler.rb
```diff
@@ -0,0 +1,13 @@
+class VolumeInstanceScheduler
```
jakolehm: `volume_instance_schduler.rb` -> `volume_instance_scheduler.rb`

server/app/models/host_node.rb
```diff
@@ -37,6 +37,7 @@ class Error < StandardError
   has_many :containers
   has_many :container_stats
   has_many :host_node_stats
+  has_many :volume_instances
```
jakolehm: Why not destroy `volume_instances` when node is removed?

server/app/models/volume_instance.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  field :name, type: String
+
+  belongs_to :host_node
+
+  belongs_to :volume
+
+  validates_uniqueness_of :name, scope: [:volume]
+
+  index({ host_node_id: 1 })
```
jakolehm: Should we index also `volume_id`?

server/app/models/host_node.rb
jnummelin: Yep, should be removed when node is removed. I'll add the dependency removal

server/app/models/volume_instance.rb
jnummelin: Yes, I'll add that

server/app/services/grid_service_instance_deployer.rb
```diff
@@ -82,4 +84,13 @@ def notify_node(node)
     rpc_client = RpcClient.new(node.node_id, 2)
     rpc_client.request('/service_pods/notify_update', [])
   end
+
+  def ensure_volume_instance(node, instance_number)
+    # TODO Maybe separate this functionality into own VolumeScheduler or such
+    self.grid_service.service_volumes.each do |sv|
+      if sv.volume
+        VolumeInstanceScheduler.new.schedule(node, sv, instance_number)
```
SpComb: This sounds more like `VolumeInstanceDeployer.new(sv, instance_number).deploy(node)`... the scheduling has already happened.

server/app/models/volume_instance.rb
```diff
@@ -0,0 +1,14 @@
...
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :name, type: String
+
+  belongs_to :host_node
+
+  belongs_to :volume
+
+  validates_uniqueness_of :name, scope: [:volume]
```
SpComb: Prevents deploying multiple service instances using a `scope: stack` volume:

```
E, [2017-03-30T18:20:51.887433 #4780] ERROR -- GridServiceInstanceDeployer: failed to deploy service instance development/null/volumes-test1-2 to node core-02
E, [2017-03-30T18:20:51.887504 #4780] ERROR -- GridServiceInstanceDeployer: 
Problem:
  Validation of VolumeInstance failed.
Summary:
  The following errors were found: Name is already taken
Resolution:
  Try persisting the document with valid data or remove the validations.
E, [2017-03-30T18:20:51.887638 #4780] ERROR -- GridServiceInstanceDeployer: /home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/persistable.rb:78:in `fail_due_to_validation!'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/persistable/creatable.rb:180:in `block in create!'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/threaded/lifecycle.rb:156:in `_creating'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/persistable/creatable.rb:175:in `create!'
/home/kontena/kontena/kontena/server/app/services/volume_instance_scheduler.rb:7:in `schedule'
/home/kontena/kontena/kontena/server/app/services/grid_service_instance_deployer.rb:92:in `block in ensure_volume_instance'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/proxy.rb:150:in `each'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/proxy.rb:150:in `method_missing'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mongoid-4.0.2/lib/mongoid/relations/embedded/many.rb:404:in `method_missing'
/home/kontena/kontena/kontena/server/app/services/grid_service_instance_deployer.rb:90:in `ensure_volume_instance'
/home/kontena/kontena/kontena/server/app/services/grid_service_instance_deployer.rb:16:in `deploy'
/home/kontena/kontena/kontena/server/app/services/grid_service_deployer.rb:100:in `block in deploy_service_instance'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/future.rb:18:in `block in new'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/thread_handle.rb:14:in `block in initialize'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:78:in `block in get_thread'
/home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/group/spawner.rb:50:in `block in instantiate'
E, [2017-03-30T18:20:51.887967 #4780] ERROR -- GridServiceDeployer: halting deploy of development/null/volumes-test1, one or more instances failed
```

SpComb:
BTW have to be careful with the `HostNode.plugins`... with Docker 1.12, `docker info` only lists the volume plugins after you run `docker volume ls`, so it depends which of `VolumeManager` or `NodeInfoWorker` is faster... it's probably going to be the `VolumeManager`, so it works.

SpComb:
If we want to support migration for existing Docker volumes, then we would need something to pick up those existing Docker volumes even if they don't have a `io.kontena.volume.id` label. Dunno how that would happen, though :confused:

SpComb:
I would consider adding some relation between `ServiceInstance` and `VolumeInstance`... ATM it's very difficult to tell if a `VolumeInstance` is still in use or not. ATM the `VolumeInstance`s of a `scope: instance` `Volume` are not destroyed if the service using them is scaled down, and there is no straightforward way to do that later on either.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -31,7 +31,11 @@ def perform
           data_container = self.ensure_data_container(service_pod)
           service_pod.volumes_from << data_container.id
         end
-        ensure_volumes(service_pod)
+
+        wait!(timeout: 30, interval: 1, message: "waiting for volumes to be created") {
```
jnummelin: rebase burns again...

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -0,0 +1,115 @@
...
+        response = request.value
+        unless response['volumes'].is_a?(Array)
+          warn "failed to get list of volumes from master: #{response['error']}"
+          return
+        end
+        debug "got volumes from master: #{response}"
+
+        volumes = response['volumes']
+
+        terminate_volumes(volumes.map {|v| v.dig('labels','io.kontena.volume.id')})
```
jnummelin: having a `volume.id` in the model makes sense. Although I couldn't figure out a case where these would actually be nil.

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -0,0 +1,115 @@
...
+
+        volumes.each do |s|
+          ensure_volume(Kontena::Models::Volume.new(s))
+        end
+      }
+    end
+
+    def populate_volumes_from_docker
+      info "syncing volumes from docker"
+      Docker::Volume.all.each do |volume|
```
jnummelin: I was thinking that this _might_ allow us to map some of the volumes to migrated grid scoped volumes. Not entirely sure yet if that's even possible/feasible.

server/app/services/grid_service_instance_deployer.rb
jnummelin: 👍

jnummelin:
> BTW have to be careful with the HostNode.plugins... with Docker 1.12, docker info only lists the volume plugins after you run docker volume ls, so it depends which of VolumeManager or NodeInfoWorker is faster... it's probably going to be the VolumeManager, so it works.

Would it make sense for `VolumeManager` to do dummy `Docker::Volume.all` when it starts and just ignore the list at that point? It doesn't make it bullet proof but would increase the probability of `NodeInfoWorker` getting volume plugins.

server/app/serializers/rpc/volume_serializer.rb
```diff
@@ -9,12 +9,11 @@ def initialize(volume_instance)
 
     def to_hash
```
SpComb: With the `RpcSerializer` these should just be `attribute ...` lines, see the `HostNodeSerializer` for an example. The `ServicePodSerializer` doesn't use that for some reason.

server/docker-compose.yml
```diff
@@ -10,8 +10,13 @@ api:
     - VAULT_IV=kfsdfopweu09fewjlhfksdjifjpoweop
     - ACME_ENDPOINT=https://acme-staging.api.letsencrypt.org/
     - INITIAL_ADMIN_CODE=initialadmincode
+    - DEBUG=1
```
SpComb: Debugging?

agent/docker-compose.yml
```diff
@@ -8,3 +8,5 @@ agent:
     - KONTENA_PEER_INTERFACE=eth1
   volumes:
     - /var/run/docker.sock:/var/run/docker.sock
+  labels:
+    - io.kontena.container.skip_logs=1
```
SpComb: Accidental? Although TBH I wouldn't mind this..

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -102,9 +102,9 @@ def volume_exist?(volume_name)
 
     def terminate_volumes(current_ids)
       Docker::Volume.all.each do |volume|
```
SpComb: Bonus points for putting a `fail if current_ids.any? {|id| id.nil?}` here... my level of paranoia just says that it's too easy to end up with `nil`s somewhere by accident, and in this kind of place they would cause a lot of damage.

server/docker-compose.yml
jnummelin: It's handy to see debug logs during development. And I figure no-one uses these compose files in any other env than dev purposes.

agent/docker-compose.yml
jnummelin: No, without this it creates a perpetual motion machine

SpComb: Yes, you need this if you do any debug logging while sending container logs.... so, having this label during development lets you do that.

However, there's a problem: if we start logging more debug messages, and you then enable agent `DEBUG` in production, without this label... it's going to explode :bomb:

jnummelin: I can take it out from this PR

agent/lib/kontena/workers/volumes/volume_manager.rb
SpComb: Perhaps have an explicit `kontena volume import volume --node core-01 --name docker-volume-name testvol` command that pre-creates a `VolumeInstance`?

server/app/services/scheduler/filter/volume_instance.rb
```diff
@@ -0,0 +1,49 @@
...
+        service.service_volumes.each do |sv|
+          if sv.volume
+            case sv.volume.scope
+            when 'instance'
+              # Try to find a node which has the volume already for the instance
+              filtered_nodes = filtered_nodes.select { |node|
+                node.volume_instances.where(name: sv.volume.name_for_service(service, instance_number)).exists?
+              }
+            when 'stack'
+              # Stack scoped volumes can be created per node
```
SpComb: Kinda questioning this assumption, it depends on the driver. This prevents using `driver: local` volumes with multiple services by scheduling all of those onto the same node.

We can fix this post-merge.

PR #2022, opened 2017-03-27
Store docker engine plugin information

jakolehm:
This PR implements storage of node (docker engine) plugin information. Volume plugin information can be used to improve volume scheduling.


SpComb:
Is this `status/merge`?

PR #2033, opened 2017-03-30
Add configurable stop timeout for services

### LINKED ISSUE #1334: Container stop/restart timeout is not configurable
`kontena service stop`, `restart` (and others like `scale` ?) use docker defaults (10s) to wait for process to terminate -- there are valid cases where services need more than 10s to exit.

`SIGTERM` --> 10s --> `SIGKILL`

jnummelin:
Agent used hard coded 10sec timeout in every container stop request. 10sec might not be enough for all services so this PR adds support for configurable timeout for services.
```
stack: test/redis-timeout
version: 0.1.0
services:
  redis:
    image: redis:3-alpine
    stop_grace_period: 20s
```

The term `stop_grace_period` is used in Docker compose so it's used also on this so things are extensible.

There's still a default 10s timeout used if no other is specified in services.

On plain service commands there's an option on this also:
```
--stop-timeout STOP_TIMEOUT   Timeout (duration) to stop a container
```
I used `stop-timeout` as that's used in Docker CLI.

fixes #1334

### TODO

- [x] API docs
- [x] docs

### DIFF
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -221,7 +221,7 @@ module Docker
     #
     # @return [Integer]
     def stop_grace_period
-      (self.labels['io.kontena.container.stop_grace_period'] || 10).to_i
+      (self.labels['io.kontena.container.stop_grace_period'] || 10).to_i
     end
 
     private
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -216,6 +216,14 @@ module Docker
       self.overlay_cidr.split('/')[1] if self.overlay_cidr
     end
 
+    # How long to wait when attempting to stop a container if it doesn’t handle
+    # SIGTERM (or whatever stop signal has been specified with stop_signal), before sending SIGKILL
+    #
+    # @return [Integer]
+    def stop_grace_period
+      self.labels['io.kontena.container.stop_grace_period'] ? self.labels['io.kontena.container.stop_grace_period'].to_i : 10
+    end
+
     private
 
     # @return [Hash]
--- a/agent/lib/docker/container.rb
+++ b/agent/lib/docker/container.rb
@@ -221,7 +221,7 @@ module Docker
     #
     # @return [Integer]
     def stop_grace_period
-      self.labels['io.kontena.container.stop_grace_period'] ? self.labels['io.kontena.container.stop_grace_period'].to_i : 10
+      (self.labels['io.kontena.container.stop_grace_period'] || 10).to_i
     end
 
     private
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -153,7 +153,7 @@ module Kontena
 
       # @param [Docker::Container] container
       def cleanup_container(container)
-        container.stop('timeout' => 10)
+        container.stop('timeout' => container.stop_grace_period)
         container.wait
         container.delete(v: true)
       end
--- a/agent/lib/kontena/service_pods/starter.rb
+++ b/agent/lib/kontena/service_pods/starter.rb
@@ -26,7 +26,7 @@ module Kontena
             self.service_id, self.instance_number,
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
-          service_container.restart('timeout' => service_container.stop_grace_period)
+          service_container.restart!('timeout' => service_container.stop_grace_period)
           
         end
 
--- a/agent/lib/kontena/service_pods/starter.rb
+++ b/agent/lib/kontena/service_pods/starter.rb
@@ -27,7 +27,10 @@ module Kontena
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
           service_container.restart!('timeout' => service_container.stop_grace_period)
-          
+          log_service_pod_event(
+            self.service_id, self.instance_number,		
+            "service:start_instance", "service instance #{service_container.name_for_humans} started successfully"		
+          )
         end
 
         Celluloid::Notifications.publish('service_pod:start', service_container)
--- a/agent/lib/kontena/service_pods/starter.rb
+++ b/agent/lib/kontena/service_pods/starter.rb
@@ -26,11 +26,8 @@ module Kontena
             self.service_id, self.instance_number,
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
-          service_container.restart!('timeout' => 10)
-          log_service_pod_event(
-            self.service_id, self.instance_number,
-            "service:start_instance", "service instance #{service_container.name_for_humans} started successfully"
-          )
+          service_container.restart('timeout' => service_container.stop_grace_period)
+          
         end
 
         Celluloid::Notifications.publish('service_pod:start', service_container)
--- a/agent/lib/kontena/service_pods/stopper.rb
+++ b/agent/lib/kontena/service_pods/stopper.rb
@@ -22,7 +22,7 @@ module Kontena
         service_container = get_container(self.service_id, self.instance_number)
         if service_container.running?
           info "stopping service: #{service_container.name_for_humans}"
-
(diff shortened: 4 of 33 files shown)


agent/lib/kontena/service_pods/starter.rb
```diff
@@ -26,11 +26,8 @@ def perform
             self.service_id, self.instance_number,
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
-          service_container.restart!('timeout' => 10)
-          log_service_pod_event(
```
SpComb: Lost the `log_service_pod_event`

agent/lib/kontena/service_pods/stopper.rb
```diff
@@ -22,16 +22,11 @@ def perform
...
         if service_container.running?
           info "stopping service: #{service_container.name_for_humans}"
-
+          service_container.stop!('timeout' => service_container.stop_grace_period)
           log_service_pod_event(
             self.service_id, self.instance_number,
             "service:stop_instance", "stopping service instance #{service_container.name_for_humans}"
           )
-          service_container.stop!('timeout' => 10)
-          log_service_pod_event(
```
SpComb: Lost the `log_service_pod_event`, and moved the `stop!` into the wrong order.

server/spec/helpers/duration_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  it 'parses seconds' do
+    expect(subject.parse_duration('93s')).to eq(93)
+  end
+
+  it 'parses unknown units to zero' do
+    expect(subject.parse_duration('1m93x')).to eq(60)
+  end
+
+  it 'parses unknown format to zero' do
```
SpComb: Is this a good idea? How does `Docker::Container.stop('timeout' =>0)` behave? Should this fail instead, and cause a validation error?

agent/lib/docker/container.rb
```diff
@@ -216,6 +216,14 @@ def overlay_suffix
       self.overlay_cidr.split('/')[1] if self.overlay_cidr
     end
 
+    # How long to wait when attempting to stop a container if it doesn’t handle
+    # SIGTERM (or whatever stop signal has been specified with stop_signal), before sending SIGKILL
+    #
+    # @return [Integer]
+    def stop_grace_period
+      self.labels['io.kontena.container.stop_grace_period'] ? self.labels['io.kontena.container.stop_grace_period'].to_i : 10
```
SpComb: Just `self.labels[...] || 10`

agent/lib/kontena/service_pods/stopper.rb
jnummelin: ugh, kdiff3 resolved most of the conflicts and seems I didn't pay enough attention :(

server/spec/helpers/duration_spec.rb
jnummelin: Docker seems to send `SIGKILL` either directly without `SIGTERM` or they just happen so close so that I do not get my test app to even log the `SIGTERM`. If I'd have to bet I'd bet for the latter one.

But yes, seems a bit dangerous to have it as zero.

jnummelin: Then again, this should not really happen as the `GridServices::Create` and `GridServices::Update` mutations will validate using the same pattern as is used to break the string into units.

agent/lib/docker/container.rb
jnummelin: there's the `to_i` "cast" which would create `0` timeout for all previously existing containers if we used `self.labels[...] || 10`

SpComb: Right, so this is okay. Although it could also be `(self.labels[...] || 10).to_i`

server/app/mutations/grid_services/common.rb
```diff
@@ -261,6 +261,7 @@ def common_validations
               integer :initial_delay, default: 10
             end
           end
+          string :stop_grace_period, matches: Duration::PATTERN
```
SpComb: Should the pattern be anchored with `/\A...\z/`? If this is supposed to support things like `1h30m`, then the pattern would also need to repeat with `/\A(...)+\z/`. This will now accept a string that also contains other gibberish, which will get ignored.

If the user enters something like `5 m 30s`, then it will accept that input, but use a 30s timeout. Configuration typos like that should lead to errors, not unexpected behavior.

server/spec/helpers/duration_spec.rb
SpComb: Agreed, using a regexp for validation should avoid this.

SpComb:
The `stop_grace_period` is missing from `kontena service show`.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -153,7 +153,7 @@ def volumes_exist?(service_pod)
 
       # @param [Docker::Container] container
       def cleanup_container(container)
-        container.stop('timeout' => 10)
+        container.stop('timeout' => container.stop_grace_period)
```
SpComb: Would it be better to use the `@service_pod.stop_grace_period` here, so changes to the service `stop_grace_period` would already apply on the immediately following deploy?

jnummelin: IMO no. One might be deploying newer version of an app that handles `SIGTERM` differently than the existing container and then stopping it with different timeout might create some issues.

jnummelin:
> The stop_grace_period is missing from kontena service show.

Fixed and added E2E specs for this :)

jakolehm:
I think we should add follow-up issue/pr that unifies attributes that can use this duration syntax.

jnummelin:
> I think we should add follow-up issue/pr that unifies attributes that can use this duration syntax.

You mean things like healthchecks etc.?

agent/lib/docker/container.rb
```diff
@@ -216,6 +216,14 @@ def overlay_suffix
       self.overlay_cidr.split('/')[1] if self.overlay_cidr
     end
 
+    # How long to wait when attempting to stop a container if it doesn’t handle
+    # SIGTERM (or whatever stop signal has been specified with stop_signal), before sending SIGKILL
+    #
+    # @return [Integer]
+    def stop_grace_period
+      (self.labels['io.kontena.container.stop_grace_period'] || 10).to_i
```
SpComb: Ctrl-space strikes again

```
W, [2017-06-28T10:25:00.735821 #1]  WARN -- Kontena::Workers::ServicePodWorker: failed to sync redis-2 at : undefined local variable or method ` 10' for #<Docker::Container:0x005578f041aa00>
W, [2017-06-28T10:25:00.738359 #1]  WARN -- Kontena::Workers::ServicePodWorker: undefined local variable or method ` 10' for #<Docker::Container:0x005578f041aa00> (NameError)
/app/lib/docker/container.rb:224:in `stop_grace_period'
/app/lib/kontena/service_pods/terminator.rb:30:in `perform'
/app/lib/kontena/workers/service_pod_worker.rb:159:in `ensure_terminated'
/app/lib/kontena/workers/service_pod_worker.rb:109:in `ensure_desired_state'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
/app/lib/kontena/workers/service_pod_worker.rb:70:in `destroy'
```

```
000015e0  5f 67 72 61 63 65 5f 70  65 72 69 6f 64 27 5d 20  |_grace_period'] |
000015f0  7c 7c c2 a0 31 30 29 2e  74 6f 5f 69 0a 20 20 20  |||..10).to_i.   |
```

SpComb: Looks like the agent is missing specs for this method.

SpComb:
:white_check_mark: tested to work with the whitespace fix: bash script that traps SIGINT/TERM takes `stop_grace_period` to stop

Issue #2024, opened 2017-03-27
Zombie Weave containers remain undead after the Docker container is destroyed


SpComb:
Because Docker does not know anything about the container Weave networks, the container `vethwepl <-> ethwe` veth pairs are not managed by Docker as part of the container lifecycle. Instead, when the last reference to the container's network namespace is released, the kernel will clean up the netns, the Weave veth pair will be destroyed, and the container's overlay network address can safely be re-allocated.

However, if any reference to the Docker container's netns is leaked outside of the Docker daemon itself, then the container's network namespace will not be destroyed by the kernel after Docker destroys the container. This leaves zombie containers connected to the Weave network, responding to pings even after the Docker container has been destroyed :ghost: 

## Symptoms
This leads to mysterious network connectivity issues when the zombie container's overlay network address is re-allocated for a different container:

* https://github.com/weaveworks/weave/issues/2433
* https://github.com/weaveworks/weave/issues/2842

The zombie network namespaces will not have any processes running or accepting connections, but the container's veth network interface will still exist within the kernel, responding to ARP queries and ICMP pings. Connections to the service container will randomly fail, as they may get routed to the zombie containers instead.  

## Known causes
There are at least two known cases that lead to leaked container network namespaces:

* https://github.com/kontena/kontena/issues/2004 via `kontena-cadvisor` bind-mounts using `rprivate` sharing
* https://github.com/docker/docker/issues/32090 via `docker-proxy` netlink sockets from `fork`

## Fixes
Fixing this issue is a combination of fixing the netns leaks, and then countermeasures to mitigate any future netns leaks:

* https://github.com/kontena/kontena/pull/2005 fix `kontena-cadvisor` to use `rshared` to propagate umounts
* https://github.com/kontena/kontena-ipam/pull/74 have `kontena-ipam` not release addresses that are still alive, preventing them from being reallocated until they stop responding and are cleaned up

SpComb:
Tracking the `kontena-ipam` workaround to hold on to zombified addresses:

* #2030 Implemented in 1.2-dev

jalberto:
Can this be causing network collisions?

SpComb:
Tracking the Docker fix for leaking netlink sockets to `docker-proxy` processes: https://github.com/docker/docker/issues/32090

* https://github.com/vishvananda/netlink/pull/209 Fixed in `github.com/vishvananda/netlink`
[...]

PR #2036, opened 2017-03-31
Release v1.1.6

kke:
* bump ipam to 0.2.2
* handle 409 zombie reponses in IpamClient
* let IpamClient raise error in case of 409 and weave actor rescue it in release


SpComb:
This needs some good release notes:

---

Workaround Weave network issues caused by zombie containers holding on to Weave overlay network addresses after the Docker container is destroyed (#2024). The Kontena IPAM will now hold on to addresses that continue to respond to pings after the container is destroyed, preventing their re-use.

Symptoms of these Weave network issues involve connection refused errors for clients connecting to newly deployed service instances, even though the service instance may respond to ping, and connections may also work from other nodes: https://github.com/weaveworks/weave/issues/2842

Upgrading to Kontena 1.1.6 will prevent any new zombies from being formed, but will not fix any existing zombie addresses that have been re-used. If you see such symptoms, then you must first upgrade to Kontena 1.1.6, and then clear out any zombie containers by either rebooting each host node, or restarting each Docker container that may have leaked netns references:
* https://github.com/kontena/kontena/issues/2004 The `kontena-cadvisor` container on each node
* https://github.com/docker/docker/issues/32090 Any service containers that publish ports and have `docker-proxy` processes running, such as any `kontena/vpn` and `kontena/lb` services.

SpComb:
And maybe also:

---

After upgrading, you may see warnings in the `kontena-agent` logs, which indicate that the workaround is working as expected:

```
WARN -- Kontena::NetworkAdapters::Weave: Failed to release container=9c9fd6f5d6d6cb7effd003af15355846bf196b483a906a1e74ae70d6ef69a257 from network=kontena at cidr=10.81.128.44/16: Skip zombie address=10.81.128.44 in pool=kontena that still responds to ping
```

PR #2045, opened 2017-04-01
Added prompt to commands that wait for input from STDIN

colinmollenhour:
The import command only works with a single value and it doesn't even always work then, it just spins forever. The update command also was broken due to the change in handling parameters. Also the design is confusing because if the user doesn't provide a value or somehow accidentally or intentionally provides an empty value it will just hang waiting for stdin input but this is not at all obvious. This change adds a -i or --stdin option to explicitly indicate that the value should be read from stdin so that if it is blank and this is not specified then it will provide a useful error message. This coincidentally fixes the `vault import` command when used with more than one value.

### DIFF
--- a/cli/lib/kontena/cli/common.rb
+++ b/cli/lib/kontena/cli/common.rb
@@ -22,6 +22,9 @@ module Kontena
         @pastel ||= Pastel.new(enabled: $stdout.tty?)
       end
 
+      # Read from STDIN. If stdin is a console, use prompt to ask.
+      # @param [String] message
+      # @param [Symbol] mode (prompt method: :ask, :multiline, etc)
       def stdin_input(message = nil, mode = :ask)
         if $stdin.tty?
           Array(prompt.send(mode, message)).join.chomp
--- a/cli/lib/kontena/cli/common.rb
+++ b/cli/lib/kontena/cli/common.rb
@@ -22,9 +22,9 @@ module Kontena
         @pastel ||= Pastel.new(enabled: $stdout.tty?)
       end
 
-      def stdin_input(message = nil)
+      def stdin_input(message = nil, mode = :ask)
         if $stdin.tty?
-          prompt.multiline(message).join.chomp
+          Array(prompt.send(mode, message)).join.chomp
         elsif !$stdin.eof?
           $stdin.read.chomp
         else
--- a/cli/lib/kontena/cli/common.rb
+++ b/cli/lib/kontena/cli/common.rb
@@ -22,6 +22,16 @@ module Kontena
         @pastel ||= Pastel.new(enabled: $stdout.tty?)
       end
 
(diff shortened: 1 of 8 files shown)


kke:
`$stdin.tty?` is false when some kind of piping is going on so the option is not needed.

I modified the PR to add a common `Kontena.stdinput` method.

cli/lib/kontena/cli/vault/update_command.rb
```diff
@@ -7,13 +7,18 @@ class UpdateCommand < Kontena::Command
 
     option ['-u', '--upsert'], :flag, 'Create secret unless already exists', default: false
     option '--silent', :flag, "Reduce output verbosity"
+    option ['-i', '--stdin'], :flag, 'Read value from stdin', default: false
```
kke: $stdin.tty? is false when some kind of piping is going on so the option is not needed.

kke:
image

Seems to work as expected.

kke:
ping @SpComb

cli/lib/kontena/cli/common.rb
```diff
@@ -22,6 +22,16 @@ def pastel
         @pastel ||= Pastel.new(enabled: $stdout.tty?)
       end
 
+      def stdin_input(message = nil)
+        if $stdin.tty?
+          prompt.multiline(message).join.chomp
```
SpComb: The `prompt.multiline` is a bit odd for things like `kontena vault write test`.... just writing something and hitting enter does not accept the value, you have to hit ^D.

Should this take some options to use different prompts depending on what is being prompted? Looking around in the [docs](https://github.com/piotrmurach/tty-prompt#23-multiline)...

* `prompt.multiline` for `kontena master config import` and `kontena vault import`?
* `prompt.mask` for `kontena vault write/update`?
* `prompt.ask` otherwise

SpComb:
@colinmollenhour we had accumulated three different PRs that all tried to fix the same thing, so we ended up merging all of them into this one... least painful that way.

SpComb:
Hangs in CLI rspecs:

```
Kontena::Cli::Vault::WriteCommand
  #execute
    sends create request
    without value parameter
      with tty
```

```ruby
      context 'with tty' do
        it 'prompts value from STDIN' do
          expect(stdin).to receive(:tty?).and_return(true)
          expect(subject.prompt).to receive(:multiline).and_return(['secret'])
          expect(client).to receive(:post).with('grids/test-grid/secrets', { name: 'mysql_password', value: 'secret'})
          expect{subject.run(['mysql_password'])}.not_to exit_with_error
        end
      end
```

It's not hitting the `prompt.multiline` mock anymore... the specs should definitely mock out the entire `subject.prompt` with a double instead, so this kind of bug would fail immediately.

PR #2046, opened 2017-04-01
Change CLI log --tail/follow options to match Docker

colinmollenhour:
In Docker "--tail" means show the last lines only (like `tail -n`) but in Kontena it means the same as Docker's `--follow` and tail's own `--follow`. This changes Kontena to follow the option names of Docker which I think makes sense given that Kontena is a Docker-based tool.

```
$ docker logs --help

Usage:  docker logs [OPTIONS] CONTAINER

Fetch the logs of a container

Options:
      --details        Show extra details provided to logs
  -f, --follow         Follow log output
      --help           Print usage
      --since string   Show logs since timestamp
      --tail string    Number of lines to show from the end of the logs (default "all")
  -t, --timestamps     Show timestamps
```

### DIFF
--- a/cli/lib/kontena/cli/helpers/log_helper.rb
+++ b/cli/lib/kontena/cli/helpers/log_helper.rb
@@ -4,7 +4,7 @@ module Kontena::Cli::Helpers
     def self.included(base)
       if base.respond_to?(:option)
         base.option ["-f", "--follow"], :flag, "Follow log output", :attribute_name => :tail, default: false
-        base.option "--tail", "LINES", "Number of lines to show from the end of the logs", :attribute_name => :lines, default: 100 do |s|
+        base.option ['--tail', '--lines'], "LINES", "Number of lines to show from the end of the logs", :attribute_name => :lines, default: 100 do |s|
           Integer(s)
         end
         base.option "--since", "SINCE", "Show logs since given timestamp"
--- a/cli/lib/kontena/cli/helpers/log_helper.rb
+++ b/cli/lib/kontena/cli/helpers/log_helper.rb
@@ -3,8 +3,8 @@ module Kontena::Cli::Helpers
 
     def self.included(base)
       if base.respond_to?(:option)
-        base.option ["-t", "--tail"], :flag, "Tail (follow) logs", default: false
-        base.option "--lines", "LINES", "Number of lines to show from the end of the logs", default: 100 do |s|
+        base.option ["-f", "--follow"], :flag, "Follow log output", :attribute_name => :tail, default: false
+        base.option "--tail", "LINES", "Number of lines to show from the end of the logs", :attribute_name => :lines, default: 100 do |s|
           Integer(s)
         end
         base.option "--since", "SINCE", "Show logs since given timestamp"


SpComb:
I don't see why we couldn't add support for the `tail -f/--follow -n/--lines 100` or `docker logs -f/--follow --tail 100` options, but we can't remove support for the existing `kontena * logs -t/--tail --lines 100` options.

@kke would it be possible to support both flavors of options in the same CLI command?

colinmollenhour:
I don't see how you can support both because `--tail` means different things in Docker and Kontena. IMHO it is better to make this breaking change sooner than later and just add a note to the release notes. Surely nobody is relying on the CLI `--tail` switch for any kind of critical production purpose? At worst it might break some shell aliases or something like that and if it is covered in the release notes then it is an easy fix. That is, the functionality is not being removed, just the naming of the CLI options. However, as someone who is frequently using both "docker logs" and "kontena logs" commands this inconsistency can be almost maddening.

colinmollenhour:
Ahh, I traced this back and it actually used to be `--follow` but was changed in #580 due to conflict in some cases with `-f` for file options.

kke:
+1 for `-f` to follow. Comes naturally as in `tail -f`.

I don't see any problem with the `--tail` param either, makes sense to me.

```ruby
base.option "--tail", "LINES"
```

can become:

```ruby
base.option ['--tail', '--lines'], "LINES"
```

to support either keyword.

I doubt anyone is scripting something with the current log following `--tail`.

cli/lib/kontena/cli/helpers/log_helper.rb
```diff
@@ -3,8 +3,8 @@ module LogHelper
 
     def self.included(base)
       if base.respond_to?(:option)
-        base.option ["-t", "--tail"], :flag, "Tail (follow) logs", default: false
-        base.option "--lines", "LINES", "Number of lines to show from the end of the logs", default: 100 do |s|
+        base.option ["-f", "--follow"], :flag, "Follow log output", :attribute_name => :tail, default: false
+        base.option "--tail", "LINES", "Number of lines to show from the end of the logs", :attribute_name => :lines, default: 100 do |s|
```
kke: Change to `['--lines', '--tail']` to keep the old `--lines` as an alternative to --tail.

kke: Also, specs are failing.

SpComb:
@jakolehm WDYT, should the current heroku-like `kontena X logs --tail --lines 1000` be changed into the more common `kontena X logs --follow --tail 1000` to match `docker logs`, `tail`?

The change to the `--tail` option would be incompatible, I don't think we can reasonably support both...

SpComb:
> I doubt anyone is scripting something with the current log following `--tail`

Disagree, there's an example that would be broken right in  #2437 (although for `kontena service events`, which also uses similar flags)

colinmollenhour:
> Disagree, there's an example that would be broken right in #2437 (although for kontena service events, which also uses similar flags)

Is that actually part of a script or just a shell example or alias? I think the point was more about actual critical production scripts. I think a simple announcement in the release notes would be sufficient for people to update their shell aliases and whatnot.

kke:
As `--tail` currently follows and needs a ctrl-c to get out, I doubt anyone is scripting with it. If it returns last X lines, it's more scriptaculous.

cli/lib/kontena/cli/helpers/log_helper.rb
SpComb: Could also consider throwing in `-n` for compatibility with `tail -f`, but maybe three different aliases is already too much.

SpComb:
Needs merge from master the fix the flaky agent spec that's failing, and then the CLI `kontena container logs` spec also needs fixing for the parameter changes:

```
  1) Kontena::Cli::Containers::LogsCommand for a single container with logs errors for an invalid --lines

     Failure/Error: expect { subject.run(["--lines=invalid", "node-1/test-mysql-1"]) }.to raise_error(Clamp::UsageError, "option '--lines': invalid value for Integer(): \"invalid\"")

       expected Clamp::UsageError with "option '--lines': invalid value for Integer(): \"invalid\"", got #<Clamp::UsageError: Unrecognised option '--lines'> with backtrace:

rspec ./spec/kontena/cli/containers/logs_command_spec.rb:48 # Kontena::Cli::Containers::LogsCommand for a single container with logs errors for an invalid --lines
```

jnummelin:
I'm ok with the  (smallish) breaking change for CLI, we just need to properly highlight that in release notes.

jakolehm:
I'm also ok with this change.

SpComb:
Can't merge this with broken specs.

colinmollenhour:
Could one of you fix the specs? I'm not a ruby dev and do not have a test environment setup or know how to set one up or have the time to figure it all out. The changes in this PR were tested by editing the ruby files from the gem install and then copy/pasting the lines into github..

kke:
Adding the `--lines` alias for `--tail` made the tests pass.

SpComb:
Interestingly clamps seems to call the `option do |s| ..` block with a value of `nil` if the option is the last argument, and the option-argument is missing... this could be abused to also support `kontena grid logs --tail`, which currently results in an ugly `TypeError`:

```
$ bundle exec bin/kontena grid logs --tail
 [error] TypeError : can't convert nil into Integer
```

colinmollenhour:
Thanks, gents!

Issue #1997, opened 2017-03-17
Server exceptions with nil backtrace


SpComb:
The server may occasionally raise an exception while logging an exception with `exc.backtrace.join("\n")`, where the `exc.backtrace` appears to be nil:

```
E, [2017-03-16T12:51:29.783964 #7] ERROR -- GridServiceInstanceDeployer: failed to deploy service instance ... to node ...
E, [2017-03-16T12:51:29.784596 #7] ERROR -- GridServiceInstanceDeployer: Connection timeout (5s)
E, [2017-03-16T12:51:29.814434 #7] ERROR -- GridServiceDeployer: Unknown error (...): NoMethodError undefined method `join' for nil:NilClass
Did you mean?  JSON
E, [2017-03-16T12:51:29.815038 #7] ERROR -- GridServiceDeployer: /app/app/services/grid_service_instance_deployer.rb:27:in `rescue in deploy'
/app/app/services/grid_service_instance_deployer.rb:17:in `deploy'
/app/app/services/grid_service_deployer.rb:112:in `block in deploy_service_instance'
```

The exception looks like an `RpcClient::TimeoutError`, raised by the [`RpcClient`](https://github.com/kontena/kontena/blob/v1.1.5.rc2/server/app/services/rpc_client.rb#L80):

```ruby
raise RpcClient::TimeoutError.new(503, "Connection timeout (#{self.timeout}s)")
```

Does raising an exception with `raise Exception.new("foo")`, as opposed to `raise Exception, "foo"`, result in a `nil` `backtrace`?

kke:
Yes I think so. I think you can/should pass the backtrace to Exception.new.

kke:
```ruby
  5 class RpcServer
 15   class Error < StandardError
 16     attr_accessor :code, :message, :backtrace
 17
 18     def initialize(code, message, backtrace = nil)
 19       self.code = code
 20       self.message = message
 21       self.backtrace = backtrace
 22     end
 23   end
```

Maybe that should be `backtrace = []` or `self.backtrace = Array(backtrace)`

SpComb:
Why does the custom `Error` class define an `attr_accessor`for `message` or `backtrace`... sounds like a bug.

SpComb:
The server has now already grown some bad workarounds for this:

```
./app/middlewares/websocket_backend.rb:    logger.error exc.backtrace.join("\n") if exc.backtrace
[...]

PR #2050, opened 2017-04-03
Improve RpcServer performance

jakolehm:
1.2.0.dev had performance regression in RpcServer message handling. This PR should fix this problem and bring rpc performance to at least same level as in 1.1.x.

Improvements:
- requests are prioritized over notifications
- if queue is full notifications are discarded
- agent sends all priority messages as requests
- `Rpc::ContainerHandler` has much better container caching -> should improve performance a lot

Fixes:
- fix EM threading issue in `RpcServer#send_message`

### DIFF
--- a/agent/lib/kontena/rpc_client.rb
+++ b/agent/lib/kontena/rpc_client.rb
@@ -36,26 +36,22 @@ module Kontena
 
     # @param [String] method
     # @param [Array] params
-    # @return [Celluloid::Future]
+    # @return [Object]
     def request(method, params)
       id = request_id
       @requests[id] = nil
-      future = Celluloid::Future.new {
-        sleep 0.01 until @client.connected?
-        @client.send_request(id, method, params)
-        time = Time.now.utc
-        until !@requests[id].nil?
-          sleep 0.01
-          raise TimeoutError.new(500, 'Request timed out') if time < (Time.now.utc - 30)
-        end
-        result, error = @requests.delete(id)
-        if error
-          raise Error.new(error['code'], error['message'], error['backtrace'])
-        end
-        result
-      }
-
-      future
+      sleep 0.01 until @client.connected?
+      @client.send_request(id, method, params)
+      time = Time.now.utc
+      until !@requests[id].nil?
+        sleep 0.01
+        raise TimeoutError.new(500, 'Request timed out') if time < (Time.now.utc - 30)
+      end
+      result, error = @requests.delete(id)
+      if error
+        raise Error.new(error['code'], error['message'])
+      end
+      result
     end
 
     def handle_response(response)
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -43,7 +43,7 @@ module Kontena::Workers
       else
         msg = check_tcp_status(ip, port, timeout)
       end
-      rpc_client.async.notification('/containers/health', [msg])
+      rpc_client.async.request('/containers/health', [msg])
 
       handle_action(msg)
     end
--- a/agent/lib/kontena/workers/service_pod_manager.rb
+++ b/agent/lib/kontena/workers/service_pod_manager.rb
@@ -42,7 +42,7 @@ module Kontena::Workers
 
     def populate_workers_from_master
       exclusive {
-        request = rpc_client.request("/node_service_pods/list", [node.id])
+        request = rpc_client.future.request("/node_service_pods/list", [node.id])
         response = request.value
         unless response['service_pods'].is_a?(Array)
           warn "failed to get list of service pods from master: #{response['error']}"
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -35,7 +35,7 @@ module Kontena::Workers::Volumes
 
     def populate_volumes_from_master
       exclusive {
-        request = rpc_client.request("/node_volumes/list", [node.id])
+        request = rpc_client.future.request("/node_volumes/list", [node.id])
         response = request.value
         unless response['volumes'].is_a?(Array)
           warn "failed to get list of volumes from master: #{response['error']}"
@@ -87,7 +87,7 @@ module Kontena::Workers::Volumes
         'volume_instance_id' => data.dig('Labels', 'io.kontena.volume_instance.id'),
         'volume_id' => data.dig('Labels', 'io.kontena.volume.id')
       }
-      rpc_client.async.notification('/node_volumes/set_state', [node.id, volume])
+      rpc_client.async.request('/node_volumes/set_state', [node.id, volume])
     end
 
     def volume_exist?(volume_name)
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -6,6 +6,9 @@ require_relative '../services/agent/node_unplugger'
 class WebsocketBackend
   KEEPALIVE_TIME = 30 # in seconds
   RPC_MSG_TYPES = %w(request notify)
+  QUEUE_SIZE = 1000
+  QUEUE_WATCH_PERIOD = 60 # once in a minute
+  QUEUE_DROP_NOTIFICATIONS_LIMIT = (QUEUE_SIZE * 0.8)
 
   attr_reader :logger
 
@@ -15,11 +18,14 @@ class WebsocketBackend
     @logger = Logger.new(STDOUT)
     @logger.level = (ENV['LOG_LEVEL'] || Logger::INFO).to_i
     @logger.progname = 'WebsocketBackend'
-    @queue = SizedQueue.new(1000)
+    @msg_counter = 0
+    @msg_dropped = 0
+    @queue = SizedQueue.new(QUEUE_SIZE)
     @rpc_server = RpcServer.new(@queue)
     @rpc_server.async.process!
     subscribe_to_rpc_channel
     watch_connections
+    watch_queue
   end
 
   def call(env)
@@ -94,6 +100,7 @@ class WebsocketBackend
   # @param [Faye::WebSocket::Event] event
   def on_message(ws, event)
     data = MessagePack.unpack(event.data.pack('c*'))
+    @msg_counter += 1
     if rpc_notification?(data)
       handle_rpc_notification(ws, data)
     elsif rpc_request?(data)
@@ -144,7 +151,10 @@ class WebsocketBackend
   # @param [Faye::WebSocket::Event] ws
   # @param [Array] data
   def handle_rpc_notification(ws, data)
-    return if @queue.size > 800 # too busy to handle notifications
+    if @queue.size > QUEUE_DROP_NOTIFICATIONS_LIMIT # too busy to handle notifications
+      @msg_dropped += 1
+      return
+    end
 
     client = client_for_ws(ws)
     if client
@@ -236,14 +246,21 @@ class WebsocketBackend
   end
 
   def watch_connections
-    Thread.new {
-      sleep 1 until EM.reactor_running?
-      EM::PeriodicTimer.new(KEEPALIVE_TIME) do
-        @clients.each do |client|
-          self.verify_client_connection(client)
-        end
+    EM::PeriodicTimer.new(KEEPALIVE_TIME) do
+      @clients.each do |client|
+        self.verify_client_connection(client)
       end
-    }
+    end
+  end
+
+  def watch_queue
+    EM::PeriodicTimer.new(QUEUE_WATCH_PERIOD) do
+      logger.warn "#{@queue.size} messages in queue" if @queue.size > QUEUE_DROP_NOTIFICATIONS_LIMIT
+      logger.warn "#{@msg_dropped} dropped notifications" if @msg_dropped > 0
+      logger.info "#{@msg_counter / QUEUE_WATCH_PERIOD} messages per second"
+      @msg_counter = 0
+      @msg_dropped = 0
+    end
   end
 
   # @param [Hash] client
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -291,4 +291,8 @@ class WebsocketBackend
       self.on_close(client[:ws])
     end
   end
+
+  def stop_rpc_server
+    Celluloid::Actor.kill(@rpc_server) if @rpc_server.alive?
+  end
 end
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -15,7 +15,9 @@ class WebsocketBackend
     @logger = Logger.new(STDOUT)
     @logger.level = (ENV['LOG_LEVEL'] || Logger::INFO).to_i
     @logger.progname = 'WebsocketBackend'
-    @rpc_server = RpcServer.new
+    @queue = SizedQueue.new(1000)
+    @rpc_server = RpcServer.new(@queue)
+    @rpc_server.async.process!
     subscribe_to_rpc_channel
     watch_connections
   end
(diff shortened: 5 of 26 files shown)


agent/lib/kontena/rpc_client.rb
```diff
@@ -36,26 +36,22 @@ def notification(method, params)
...
-        result
-      }
-
-      future
+      sleep 0.01 until @client.connected?
+      @client.send_request(id, method, params)
+      time = Time.now.utc
+      until !@requests[id].nil?
+        sleep 0.01
+        raise TimeoutError.new(500, 'Request timed out') if time < (Time.now.utc - 30)
```
SpComb: This could use the `WaitHelper.wait_until!`, and that would give us lots of nice debug logging on slow requests.

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -42,7 +42,7 @@ def on_update_notify(_, _)
 
     def populate_workers_from_master
       exclusive {
-        request = rpc_client.request("/node_service_pods/list", [node.id])
+        request = rpc_client.future.request("/node_service_pods/list", [node.id])
```
SpComb: Equivalent to just `response = rpc_client.request(...)` without the `future` and `request.value`?

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -35,7 +35,7 @@ def on_update_notify(_, _)
 
     def populate_volumes_from_master
       exclusive {
-        request = rpc_client.request("/node_volumes/list", [node.id])
+        request = rpc_client.future.request("/node_volumes/list", [node.id])
```
SpComb: Same thing here, does this need to be a `future`?

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -43,7 +43,7 @@ def check_status
       else
         msg = check_tcp_status(ip, port, timeout)
       end
-      rpc_client.async.notification('/containers/health', [msg])
+      rpc_client.async.request('/containers/health', [msg])
```
SpComb: I hope we have some reasonable bound on the number of healthchecks... but it's going to be less than the logs, for sure.

server/app/middlewares/websocket_backend.rb
```diff
@@ -135,16 +137,18 @@ def handle_rpc_response(data)
...
     if client
-      @rpc_server.async.handle_request(ws, client[:grid_id].to_s, data)
+      @queue << [ws, client[:grid_id].to_s, data]
     end
   end
 
   # @param [Faye::WebSocket::Event] ws
   # @param [Array] data
   def handle_rpc_notification(ws, data)
+    return if @queue.size > 800 # too busy to handle notifications
```
SpComb: Needs logging of drops... we really have to make these kind of overload situations easier to diagnose as well.

```ruby
if @queue.size > 800
  @dropped += 1
  return
end
```

And then periodically warn on `@dropped > 0` and reset it?

server/app/services/rpc/container_handler.rb
```diff
@@ -115,18 +109,36 @@ def event(data)
...
+    end
+
+    def flush_stats
+      @db_session[:container_stats].insert(@stats.dup)
+      @stats.clear
+    end
+
+    def gc_cache
+      if @cached_containers.keys.size > @containers_cache_size
+        (@containers_cache_size / 5).times { @cached_containers.shift }
```
SpComb: It seems to me as if `Hash#shift` is deterministic: it will always remove the oldest value from the hash. That would mean that this will behave as a kind of ordered queue, which is okay, I guess, in that it ensures that it can't get clogged with dead containers.

Ideally this would use some more effective cache... some basic LRU would be more effective at keeping a small number of very heavy-logging containers in the cache.... this cache will quickly flush out any heavy loggers if you also have lots of quieter ones?

server/app/services/rpc_server.rb
```diff
@@ -41,20 +74,13 @@ def handle_request(ws_client, grid_id, message)
...
-        @handlers[grid_id].delete(handler)
       rescue RpcServer::Error => exc
-        send_message(ws_client, [1, msg_id, {code: exc.code, message: exc.message, backtrace: exc.backtrace}, nil])
+        send_message(ws_client, [1, msg_id, {code: exc.code, message: exc.message}, nil])
         @handlers[grid_id].delete(handler)
       rescue => exc
         error "#{exc.class.name}: #{exc.message}"
-        debug exc.backtrace.join("\n")
-        send_message(ws_client, [1, msg_id, {code: 500, message: "#{exc.class.name}: #{exc.message}", backtrace: exc.backtrace}, nil])
+        debug exc.backtrace.join("\n") if exc.backtrace
```
SpComb: The #1997 [cancer is spreading](https://github.com/kontena/kontena/issues/1997#issuecomment-291082249) :unamused:

server/app/services/rpc_server.rb
```diff
@@ -93,15 +116,19 @@ def handling_actor(grid_id, name)
...
 
     @handlers[grid_id][name]
   end
 
   # @param [Faye::Websocket] ws
-  # @param [Array] message
+  # @param [Object] message
   def send_message(ws, message)
-    ws.send(MessagePack.dump(message).bytes)
+    EM.next_tick {
```
SpComb: Just use `ws.send_message(ws, message)`, it already encapsulates the EM threading stuff.

server/app/services/rpc_server.rb
```diff
@@ -93,15 +116,19 @@ def handling_actor(grid_id, name)
...
+      ws.send(MessagePack.dump(message.as_json).bytes)
```
SpComb: Confused by the `message.as_json`... the agent isn't expecting a JSON-encoded string, it should just be the binary `MessagePack.dump` of the actual ruby hash from the `RpcSerializer.to_hash`? What does this `.as_json` do?

server/app/services/rpc_server.rb
```diff
@@ -71,10 +97,7 @@ def handle_notification(grid_id, message)
     if actor = handling_actor(grid_id, handler)
```
SpComb: It's not an actor anymore.

server/app/services/rpc_server.rb
```diff
@@ -25,8 +28,38 @@ def initialize(code, message, backtrace = nil)
...
+    if @queue.size > QUEUE_WARN_LIMIT
+      warn "#{@queue.size} messages in queue"
+      info "#{@counter / REPORT_EVERY} requests per second"
+    end
+    @counter = 0
+  end
+
+  def process!
+    @processing = true
+    defer {
```
SpComb: Hmm the `defer` feels a little scary, all of the `@handlers` state is now mutated by the defer thread, not the actor thread. Maybe just keep this in the main actor thread, and perhaps move the queue size reporting stuff to the `WebsocketBackend` instead?

server/app/services/rpc_server.rb
jakolehm: MessagePack will explode if it receives object that is not transformed to some json-ish format. That's why. This is not mandatory, can be removed.

server/app/services/rpc_server.rb
jakolehm: This is `Faye::Websocket`, it does not have `#send_message`?

server/app/services/rpc_server.rb
jakolehm: Agreed, I'll try to refactor this reporting to `WebsocketBackend` (although numbers there are different).

server/app/services/rpc_server.rb
SpComb: So `.as_json` returns a ruby hash, not a JSON string?

server/app/services/rpc_server.rb
SpComb: Oh, right. So the existing code was violating the EM threading, writing to the WS from the `RpcServer` actor thread? Hmm... :boom:

Perhaps this actor should be passed the `WebsocketBackend` instead, with some explicit comment about how `send_message` is okay for calling by other threads... mixing code for three different threads into the one Actor doesn't feel like a good idea.

server/app/services/rpc_server.rb
jakolehm: Yes, it returns a ruby hash/array. Classes can override it if necessary. `.to_json` would return a JSON string and that's not really good idea here.

server/app/services/rpc/container_handler.rb
jnummelin: I agree that LRU would work nicely here

jakolehm: Yes, LRU would be nice.. but I would say that is a separate PR :)

agent/lib/kontena/workers/service_pod_manager.rb
jakolehm: Maybe, don't want to change these here.. tried to keep consumer code as close as possible to original.

server/app/services/rpc_server.rb
jakolehm: I don't like the idea of passing whole `WebsocketBackend`. It's too "fat" and has references to every websocket client connection.

server/app/services/rpc/container_handler.rb
```diff
@@ -14,9 +13,10 @@ def initialize(grid)
       @grid = grid
       @logs = []
       @stats = []
-      @cached_container = nil
-      @logs_buffer_size = 5
+      @cached_containers = {}
+      @logs_buffer_size = 10
```
SpComb: Because it holds on to the buffered logs indefinitely, this change to increase the logs buffer size can break things like the https://github.com/kontena/kontena/pull/1874#issuecomment-284659342 e2e specs, where there's nothing else getting logged :(

IMO if you want to increase the logs buffering any more, then it needs to also flush them out after e.g. 1s, or this will start to behave too much like a bug from the perspective of users..

server/app/services/rpc/container_handler.rb
SpComb: Yes, this is good enough.

server/app/services/rpc_server.rb
```diff
@@ -25,8 +27,27 @@ def initialize(code, message, backtrace = nil)
...
+  # @param [SizedQueue] queue
+  def initialize(queue)
+    @queue = queue
     @handlers = {}
+    @counter = 0
+    @processing = false
+  end
+
+  def process!
+    sleep 0.1 # avoid busy loop block in actor
```
SpComb: I don't understand this... this sleep isn't in the busy-loop?

server/app/services/rpc/container_handler.rb
jakolehm: I agree. Reverted.

server/app/services/rpc_server.rb
jakolehm: It's not actually needed... pushed a fix for this.

SpComb: I still don't get it... comments should make the code less confusing, not more confusing :D 

Is that sleep supposed to be inside the `while` loop?

jakolehm: Sleep and comment removed. Less confusing ;)

jnummelin:
Works as expected:
```
api_1      | W, [2017-04-03T12:14:23.723361 #19]  WARN -- WebsocketBackend: 221 dropped notifications
api_1      | I, [2017-04-03T12:14:23.723446 #19]  INFO -- WebsocketBackend: 353 messages per second
api_1      | I, [2017-04-03T12:14:26.033611 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:14:26.177458 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:14:26.278395 #17]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:15:23.678795 #19]  INFO -- WebsocketBackend: 327 messages per second
api_1      | I, [2017-04-03T12:15:26.039578 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:15:26.183056 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:15:26.285502 #17]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:16:23.634248 #19]  INFO -- WebsocketBackend: 351 messages per second
api_1      | I, [2017-04-03T12:16:26.046294 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:16:26.186599 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:16:26.293858 #17]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:17:23.590140 #19]  INFO -- WebsocketBackend: 303 messages per second
api_1      | I, [2017-04-03T12:17:26.053135 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:17:26.192397 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:17:26.300380 #17]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:18:23.545834 #19]  INFO -- WebsocketBackend: 493 messages per second
api_1      | I, [2017-04-03T12:18:26.057704 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:18:26.199769 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:18:26.301133 #17]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:19:23.501267 #19]  INFO -- WebsocketBackend: 1487 messages per second
api_1      | I, [2017-04-03T12:19:26.055644 #13]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:19:26.207529 #9]  INFO -- WebsocketBackend: 0 messages per second
api_1      | I, [2017-04-03T12:19:26.299376 #17]  INFO -- WebsocketBackend: 0 messages per second
```

SpComb:
BTW the PR title isn't exactly accurate, the request > notification prioritization stuff is a new feature, not just a regression fix over 1.1?

jakolehm:
Not sure why travis fails like that... "works on my machine".

jakolehm:
Mergeable if Travis agrees.

server/spec/middlewares/websocket_backend_spec.rb
```diff
@@ -4,10 +4,23 @@
   let(:app) { spy(:app) }
   let(:subject) { described_class.new(app) }
 
+  before(:all) do
+    EM.run
```
SpComb: `EM.run` blocks the calling thread: http://www.rubydoc.info/github/eventmachine/eventmachine/EventMachine#run-class_method

SpComb: Basic pattern for running EM specs should be like on the agent: https://github.com/kontena/kontena/blob/master/agent/spec/lib/kontena/websocket_client_spec.rb#L13-L18

```
ruby
  around(:each, :em => true) do |example|
    EM.run {
      example.run
      EM.stop
    }
end
```

jakolehm: I had it like that and travis did not like it.. that's why it's now before/after :all .

SpComb:
`rspec --seed 36278` consistently fails in the `services/cloud/websocket_client_spec.rb`, it's running some deferred `RpcServer#send_message` call:

```
Failures:

  1) Cloud::WebsocketClient#on_message when instance is leader on rpc request requests RPC::Server::Api
     Failure/Error: EM.run {
       Double :ws_client was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
     # ./app/services/rpc_server.rb:118:in `block in send_message'
     # ./vendor/bundle/ruby/2.3.0/gems/eventmachine-1.2.3/lib/eventmachine.rb:977:in `block in run_deferred_callbacks'
     # ./vendor/bundle/ruby/2.3.0/gems/eventmachine-1.2.3/lib/eventmachine.rb:974:in `times'
     # ./vendor/bundle/ruby/2.3.0/gems/eventmachine-1.2.3/lib/eventmachine.rb:974:in `run_deferred_callbacks'
     # ./vendor/bundle/ruby/2.3.0/gems/eventmachine-1.2.3/lib/eventmachine.rb:194:in `run_machine'
     # ./vendor/bundle/ruby/2.3.0/gems/eventmachine-1.2.3/lib/eventmachine.rb:194:in `run'
     # ./spec/services/cloud/websocket_client_spec.rb:87:in `block (5 levels) in <top (required)>'

```

SpComb:
The spec failure happens if the `RpcServer` specs run in between two other EM-enabled specs: `spec/middlewares/websocket_backend_spec.rb` -> `spec/services/rpc_server_spec.rb` `spec/services/cloud/websocket_client_spec.rb`, which leaks some `RpcServer#send_message` -> `EM.next_tick{ .. }` stuff that explodes once something touchs `EM.run`.

The `RpcServer` specs should mock out the `EM.next_tick`.

PR #2053, opened 2017-04-03
Make stack variable yes/no prompts honor default value

### LINKED ISSUE #1854: stack variable boolean default does not matter for prompt
```
  mongodb_lock:
    type: boolean
    default: false
    from:
      prompt:
```

results in:

```
> Enable mongodb_oplog : (Y/n)
```

when it should say:

```
> Enable mongodb_oplog : (y/N)
```

kke:
Fixes #1854


SpComb:
Still seems to prompt with `(Y/n)` by default

```yaml
stack: test/test
variables:
  mongodb_lock:
    type: boolean
    default: false
    from:
      prompt:
services:
  foo:
    image: redis
```

#### `kontena stack validate`
```
> Enable mongodb_lock : (Y/n) 
```

SpComb:
Although the resulting value on hitting Enter is actually `false` now:

```
variables:
  mongodb_lock:
    label: mongodb_lock
    type: boolean
    from:
      prompt: 
      env: mongodb_lock
    to:
      env: mongodb_lock
    errors: {}
    value: false
```

SpComb:
The existing code in git master is at consistent in that it prompts with `(Y/n)` and sets it to `true` if you hit enter. Prompting with `(Y/n)` but setting it to `false` if you hit enter is still buggy.

kke:
image

image

true:

image

image

SpComb:
Odd, it's different for me:
2017-04-03-151926_1278x629_scrot

Also with your example:
2017-04-03-152055_1278x629_scrot

`Gemfile.lock` shows `tty-prompt (0.11.0)`

kke:
Strange indeed. CLI shouldn't have Gemfile.lock in the repo anyway. 

Also the `prompt.yes?` defaulting syntax hasn't changed since 0.11.

SpComb:
Bumping to 1.2.1 until we figure out why the prompt seems to be wrong on Linux....

SpComb:
Perhaps this should use [`prompt.no?`]I(https://github.com/piotrmurach/tty-prompt#25-yesno) for the `default: false` case?

kke:
Uses `no?` now.

SpComb:
Eh... now it correctly prompts with `> Enable mongodb_lock : (y/N)`, but then if I just hit enter, the result is `value: true` :/

kke:
Heh, that is because `no?` returns true when you choose no. Reverting.

kke:
`prompt.yes?('foo', default: false)` works as expected.

The original `option.default == false ? false : true` did not.

SpComb:
Same thing as before, it's prompting with `> Enable mongodb_lock : (Y/n)` but then making the value `false` when you hit enter.... some bug with `tty-prompt (0.11.0)` on Ubuntu 16.04 `gnome-terminal` vs OSX terminal?

SpComb:
This works, but the default for a boolean variable is now `false`:

```ruby
       def bool
        prompt.yes?(question_text, default: !!option.default, type: option.default ? :yes : :no)
       end
```

The `tty-prompt-0.11.0` `TTY::Prompt::ConfirmQuestion` code is really weird. I think the use of `default?` is wrong, it should be `default`?

```
      # Check if default value is set
      #
      # @return [Boolean]
      #
      # @api public
      def default?
        @default != UndefinedSetting
      end
...
      # @api private
      def setup_defaults
        return if suffix? && positive?

        if suffix? && !positive?
          parts = @suffix.split('/')
          @positive = parts[0]
          @negative = parts[1]
          @convert = conversion
        elsif !suffix? && positive?
          @suffix = create_suffix
          @convert = conversion
        else
          create_default_labels
          @convert  = :bool
        end
      end

      def create_default_labels
        if is?(:yes)
          @suffix   = default? ? 'Y/n' : 'y/N'
          @positive = default? ? 'Yes' : 'yes'
          @negative = default? ? 'no' : 'No'
        else
          @suffix   = default? ? 'y/N' : 'Y/n'
          @positive = default? ? 'Yes' : 'yes'
          @negative = default? ? 'No'  : 'no'
        end
      end

      # @api private
      def create_suffix
        result = ''
        if is?(:yes)
          result << "#{default? ? @positive.capitalize : @positive.downcase}"
          result << '/'
          result << "#{default? ? @negative.downcase : @negative.capitalize}"
        else
          result << "#{default? ? @positive.downcase : @positive.capitalize}"
          result << '/'
          result << "#{default? ? @negative.capitalize : @negative.downcase}"
        end
      end
```

SpComb:
Seems to be fixed in `tty-prompt` 0.12.0, the Gemfile needs update.

cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
```diff
@@ -50,7 +50,7 @@ def enum
       end
 
       def bool
-        prompt.yes?(question_text, default: option.default == false ? false : true)
```
SpComb: I think we should keep `default = nil` -> `default: true`?

cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
```diff
@@ -50,7 +50,7 @@ def enum
...
+        prompt.yes?(question_text, default: option.default.nil? || option.default)
```
SpComb: It seems like `default: false` in the YAML turns into `option.default = nil` :speak_no_evil:

kke:
Confused 😕 :
```
(byebug) opts
{"default"=>false, "from"=>"prompt"}
(byebug) opts['from']
"prompt"
(byebug) opts['default']
nil
```

cli/lib/kontena/cli/stacks/yaml/opto/prompt_resolver.rb
SpComb: There's some weird whitespace here between the `|| option`.

PR #2056, opened 2017-04-04
Fix possible thread leaks in WebsocketBackend

jakolehm:
There might be potential leaks in WebsocketBackend. This PR tries to make those spots a little more safer.

### DIFF
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -86,7 +86,7 @@ class WebsocketBackend
       end
 
       logger.info "node opened connection: #{node.name || node_id}, labels: #{labels}"
-      node_plugger.plugin!
+      EM.defer { node_plugger.plugin! }
     else
       logger.error 'invalid grid token, closing connection'
       ws.close(4001)
@@ -186,14 +186,14 @@ class WebsocketBackend
   # @param [Faye::WebSocket] ws
   # @return [Hash,NilClass]
   def client_for_ws(ws)
-    @clients.find{|c| c[:ws] == ws}
+    @clients.find{ |c| c[:ws] == ws }
   end
 
   ##
   # @param [String] id
   # @return [Hash,NilClass]
   def client_for_id(id)
-    @clients.find{|c| c[:id] == id}
+    @clients.find{ |c| c[:id] == id }
   end
 
   # @param [String] agent_version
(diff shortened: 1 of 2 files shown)


server/app/middlewares/websocket_backend.rb
```diff
@@ -86,7 +86,7 @@ def on_open(ws, req)
       end
 
       logger.info "node opened connection: #{node.name || node_id}, labels: #{labels}"
-      node_plugger.plugin!
+      EM.defer { node_plugger.plugin! }
```
SpComb: Should the `on_close` -> `Agent::NodeUnplugger.unplug!` also be moved into a defer thread?

server/app/middlewares/websocket_backend.rb
```diff
@@ -237,23 +237,22 @@ def subscribe_to_rpc_channel
...
   # @param [Hash] msg
   def on_rpc_message(msg)
-    client = client_for_id(msg['id'])
-    if client
-      self.send_message(client[:ws], msg['message'])
-    end
+    EM.next_tick{
+      client = client_for_id(msg['id'])
+      if client
+        self.send_message(client[:ws], msg['message'])
```
SpComb: This now `EM.next_tick`'s twice... once in the `on_rpc_message`, and then again in `send_message`... but I suppose that doesn't hurt.

server/app/middlewares/websocket_backend.rb
```diff
@@ -237,23 +237,22 @@ def subscribe_to_rpc_channel
...
+      end
+    }
   rescue => exc
```
SpComb: Does `EM.next_tick` propagate exceptions? I doubt it, and I suspect this doesn't rescue anything anymore?

server/app/middlewares/websocket_backend.rb
```diff
@@ -237,23 +237,22 @@ def subscribe_to_rpc_channel
...
     logger.error "on_rpc_message: #{exc.message}"
   end
 
   def watch_connections
-    Thread.new {
-      sleep 1 until EM.reactor_running?
```
SpComb: I wonder why this was in a separate thread with the sleep before... I assume it's perfectly fine to install a new EM timer before the EM thread is running?

Interestingly, it seems like puma actually waits for the `on_worker_boot` hook -> require `server/app/initializers/eventmachine.rb` -> thread to call `EM.run` before it will proceed with the worker boot, and thus EM will already be running when this `WebsocketBackend#initialize` -> `watch_connections` happens, even with a `sleep` to delay the `EM.run`.
atm the `WebsocketBackend#initialize` is called by puma after running the `on_worker_boot` hook which loads the initializers that start EM

server/app/middlewares/websocket_backend.rb
SpComb: Letting the `EM.next_tick { ... }` raise seems to cause weird things (the `MongoPubsub` actor dies without any errors logged)... move the `EM.next_tick{ ... }` into the `subscribe_to_rpc_channel` block, or move this rescue into the `EM.next_tick` block.

SpComb: Well, I suppose that all of the `EM.next_tick { ... }` blocks would need some fixing like this... not sure if part of this PR or not.

PR #2061, opened 2017-04-04
Send both legacy & new driver information from a node

jakolehm:
Also adds volume plugin information to `k node show <name>` output.

Example:

```
$ kontena node show core-01
core-01:
  id: YZCH:NO2R:UB3N:5B4M:KEZD:OTEY:NY7A:XFKX:HAHU:BV2R:WCCD:HAGJ
  agent version: 1.2.0.dev
  docker version: 1.13.1
  connected: yes
  last connect: 2017-04-04T10:15:21.818Z
  last seen: 2017-04-04T12:27:58.007Z
  public ip: 82.181.164.219
  private ip: 192.168.66.101
  overlay ip: 10.81.0.1
  os: Container Linux by CoreOS 1367.5.0 (Ladybug)
  kernel: 4.10.4-coreos-r1
  drivers:
    storage: overlay
    volume: rexray/s3fs:latest,local
  initial node: yes
  labels:
  stats:
    cpus: 1
    load: 0.0 0.01 0.0
    memory: 0.4 of 0.97 GB
    filesystem:
      - /var/lib/docker: 2.53 of 15.57 GB
```

### DIFF
--- a/agent/lib/kontena/workers/node_info_worker.rb
+++ b/agent/lib/kontena/workers/node_info_worker.rb
@@ -123,7 +123,11 @@ module Kontena::Workers
 
     # @return [Array<Hash>]
     def plugins
-      JSON.parse(Docker.connection.get('/plugins'))
+      if docker_api_version >= 1.26
+        JSON.parse(Docker.connection.get('/plugins'))
+      else
+        []
+      end
     rescue => exc
       warn "cannot fetch docker engine plugins: #{exc.message}"
       []
@@ -386,6 +390,16 @@ module Kontena::Workers
     # @return [Hash]
     def docker_info
       @docker_info ||= Docker.info
+    end
+
+    # @return [Hash]
+    def docker_version
+      @docker_version ||= Docker.version
+    end
+
+    # @return [Float]
+    def docker_api_version
+      docker_version["ApiVersion"].to_f
     end
   end
 end
--- a/agent/lib/kontena/workers/node_info_worker.rb
+++ b/agent/lib/kontena/workers/node_info_worker.rb
@@ -95,7 +95,8 @@ module Kontena::Workers
       plugins.each do |plugin|
         config = plugin['Config']
         if config.dig('Interface', 'Types').include?('docker.volumedriver/1.0')
-          drivers << { name: plugin['Name'] } if plugin['Enabled']
+          name, version = plugin['Name'].split(':')
+          drivers << { name: name, version: version } if plugin['Enabled']
         end
       end
       docker_info.dig('Plugins', 'Volume').to_a.each do |plugin|
@@ -111,7 +112,8 @@ module Kontena::Workers
       plugins.each do |plugin|
         config = plugin['Config']
         if config.dig('Interface', 'Types').include?('docker.networkdriver/1.0')
-          drivers << { name: plugin['Name'] } if plugin['Enabled']
+          name, version = plugin['Name'].split(':')
+          drivers << { name: name, version: version } if plugin['Enabled']
         end
       end
       docker_info.dig('Plugins', 'Network').to_a.each do |plugin|
--- a/agent/lib/kontena/workers/node_info_worker.rb
+++ b/agent/lib/kontena/workers/node_info_worker.rb
@@ -76,12 +76,57 @@ module Kontena::Workers
 
     def publish_node_info
       debug 'publishing node information'
-      docker_info['PublicIp'] = self.public_ip
-      docker_info['PrivateIp'] = self.private_ip
-      docker_info['AgentVersion'] = Kontena::Agent::VERSION
-      rpc_client.async.request('/nodes/update', [docker_info])
+      node_info = docker_info.dup
+      node_info['PublicIp'] = self.public_ip
+      node_info['PrivateIp'] = self.private_ip
+      node_info['AgentVersion'] = Kontena::Agent::VERSION
+      node_info['Plugins'] = {
+        'Volume' => volume_drivers,
+        'Network' => network_drivers
+      }
+      rpc_client.async.request('/nodes/update', [node_info])
     rescue => exc
       error "publish_node_info: #{exc.message}"
+    end
+
+    # @return [Array<Hash>]
+    def volume_drivers
+      drivers = []
+      plugins.each do |plugin|
+        config = plugin['Config']
+        if config.dig('Interface', 'Types').include?('docker.volumedriver/1.0')
+          drivers << { name: plugin['Name'] } if plugin['Enabled']
+        end
+      end
+      docker_info.dig('Plugins', 'Volume').to_a.each do |plugin|
+        drivers << { name: plugin }
+      end
+
+      drivers
+    end
+
+    # @return [Array<Hash>]
+    def network_drivers
+      drivers = []
+      plugins.each do |plugin|
+        config = plugin['Config']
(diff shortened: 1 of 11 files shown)


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -76,14 +76,59 @@ def configure_statsd(node)
...
       debug 'publishing node information'
-      docker_info['PublicIp'] = self.public_ip
-      docker_info['PrivateIp'] = self.private_ip
-      docker_info['AgentVersion'] = Kontena::Agent::VERSION
-      rpc_client.async.request('/nodes/update', [docker_info])
+      node_info = docker_info.dup
+      node_info['PublicIp'] = self.public_ip
+      node_info['PrivateIp'] = self.private_ip
+      node_info['AgentVersion'] = Kontena::Agent::VERSION
+      node_info['Plugins'] = {
```
SpComb: Is it plugins or drivers? :) :bikeshed:

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -76,14 +76,59 @@ def configure_statsd(node)
...
+      docker_info.dig('Plugins', 'Network').to_a.each do |plugin|
+        drivers << { name: plugin }
+      end
+
+      drivers
+    end
+
+    # @return [Array<Hash>]
+    def plugins
+      JSON.parse(Docker.connection.get('/plugins'))
```
SpComb: The `/plugins` endpoint seems to have been added in Docker API version 1.24 (per the docs), and we're still using 1.22. Should we be bumping the `Docker::API_VERSION`?

server/app/services/scheduler/filter/volume_plugin.rb
```diff
@@ -19,7 +19,8 @@ def for_service(service, instance_number, nodes)
         }.compact
 
         nodes = nodes.select { |node|
-          (needed_drivers - node.plugins['volume']).empty?
+          volume_plugins = node.plugins['volume'].map { |v| v['name'] }
```
SpComb: AFAIK if the plugin name is something like `rexray/s3fs:latest`, then the volume driver name is just `rexray/s3fs`? Or was it even just `s3fs`?

This would read a lot better if it was

```ruby
volume_drivers = node.volume_plugins.map { |v| v['name'].split(':')[0] }
```

Notice how the "plugins" turn into "drivers"... maybe the agent could parse out the `name` and `version` fields separately?

agent/lib/kontena/workers/node_info_worker.rb
SpComb: This will fail with a warning on Docker 1.12 without experimental:

```
W, [2017-04-04T12:43:40.985256 #1]  WARN -- Kontena::Workers::NodeInfoWorker: cannot fetch docker engine plugins: page not found
```

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -76,14 +76,63 @@ def configure_statsd(node)
...
+      docker_info.dig('Plugins', 'Network').to_a.each do |plugin|
+        drivers << { name: plugin }
+      end
+
+      drivers
+    end
+
+    # @return [Array<Hash>]
+    def plugins
+      if docker_api_version >= 1.26
```
jnummelin: `/plugins` was introduced in 1.24: https://docs.docker.com/engine/api/v1.24/#plugins-experimental

cli/lib/kontena/cli/nodes/show_command.rb
```diff
@@ -23,8 +23,10 @@ def execute
       puts "  private ip: #{node['private_ip']}"
       puts "  overlay ip: #{node['overlay_ip']}"
       puts "  os: #{node['os']}"
-      puts "  driver: #{node['driver']}"
       puts "  kernel: #{node['kernel_version']}"
+      puts "  drivers:"
+      puts "    storage: #{node['driver']}"
+      puts "    volume: #{node.dig('plugins', 'volume').to_a.map{|v| v['name']}.join(',')}"
```
jnummelin: should this be `node.volume_drivers`?

agent/lib/kontena/workers/node_info_worker.rb
jakolehm: How many users will run Docker Engine in experimental mode? :)

jnummelin: Probably not many. :) 1.25 seems to have removed the experimental note

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -76,14 +76,63 @@ def configure_statsd(node)
...
       error "publish_node_info: #{exc.message}"
     end
 
+    # @return [Array<Hash>]
+    def volume_drivers
+      drivers = []
+      plugins.each do |plugin|
+        config = plugin['Config']
+        if config.dig('Interface', 'Types').include?('docker.volumedriver/1.0')
+          drivers << { name: plugin['Name'] } if plugin['Enabled']
```
SpComb: This stores the full `{ "Name": "rexray/s3fs:latest"}` string as the driver name. I think this should split out the `name: 'rexray/s3fs'` and `version: 'latest'` fields.

That would make it easier to match the expected `driver: rexray/s3fs` in the server scheduler's volume plugin filter...

server/app/models/host_node.rb
```diff
@@ -29,6 +29,8 @@ class Error < StandardError
   field :agent_version, type: String
   field :docker_version, type: String
   field :plugins, type: Hash, default: {}
```
SpComb: The `plugins` field is abandoned, and not getting updated anymore? Should we remove it?

server/app/models/host_node.rb
```diff
@@ -29,6 +29,8 @@ class Error < StandardError
...
+  field :volume_drivers, type: Array, default: []
```
SpComb: Some comment like `# [ { 'name' => String } ]` describing the array structure would be nice. Even better would be an `embeds_many :host_node_drivers`.

server/app/services/scheduler/filter/volume_plugin.rb
```diff
@@ -19,7 +19,8 @@ def for_service(service, instance_number, nodes)
         }.compact
 
         nodes = nodes.select { |node|
-          (needed_drivers - node.plugins['volume']).empty?
+          volume_drivers = node.volume_drivers.map { |v| v['name'] }
```
SpComb: ATM the `volume_drivers` `name` is the full plugin name with a version tag, like `rexray/s3fs:latest`. This means that creating a volume with `kontena volume create --driver rexray/s3fs` will fail to deploy... the `rexray/s3fs:latest` doesn't match `rexray/s3fs`.

jakolehm: @jnummelin what do you think? Maybe separate version would make sense here?

agent/lib/kontena/workers/node_info_worker.rb
jakolehm: @jnummelin what do you think? Maybe separate version makes sense here?

jnummelin: yes, separate version makes sense for informational purposes

PR #2067, opened 2017-04-04
Refactor agent RpcClient

jakolehm:


SpComb:
Not sure I like the non-actor alternative any more... my thinking for this was something along the lines of having a `Kontena::Helpers::RpcHelper` wrapper to handle the raise within the calling actor, thus avoiding crashing the `RpcClient` actor:

```ruby
module Kontena::Helpers::RpcHelper
  def rpc_request(method, params)
    result, error = rpc_client.request(method, params)

    raise error if error
    return result
  end
end
```

Remove the `raise` stuff from the `RpcClient#request` and just have it return the `result, error = @requests.delete(id)` directly.

jakolehm:
#2008 seems to fix most of the issues, closing this one.

PR #2077, opened 2017-04-05
Refactor stacks api to always require extenal name for a volume

jnummelin:
Now the stacks api requires always the external name to be given:
```
"volumes": [
  {"name": "vol", "external":"ext-name"}
]
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -110,6 +110,11 @@ module Kontena::Cli::Stacks
     def generate_volumes(yaml_volumes)
       return [] unless yaml_volumes
       yaml_volumes.map do |name, config|
+        if config['external'].is_a?(TrueClass)
+          config['external'] = name
+        elsif config['external']['name']
+          config['external'] = config['external']['name']
+        end
         config.merge('name' => name)
       end
     end
--- a/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
@@ -92,21 +92,7 @@ module Kontena::Cli::Stacks
             yaml['volumes'].each do |volume, options|
               if options.is_a?(Hash)
                 option_errors = validate_volume_options(options)
-                if option_errors.valid?
-                  if !options.key?('driver') && options.key?('driver_opts')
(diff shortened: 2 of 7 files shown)


SpComb:
Can we have the CLI support the `external: true`?

ATM it looks like the `test/spec/fixtures/stack/volumes/redis-simple.yml` spec is going to be broken.

jnummelin:
forgot to put the CLI on the todo list

jnummelin:
CLI side fixed to always send `external: name` for volumes. E2E cases pass.

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -110,6 +110,11 @@ def require_config_file(filename)
     def generate_volumes(yaml_volumes)
       return [] unless yaml_volumes
       yaml_volumes.map do |name, config|
+        if config['external'].is_a?(TrueClass)
+          config['external'] = name
+        elsif config['external']['name']
```
SpComb: Do we support `external: false`? It seems to pass the options validations and explodes here...

```
NoMethodError: undefined method `[]' for false:FalseClass
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:115:in `block in generate_volumes'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:112:in `each'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:112:in `map'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:112:in `generate_volumes'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:79:in `stack_from_reader'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:101:in `stack_read_and_dump'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/install_command.rb:23:in `execute'
```

jnummelin: Nope, `external: false` isn't supported. but it should not 💣

SpComb: Might also make sense to support `external: false`... but it should be mapped to `{"external": null }` (or better yet, omit it entirely) in the API.

PR #2084, opened 2017-04-06
Only cleanup nodes labeled as ephemeral

### LINKED ISSUE #2025: Node went offline and came back online without labels
Another issue seemingly randomly occurred over the weekend.. I'm beginning to think the best way to run Kontena is to use it to launch all of your containers and then just kill the master until you need it again to deploy an upgrade.  :/

You can see in the attached logs where there was a node disconnection and then containers were removed. The services all have affinities and although the node is currently connected it is now missing the "app" and "frontend" labels that it had before. Looks like there are some unexpected Ruby exceptions, but I'm not a strong Ruby debugger so it wasn't obvious to me what the exact cause was.

Master/Node Version: 1.1.5
[kontena-server-api.txt](https://github.com/kontena/kontena/files/873032/kontena-server-api.txt)
[kontena-agent.txt](https://github.com/kontena/kontena/files/873033/kontena-agent.txt)

SpComb:
Fixes #2025

The server `NodeCleanupJob` will first mark host nodes as disconnected after 1 minute of no websocket ping-pongs, and then destroy any disconnected nodes after 1 hour, unless they are initial nodes or have stateful containers. This has some side-effects:

* #2025 Any labels set with `kontena node label add` are forgotten, and will be gone if the node later re-connects
* Any `VolumeInstance`s for that node are destroyed, which currently means that the host node will terminate them after reconnecting

This PR bumps the default destroy timeout to 6h, and only cleans up host nodes that have an explicit `emphemeral` label set.

### DIFF
--- a/docs/using-kontena/nodes.md
+++ b/docs/using-kontena/nodes.md
@@ -23,6 +23,8 @@ The agents will ping the server every 30 seconds, and expect a response within 2
 The server will ping each agent every 30 seconds, and expect a response within 5 seconds.
 
 Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy#Affinity) or the [grid default affinity](grids#Default Affinity).
+When a host node comes online, existing grid services will be re-scheduled by the server to deploy new [`daemon`-strategy](deploy#Daemon) instances or re-balance other service instances onto the new node.
+The re-schduling of grid services will happen within 20 seconds of the node coming online.
 
 ### Offline nodes
 
--- a/docs/using-kontena/nodes.md
+++ b/docs/using-kontena/nodes.md
@@ -22,8 +22,8 @@ Nodes will be considered as online so long as they have an active Websocket conn
 The agents will ping the server every 30 seconds, and expect a response within 2 seconds.
 The server will ping each agent every 30 seconds, and expect a response within 5 seconds.
 
-Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy.md#Affinity) or the [grid default affinity](grids.md#Default Affinity).
-When a host node comes online, existing grid services will be re-scheduled by the server to deploy new [`daemon`-strategy](deploy.md#Daemon) instances or re-balance other service instances onto the new node.
+Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy.md#affinity) or the [grid default affinity](grids.md#default-affinity).
+When a host node comes online, existing grid services will be re-scheduled by the server to deploy new [`daemon`-strategy](deploy.md#daemon) instances or re-balance other service instances onto the new node.
 The re-schduling of grid services will happen within 20 seconds of the node coming online.
 
 ### Offline nodes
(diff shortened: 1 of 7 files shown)


SpComb:
Design review: Ideally there should be some explicit "transient" flag used for things like auto-scaled host nodes that are expected to disappear. By default, the server should not destroy host nodes without an explicit DELETE request. However, I'm not sure how that flag should be configured? A label?

SpComb:
Also fixing the "node comes back after kontena node rm" -> "you need to terminate it first" issue while I'm here.

SpComb:
Trying to figure out where to document this new label...

SpComb:
Added a new `Using Kontena > Nodes` page to the docs.

SpComb:
Suggestion: have the server exclude `ephemeral` nodes from being initial nodes, or scheduling any stateful containers?

docs/using-kontena/nodes.md
```diff
@@ -0,0 +1,45 @@
...
+Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy#Affinity) or the [grid default affinity](grids#Default Affinity).
+
+The `kontena node rm` command can not be used to remove online nodes. The node must first be terminated (using the `kontena <provider> node terminate` command), and can then be removed once offline.
+
+## Offline nodes
+
+If the agent's Websocket connection to the master is disconnected or times out, the server will mark the nodes as as offline.
+
+Offline nodes will not have any new service instances scheduled to them. Any services with instances deployed to any offline nodes will be re-scheduled by the server, moving the instances to the remaining online nodes.
+
```
jnummelin: re-scheduled after a grace period

SpComb: Right... and the `NodePlugger` doesn't immediatel reschedule services either, anymore.

docs/using-kontena/nodes.md
```diff
@@ -0,0 +1,45 @@
...
+## Connecting nodes
+
+Nodes do not need to be explicitly created. The Kontena Master will automatically create new grid nodes for any `kontena-agent` connecting to the Kontena Master with the correct [grid token](grids#Grid Token).  
+
+## Online nodes
+
+Nodes will be considered as online so long as they have an active Websocket connection to the Kontena Master.
+Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy#Affinity) or the [grid default affinity](grids#Default Affinity).
+
+The `kontena node rm` command can not be used to remove online nodes. The node must first be terminated (using the `kontena <provider> node terminate` command), and can then be removed once offline.
```
jakolehm: We should add note that node can be terminated also via other tools (like cloud provider sdk/web ui)

docs/using-kontena/nodes.md
jakolehm: Scheduler does not track containers anymore.

jnummelin: Currently with the new scheduler there's no grace period AFAICT

docs/using-kontena/nodes.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Online nodes
+
+Nodes will be considered as online so long as they have an active Websocket connection to the Kontena Master.
+The agents will ping the server every 30 seconds, and expect a response within 2 seconds.
+The server will ping each agent every 30 seconds, and expect a response within 5 seconds.
+
+Grid service instances can be deployed to any online node, unless restricted using [service affinity filters](deploy#Affinity) or the [grid default affinity](grids#Default Affinity).
+When a host node comes online, existing grid services will be re-scheduled by the server to deploy new [`daemon`-strategy](deploy#Daemon) instances or re-balance other service instances onto the new node.
+The re-schduling of grid services will happen within 20 seconds of the node coming online.
```
jakolehm: re-scheduling

PR #2086, opened 2017-04-06
Refuse to remove an online node

SpComb:
Fixing the "node comes back after `kontena node rm`" -> "you need to terminate it first" issue.

Also means that typo'd `kontena node rm asdf` will fail with `Not Found` before it bothers to prompt you.

### DIFF
--- a/cli/lib/kontena/cli/nodes/remove_command.rb
+++ b/cli/lib/kontena/cli/nodes/remove_command.rb
@@ -10,6 +10,13 @@ module Kontena::Cli::Nodes
       require_api_url
       require_current_grid
       token = require_token
+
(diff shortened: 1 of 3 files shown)


SpComb:
Unsure how this affects the `kontena <provider> node terminate` commands. Will the terminated nodes be offline before they are removed?

jakolehm:
They most likely wont be offline when cloud provider says that machine is terminated. So if we take this in we need to change plugins to wait for offline status.

SpComb:
> They most likely wont be offline when cloud provider says that machine is terminated. So if we take this in we need to change plugins to wait for offline status.

I could drop the server mutations validation, and just keep the CLI validation? The plugins use the API directly.

jakolehm:
Sounds like a plan.

PR #2095, opened 2017-04-06
Do nothing if secret value does not change on update

### LINKED ISSUE #2094: Secret update triggers update of linked service even value does not change

jakolehm:
Fixes #2094

Also removes manual deployment triggering, it's not needed anymore because scheduler will notice timestamp change.

### DIFF
--- a/server/app/mutations/grid_secrets/update.rb
+++ b/server/app/mutations/grid_secrets/update.rb
@@ -10,6 +10,8 @@ module GridSecrets
     end
 
     def execute
+      return grid_secret if grid_secret.value == value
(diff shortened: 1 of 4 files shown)


server/app/mutations/grid_secrets/update.rb
```diff
@@ -10,6 +10,8 @@ class Update < Mutations::Command
     end
 
     def execute
+      return grid_secret if grid_secret.value == value
```
SpComb: Use the mongoid `grid_secret.changed?` after the assignment?

PR #2096, opened 2017-04-06
Use autoload to load missing constants

### LINKED ISSUE #2088: Cli lazy loading breaks some commands
Because they don't require all necessary files.

Example:
```
$ kontena stack logs redis-sentinel
 [error] NameError : uninitialized constant Kontena::Cli::Helpers
```

kke:
Fixes #2088 

Perhaps killing a bug with a cruise missile.


SpComb:
Feels a bit too big for 1.2, if we're going to release the RC today.

kke:
Some of the autoloads are also in #2335

PR #2097, opened 2017-04-06
Fix service update changes detection

### LINKED ISSUE #2090: GridService::Update mutation modifies embedded service_volumes always
Even there are no changes. This causes side-effects because it will mark service as modified.

SpComb:
Fixes  #2090

* Log all `GridServiceDeploy`s to show why they are happening
* Refactor `StackDeployWorker` to use `WaitHelper` for better logging
* Log changed `GridService` fields during `GridServices::Update`
* Spec `GridServices::Update#build_service_*` methods
* Fix `GridServices::Update` to preserve existing embedded `GridServiceLink` and `ServiceVolume` objects

## Issues

* #2109 Mongoid does not give `changed?` when removing embedded objects from the parent. This PR includes skipped specs for these cases.


server/app/mutations/grid_services/update.rb
```diff
@@ -46,6 +47,15 @@ def validate
       validate_volumes(self.volumes)
     end
 
+    # List changed fields of model
+    # @param document [Mongoid::Document]
+    # @return [String] field, embedded{field}
+    def changed(document)
```
SpComb: Should move this into some model-level place, it doesn't belong here.

jakolehm: Yes.

SpComb:
This still needs testing if the secrets/volumes/links removal cases still trigger `updated_at` on save, even though `changed?` does not bump the service `revision`.

SpComb:
The issue with removing embedded objects affects all of `changed?`, `revision` and also the `updated_at` timestamp, IOW, `stack upgrade` will not be able to remove secrets, links, volumes or hooks from service containers until a `kontena service deploy --force`.

SpComb:
@jakolehm merge this with the known issues requiring a `kontena service deploy --force`?

jakolehm:
@SpComb yes.

PR #2099, opened 2017-04-06
Volume show command & API

jnummelin:
```
$ k volume show foo
foo:
  id: test/foo
  created: 2017-04-05T22:11:34.201Z
  scope: instance
  driver: local
  driver_opts:
  instances:
    name: redis.foo-2
    node: moby
    name: redis.foo-1
    node: moby
  services:
    - test/null/redis

$ k volume show foobar
foobar:
  id: test/foobar
  created: 2017-04-05T22:33:32.327Z
  scope: instance
  driver: local
  driver_opts:
    foo: bar
  instances:
  services:
```

### DIFF
--- a/cli/lib/kontena/cli/volumes/show_command.rb
+++ b/cli/lib/kontena/cli/volumes/show_command.rb
@@ -1,0 +1,34 @@
+
+module Kontena::Cli::Volumes
+  class ShowCommand < Kontena::Command
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+
+    banner "Show details of a volume"
+
+    parameter 'VOLUME', 'Volume'
+
+    requires_current_master
+    requires_current_master_token
+
+    def execute
+      vol = client.get("volumes/#{current_grid}/#{volume}")
+      puts "#{vol['name']}:"
+      puts "  id: #{vol['id']}"
+      puts "  created: #{vol['created_at']}"
+      puts "  scope: #{vol['scope']}"
+      puts "  driver: #{vol['driver']}"
+      puts "  driver_opts:"
+      vol['driver_opts'].each do |k,v|
+        puts "    #{k}: #{v}"
+      end
+      puts "  instances:"
+      vol['instances'].each do |instance|
+        puts "    name: #{instance['name']}"
+        puts "    node: #{instance['node']}"
+      end
+
+    end
+
+  end
+end
--- a/cli/lib/kontena/cli/volumes/show_command.rb
+++ b/cli/lib/kontena/cli/volumes/show_command.rb
@@ -24,8 +24,8 @@ module Kontena::Cli::Volumes
       end
       puts "  instances:"
       vol['instances'].each do |instance|
-        puts "    name: #{instance['name']}"
-        puts "    node: #{instance['node']}"
+        puts "    - name: #{instance['name']}"
+        puts "      node: #{instance['node']}"
       end
       puts "  services:"
       vol['services'].each do |service|
--- a/cli/lib/kontena/cli/volumes/show_command.rb
+++ b/cli/lib/kontena/cli/volumes/show_command.rb
@@ -27,6 +27,10 @@ module Kontena::Cli::Volumes
         puts "    name: #{instance['name']}"
         puts "    node: #{instance['node']}"
       end
+      puts "  services:"
+      vol['services'].each do |service|
+        puts "    - #{service}"
+      end
 
     end
 
--- a/cli/lib/kontena/cli/volumes/show_command.rb
+++ b/cli/lib/kontena/cli/volumes/show_command.rb
@@ -29,7 +29,7 @@ module Kontena::Cli::Volumes
       end
       puts "  services:"
       vol['services'].each do |service|
-        puts "    - #{service}"
+        puts "    - #{service['id']}"
       end
 
     end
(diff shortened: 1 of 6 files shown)


server/app/models/volume.rb
```diff
@@ -34,4 +34,16 @@ def name_for_service(service, instance_number)
     end
   end
 
+  def services
+    services = []
+    self.grid.grid_services.each do |service|
+      service.service_volumes.each do |sv|
```
SpComb: Something like `service.service_volumes.where(volume: self)`? Is there any direct way to query embedded documents?

server/app/views/v1/volumes/_volume.json.jbuilder
```diff
@@ -5,3 +5,21 @@ json.name volume.name
...
 json.driver volume.driver
 json.driver_opts volume.driver_opts
+json.instances volume.volume_instances.map { |v|
+  {
+    node: v.host_node.name,
+    name: v.name
+  }
+}
+
+# services = []
```
SpComb: EDEADCODE

server/app/views/v1/volumes/_volume.json.jbuilder
```diff
@@ -5,3 +5,21 @@ json.name volume.name
...
+#
+# volume.grid.grid_services.each do |service|
+#   service.service_volumes.each do |sv|
+#     if sv.volume == volume
+#       services << service.to_path
+#     end
+#   end
+# end
+
+json.services volume.services.map {|s| s.to_path}
```
SpComb: This should be like `[{"id": "grid/stack/service"}]`

server/app/models/volume.rb
jnummelin: Well, not embedded docs themselves, but there's a way to query all services based on some attributes on the embedded doc values. Which actually work here nicely:
```
  def services
    self.grid.grid_services.where("service_volumes.volume_id" => self.id)
  end
```

server/app/views/v1/volumes/_volume.json.jbuilder
```diff
@@ -5,3 +5,13 @@ json.name volume.name
...
 json.driver volume.driver
 json.driver_opts volume.driver_opts
+json.instances volume.volume_instances.map { |v|
+  {
+    node: v.host_node.name,
+    name: v.name
+  }
+}
+
+json.services volume.services.map { |s|
```
SpComb: Shouldn't these jbuilder objects be like

```ruby
json.services volume.services.to_a do |volume_service|
  json.id volume_service.to_path
end
```

cli/lib/kontena/cli/volumes/show_command.rb
```diff
@@ -0,0 +1,38 @@
...
+      puts "  created: #{vol['created_at']}"
+      puts "  scope: #{vol['scope']}"
+      puts "  driver: #{vol['driver']}"
+      puts "  driver_opts:"
+      vol['driver_opts'].each do |k,v|
+        puts "    #{k}: #{v}"
+      end
+      puts "  instances:"
+      vol['instances'].each do |instance|
+        puts "    name: #{instance['name']}"
```
SpComb: This needs to be formatted as a list of YAML hashes, like

```yaml
  instances:
    - name: redis-data
      host: core-01
    - name: redis-data
      host: core-02
```

instead of the current

```
  instances:
    name: redis-data
    node: core-01
    name: redis-data
    node: core-02
```

jakolehm:
I think output would make more sense this way:
```
foo:
  id: test/foo
  created: 2017-04-05T22:11:34.201Z
  scope: instance
  driver: local
  driver_opts:
  instances:
    redis.foo-2:
      node: moby
      services:
        - test/null/redis
    redis.foo-1:
      node: moby
      services:
        - test/null/redis
```

SpComb:
> ```
> instances:
>   redis.foo-2:
>     node: moby
>     services:
>       - test/null/redis

If we go for this level of detail... then it would need to show all the service *instances* using that specific volume instance, like `test/null/redis-2`.

server/app/views/v1/volumes/_volume.json.jbuilder
jnummelin: both ways produce same results. But I'll change this to ^ this format for consistency sake.

server/app/views/v1/volumes/_volume.json.jbuilder
```diff
@@ -5,3 +5,10 @@ json.name volume.name
 json.scope volume.scope
 json.driver volume.driver
 json.driver_opts volume.driver_opts
+json.instances volume.volume_instances.each do |v|
+    json.node v.host_node.name
+    json.name v.name
+end
+json.services volume.services.each do |s|
```
jakolehm: Should we add `.includes(:some_models)` here so that it does not need to query related objects on each iteration?

jnummelin:
Currently there's no relation between `ServiceInstance` and `VolumeInstance`. I'd propose we add that in separate PR and then enhance the output.

server/app/views/v1/volumes/_volume.json.jbuilder
jnummelin: `volume.services` is a method call to already query all services which have a `ServiceVolume` that uses this volume:
```
  def services
    self.grid.grid_services.where("service_volumes.volume_id" => self.id)
  end
```
And it doesn't need any relations as least for now.

jnummelin:
Refactored the view to include relevant models in queries

PR #2108, opened 2017-04-07
CLI: mark volumes commands as experimental

SpComb:
2017-04-07-144735_1278x541_scrot

### DIFF
--- a/cli/lib/kontena/cli/volume_command.rb
+++ b/cli/lib/kontena/cli/volume_command.rb
@@ -1,6 +1,8 @@
 
 class Kontena::Cli::VolumeCommand < Kontena::Command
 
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/volume_command.rb
```diff
@@ -1,6 +1,8 @@
 
 class Kontena::Cli::VolumeCommand < Kontena::Command
 
+  warn Kontena.pastel.yellow('[EXPERIMENTAL] The `kontena volume` commands are still experimental in Kontena 1.2, and may change in future releases')
```
jakolehm: Should we get that version from `Kontena.version`?

PR #2124, opened 2017-04-10
Do not log entire yield value from wait_helper

### LINKED ISSUE #2110: Wait helper should truncate logged yield
Now it logs a "bit" too long lines for rpc responses.

jnummelin:
fixes #2110 

WaitHelper logged the whole value from `yield` which could leak secrets if it logs slow RPC calls for example.

This PR makes it log only the class name of the yielded value.


jnummelin:
@SpComb should we do the same on master side? AFAIK master does not anything sensitive using `wait_until`

jnummelin:
we should really try to find out a way to share this piece between agent and master, copy-pasta is just bad. :)

SpComb:
> we should really try to find out a way to share this piece between agent and master, copy-pasta is just bad. :)

Indeed. Until then we shouldn't let them diverge, or fixing that will become harder.

PR #2129, opened 2017-04-10
Migrate service instance also from volume containers

### LINKED ISSUE #2121: CreateGridServiceInstance migration should check also volume containers
Just in case, cleanup might have removed normal containers.

jakolehm:
Fixes #2121

### DIFF
--- a/server/db/migrations/20_create_grid_service_instance.rb
+++ b/server/db/migrations/20_create_grid_service_instance.rb
@@ -1,7 +1,12 @@
 class CreateGridServiceInstance < Mongodb::Migration
   def self.up
     GridServiceInstance.create_indexes
-    Container.unscoped.where(container_type: 'container').includes(:grid_service).each do |c|
+    migrate_containers('volume')
+    migrate_containers('container')
+  end
+
+  def self.migrate_containers(container_type)
+    Container.unscoped.where(container_type: container_type).includes(:grid_service).each do |c|
       if c.grid_service && c.grid_service.grid_service_instances.find_by(instance_number: c.instance_number).nil?
         c.grid_service.grid_service_instances.create!(
           host_node_id: c.host_node_id,
--- a/server/spec/db/migrations/20_create_grid_service_instance_spec.rb
+++ b/server/spec/db/migrations/20_create_grid_service_instance_spec.rb
@@ -1,0 +1,48 @@
+require_relative '../../../db/migrations/20_create_grid_service_instance'
+
+describe CreateGridServiceInstance do
+  let(:grid) { Grid.create!(name: 'test-grid') }
+  let(:node) do
+    HostNode.create!(
+      node_id: SecureRandom.uuid, grid: grid, connected: true
+    )
+  end
+  let(:service) do
+    GridService.create!(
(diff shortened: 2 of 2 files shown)


server/db/migrations/20_create_grid_service_instance.rb
```diff
@@ -1,7 +1,12 @@
 class CreateGridServiceInstance < Mongodb::Migration
   def self.up
     GridServiceInstance.create_indexes
-    Container.unscoped.where(container_type: 'container').includes(:grid_service).each do |c|
+    migrate_containers('volume')
+    migrate_containers('container')
```
SpComb: Thinking about ordering... can there be multiple volume/containers for a given service instance across different host nodes, and the `GridServiceInstance` might get be created with a different host node depending on the ordering of the containers here?

With `volume` containers first... could there be old volume containers left over from stateful services that got rescheduled away from an offline node, and this migration might revert the container from where it is currently running, back to some old node where it used to be running?

Should we migrate the `container` or `volume` types first, or sort by something like the `Container.updated_at`?

jakolehm: Only stateful services have volume containers. And stateful service instances have always been scheduled to a node where volume container is living. So this should be safe.

SpComb: I don't really understand the ordering here, because I'm not sure how the `volume` vs `container` types behaves in regards to having multiple `Container` objects for the same service instance on different host nodes... is there some specific reason to migrate both `volume` and `container` types separately, and in this specific order? As opposed to just `Container.unscoped.includes(:grid_service).each`?

jakolehm: The reason to do these in separate steps is to make it clear that volume is the primary source of scheduling decision.

jakolehm: And I don't really understand why this is a problem? :)

SpComb: > And I don't really understand why this is a problem? :)

I'm just unsure how it works exactly.  If there's multiple `volume` containers for the same instance on different nodes, but only one `container` type container, then this migration change could move the stateful container from the host node where it is currently running to some other host node where it used to be running?

jnummelin: In theory yes, but I'm not able to figure out a scenario where we would have ended up in situation where there is multiple volume containers for a given instance.

jakolehm: So can we approve this or not?

jnummelin: IMO yes

SpComb: I'll merge it if someone approves it, but I don't know if this change makes the migration better or worse.

jakolehm: It makes it better, reason: without actually checking volume containers we might miss information during migration. And this might cause service instance to be scheduled to a node that does not have data volume container -> data is lost.

PR #2135, opened 2017-04-10
Check volume driver match when ensuring volume existence

### LINKED ISSUE #2089: Agent should check that volume driver match before reusing it
Now it's possible that agent reuses a volume from a different driver if volume name matches.

jnummelin:
fixes #2089 

Volume manager did not check volume driver match when it checked volume existence. Now it does :)

## After
With existence of a expected volume but with a wrong driver:
```
$ k service deploy redis
⊗ Failed to deploy instance test/null/redis-1 to node moby: GridServiceInstanceDeployer::ServiceError: Kontena::Workers::Volumes::VolumeManager::DriverMismatchError: Volume driver not as expected. Expected foo, existing volume has local
```

### DIFF
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -149,7 +149,7 @@ module Kontena
         volume_manager = Celluloid::Actor[:volume_manager]
         service_pod.volumes.each do |volume|
           if volume['name']
-            return false unless volume_manager.volume_exist?(volume['name'])
+            return false unless volume_manager.volume_exist?(volume['name'], volume['driver'])
           end
         end
         true
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -70,9 +70,8 @@ module Kontena::Workers::Volumes
     # @param [Kontena::Models::Volume] volume
     def ensure_volume(volume)
       debug "ensuring volume existence: #{volume.inspect}"
-
       begin
-        unless __volume_exist?(volume.name, volume.driver)
+        unless volume_exist?(volume.name, volume.driver)
           info "creating volume"
           v = Docker::Volume.create(volume.name, {
             'Driver' => volume.driver,
@@ -103,7 +102,7 @@ module Kontena::Workers::Volumes
     # @param [String] name of the volume
     # @param [String] driver to expect on the volume if already existing
     # @raise [DriverMismatchError] If the volume is found but using a different driver than expected
-    def __volume_exist?(volume_name, driver)
+    def volume_exist?(volume_name, driver)
       begin
         debug "volume #{volume_name} exists"
         volume = Docker::Volume.get(volume_name)
@@ -115,12 +114,6 @@ module Kontena::Workers::Volumes
       rescue Docker::Error::NotFoundError
         debug "volume #{volume_name} does NOT exist"
         false
-      end
-    end
-
-    def volume_exist?(volume_name, driver)
-      begin
-        __volume_exist?(volume_name, driver)
       rescue => error
         abort error
       end
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -9,6 +9,9 @@ module Kontena::Workers::Volumes
     include Kontena::Helpers::WaitHelper
     include Kontena::Observer
 
+    class DriverMismatchError < StandardError
+    end
+
     attr_reader :node
 
     def initialize(autostart = true)
@@ -94,10 +97,20 @@ module Kontena::Workers::Volumes
       rpc_client.async.request('/node_volumes/set_state', [node.id, volume])
     end
 
-    def volume_exist?(volume_name)
+    # Checks if given volume exists with the expected driver
+    #
+    # @param [String] name of the volume
+    # @param [String] driver to expect on the volume if already existing
+    # @raise [DriverMismatchError] If the volume is found but using a different driver than expected
+    def volume_exist?(volume_name, driver)
       begin
         debug "volume #{volume_name} exists"
-        true if Docker::Volume.get(volume_name)
+        volume = Docker::Volume.get(volume_name)
+        if volume && volume.info['Driver'] == driver
+          return true
+        elsif volume && volume.info['Driver'] != driver
+          raise DriverMismatchError.new("Volume driver not as expected. Expected #{driver}, existing volume has #{volume.info['Driver']}")
+        end
       rescue Docker::Error::NotFoundError
         debug "volume #{volume_name} does NOT exist"
         false
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -70,16 +70,17 @@ module Kontena::Workers::Volumes
     # @param [Kontena::Models::Volume] volume
     def ensure_volume(volume)
       debug "ensuring volume existence: #{volume.inspect}"
+
       begin
-        Docker::Volume.get(volume.name)
-      rescue Docker::Error::NotFoundError
-        info "creating volume"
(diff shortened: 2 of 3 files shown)


agent/lib/kontena/service_pods/creator.rb
```diff
@@ -149,7 +149,7 @@ def volumes_exist?(service_pod)
         volume_manager = Celluloid::Actor[:volume_manager]
         service_pod.volumes.each do |volume|
           if volume['name']
-            return false unless volume_manager.volume_exist?(volume['name'])
```
SpComb: This will crash the `VolumeManager` actor if the volume driver mismatches, which isn't great. That's a external configuration error/conflict, and not something that restarting the `VolumeManager` will fix. The exception is good, though, because the `ServicePods::Create` should fail.

It seems like there's a [`Celluloid::abort`](http://www.rubydoc.info/gems/celluloid/Celluloid#abort-instance_method) that can be used to raise an error in the calling actor without crashing the manager actor. That would probably need a separate wrapper function in the `VolumeManager` to rescue and `abort` with the error?

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -94,10 +97,20 @@ def sync_volume_to_master(docker_volume)
...
+    # @raise [DriverMismatchError] If the volume is found but using a different driver than expected
+    def volume_exist?(volume_name, driver)
       begin
         debug "volume #{volume_name} exists"
-        true if Docker::Volume.get(volume_name)
+        volume = Docker::Volume.get(volume_name)
+        if volume && volume.info['Driver'] == driver
+          return true
+        elsif volume && volume.info['Driver'] != driver
+          raise DriverMismatchError.new("Volume driver not as expected. Expected #{driver}, existing volume has #{volume.info['Driver']}")
```
SpComb: I think the `ensure_volume` would be the right place to handle this error... change it to call `volume_exist?` and log it?

Then the agent will immediately warn about these volume conflicts earlier, and we could even later extend it with some kind of error reporting, and perhaps even have the scheduler avoid this node or something?

jnummelin:
Changed the `VolumeManager` not to crash on errors. In case of volume driver mismatch it will use Celluloid::abort to raise the error on caller side which eventually is seen on cli also:
```
$ k service deploy redis
⊗ Failed to deploy instance e2e/null/redis-1 to node moby: GridServiceInstanceDeployer::ServiceError: Kontena::Workers::Volumes::VolumeManager::DriverMismatchError: Volume driver not as expected. Expected foo, existing volume has local
```

On vol manager itself, it just logs the errors and goes on with it's business:
```
kontena-agent | D, [2017-04-12T09:06:12.782772 #1] DEBUG -- Kontena::Workers::Volumes::VolumeManager: ensuring volume existence: #<Kontena::Models::Volume:0x00557c66df1fc8 @volume_instance_id="58edeaf576f237020b000000", @name="foo", @labels={"io.kontena.volume_instance.id"=>"58edeaf576f237020b000000", "io.kontena.volume.id"=>"58edea4b76f237000d00003e"}, @driver="foo", @driver_opts={}>
kontena-agent | D, [2017-04-12T09:06:12.782817 #1] DEBUG -- Kontena::Workers::Volumes::VolumeManager: volume foo exists
kontena-agent | E, [2017-04-12T09:06:12.787997 #1] ERROR -- Kontena::Workers::Volumes::VolumeManager: Kontena::Workers::Volumes::VolumeManager::DriverMismatchError: Volume driver not as expected. Expected foo, existing volume has local
kontena-agent | E, [2017-04-12T09:06:12.788196 #1] ERROR -- Kontena::Workers::Volumes::VolumeManager: /app/lib/kontena/workers/volumes/volume_manager.rb:113:in `__volume_exist?'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:75:in `ensure_volume'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:56:in `block (2 levels) in populate_volumes_from_master'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:55:in `each'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:55:in `block in populate_volumes_from_master'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:40:in `populate_volumes_from_master'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:30:in `block in start'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:29:in `loop'
kontena-agent | /app/lib/kontena/workers/volumes/volume_manager.rb:29:in `start'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/async.rb:7:in `dispatch'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
kontena-agent | /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
```

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -94,16 +98,34 @@ def sync_volume_to_master(docker_volume)
...
+        end
       rescue Docker::Error::NotFoundError
         debug "volume #{volume_name} does NOT exist"
         false
       end
     end
 
+    def volume_exist?(volume_name, driver)
```
jakolehm: why do we need two methods?

SpComb: This is the non-crashing wrapper that does `Celluloid::abort` instead of crashing when called as a remote method... the internal `__volumes_exist?` is also called locally within this actor, and it should raise.

How does `Celluloid::abort` behave if called locally by the same actor? If it just raises, then I guess we don't need a wrapper function for this?

SpComb: Needs some comment mentioning that this is intended for external callers and will abort with `DriverMismatchError`.

jnummelin:
Did some local testing and `abort` is safe to use within the actor itself. So no need for wrapper method.

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -68,15 +71,15 @@ def populate_volumes_from_docker
...
+        unless volume_exist?(volume.name, volume.driver)
+          info "creating volume"
+          v = Docker::Volume.create(volume.name, {
+            'Driver' => volume.driver,
+            'DriverOpts' => volume.driver_opts,
+            'Labels' => volume.labels
+          })
+          sync_volume_to_master(v)
+        end
       rescue => exc
```
SpComb: I'm guessing that this wasn't rescuing `Docker::Volume.create` errors from the earlier `rescue Docker::Error::NotFoundError` before, leading to actor crashes?

jnummelin: yep

PR #2136, opened 2017-04-11
Use tty-table for volume ls

### LINKED ISSUE #2083: Cli: kontena volume ls cuts long volume names
```
$ kontena volume ls
NAME                      SCOPE                     DRIVER                    CREATED AT               
data-1a                   instance                  rexray/ebs                2017-04-05T18:06:45.302Z 
edge-redis-sentinel-data  instance                  rexray/ebs                2017-04-06T05:15:10.835Z 
edge-redis-sentinel-monit instance                  rexray/ebs                2017-04-06T05:15:19.683Z
```
(last one is `edge-redis-sentinel-monitor`)

jnummelin:
`kontena volume ls` cuts long volume names. This PR takes `tty-table` gem into use which provides dynamic-size table functionality and makes these tabular form cli output much nicer to write.

If this is ok to use, we can utilise it in other tabular command outputs.

fixes #2083 

## TODO

Test that it actually works on Win platforms too, some of the tty stuff has had bad track record in the past.

### DIFF
--- a/cli/lib/kontena_cli.rb
+++ b/cli/lib/kontena_cli.rb
@@ -107,7 +107,6 @@ end
 require 'ruby_dig'
 require 'shellwords'
 require "safe_yaml"
-require 'tty-table'
 SafeYAML::OPTIONS[:default_mode] = :safe
 require_relative 'kontena/cli/version'
 require_relative 'kontena/cli/common'
--- a/cli/lib/kontena_cli.rb
+++ b/cli/lib/kontena_cli.rb
@@ -107,6 +107,7 @@ end
 require 'ruby_dig'
 require 'shellwords'
 require "safe_yaml"
+require 'tty-table'
 SafeYAML::OPTIONS[:default_mode] = :safe
 require_relative 'kontena/cli/version'
 require_relative 'kontena/cli/common'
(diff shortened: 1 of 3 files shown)


kke:
Would be awesome to have some common decorator thing. 

Something like: 
`decoputs "Using grid :grid", grid: current_master.grid` (would colorize grid with some predefined grid color)

```
puts Kontena::Decorate([{name: 'foo', status: 'online'}, {name: 'bar', status: 'offline'}])
+------+-------+
| name | status|
+------+-------+
| foo  | online|
...
```

I don't expect tty-table to have any issues with windows. The earlier windows issues were in tty-prompt when reading keypresses, now I think it's one of the only ruby libraries that can do that properly in windows (using win32api).

The windows terminal does not display unicode properly, so perhaps we need `Kontena.glyph(:circled_dot)` which would give `+` or `⊙` depending on `use_unicode?`, but this PR seems to use the `:basic` renderer which doesn't draw any graphics.

cli/lib/kontena_cli.rb
```diff
@@ -107,6 +107,7 @@ def colorize(color_sym)
 require 'ruby_dig'
 require 'shellwords'
 require "safe_yaml"
+require 'tty-table'
```
kke: Perhaps better to require it where it's needed to avoid increasing spin up lag.

jnummelin:
@kke moved the require to `volume ls` command, PTAL.

PR #2138, opened 2017-04-11
Fix agent to raise on service container start, stop, restart errors

### LINKED ISSUE #2134: Agent does not report Docker container start errors
This is because it uses the [`Docker::Container#start`](https://github.com/swipely/docker-api/blob/v1.32.0/lib/docker/container.rb#L196-L203) method, which ignores any of the errors raised by `Docker::Container#start!`:

```ruby
  # #start! and #kill! both perform the associated action and
  # return the Container. #start and #kill do the same,
  # but rescue from ServerErrors.
  [:start, :kill].each do |method|
    define_method(method) do |*args|
      begin; public_send(:"#{method}!", *args); rescue ServerError; self end
    end
end
```

...who thought that was a good idea?

SpComb:
Fixes #2134

* Have the agent raise Docker API errors on service pod actions

* Add a spec for service deploy errors using `-v /dev/null/wtf:/dev/wtf`, which will fail to start

### Before

Testing with `kontena service create -v /dev/null/wtf:/dev/wtf test-fail redis`:

#### `kontena service deploy test-fail`
```
[error] Kontena::Errors::StandardErrorArray : halting deploy of e2e/null/test-fail, one or more instances failed:
	⊗ Failed to deploy instance e2e/null/test-fail-1 to node localhost: GridServiceInstanceDeployer::StateError: Service instance is not running, but stopped
```

#### `kontena service show test-fail`
```
e2e/null/test-fail:
  ...
  desired_state: running
  ...
  instances:
    test-fail/1:
      ...
      state: stopped
      containers:
        test-fail-1 (on localhost):
          ...
          status: stopped
          reason: invalid header field value "oci runtime error: container_linux.go:247: starting container process caused \"process_linux.go:359: container init caused \\\"rootfs_linux.go:53: mounting \\\\\\\"/dev/null/wtf\\\\\\\" to rootfs \\\\\\\"/var/lib/docker/overlay/11737453df3e9f59f1c0d8e97c4d4fb833d4b160a864d00b91b6f7f14455cb4a/merged\\\\\\\" at \\\\\\\"/dev/wtf\\\\\\\" caused \\\\\\\"stat /dev/null/wtf: not a directory\\\\\\\"\\\"\"\n"
          exit code: 128
```

#### `kontena service deploy test-fail`

```
 [error] Kontena::Errors::StandardErrorArray : halting deploy of e2e/null/test-fail, one or more instances failed:
	⊗ Failed to deploy instance e2e/null/test-fail-1 to node localhost: GridServiceInstanceDeployer::ServiceError: Docker::Error::ServerError: invalid header field value "oci runtime error: container_linux.go:247: starting container process caused \"process_linux.go:359: container init caused \\\"rootfs_linux.go:53: mounting \\\\\\\"/dev/null/wtf\\\\\\\" to rootfs \\\\\\\"/var/lib/docker/overlay/98aafa6d430feaa89c3aea9e9df186b3717243230fc613dc25beec85970bb941/merged\\\\\\\" at \\\\\\\"/dev/wtf\\\\\\\" caused \\\\\\\"stat /dev/null/wtf: not a directory\\\\\\\"\\\"\"\n"
```

### DIFF
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -64,8 +64,8 @@ module Kontena
           Celluloid::Notifications.publish('lb:remove_config', service_container.service_name_for_lb)
         end
 
-        service_container.start
-        info "service started: #{service_pod.name_for_humans}"
+        out = service_container.start!
+        info "service started: #{service_pod.name_for_humans}: #{out}"
         log_service_pod_event("service:create_instance", "service #{service_pod.name_for_humans} instance started")
 
         Celluloid::Notifications.publish('service_pod:start', service_pod.name)
--- a/agent/lib/kontena/service_pods/restarter.rb
+++ b/agent/lib/kontena/service_pods/restarter.rb
@@ -22,7 +22,7 @@ module Kontena
         service_container = get_container(self.service_id, self.instance_number)
         if service_container.running?
           info "restarting service: #{service_container.name_for_humans}"
-          service_container.restart('timeout' => 10)
+          service_container.restart!('timeout' => 10)
           log_service_pod_event(
             self.service_id, self.instance_number,
             "service:restart_instance", "service #{service_container.name_for_humans} instance restarted successfully"
--- a/agent/lib/kontena/service_pods/starter.rb
+++ b/agent/lib/kontena/service_pods/starter.rb
@@ -26,7 +26,7 @@ module Kontena
             self.service_id, self.instance_number,
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
-          service_container.restart('timeout' => 10)
+          service_container.restart!('timeout' => 10)
           log_service_pod_event(
             self.service_id, self.instance_number,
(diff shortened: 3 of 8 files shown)


SpComb:
I figure this should be fine re the `ServicePodWorker`, which can handle these errors, but I'm not so sure about the `Kontena::Workers::ContainerHealthCheckWorker#restart_container`, it's now more likely to raise an exception within the `Celluloid::defer { ... }` thread that it's called from... any side-effects from that?

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -64,8 +64,8 @@ def perform
           Celluloid::Notifications.publish('lb:remove_config', service_container.service_name_for_lb)
         end
 
-        service_container.start
-        info "service started: #{service_pod.name_for_humans}"
+        out = service_container.start!
+        info "service started: #{service_pod.name_for_humans}: #{out}"
```
SpComb: This extra logging was only actually meant for development testing...

PR #2140, opened 2017-04-11
Reduce already initialized constant warnings in api client

kke:
The client library is loaded multiple times in some cases. Defining the constants with `||=` will get rid of `already initialized constant` warnings.

Before:
image
After:
image


SpComb:
Shouldn't ruby's `require` only load the class once? How is kosh loading it that's causing it to get loaded twice?

kke:
First assumption:

require is somehow gem scoped.

Gem 1 loads Kontena::Client
Gem 2 loads Kontena::Client and gets warnings.

kke:
I think it's `require` vs `require_relative` and perhaps some symlinks.

https://bugs.ruby-lang.org/issues/10222

SpComb:
> I think it's require vs require_relative and perhaps some symlinks.

The CLI code seems to use a mixture of both `require` and `require_relative` to load it... could we avoid this issue by fixing those up?

```ruby
./lib/kontena/cli/services/services_helper.rb:require 'kontena/client'
./lib/kontena/scripts/completer.rb:require 'kontena/client'
./lib/kontena/stacks_client.rb:require_relative 'client'
./lib/kontena_cli.rb:require_relative 'kontena/client'
```

jakolehm:
`require` and `require_relative` only loads file once but cli plugin manager loads plugins with a `load`, see: https://github.com/kontena/kontena/blob/master/cli/lib/kontena/plugin_manager.rb#L170

kke:
Plugin manager however does not load kontena/client using `load`.

jakolehm:
Yes but if even one of those plugins is requiring `kontena/client` then it will trigger multiple requires?

kke:
No.

require and require_relative look in $LOADED_FEATURES (`$"`) for an absolute path. If that path is there, the file will not be loaded.

With symlinks or perhaps with something like `require_relative '../foo'` the absolute path may differ in different contexts.

kke:
In my case the kontena/client is loaded first by 'kontena_cli.rb' using `require_relative` and the second time by 'services_helper' using `require`.

First load thinks the path is:
 /usr/local/Cellar/rbenv/1.1.0/versions/2.4.0/lib/ruby/gems/2.4.0/gems/kontena-cli-1.1.5/lib/kontena/client.rb
Second load thinks the path is:
 /usr/local/opt/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/kontena-cli-1.1.5/lib/kontena/client.rb

The `Pathname.new(X).realpath` for both point to /usr/local/Cellar/rbenv/1.1.0/versions/2.4.0/lib/ruby/gems/2.4.0/gems/kontena-cli-1.1.5/lib/kontena/client.rb

jakolehm:
Maybe we should fix those requires then?

PR #2141, opened 2017-04-11
Bring scheduler node offline grace period back

jakolehm:
Previous versions had a grace period, this PR brings that functionality back to 1.2 series.


SpComb:
@jakolehm The `docs/using-kontena/nodes.md` also needs an update after this change.

SpComb:
@jakolehm The `docs/using-kontena/nodes.md` also needs an update after this change.

Issue #1948, opened 2017-03-07
Nodes randomly go offline and don't reconnect..


colinmollenhour:
I have a master that is in one datacenter and seven nodes in another datacenter. The `kontena node ls` command shows currently two nodes offline. Restarting kontena-agent service on the nodes causes them to reconnect and come back online, but as the connection is not actually broken they should have reconnected themselves.

```
$ kontena node ls
  Name                                                                   Version    Status     Initial    Labels
⊗ app2                                                                   1.1.2      online     1 / 4      region=chi,app,frontend
⊝ app1                                                                   1.1.2      offline    2 / 4      region=chi,app,frontend
⊝ admin1                                                                 1.1.2      offline    3 / 4      region=chi,admin,app,nfs
⊗ db2                                                                    1.1.2      online     4 / 4      region=chi,db,db-node
⊛ app3                                                                   1.1.2      online     -          region=chi,app,frontend
⊛ db3                                                                    1.1.2      online     -          region=chi,db,db-node
⊛ db1                                                                    1.1.2      online     -          region=chi,db,db-node
```

```
$ docker logs --tail 100 kontena-agent
----------------/// Lots of lines like:
I, [2017-03-07T12:58:36.632891 #1]  INFO -- Kontena::ServicePods::Creator: creating service: redis-session.lb-2
I, [2017-03-07T12:58:37.228617 #1]  INFO -- Kontena::Workers::ImagePullWorker: pulled image: quay.io/xxx/xxx:galera-lb-latest
----------------/// and then this:
I, [2017-03-07T12:59:08.453206 #1]  INFO -- Kontena::WebsocketClient: connection closed with code: 1000
I, [2017-03-07T12:59:08.555961 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://mgr.sys.bss-llc.com
I, [2017-03-07T12:59:13.710580 #1]  INFO -- Kontena::WebsocketClient: connection established
E, [2017-03-07T12:59:18.333473 #1] ERROR -- Kontena::NetworkAdapters::WeaveExecutor: weaveexec exit 1: ["--local", "launch-router", "--ipalloc-range", "", "--dns-domain", "kontena.local", "--password", "<redacted>", "--trusted-subnets", "10.36.18.0/24"]
weave is already running; you can stop it with 'weave stop-router'.

[...]

PR #2142, opened 2017-04-11
Improve websocket timeouts and node connection open/close logging

SpComb:
Helps diagnose node connectivity issues like in #1948

The default 2s keepalive ping timeout on the agent websocket might also be a bit optimistic, and it leads to to grid instability if the agents start bouncing.

* Bump default websocket ping timeouts to 5s, configurable via `KONTENA_WEBSOCKET_TIMEOUT`

* Log error with timeout delay if timeout exceeded
* Log warning with ping delay if threshold (timeout / 2) exceeded
* Log debug with ping delay

### Server
* Log warnings when marking stale nodes as disconnected
* Improve websocket timeout, close and node unplug handling and logging

  * Log different warnings when disconnecting a client, vs when a client disconnects

### Agent
* Fix agent `LogWorker` to also stop log streaming when the server closes the connection

## TODO
* Better server websocket specs


agent/lib/kontena/websocket_client.rb
```diff
@@ -15,7 +15,8 @@ module Kontena
   class WebsocketClient
     include Kontena::Logging
 
-    KEEPALIVE_TIME = 30
+    KEEPALIVE_INTERVAL = 30.0 # seconds
+    PING_TIMEOUT = 5.0 # seconds
```
SpComb: Should this be a configurable `KONTENA_WEBSOCKET_TIMEOUT` env?

server/app/jobs/node_cleanup_job.rb
```diff
@@ -25,7 +25,8 @@ def perform
   end
 
   def cleanup_stale_connections
-    HostNode.where(:last_seen_at.lt => NODE_DISCONNECT_TIMEOUT.ago).each do |node|
+    HostNode.where(:connected => true, :last_seen_at.lt => NODE_DISCONNECT_TIMEOUT.ago).each do |node|
```
SpComb: Is this `HostNode.where(:connected => true)` legit? Before, it was re-running this loop on all disconnected nodes every minute.

SpComb:
Still working on better websocket specs, but otherwise reviewable.

agent/spec/spec_helper.rb
```diff
@@ -74,6 +74,11 @@
     end
   end
 
+  config.around :each do |ex|
+    Timeout.timeout(5.0) do
```
jakolehm: Why we need this?

agent/lib/kontena/websocket_client.rb
```diff
@@ -171,6 +172,7 @@ def on_close(event)
         handle_invalid_version
       end
       info "connection closed with code #{event.code}"
+      notify_actors('websocket:close', nil)
```
SpComb: The agent was not notifying anything if the server closed the connection, as opposed to the `websocket:disconnect` notification when the agent keepalive ping times out and the agent decides to close the connection.

agent/lib/kontena/workers/log_worker.rb
```diff
@@ -24,8 +24,9 @@ def initialize(autostart = true)
       @workers = {}
       @etcd = Etcd.client(host: '127.0.0.1', port: 2379)
       subscribe('container:event', :on_container_event)
-      subscribe('websocket:connected', :on_connect)
+      subscribe('websocket:connected', :on_connect) # from master_info RPC
       subscribe('websocket:disconnect', :on_disconnect)
+      subscribe('websocket:close', :on_disconnect)
```
SpComb: The `on_disconnect` logs something, but the `async.stop` doesn't actually do anything unless it's `queue_processing?`.

server/app/middlewares/websocket_backend.rb
```diff
@@ -9,7 +9,9 @@ class WebsocketBackend
   WATCHDOG_THRESHOLD = 1.0.seconds
   WATCHDOG_TIMEOUT = 60.0.seconds
 
-  KEEPALIVE_TIME = 30 # in seconds
+  KEEPALIVE_TIME = 30.seconds
+  PING_TIMEOUT = Kernel::Float(ENV['KONTENA_WEBSOCKET_TIMEOUT'] || 5.seconds)
```
jakolehm: Server envs don't have `KONTENA_` prefix.. should we remove it here?

agent/spec/spec_helper.rb
SpComb: The EM specs need to wait for some expected method to get called before they can `EM.stop`. If that method doesn't get called, then the specs just time out. Typically that happens because when some spec double expectation fails, the `WebsocketClient` rescues and logs it as an error, without propagating it out and letting `EM.run { ... }` fail.

server/app/middlewares/websocket_backend.rb
SpComb: Should I also have the agent use `WEBSOCKET_TIMEOUT`?

agent/spec/spec_helper.rb
SpComb: So this is a safeguard that broken specs don't just hang indefinitely, but fail with an error. If the specs just hang, then it's very hard to tell from e.g. the travis logs which spec got stuck.

agent/lib/kontena/websocket_client.rb
jnummelin: what's the difference from other actors point-of-view between `websocket:close` and `websocket:disconnect`? I think in both cases they should stop trying to send data to master, right?

server/app/middlewares/websocket_backend.rb
jnummelin: +1 for `WEBSOCKET_TIMEOUT` on both sides

server/app/services/logging.rb
```diff
@@ -2,16 +2,10 @@
...
-    else
-      @logger = Logger.new(log_target)
-      @logger.progname = 'API'
-      @logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
-    end
-    @logger
+  def self.initialize_logger(log_target = STDOUT, log_level = Logger::INFO)
+    @logger = Logger.new(log_target)
+    @logger.progname = 'API'
+    @logger.level = ENV["DEBUG"] ? Logger::DEBUG : log_level
```
jnummelin: missing "return" of `@logger`

agent/lib/kontena/websocket_client.rb
SpComb: The `websocket:disconnect` happens immediately when the agent decides that it wants to `close` the websocket. The `websocket:close` from `on :close` might come 30s later, and any messages that are sent between those two are presumably just going to get lost.

OTOH if the server shuts down or closes the connection, then there is never going to be a `websocket:disconnect`, but now there is going to be a `websocket:close`.

Are the notification names are confusing enough? There's also `websocket:connect` vs `websocket:connected`, which are very different and come from different places.. :)

SpComb:
Having some trouble with my WIP server `WebsocketBackend` specs, needs more and better debug logging to figure out what's going wrong.

agent/lib/kontena/websocket_client.rb
jnummelin: Yeah, from ws point of view they are different things. I'm just thinking if the other actors need to know these details?

server/app/jobs/node_cleanup_job.rb
jnummelin: It still runs this loop every minute?

server/app/middlewares/websocket_backend.rb
```diff
@@ -168,26 +172,40 @@ def handle_rpc_notification(ws, data)
...
-      if node
-        Agent::NodeUnplugger.new(node).unplug!
-        logger.info "node closed connection: #{node.name || node.node_id}"
-      end
-      @clients.delete(client)
+      logger.info "node #{client[:id]} connection closed"
+      unplug_client(client)
+    else
+      logger.debug "ignore close of unplugged client"
     end
```
jnummelin: Why don't this call `ws.close` anymore?

server/app/jobs/node_cleanup_job.rb
SpComb: After it sets the `HostNode.connected = false`, it shouldn't re-log the `Disconnecting stale node ...` warning anymore for each `HostNode`.

The reason for this change is to avoid re-logging that new warning every minute for each disconnected node.

jnummelin: > Now it only runs it once when first marking a connected node as disconnected.
Got me sidetracked to think I missed something about the loop itself :)

server/app/middlewares/websocket_backend.rb
SpComb: This used to call `ws.close` because the ping timeout handler called `on_close` to disconnect the node. Those now call `close_client`, so now this is only called from the `ws.on :close`.

I don't think it makes any sense to close the websocket from the websocket close handler. I guess `Faye::Websocket` just used to ignore it in that case.

jnummelin: That's what I kind of figured, just wanted to make sure :)

agent/lib/kontena/websocket_client.rb
SpComb: You mean have both of these cases notify `websocket:close`? I suppose, but then it would just sometimes get two `websocket:close` notifications when disconnecting. Now it might get a `websocket:disconnect`, and then finally it will always get a `websocket:close`.

The only thing subscribing to these notifications is the `LogWorker`.

PR #2143, opened 2017-04-11
Improve agent resource usage

jakolehm:
With following changes:

- change `ServicePodManager` every to loop -> does not queue multiple tasks
- throttle worker creation
- `ServicePodWorker` only checks state from Docker Engine after it receives an event from Docker / or worker receives a new state/deploy_rev from manager

### DIFF
--- a/agent/lib/kontena/workers/service_pod_manager.rb
+++ b/agent/lib/kontena/workers/service_pod_manager.rb
@@ -32,8 +32,8 @@ module Kontena::Workers
       subscribe('service_pod:update', :on_update_notify)
       subscribe('service_pod:event', :on_pod_event)
       loop do
-        sleep 30
         populate_workers_from_master
+        sleep 30
       end
     end
 
@@ -70,6 +70,7 @@ module Kontena::Workers
       warn "failed to get list of service pods from master: #{exc}"
     rescue => exc
       error exc.message
+      error exc.backtrace.join("\n")
     end
 
     def populate_workers_from_docker
--- a/agent/lib/kontena/workers/service_pod_manager.rb
+++ b/agent/lib/kontena/workers/service_pod_manager.rb
@@ -31,7 +31,8 @@ module Kontena::Workers
 
       subscribe('service_pod:update', :on_update_notify)
       subscribe('service_pod:event', :on_pod_event)
-      every(30) do
+      loop do
+        sleep 30
         populate_workers_from_master
       end
     end
@@ -62,10 +63,13 @@ module Kontena::Workers
 
         service_pods.each do |s|
           ensure_service_worker(Kontena::Models::ServicePod.new(s))
+          sleep 0.05
         end
       }
     rescue Kontena::RpcClient::Error => exc
       warn "failed to get list of service pods from master: #{exc}"
+    rescue => exc
+      error exc.message
     end
 
     def populate_workers_from_docker
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -7,24 +7,49 @@ require_relative '../helpers/event_log_helper'
 module Kontena::Workers
   class ServicePodWorker
     include Celluloid
+    include Celluloid::Notifications
     include Kontena::Logging
     include Kontena::ServicePods::Common
     include Kontena::Helpers::RpcHelper
     include Kontena::Helpers::EventLogHelper
 
     attr_reader :node, :prev_state, :service_pod
-    attr_accessor :service_pod
+    attr_accessor :service_pod, :container_state_changed
 
     def initialize(node, service_pod)
       @node = node
       @service_pod = service_pod
       @prev_state = nil # sync'd to master
+      @container_state_changed = true
+      subscribe('container:event', :on_container_event)
     end
 
     # @param [Kontena::Models::ServicePod] service_pod
     def update(service_pod)
-      @service_pod = service_pod
-      apply
+      if needs_apply?(service_pod)
+        @service_pod = service_pod
+        apply
+      else
+        @service_pod = service_pod
+      end
+    end
+
+    # @param [Kontena::Models::ServicePod] service_pod
+    # @return [Boolean]
+    def needs_apply?(service_pod)
+      @container_state_changed ||
+        @service_pod.desired_state != service_pod.desired_state ||
+          @service_pod.deploy_rev != service_pod.deploy_rev
+    end
+
+    # @param [String] topic
+    # @param [Docker::Event] e
+    def on_container_event(topic, e)
+      attrs = e.actor.attributes
+      if attrs['io.kontena.service.id'] == @service_pod.service_id &&
+          attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number
+        @container_state_changed = true
+      end
     end
 
     def destroy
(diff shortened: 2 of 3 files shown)


agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -62,10 +63,13 @@ def populate_workers_from_master
 
         service_pods.each do |s|
           ensure_service_worker(Kontena::Models::ServicePod.new(s))
+          sleep 0.05
```
SpComb: 20 service pods / second limit

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -62,10 +63,13 @@ def populate_workers_from_master
...
         end
       }
     rescue Kontena::RpcClient::Error => exc
       warn "failed to get list of service pods from master: #{exc}"
+    rescue => exc
+      error exc.message
```
SpComb: What errors is this rescuing? Should it crash and restart instead?

At least log the backtrace.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -40,6 +65,7 @@ def apply
           warn "failed to sync #{service_pod.name} at #{service_pod.deploy_rev}: #{error}"
           sync_state_to_master(current_state, error)
         else
+          @container_state_changed = false
```
SpComb: BTW this `ensure_desired_state` -> `@container_state_changed = false` is going to immediately get followed by a `on_container_event` `@container_state_changed = true` if the ensure did something:

```
I, [2017-04-12T07:28:20.429746 #1]  INFO -- Kontena::ServicePods::Creator: service started: redis-1
D, [2017-04-12T07:28:20.501609 #1] DEBUG -- Kontena::Workers::ServicePodWorker: mark as changed on event: create
D, [2017-04-12T07:28:20.503124 #1] DEBUG -- Kontena::Workers::ServicePodWorker: mark as changed on event: start
```

Then the second time around, I'll rerun `ensure_desired_state`, but that shouldn't do anything, so it'll settle down:

```
D, [2017-04-12T07:28:43.772623 #1] DEBUG -- Kontena::Workers::ServicePodWorker: apply with changes
D, [2017-04-12T07:28:43.772901 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-1: running
D, [2017-04-12T07:28:43.820727 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state is in-sync: running
```

So it's not broken, just... behaves in a slightly surprising way.

SpComb: This also kinda breaks without #2138 to raise on container start errors:

`kontena service create -v /dev/null/wtf:/dev/wtf test-fail redis`
`kontena service deploy test-fail`

```
D, [2017-04-12T07:34:56.511908 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of test-fail-1: running
I, [2017-04-12T07:34:56.514946 #1]  INFO -- Kontena::Workers::ServicePodWorker: creating test-fail-1
I, [2017-04-12T07:34:56.515191 #1]  INFO -- Kontena::ServicePods::Creator: creating service: test-fail-1

D, [2017-04-12T07:34:59.044802 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: test-fail-1
D, [2017-04-12T07:34:59.093530 #1] DEBUG -- Kontena::ServicePods::Creator: container created: test-fail-1
I, [2017-04-12T07:34:59.365626 #1]  INFO -- Kontena::ServicePods::Creator: service started: test-fail-1
D, [2017-04-12T07:34:59.366491 #1] DEBUG -- Kontena::Workers::ServicePodWorker: test-fail-1 mark as un-changed after ensure without error
D, [2017-04-12T07:34:59.385622 #1] DEBUG -- Kontena::Workers::ServicePodWorker: test-fail-1 mark as changed on event: create

D, [2017-04-12T07:35:10.759087 #1] DEBUG -- Kontena::Workers::ServicePodWorker: test-fail-1 update with changes -> apply
D, [2017-04-12T07:35:10.759671 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of test-fail-1: running
I, [2017-04-12T07:35:10.804600 #1]  INFO -- Kontena::Workers::ServicePodWorker: starting test-fail-1
I, [2017-04-12T07:35:10.856632 #1]  INFO -- Kontena::ServicePods::Starter: starting service: /test-fail-1
D, [2017-04-12T07:35:11.071737 #1] DEBUG -- Kontena::Workers::ServicePodWorker: test-fail-1 mark as un-changed after ensure without error

D, [2017-04-12T07:35:41.083002 #1] DEBUG -- Kontena::Workers::ServicePodWorker: test-fail-1 update without changes
```

And the `test-fail-1` container is and remains stopped, the worker will not try and re-start it anymore.

SpComb: > This also kinda breaks without #2138 to raise on container start errors:

It's not fatally broken, though, because the `kontena service deploy` still fails with a `GridServiceInstanceDeployer::StateError: Service instance is not running, but stopped`, and a further `kontena service deploy` will still bump the `deploy_rev` and have the worker attempt to re-start it.

SpComb: Logging used here:

```patch
diff --git a/agent/lib/kontena/workers/service_pod_worker.rb b/agent/lib/kontena/workers/service_pod_worker.rb
index c1e8a7f..f514ee7 100644
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -27,9 +27,11 @@ module Kontena::Workers
     # @param [Kontena::Models::ServicePod] service_pod
     def update(service_pod)
       if needs_apply?(service_pod)
+        debug "#{@service_pod.name} update with changes -> apply"
         @service_pod = service_pod
         apply
       else
+        debug "#{@service_pod.name} update without changes"
         @service_pod = service_pod
       end
     end
@@ -48,6 +50,7 @@ module Kontena::Workers
       attrs = e.actor.attributes
       if attrs['io.kontena.service.id'] == @service_pod.service_id &&
           attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number
+        debug "#{@service_pod.name} mark as changed on event: #{e.action}"
         @container_state_changed = true
       end
     end
@@ -65,6 +68,8 @@ module Kontena::Workers
           warn "failed to sync #{service_pod.name} at #{service_pod.deploy_rev}: #{error}"
           sync_state_to_master(current_state, error)
         else
+          debug "#{@service_pod.name} mark as un-changed after ensure without error"
+
           @container_state_changed = false
           sync_state_to_master(current_state)
```

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -31,7 +31,8 @@ def start
 
       subscribe('service_pod:update', :on_update_notify)
       subscribe('service_pod:event', :on_pod_event)
-      every(30) do
```
jakolehm: Maybe it's safe to revert back to `every`.  There is still exclusive block that should block if something weird happens with task scheduling.

jakolehm: Actually we should probably do similar loop here as in volume manager.. or change volume manager to do `every`. /cc @jnummelin

agent/lib/kontena/workers/service_pod_manager.rb
jakolehm: Just wanted to make sure that this does not crash whole manager because now we are using loop. I'll add backtrace.

SpComb: It would crash the whole actor even with the `every` loop?

jakolehm: Oh, right. Confusing this with concurrent-ruby where TimerTask does not terminate actor.

PR #2144, opened 2017-04-11
Fix websocket node connection plug, unplug races

### LINKED ISSUE #1948: Nodes randomly go offline and don't reconnect..

SpComb:
Fixes #1948

Multiple server workers handling agent websocket connections can race between `NodePlugger.plugin!` and `NodeUnplugger.unplug!` leaving the agent websocket connection active, but the `HostNode` disconnected: https://github.com/kontena/kontena/issues/1948#issuecomment-293255702

Includes an `Kontena-Connected-At` header in the websocket connection with an increasing timestamp, and check that against the stored `HostNode.connected_at` timestamp when plugging and unplugging.


server/app/services/agent/node_plugger.rb
```diff
@@ -29,7 +31,15 @@ def plugin!
     end
 
     def update_node
-      node.set(connected: true, last_seen_at: Time.now.utc)
+      result = HostNode.where(:id => node.id).any_of(
+        {:connected_at => nil}, {:connected_at.lt => @connected_at}
+      ).update(connected: true, last_seen_at: Time.now.utc, connected_at: @connected_at)
```
SpComb: Is this atomic?

server/app/models/host_node.rb
```diff
@@ -28,6 +28,7 @@ class HostNode
   field :last_seen_at, type: Time
   field :agent_version, type: String
   field :docker_version, type: String
+  field :connected_at, type: String
```
jnummelin: Why a `String`and not `Time`? This is a timestamp, right?

server/app/services/agent/node_plugger.rb
jnummelin: Depends on what you want to do. :) AFAIK this will fire first the query and after that the update. If you want that to be one single operation you'd have to use `find_and_modify`.

I'm not sure how this behaves now when `connected_at` is a string. Feels a bit fishy though. :)

SpComb:
Tentatively bumping this to 1.2.1. The existing 1.2 changes should be enough to mitigate #1948 by having the racing worker to close the disconnected agent's websocket connection, allowing the agent to reconnect.

OTOH if we delay this to 1.2.1, then we have to deal with 1.2.0 agents that don't send the new `Kontena-Connected-At` header.

SpComb:
Another relevant thing here is cases like  #2185, where multiple agents are connecting with the same Docker ID, so they also race eachother for the same `HostNode`. It would be useful to somehow be able to complain loudly about this case, making it easier to notice.

SpComb:
Yet another relevant case for the agent `Connected-At` timestamp is cases like  #2298, where the host and master node clocks are out of sync. The master/agent should check eachother's timestamp offsets when connecting to warn about clock offsets.

SpComb:
Added bi-directional `Kontena-Connected-At` headers, with logging of the time offsets.

```
I, [2017-05-16T14:42:45.489795 #1]  INFO -- Kontena::WebsocketClient: connecting to master ws://192.168.66.1:9292 at 2017-05-16 14:42:45 UTC
```
```
I, [2017-05-16T17:42:45.586195 #27170]  INFO -- WebsocketBackend: node CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q opened connection at 2017-05-16 14:42:45 UTC, 0.59s ago
```
```
I, [2017-05-16T14:42:45.579723 #1]  INFO -- Kontena::WebsocketClient: connection established with master in 0.09s at 2017-05-16 14:42:45 UTC, 0.58s ago
```

This information could be used to log warnings or abort the connection of the clock offsets exceed some threshold when the node connects... OTOH there should also be something to check them continuously... unfortunately the websocket `ping`'s can't be used for that, since the `faye-websocket-driver` does not expose the `ping` events to the receiver.

SpComb:
Added a check for clock offsets in the server `WebsocketBackend#on_open`

### Agent clock behind
```
I, [2017-05-16T15:25:01.211263 #1]  INFO -- Kontena::WebsocketClient: connecting to master ws://192.168.66.1:9292 at 2017-05-16 15:25:01 UTC
I, [2017-05-16T15:25:01.283493 #1]  INFO -- Kontena::WebsocketClient: connection established with master in 0.07s at 2017-05-16 15:28:14 UTC, -192.72s ago
```
```
W, [2017-05-16T18:27:58.096522 #32407]  WARN -- WebsocketBackend: node core-02 connected too far in the past at 2017-05-16 15:24:45 UTC, 192.80s ago
```
```
E, [2017-05-16T15:25:01.285769 #1] ERROR -- Kontena::WebsocketClient: master does not accept our clock: clock offset 192.80s
I, [2017-05-16T15:25:01.285904 #1]  INFO -- Kontena::WebsocketClient: connection closed with code 4020
Shutting down...
```

### Agent clock ahead
```
I, [2017-05-16T15:30:22.835415 #1]  INFO -- Kontena::WebsocketClient: connecting to master ws://192.168.66.1:9292 at 2017-05-16 15:30:22 UTC
I, [2017-05-16T15:30:22.950375 #1]  INFO -- Kontena::WebsocketClient: connection established with master in 0.11s at 2017-05-16 15:28:42 UTC, 100.44s ago
```
```
W, [2017-05-16T18:28:42.515161 #32411]  WARN -- WebsocketBackend: node core-02 connected too far in the future at 2017-05-16 15:30:22 UTC, 100.32s ahead
D, [2017-05-16T18:28:42.533886 #32411] DEBUG -- WebsocketBackend: ignore close of unplugged client
```
```
E, [2017-05-16T15:30:22.955711 #1] ERROR -- Kontena::WebsocketClient: master does not accept our clock: clock offset -100.32s
I, [2017-05-16T15:30:22.955864 #1]  INFO -- Kontena::WebsocketClient: connection closed with code 4020
Shutting down...
```

SpComb:
Added an atomic  `HostNode` check `connected_at` -> update `last_seen_at` to the server `WebsocketBackend` keepalive ping-pong loop. This can now detect and warn on connections from multiple agents with the same `HostNode.node_id`, which can be used to diagnose cases like #2185 (cloned host nodes with identical Docker IDs)

Running two `kontena-agent` containers on the same host node to simulate such a scenario:

```
D, [2017-05-16T19:38:36.396551 #11443] DEBUG -- WebsocketBackend: keepalive ping 0.01s of 5.00s timeout from client XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5
D, [2017-05-16T19:38:36.932144 #11451] DEBUG -- WebsocketBackend: keepalive ping 0.00s of 5.00s timeout from client CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q
D, [2017-05-16T19:38:37.934792 #11447] DEBUG -- WebsocketBackend: keepalive ping 0.00s of 5.00s timeout from client CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q
W, [2017-05-16T19:38:37.943616 #11447]  WARN -- WebsocketBackend: Close conflicting connection for node core-02 connected at 2017-05-16 16:37:52 UTC, node has re-connected at 2017-05-16T16:38:10+00:00
I, [2017-05-16T19:38:37.947275 #11447]  INFO -- Agent::NodeUnplugger: disconnect node core-02
W, [2017-05-16T19:38:37.954438 #11447]  WARN -- Agent::NodeUnplugger: Skip node development/core-02 connected at 2017-05-16 16:37:52 UTC, because node has re-connected at 2017-05-16T16:38:10+00:00
D, [2017-05-16T19:38:37.974977 #11447] DEBUG -- WebsocketBackend: ignore close of unplugged client
```

```
W, [2017-05-16T16:38:37.983234 #1]  WARN -- Kontena::WebsocketClient: connection closed with code 4041: host node core-02 connection conflict with new connection at 2017-05-16T16:38:10+00:00
```

However, there's still some work here.. the server `WebsocketBackend@clients` behaves oddly in terms of RPCs if there are multiple clients for the same host node in the same puma worker.

agent/lib/kontena/websocket_client.rb
```diff
@@ -126,10 +129,20 @@ def send_request(id, method, params)
...
+      server_version = ws.headers['Kontena-Version']
+
       @connecting = false
-      notify_actors('websocket:open', event)
+      connecting_dt = Time.now - @connecting_at
+      @connected = true
+      @connected_at = Time.parse(@ws.headers['Kontena-Connected-At'])
+      connected_dt = Time.now - @connected_at
+
+      info "connected to master #{@api_uri} server version #{server_version} in #{'%.2fs' % connecting_dt} at #{@connected_at}, #{'%.2fs' % connected_dt} ago"
```
SpComb: The `@connecting_at` is the agent timestamp, and the `@connected_at` is the server timestamp, so this attempts to compare the two to log any clock offsets.

agent/lib/kontena/websocket_client.rb
```diff
@@ -166,12 +179,15 @@ def on_close(event)
...
-      if event.code == 4001
+
+      case event.code
+      when 4001
         handle_invalid_token
-      elsif event.code == 4010
-        handle_invalid_version
+      when 4010
+        handle_invalid_version(event.reason)
+      else
```
SpComb: What other codes should be handled specially? Should the agent abort on 404x close events as well (conflicts on the server `HostNode` state)? It should definitely just reconnect on the 4030 ping timeout, though.

agent/lib/kontena/websocket_client.rb
```diff
@@ -126,10 +129,20 @@ def send_request(id, method, params)
...
-      info 'connection established'
-      @connected = true
+
+      server_version = ws.headers['Kontena-Version']
+
       @connecting = false
-      notify_actors('websocket:open', event)
+      connecting_dt = Time.now - @connecting_at
+      @connected = true
+      @connected_at = Time.parse(@ws.headers['Kontena-Connected-At'])
```
SpComb: This is not backwards-compatible with old servers... they will still accept the request and open up the websocket without sending any `Kontena-Connected-At` headers. Need to deal with that case... likely to end up in a `on :close, code: 4010` very soon.

agent/lib/kontena/websocket_client.rb
```diff
@@ -182,9 +198,9 @@ def handle_invalid_token
       EM.next_tick { abort('Shutting down ...') }
     end
 
-    def handle_invalid_version
+    def handle_invalid_version(reason)
       agent_version = Kontena::Agent::VERSION
-      error "master does not accept our version (#{agent_version}), shutting down ..."
+      error "master does not accept our version: #{reason}"
```
SpComb: Also not backwards-compatible with older server version... they don't send a `close reason: ...`.

server/app/middlewares/websocket_backend.rb
```diff
@@ -9,8 +9,10 @@ class WebsocketBackend
   WATCHDOG_THRESHOLD = 1.0.seconds
   WATCHDOG_TIMEOUT = 60.0.seconds
 
+  STRFTIME = '%F %T.%NZ'
   KEEPALIVE_TIME = 30.seconds
   PING_TIMEOUT = Kernel::Float(ENV['WEBSOCKET_TIMEOUT'] || 5.seconds)
+  CLOCK_SKEW = Kernel::Float(ENV['KONTENA_CLOCK_SKEW'] || 1.seconds)
```
SpComb: The acceptable clock skew into the future can be very small (1s), but the acceptable clock skew into the past is `CLOCK_SKEW + PING_TIMEOUT`, because the websocket open will always have some network delay.

server/app/middlewares/websocket_backend.rb
```diff
@@ -36,10 +38,29 @@ def initialize(app)
...
+    Gem::Dependency.new('', "~> #{self.our_version}").match?('', agent_version)
+  end
+
   def call(env)
     if Faye::WebSocket.websocket?(env)
       req = Rack::Request.new(env)
-      ws = Faye::WebSocket.new(env)
+      ws = Faye::WebSocket.new(env, nil,
+        headers: {
+          'Kontena-Version' => Server::VERSION,
```
SpComb: Considering using the websocket request headers to replace the `/agent/master_info` RPC, by having the request handler lookup and validate the node etc before actually opening the websocket.

The agent would use this to update its `/etc/kontena.env`?

server/app/middlewares/websocket_backend.rb
```diff
@@ -66,43 +87,84 @@ def call(env)
...
-    logger.error "#{exc.class.name}: #{exc.message}"
+    logger.error exc
+  end
+
+  # @param [Faye::WebSocket] ws
+  # @param [HostNode] node
+  def handle_invalid_agent_version(ws, node, version)
+    logger.warn "node #{node} agent version #{version} is not compatible with server version #{Server::VERSION}"
+
+    ws.close(4010, "agent version #{version} is not compatible with server version #{Server::VERSION}")
```
SpComb: This is missing the `node_plugger.send_master_info` + `EventMachine::Timer.new(1) do ws.close` that was there before...

server/app/middlewares/websocket_backend.rb
```diff
@@ -66,43 +87,84 @@ def call(env)
...
+  # Add to @clients, and mark client HostNode as connected
+  #
+  # @param [Hash] client
+  # @param [HostNode] node
+  def plugin_client(client, node)
+    @clients << client
+
+    node_plugger = Agent::NodePlugger.new(node, client[:connected_at])
+
+    EM.defer { node_plugger.plugin! }
```
SpComb: The `EM.defer { ... }` here is problematic, because if the `NodePlugger#update_node` detects a newer `connected_at`, then this should really fail the connection immediately, and not append it to `@clients` etc... but with the async call, that is difficult.

server/app/services/agent/node_plugger.rb
```diff
@@ -31,7 +30,15 @@ def plugin!
...
     def update_node
-      node.set(connected: true, last_seen_at: Time.now.utc)
+      result = HostNode.where(:id => node.id).any_of(
+        {:connected_at => nil}, {:connected_at.lt => @connected_at}
+      ).find_and_modify({:$set => {connected: true, last_seen_at: Time.now.utc, connected_at: @connected_at}}, {new: true})
+
+      if result
+        info "Connected node #{@node.to_path} at #{@connected_at}"
+      else
+        warn "Skip node #{@node.to_path} connect at #{@connected_at}, because node has re-connected at #{@node.connected_at}"
```
SpComb: This should probably also reject the websocket connection, but that is not possible to do right now.

server/app/models/host_node.rb
```diff
@@ -28,6 +28,7 @@ class HostNode
   field :last_seen_at, type: Time
   field :agent_version, type: String
   field :docker_version, type: String
+  field :connected_at, type: DateTime
```
SpComb: The `:connected_at.lt => Time.parse(...)` comparisons do not seem to work at all if this is `type: Time`.

server/app/middlewares/websocket_backend.rb
```diff
@@ -246,6 +246,7 @@ def our_version
   # @param [HostNode] node
   def handle_invalid_agent_version(ws, node)
     node.set(connected: false, last_seen_at: Time.now.utc)
+    # XXX: delay to give the Agent::NodePlugger.send_master_info -> MongoPubSub time to call send_message before the websocket gets closed... hopefully?
     EventMachine::Timer.new(1) do
       ws.close(4010) if ws
```
SpComb: If I'm reading this right, this feels pretty evil... if the `Agent::NodePlugger.send_master_info  -> `MongoPubSub` is slow, the websocket might get closed before the `/agent/master_info` RPC gets sent.

SpComb:
I backed out the breaking websocket protocol changes. Also fixed up the specs to now pass... but the server `WebsocketBackend#on_open` doesn't have any, so I need to write some :confused: 

@jakolehm @jnummelin does the `Kontena-Connected-At`/`HostNode.connected_at` design look reasonable, with the new `CLOCK_SKEW` checks and close on node connection conflicts during keepalive checks?

Also started some [WIP branch](https://github.com/kontena/kontena/compare/fix/server-websocket-node-unplug-race...feature/server-hostnode-websocket-health) to report the `WebsocketBackend` connection errors to the CLI, but that still needs more changes, want to get these in first.

server/app/middlewares/websocket_backend.rb
```diff
@@ -335,25 +369,33 @@ def on_pong(client, delay)
...
-        node.set(last_seen_at: Time.now.utc)
-      else
-        logger.warn "Close connection of disconnected node #{node.name || node.node_id}"
-        close_client(client)
+      connected_node = HostNode.where(id: node.id, connected_at: client[:connected_at])
+        .find_one_and_update({:$set => {last_seen_at: Time.now.utc}})
+
+      if !connected_node
+        logger.warn "Close conflicting connection for node #{node} connected at #{client[:connected_at]}, node has re-connected at #{node.connected_at}"
+        close_client(client, 4041, "host node #{node} connection conflict with new connection at #{node.connected_at}")
```
SpComb: This check should trigger if there are multiple active connections with the same Docker/Node ID, such as for cloned VMs. Tested this by running two agents on the same host node.

How those agents will then behave is rather odd, it depends on whether or not they connect to the same server puma worker or not... If they both connect to the same worker, then both clients end up in the same `WebsocketBackend@clients` array, and the `client_for_id` used to send RPCs to the agents will pick the first one to have connected... so after a reconnect, they don't get their `master_info` or `node_info` RPCs, those go to the conflicting agent.

server/app/services/agent/node_plugger.rb
```diff
@@ -1,37 +1,35 @@
...
 
-    def update_node
-      node.set(connected: true, last_seen_at: Time.now.utc)
+    # @raise [RuntimeError] Node ... has already re-connected at ...
+    def update_node!(connected_at)
+      connected_node = HostNode.where(:id => node.id)
+        .any_of({:connected_at => nil}, {:connected_at.lt => connected_at})
+        .find_one_and_update({:$set => {connected: true, last_seen_at: Time.now.utc, connected_at: connected_at}})
+
+      fail "Node #{@node} has already re-connected at #{@node.connected_at}" unless connected_node
```
SpComb: If this triggers, the conflicting websocket connection still gets opened, but the agent just never gets the `master_info`... then on the next keepalive interval, it will get closed by the `WebsocketBackend`. Ugly as F***

SpComb: There's possibly also some race condition here, if the `WebsocketBackend#verify_client_connection ping` -> `on_pong` happens after the `on :connect`, but before the `EM.defer { node_plugger.plugin! }` runs... probably rather unlikely.

server/app/middlewares/websocket_backend.rb
```diff
@@ -66,43 +73,70 @@ def call(env)
...
-      logger.error 'invalid grid token, closing connection'
-      ws.close(4001)
+    node_plugger = Agent::NodePlugger.new(node)
+
+    # check version
+    agent_version = req.env['HTTP_KONTENA_VERSION'].to_s
+
+    unless self.valid_agent_version?(agent_version)
+      logger.warn "node #{node} agent version #{version} is not compatible with server version #{Server::VERSION}"
+      node_plugger.send_master_info
```
SpComb: Broken because the websocket is not yet in `@clients`, so the MongoPubsub RPC does not get sent.

SpComb: Fixed by sending the `/agent/master_info` RPC directly within the same `WebsocketBackend` EM thread, without going through the `NodePlugger` or `MongoPubsub`.

SpComb:
The server `WebsocketBackend` should have reasonable spec coverage now.

agent/lib/kontena/websocket_client.rb
jnummelin: I think it makes sense to abort on 404x conditions. I.e. the docker ID conflict will not resolve by itself and a hint in logs would help a lot.

jnummelin:
> does the Kontena-Connected-At/HostNode.connected_at design look reasonable, with the new CLOCK_SKEW checks and close on node connection conflicts during keepalive checks?

IMO yes. The `CLOCK_SKEW` check is nice addition, as we saw in one case skewed clocks create quite a bit of mayhem :)

server/app/middlewares/websocket_backend.rb
```diff
@@ -66,43 +73,67 @@ def call(env)
...
+
+    unless self.valid_agent_version?(agent_version)
+      logger.warn "node #{node} agent version #{agent_version} is not compatible with server version #{Server::VERSION}"
+      handle_invalid_agent_version(ws, node, agent_version)
+      return
     end
+
+    # check clock after version check, because older agent versions do not send this header
+    connected_at = Time.parse(req.env['HTTP_KONTENA_CONNECTED_AT'])
+    connected_dt = Time.now - connected_at
```
jnummelin: Should the server side connection time be "recorded" earlier? Before hitting this check there has been quite a few DB operations in the path. On a busy master that might cause enough delay to give false clock skew disconnects.

jnummelin: Maybe not a real issue as the tolerated skew in the past is 6s (CLOCK_SKEW+PING_TIMEOUT)

server/app/middlewares/websocket_backend.rb
```diff
@@ -236,19 +267,30 @@ def our_version
...
+
+  # Must be called in EM thread.
+  #
+  # @param [Faye::Websocket] ws
   # @param [Array] message
   def send_message(ws, message)
-    EM.next_tick {
-      ws.send(MessagePack.dump(message).bytes)
-    }
+    ws.send(MessagePack.dump(message).bytes)
```
jakolehm: Why did we remove `next_tick`?

server/app/middlewares/websocket_backend.rb
```diff
@@ -335,25 +377,33 @@ def on_pong(client, delay)
...
     node = HostNode.find_by(node_id: client[:id])
+
     if node
-      if node.connected?
-        node.set(last_seen_at: Time.now.utc)
-      else
-        logger.warn "Close connection of disconnected node #{node.name || node.node_id}"
-        close_client(client)
+      connected_node = HostNode.where(id: node.id, connected_at: client[:connected_at])
+        .find_one_and_update({:$set => {last_seen_at: Time.now.utc}})
```
jakolehm: Rebase needed because Mongoid 5 behaves differently.

SpComb: Which part? I changed this to use `find_one_and_update`?

jakolehm: Oh, sorry.. comment is in a wrong place.. just look for `{new: true}` , we need to change that with latest mongoid.

SpComb: Found it in the `NodeUnplugger`, fixed.

server/app/middlewares/websocket_backend.rb
SpComb: Because this is now called from `handle_invalid_agent_version` -> `send_master_info` -> `send_rpc_notify` immediately before calling `ws.close`, so the `ws.send` needs to happen immediately. With the `EM.next_tick` here, the `ws.close` in the invalid agent version case would happen before the `ws.send` of the `master_info`.

The only call to `WebsocketBackend#send_message` that I can find is in the local `on_rpc_message`, and that got changed to use `EM.next_tick` in #2056, so this `ws.send` was actually being deferred twice.

The `RpcServer` has its own `send_message` method that uses `EM.next_tick` and calls `ws.send` directly.

jakolehm: Maybe it's ok then, it's just smells a bit that caller needs to know this detail. Maybe this method should be protected/private?

SpComb: It has a comment that says it must be called from the EM thread... also tried making almost all of the methods private, but then that makes the specs harder :)

jnummelin: You can always do some monkey patching during specs if private methods hinder you speccing:
```
subject.singleton_class.class_eval{public :foo}
expect(subject.foo).to eq('foo')
```

Of course that's not too nice and leaves the possibility of specs passing but actual code failing to call private methods.

jnummelin: @SpComb Planning to work on this area still or should we just rely on the comment on the method?

SpComb: I'm not sure how to use `protected/private` for this... there's a lot of other methods that have similar "only call in the EM thread" semantics, but marking all of those as private/protected gets ugly quickly with the specs, which call things like the `on_open` etc methods directly.

SpComb: Tested the new `handle_invalid_agent_version` logic to make sure it  still works:

```
core@core-02 ~/share $ cat VERSION 
1.2.2.dev
$ cat kontena.env 
KONTENA_VERSION=1.2.2.dev
$ docker-compose run agent
...
I, [2017-05-26T12:22:04.557960 #1]  INFO -- Kontena::WebsocketClient: connecting to master at ws://192.168.66.1:9292
I, [2017-05-26T12:22:04.680311 #1]  INFO -- Kontena::WebsocketClient: connection established
E, [2017-05-26T12:22:04.685189 #1] ERROR -- Kontena::WebsocketClient: master does not accept our version (1.2.2.dev): agent version 1.2.2.dev is not compatible with server version 1.3.0.pre1
D, [2017-05-26T12:22:04.685908 #1] DEBUG -- Kontena::RpcServer: rpc notification: Kontena::Rpc::AgentApi#master_info [{"version"=>"1.3.0.pre1"}]
D, [2017-05-26T12:22:04.687231 #1] DEBUG -- Kontena::Workers::NodeInfoWorker: publishing node information
I, [2017-05-26T12:22:04.686754 #1]  INFO -- Kontena::Workers::LogWorker: stopped log streaming
I, [2017-05-26T12:22:04.696142 #1]  INFO -- Kontena::Workers::LogWorker: started log streaming
Shutting down ...
I, [2017-05-26T12:22:04.854881 #1]  INFO -- Kontena::Workers::ContainerHealthCheckWorker: stopped to check status from /ingress-lb.lb-2
$ cat kontena.env 
KONTENA_VERSION=1.3.0.pre1
```

It seems to work now, without the magic 1s timer for `ws.close` that was there before.

jnummelin: True, the comment is enough in this case then. Better to have well tested code.

PR #2146, opened 2017-04-11
Add service pod caching on Rpc::NodeServicePodHandler

jnummelin:
The serializer does quite many queries (~30) while "transforming" the service instance into pod hash to be sent to agent. This slows down the pod rpc quite much and causes significant load into the server and mongo. To avoid unnecessary serialization and thus queries the pod hashes are now cache in [LruRedux](https://github.com/SamSaffron/lru_redux) cache.

The cache key used is a composite of `service_instance.id`, `service_instance.desired_state` and `service_instance.deploy_rev`. That should make the cached pod hash unique per deployment. Desired state is needed since `GridServiceInstance.desired_state` can change `running` -> `stopped` -> `running` in the `GridServiceInstanceDeployer` if it moves the instance to a different node while the `deploy_rev` remains.

With the caching in place, the rpc call execution time dropped from ~80ms to ~8ms for single node setup with ~30 instances running. (Docker for Mac)

At the moment there's no cache cleanup happening but the size should limit memory usage nicely as the cache will evict least recently used items as the cache get full. The cleanup would have to be TTL based with some backgroud worker doing the cleanup. `ObjectSpace.memsize_of` shows about 2kb mem usage per pod hash so should not make mem usage grow that much even if we don't clean it up in the background.

### DIFF
--- a/server/app/services/rpc/node_service_pod_handler.rb
+++ b/server/app/services/rpc/node_service_pod_handler.rb
@@ -12,7 +12,8 @@ module Rpc
       # TODO Do we need any other attrbutes for the composed key?
       cache_key = {
         id: service_instance.id,
-        deploy_rev: service_instance.deploy_rev
+        deploy_rev: service_instance.deploy_rev,
+        desired_state: service_instance.desired_state
       }
       @@lru_cache.getset(cache_key) {
         debug "pod_cache miss"
--- a/server/app/services/rpc/node_service_pod_handler.rb
+++ b/server/app/services/rpc/node_service_pod_handler.rb
@@ -2,7 +2,7 @@ module Rpc
   class NodeServicePodHandler
     include Logging
 
-    @@lru_cache ||= LruRedux::ThreadSafeCache.new(100)
+    @@lru_cache ||= LruRedux::ThreadSafeCache.new(1000)
 
     def initialize(grid)
       @grid = grid
--- a/server/app/services/rpc/node_service_pod_handler.rb
+++ b/server/app/services/rpc/node_service_pod_handler.rb
@@ -4,7 +4,7 @@ module Rpc
 
     def initialize(grid)
       @grid = grid
-      @lru_cache ||= LruRedux::ThreadSafeCache.new(1000)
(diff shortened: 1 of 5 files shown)


server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -1,21 +1,39 @@
 module Rpc
   class NodeServicePodHandler
+    include Logging
+
+    @@lru_cache ||= LruRedux::ThreadSafeCache.new(1000)
```
jakolehm: Does this really need to be a class variable?

jnummelin: Doesn't have to be, but there can be multiple instances of `NodeServicePodHandler`, one per grid. So I thought it would keep the cache mem usage at minimum by having it as class variable.

jakolehm: Usually class variables are a bit code smell. Just look how you need to access this in specs. If we really need to have this at class level maybe then just class instance variable with proper accessor method?

But I think cache would be more predictable if we just create cache object per grid instance.

jnummelin: sure, couple megs cache per grid should not be too big "overhead". Doubt many people are actually operating many grids per master anyway, usually only one or couple at max.

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -1,21 +1,38 @@
 module Rpc
   class NodeServicePodHandler
+    include Logging
 
     def initialize(grid)
       @grid = grid
+      @lru_cache ||= LruRedux::ThreadSafeCache.new(1000)
```
jakolehm: Just `=`

SpComb:
This does have the side-effect of `GridService` updates no longer being visible to all agents until the service actually gets deployed. Good or bad?

Some agents might get newer service revisions while others are still on old cached service revisions, but that can probably also happen ATM.

jnummelin:
AFAIK the existing model has the same feature.

PR #2147, opened 2017-04-12
ContainerInfoWorker fixes

jakolehm:
- worker starts to send container info only after websocket is connected
- coroner now sweeps all containers and sends list of removed containers to master

### DIFF
--- a/agent/lib/kontena/actors/container_coroner.rb
+++ b/agent/lib/kontena/actors/container_coroner.rb
@@ -7,14 +7,15 @@ module Kontena::Actors
     INVESTIGATION_PERIOD = 20
 
     # @param [String] node_id
-    def initialize(node_id)
+    # @param [Boolean] autostart
+    def initialize(node_id, autostart = true)
       @node_id = node_id
-      @processing = false
+      async.process if autostart
     end
 
     def process
       prev = nil
-      while @processing do
+      loop do
         ids = all_containers.map { |c| c.id }
         if prev
           diff = prev - ids
@@ -39,23 +40,6 @@ module Kontena::Actors
     # @return [Array<Docker::Container>]
     def all_containers
       Docker::Container.all(all: true)
-    end
-
-    def start
-      return if @processing
-
-      @processing = true
-      async.process
-    end
-
-    def stop
-      return unless @processing
-
-      @processing = false
-    end
-
-    def processing?
-      @processing == true
     end
   end
 end
--- a/agent/lib/kontena/actors/container_coroner.rb
+++ b/agent/lib/kontena/actors/container_coroner.rb
@@ -29,7 +29,10 @@ module Kontena::Actors
     end
 
     def report(data)
-      rpc_client.async.request('/containers/cleanup', [@node_id, data])
+      rpc_request('/containers/cleanup', [@node_id, data])
+    rescue => exc
+      warn "failed to send report: #{exc.message}"
+      retry
     end
 
     # @return [Array<Docker::Container>]
@@ -48,6 +51,10 @@ module Kontena::Actors
       return unless @processing
 
       @processing = false
+    end
+
+    def processing?
+      @processing == true
     end
   end
 end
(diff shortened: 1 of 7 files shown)


server/app/services/rpc/container_handler.rb
```diff
@@ -32,6 +32,18 @@ def save(data)
...
 
+    # @param [String] node_id
+    # @param [Array<String>] ids
+    def cleanup(node_id, ids)
+      node = @grid.host_nodes.find_by(node_id: node_id)
+      if node
+        @grid.containers.where(
+          :host_node_id => node.id,
+          :container_id.in => ids
+        ).delete
```
jnummelin: delete vs. destroy?

This won't delete `type: 'volume` containers as the default scope limits only to "normal" containers. I think this should use  `@grid.containers.unscoped....`

agent/lib/kontena/actors/container_coroner.rb
```diff
@@ -2,53 +2,59 @@ module Kontena::Actors
...
-    def investigate
-      exists = Docker::Container.get(@container.id) rescue nil
-      unless exists
-        confirm
-      end
+    def report(data)
+      rpc_request('/containers/cleanup', [@node_id, data])
+    rescue => exc
+      warn "failed to send report: #{exc.message}"
+      retry
```
jnummelin: This might create while-true loop in some cases. maybe bit of back-off?

agent/lib/kontena/workers/container_info_worker.rb
```diff
@@ -38,22 +44,26 @@ def on_container_event(topic, event)
...
 
     def on_container_publish_info(topic, container)
-      self.publish_info(container)
+      publish_info(container)
     end
 
     def on_websocket_connected(topic, data)
-      self.publish_all_containers
+      publish_all_containers
+      container_coroner.start
```
jnummelin: Could this be missed in some cases? For example if `ContainerInfoWorker` crashes for some random reason, when it restarts it won't see the `websocket:connected` event at all and thus the coroner never gets started.

jakolehm: Yes, it could be missed. Any ideas how to improve?

server/app/services/rpc/container_handler.rb
jakolehm: Good point.

jakolehm:
@jnummelin did changes based on your review.

agent/lib/kontena/workers/container_info_worker.rb
jnummelin: I think we'd need to use the observer thingy to observe the state of the websocket or something. That way the observable will "notify" the observer in the startup. Not sure though if it warrants it's own PR as this same "flaw" was there even before and seems to be a bit bigger refactoring.

agent/lib/kontena/workers/container_info_worker.rb
```diff
@@ -7,29 +7,35 @@ class ContainerInfoWorker
...
 
-    attr_reader :queue
+    attr_reader :container_coroner
 
     # @param [Boolean] autostart
     def initialize(autostart = true)
       subscribe('container:event', :on_container_event)
       subscribe('container:publish_info', :on_container_publish_info)
       subscribe('websocket:connected', :on_websocket_connected)
+      subscribe('websocket:disconnect', :on_websocket_disconnect)
```
SpComb: Hmm... needs to also consider `websocket:close` from #2142

agent/lib/kontena/workers/container_info_worker.rb
```diff
@@ -38,22 +44,26 @@ def on_container_event(topic, event)
...
     end
 
     def on_websocket_connected(topic, data)
-      self.publish_all_containers
+      publish_all_containers
+      container_coroner.start
+    end
+
+    def on_websocket_disconnect(topic, data)
+      container_coroner.stop
```
SpComb: Why do we need to stop the coroner... wouldn't it be better to just let it sit in its `report` retry-loop until the websocket reconnects? That way it can report on those destroyed containers that we missed out publishing info for?

jakolehm:
@SpComb @jnummelin now coroner is working around the clock.

PR #2166, opened 2017-04-18
Refactor node stats to NodeStatsWorker

### LINKED ISSUE #2165: Unresolvable statsd endpoint crashes NodeInfoWorker
```
I, [2017-04-17T10:31:56.996052 #1]  INFO -- Kontena::Workers::NodeInfoWorker: exporting stats via statsd to udp://this-does-not-exist.kontena.local:8125
E, [2017-04-17T10:31:56.998093 #1] ERROR -- : Actor crashed!
SocketError: getaddrinfo: Name does not resolve
```

jakolehm:
NodeInfoWorker has grown too big, this PR splits stats related pieces to a separate worker.

Fixes #2165

### DIFF
--- a/agent/lib/kontena/workers/node_info_worker.rb
+++ b/agent/lib/kontena/workers/node_info_worker.rb
@@ -1,11 +1,8 @@
 require 'net/http'
-require 'vmstat'
-
 require_relative '../models/node'
 require_relative '../helpers/node_helper'
 require_relative '../helpers/iface_helper'
 require_relative '../helpers/rpc_helper'
-require_relative '../helpers/stats_helper'
 
 module Kontena::Workers
   class NodeInfoWorker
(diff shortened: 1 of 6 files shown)


agent/lib/kontena/workers/node_stats_worker.rb
```diff
@@ -0,0 +1,300 @@
...
+        @node = node
+      end
+
+      start_publish_loop
+    end
+
+    def start_publish_loop
+      loop do
+        sleep PUBLISH_INTERVAL
+        self.publish_node_stats if rpc_client.connected?
```
SpComb: The `if rpc_client.connected?` check is new? This also stops statsd metrics if the RPC is down.

jakolehm: Yes, it's new and good point that we probably should not stop statsd sending. I just tried to protect rpc_client here a bit.

jakolehm: Fixed.

agent/lib/kontena/workers/node_stats_worker.rb
```diff
@@ -116,7 +116,7 @@ def publish_node_stats
         network: network_traffic,
         time: Time.now.utc.to_s
       }
-      rpc_client.async.notification('/nodes/stats', [data])
+      rpc_client.async.notification('/nodes/stats', [data]) if rpc_client.connected?
```
SpComb: How much effect does the `... if rpc_client.connected?` have? The `RpcClient#notification` -> `WebsocketClient#send_message` already skips sending if the `@ws` is nil, and AFAICT that ends up being exactly the same thing as `rpc_client.connected?`... so this just introduces an extra sync actor call of overhead?

jakolehm: Again good point. Probably this check is not necessary at all.

agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -16,19 +13,13 @@ class NodeInfoWorker
...
     # @param [Boolean] autostart
     def initialize(autostart = true)
-      @statsd = nil
-      @stats_since = Time.now
-      @container_seconds = 0
-      @previous_cpu = Vmstat.cpu
-      @previous_network = Vmstat.network_interfaces
       subscribe('websocket:connected', :on_websocket_connected)
       subscribe('agent:node_info', :on_node_info)
       subscribe('container:event', :on_container_event)
```
SpComb: Unused now

PR #2172, opened 2017-04-18
Don't crash ImagePullWorker if pull fails

jakolehm:


SpComb:
The image pull error case seem to somehow still let the `Weave` actor crash:

```
I, [2017-04-18T12:29:36.815896 #1]  INFO -- Kontena::ServicePods::Creator: creating service: redis-fail-1
D, [2017-04-18T12:29:36.918174 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-2: running
W, [2017-04-18T12:29:36.976806 #1]  WARN -- Kontena::Workers::ServicePodWorker: failed to sync redis-fail-1 at 2017-04-18 12:28:45 UTC: No such image: redis:nonexist

E, [2017-04-18T12:29:36.976446 #1] ERROR -- : Actor crashed!
Docker::Error::NotFoundError: No such image: redis:nonexist

	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:46:in `rescue in request'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:38:in `request'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/image.rb:126:in `get'
	/app/lib/kontena/network_adapters/weave.rb:131:in `modify_create_opts'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
I, [2017-04-18T12:29:36.991937 #1]  INFO -- Kontena::NetworkAdapters::Weave: initialized
```

jakolehm:
@SpComb it raises error to the caller.. so probably it's caller responsibility to not crash here?

SpComb:
> @SpComb it raises error to the caller.. so probably it's caller responsibility to not crash here?

It looks like the first time it tries to create the `redis-fail` container, it aborts and the `ServicePods::Creator` fails, and does not continue to the `config_container` -> `Weave#modify_create_opts`. But the second time when the `ServicePodWorker` retries the create, the `ensure_image` does not fail, so it continues to the `Weave#modify_create_opts` which crashes on the missing image.

agent/lib/kontena/workers/image_pull_worker.rb
```diff
@@ -28,7 +28,8 @@ def pull_image(image, deploy_rev, creds)
       else
         info "pulling image with credentials: #{image}"
       end
-      update_image_cache("#{image}:#{deploy_rev}")
+      cache_key = "#{image}:#{deploy_rev}"
+      image_cache[cache_key] = Time.now.utc
```
SpComb: Setting the cache before pulling the image is still broken in the case of two concurrent instance deploys onto the same host node... there will be two `ServicePodWorker` actors running `ServicePods::Create` that both call into `ensure_image` with the same image and deploy_rev... while the first task sleeps, the second one will pick up the "cached" image and proceed to :boom:

agent/lib/kontena/workers/image_pull_worker.rb
```diff
@@ -4,6 +4,8 @@ class ImagePullWorker
     include Celluloid
     include Kontena::Logging
 
+    exclusive :ensure_image
```
SpComb: The exclusive/caching logic is split across a number of places and hard to follow, wouldn't something like this be equivalent and easier to follow?

```ruby
    def ensure_image(image, deploy_rev, creds = nil)
      exclusive {
        unless fresh_pull?("#{image}:#{deploy_rev}")
          pull_image(image, deploy_rev, creds)

          update_image_cache("#{image}:#{deploy_rev}")
        end
      }
    end
```

jakolehm: We should remove entry from the cache if pull fails, how would you handle it here?

SpComb: By not adding it if the `pull_image` fails with an exception.

jakolehm: Oh, right. I agree that your example is simpler. But we can probably refactor this later.

PR #2131, opened 2017-04-10
[WIP] Re-add dns addresses for containers after weave gets re-created

jnummelin:
A bit dirty but there's (or at least was) bigger refactoring pending on weave startup logic so didn't want to refactor too much.

fixes #2079

Issue #1923, opened 2017-03-02
On weave upgrades, the Agent will attach containers before re-launching 


SpComb:
This can lead to races between the `weaveexec attach` and `docker rm && weaveexec launch-router`

```
kontena-agent | I, [2017-03-02T09:50:02.348968 #1]  INFO -- Kontena::Workers::WeaveWorker: initialized
kontena-agent | D, [2017-03-02T09:50:02.817707 #1] DEBUG -- NetworkAdapters::WeaveExecutor: weaveexec stream: ["--local", "ps"]
kontena-agent | D, [2017-03-02T09:50:02.825089 #1] DEBUG -- Workers::WeaveWorker: Scanned 4 existing containers for potential migration: {"fe5af6db7c10"=>["10.81.128.92/16"], "a746a4256299"=>["10.81.128.5/16"], "ba04a7ced236"=>["10.81.128.28/16"], "653e4574f253"=>["10.81.128.3/16"]}
kontena-agent | I, [2017-03-02T09:50:02.825367 #1]  INFO -- Kontena::Workers::WeaveWorker: attaching network to existing containers
kontena-agent | D, [2017-03-02T09:50:02.831979 #1] DEBUG -- Workers::WeaveWorker: skip start for container=/kontena-agent without overlay_cidr
kontena-agent | I, [2017-03-02T09:50:02.863790 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=fe5af6db7c101a1314bfe68a7f020b11083378caa5a95e4db38580eb32af2559 at cidr=10.81.128.92/16
kontena-agent | I, [2017-03-02T09:50:02.964824 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=a746a42562999f5cbd9984e8aea6403f69f5dce783622f643e7514371b88f443 at cidr=10.81.128.5/16
kontena-agent | I, [2017-03-02T09:50:03.007447 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=ba04a7ced236046c8886fe852e5ffe12d977c3c3b480cc2dee547f0398129b37 at cidr=10.81.128.28/16
kontena-agent | D, [2017-03-02T09:50:03.018868 #1] DEBUG -- Workers::WeaveWorker: skip start for container=/kontena-cadvisor without overlay_cidr
[...]

PR #1398, opened 2016-11-24
Agent: Cache weave ps at WeaveWorker startup for container migration

SpComb:
Calling `weaveexec ps` for large numbers of containers is surprisingly slow, migrating >100 service containers can take ~6 minutes:

```
I, [2016-11-23T14:31:37.416612 #1]  INFO -- Kontena::NetworkAdapters::Weave: Migrate container=e1c4e9b032a37f407b2ccdcec84bd95a8aa1f068a2a27949a390d345ab739b50 to cidr=10.81.10.95/16
W, [2016-11-23T14:31:38.106023 #1]  WARN -- Kontena::NetworkAdapters::Weave: Migrate container=e1c4e9b032a37f407b2ccdcec84bd95a8aa1f068a2a27949a390d345ab739b50 from cidr=10.81.10.95/19
I, [2016-11-23T14:31:39.330565 #1] INFO -- Kontena::NetworkAdapters::Weave: Attach container=e1c4e9b032a37f407b2ccdcec84bd95a8aa1f068a2a27949a390d345ab739b50 at cidr=10.81.10.95/16
I, [2016-11-23T14:31:39.687663 #1]  INFO -- Kontena::NetworkAdapters::Weave: Migrate container=200bde7335ed9f258379316755b9dc6af6f9cb3cab80687e39660723990f1052 to cidr=10.81.10.92/16
W, [2016-11-23T14:31:40.497002 #1]  WARN -- Kontena::NetworkAdapters::Weave: Migrate container=200bde7335ed9f258379316755b9dc6af6f9cb3cab80687e39660723990f1052 from cidr=10.81.10.92/19
I, [2016-11-23T14:31:41.836367 #1] INFO -- Kontena::NetworkAdapters::Weave: Attach container=200bde7335ed9f258379316755b9dc6af6f9cb3cab80687e39660723990f1052 at cidr=10.81.10.92/16
...
I, [2016-11-23T14:37:13.900166 #1]  INFO -- Kontena::NetworkAdapters::Weave: Migrate container=577c45919691757dd6b86969d1cf4e778f330fd66201a40bda9c48755d06732b to cidr=10.81.1.8/16
W, [2016-11-23T14:37:14.480095 #1]  WARN -- Kontena::NetworkAdapters::Weave: Migrate container=577c45919691757dd6b86969d1cf4e778f330fd66201a40bda9c48755d06732b from cidr=10.81.1.8/19
[...]

PR #2173, opened 2017-04-18
Agent: WeaveWorker reload containers in both weave restart and re-create cases

### LINKED ISSUE #2158: Service *.kontena.local DNS aliases missing after kontena grid trusted-subnet changes
Using `kontena grid trusted-subnet add/remove` will re-create the weave container on each host node to update the `--trusted-subnet` configuration. This causes weave to forget all of the `$service.$grid.kontena.local` and `$service.$stack.kontena.local `DNS aliases. The agent currently does not re-register these DNS names when weave is re-created.

Workaround is to either `docker restart kontena-agent`, or `docker restart weave`, on each host node having the containers that are missing their service aliaes. This triggers a `kontena-agent` `WeaveWorker#start`, which re-registers each container's service aliases.

SpComb:
Fixes #2158

Extends the existing `WeaveWorker` logic that re-runs start` logic on all weave container `start` events, not just `restart` events. This fixes the missing container DNS names on weave re-configurations by both re-attaching all containers and re-registering their DNS names after the weave container is re-created.

The difference between the two is that `restart` events would only happen if the weave router crashes, and Docker restarts it. If the `Weave` worker decides to re-create the container to change the configuration, then it will not be a `restart` event, but rather a `destroy` event followed by `create` and `start`. Currently the attached containers remain attached, but weave forgets the DNS registrations.

There is some trickyness going on here, because this will avoid triggering on the initial `start` event when the `Weave` actor is first creating the `weave` router container, as it only subscribes to `container:event` notifications after the `Weave` actor signals `network_adapter:start`, so this logic should not kick in before that.

### DIFF
--- a/agent/lib/kontena/workers/weave_worker.rb
+++ b/agent/lib/kontena/workers/weave_worker.rb
@@ -29,6 +29,18 @@ module Kontena::Workers
       self.start
     end
 
+    # Ensure that all service containers are attached and registered in Weave DNS.
+    #
+    # Only called after the network_adapter is running?
+    #
+    # This is called in multiple cases:
+    # * the agent was installed, and weave was just launched... it is unlikely for there to be any deployed containers waiting to be attached yet
+    # * the agent was rebooted, and weave was restarted by Docker, but none of the autostarted containers are attached yet
+    # * the agent was restarted, so weave is still running, and the containers might already be attached
+    # * the agent was upgraded, so weave might have been re-created, and some containers might be attached with the wrong netmask
+    # * the worker has crashed and restarted, so we might have missed some container events, and some containers might not be attached
+    # * the weave router container has crashed and restarted, so it will have forgotten Weave DNS names, but the containers will still be attached
+    # * the weave router container was re-created to update the configuration, so it will have forgotten Weave DNS names, but the containers will still be attached
     def start
       @migrate_containers = network_adapter.get_containers
       debug "Scanned #{@migrate_containers.size} existing containers for potential migration: #{@migrate_containers}"
@@ -39,7 +51,6 @@ module Kontena::Workers
       Docker::Container.all(all: false).each do |container|
         self.start_container(container)
       end
-      
     end
 
     def on_dns_add(topic, event)
@@ -54,17 +65,16 @@ module Kontena::Workers
       return unless started?
 
       if event.status == 'start'
-        container = Docker::Container.get(event.id) rescue nil
-        if container
-          self.start_container(container)
-        else
-          warn "skip start event for missing container=#{event.id}"
-        end
-      elsif event.status == 'restart'
         if network_adapter.router_image?(event.from)
+          info "Restart after weave restart"
+
           wait_weave_running?
 
           self.start
+        elsif container = Docker::Container.get(event.id) rescue nil
+          self.start_container(container)
+        else
+          warn "skip start event for missing container=#{event.id}"
         end
       elsif event.status == 'destroy'
         self.on_container_destroy(event)
--- a/agent/spec/lib/kontena/workers/weave_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/weave_worker_spec.rb
@@ -28,6 +28,8 @@ describe Kontena::Workers::WeaveWorker do
     before(:each) do
       allow(network_adapter).to receive(:running?).and_return(true)
       allow(subject.wrapped_object).to receive(:started?).and_return(true)
+      allow(network_adapter).to receive(:router_image?).and_return(false)
+      allow(network_adapter).to receive(:router_image?).with('weaveworks/weave:1.4.5').and_return(true)
     end
 
     it 'calls #weave_attach on start event' do
@@ -42,9 +44,14 @@ describe Kontena::Workers::WeaveWorker do
       subject.on_container_event('topic', event)
     end
 
-    it 'calls #start on weave restart event' do
+    it 'does not call #start on weave restart event' do
       event = spy(:event, id: 'foobar', status: 'restart', from: 'weaveworks/weave:1.4.5')
-      expect(network_adapter).to receive(:router_image?).with('weaveworks/weave:1.4.5').and_return(true)
+      expect(subject.wrapped_object).not_to receive(:start)
+      subject.on_container_event('topic', event)
+    end
+
+    it 'calls #start on weave start event' do
+      event = spy(:event, id: 'foobar', status: 'start', from: 'weaveworks/weave:1.4.5')
       expect(subject.wrapped_object).to receive(:start).once
       subject.on_container_event('topic', event)
     end


agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -81,12 +81,6 @@ def adapter_image?(image)
       false
     end
 
-    def router_image?(image)
```
SpComb: Moved this to the `WeaveWorker`, because:

* It was the only caller for it 
* It's now called for every `container:event`

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -45,22 +45,31 @@ def on_dns_add(topic, event)
       add_dns(event[:id], event[:ip], event[:name])
     end
 
+    # Was container event from the weave router?
+    def router_image?(image)
+      image.to_s == "#{Kontena::NetworkAdapters::Weave::WEAVE_IMAGE}:#{Kontena::NetworkAdapters::Weave::WEAVE_VERSION}"
```
jakolehm: This looks ugly, maybe we should refactor `weave_image` etc methods to a helper?

SpComb: Moved those bits into the `WeaveHelper` module. Didn't touch the `WeaveExecutor`.

SpComb: There's more of these that need would need refactoring (the `EventWorker` also has a `Actor[:network_adapter].adapter_image?(event.from)` which should be fixed), but I don't want to bloat this PR any more...

SpComb:
The alternative to this would be to change the `Weave` actor to re-notify `network_adapter:start` after upgrades.

SpComb:
Split the `network_adapter.router_image?` and friends refactoring into #2207

This now just results in slightly more `network_adapter.router_image?` calls.

SpComb:
Compared to #2131, this is simpler, but it does more work, because it will also re-attach every container when re-creating weave. This also happens on weave restarts.

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -48,19 +48,21 @@ def on_dns_add(topic, event)
     # @param [String] topic
     # @param [Docker::Event] event
     def on_container_event(topic, event)
-      if event.status == 'start'
+      if network_adapter.router_image?(event.from)
```
jnummelin: This actor sync call will now be made for each container event, could it exhaust celluloid?

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -48,19 +48,21 @@ def on_dns_add(topic, event)
...
+        if event.status == 'start' || event.status == 'restart'
```
jnummelin: In case `docker restart xyz` you will actually get both `start` and `restart` events:
```
2017-04-26T12:42:16.871540235+03:00 container start 5266dd1ce9128826f16dace277110b263f79587519608c71cc21df6d0b19e450 (image=nginx:1-alpine, io.kontena.container.deploy_rev=2017-04-21 10:41:25 UTC, io.kontena.container.name=nginx-1, io.kontena.container.overlay_cidr=10.81.128.63/16, io.kontena.container.overlay_network=kontena, io.kontena.container.pod=nginx-1, io.kontena.container.service_revision=1, io.kontena.container.type=container, io.kontena.grid.name=e2e, io.kontena.service.id=58f9e1cf6cc0b100090004a9, io.kontena.service.instance_number=1, io.kontena.service.name=nginx, io.kontena.stack.name=null, name=nginx-1)
2017-04-26T12:42:16.871768064+03:00 container restart 5266dd1ce9128826f16dace277110b263f79587519608c71cc21df6d0b19e450 (image=nginx:1-alpine, io.kontena.container.deploy_rev=2017-04-21 10:41:25 UTC, io.kontena.container.name=nginx-1, io.kontena.container.overlay_cidr=10.81.128.63/16, io.kontena.container.overlay_network=kontena, io.kontena.container.pod=nginx-1, io.kontena.container.service_revision=1, io.kontena.container.type=container, io.kontena.grid.name=e2e, io.kontena.service.id=58f9e1cf6cc0b100090004a9, io.kontena.service.instance_number=1, io.kontena.service.name=nginx, io.kontena.stack.name=null, name=nginx-1)
```

SpComb: Seems like both restart and re-create cases will include a `start` event, so let's just stick to that.

agent/lib/kontena/workers/weave_worker.rb
SpComb: Refactored as a local method call in #2207... but changed this to only do this for `start` events now, so it shouldn't matter as much.

SpComb:
Noticed another problem with the existing `WeaveWorker' code: re-calling `start` will also re-subscribe to `container:event`, so the worker starts seeing everything twice :(

```
I, [2017-04-26T11:25:17.126175 #1]  INFO -- Kontena::Workers::WeaveWorker: Restart after weave restart
I, [2017-04-26T11:25:17.127935 #1]  INFO -- Kontena::Workers::WeaveWorker: Restart after weave restart
D, [2017-04-26T11:25:17.824282 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.7s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:25:17.985824 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.9s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:25:18.411287 #1] DEBUG -- Kontena::NetworkAdapters::WeaveExecutor: weaveexec stream: ["--local", "ps"]
D, [2017-04-26T11:25:19.016518 #1] DEBUG -- Kontena::NetworkAdapters::WeaveExecutor: weaveexec stream: ["--local", "ps"]
```

```
I, [2017-04-26T11:31:32.186142 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=067cc8d4aace21339d38f4bac43fff62b6aef7c55dee1b47fcfb164484a05772 at cidr=10.81.128.99/16 I, [2017-04-26T11:31:32.198249 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=067cc8d4aace21339d38f4bac43fff62b6aef7c55dee1b47fcfb164484a05772 at cidr=10.81.128.99/16
```

Affects existing agent code for weave `restart` events, introduced by #1398 in Kontena 1.0

SpComb:
The new behavior to react to `weave` `start` `container:event` notifications needs some consideration of how those are timed relative to the `network_adapter:start` notifications. If they were to arrive in the `network_adapter:start` -> `container:event` order, then this `WeaveWorker` would run `start_all_containers` twice. Not fatal, but best avoided?

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -49,17 +66,16 @@ def on_dns_add(topic, event)
...
-          warn "skip start event for missing container=#{event.id}"
-        end
-      elsif event.status == 'restart'
         if network_adapter.router_image?(event.from)
+          info "Restart after weave restart"
+
           wait_weave_running?
 
-          self.start
+          self.start_all_containers
```
SpComb: Note how the `on_container_event` now calls `start_all_containers`, instead of `start`, to avoid the #2225 re-`subscribe` issue.

jnummelin:
IMO we should try to avoid the possibility to run `start_all_containers` unnecessarily, it's pretty expensive when there are lot of containers.

SpComb:
> IMO we should try to avoid the possibility to run `start_all_containers` unnecessarily, it's pretty expensive when there are lot of containers.

In terms of the potential race condition between the `network_adapter:started` and the Docker weave `container:event` `start` notifications, the `Weave#launch` also runs other `weaveexec` containers between the `launch-router` that creates and starts the `weave` container, and before notifying `network_adapter:started`, so I would consider it highly unlikely for the Docker container weave `start` event to come after the `network_adapter:started` notification. Not impossible though, and I don't think there's any way to really avoid the races between the `Weave` actor `network_adapter:start` notifications and the Docker weave `container:event` `start` notifications.

However, even that unlikely race condition would just lead to two sequential to `start_all_containers`, which should AFAIK be harmless, just result in a lot of unnecessary `weaveexec attach` calls. So not IMO worth worrying too much about?

SpComb:
> @SpComb Compared to #2131, this is a smaller change, but it does more work, because it will also re-attach every container when re-creating weave. This is similar to what also happens on agent restarts, and if weave crashes and gets restarted by Docker.

> @jnummelin  I was trying to avoid the migration logic in the "full" network restart in this case as it might be bit expensive if there's lot of containers.

We could optimize `start_all_containers` by enhancing the `WeaveWorker@migrate_containers` caching to not re-attach containers that are already attached, but that then needs to be careful to not race with the service container restart events and leave the containers unattached.

The important question is whether or not we have an issue with `attach_overlay` of service containers failing during weave restart/re-create events. If weave getting restarted/re-created can result in concurrent attaches of starting service containers failing, then it would make sense to go to the effort of re-attaching containers after weave restarts/re-creates, in order to not leave those service containers detached.

So far, we have not yet seen attaches failing during restart events. In fact, there's some indication of them succeeding despite weave being restarting over in #1923:

```
kontena-agent | D, [2017-03-02T09:50:04.801681 #1] DEBUG -- NetworkAdapters::WeaveExecutor: weaveexec ok: ["--local", "attach", "10.81.128.92/16", "--rewrite-hosts", "fe5af6db7c101a1314bfe68a7f020b11083378caa5a95e4db38580eb32af2559"]
kontena-agent | 10.81.128.92
kontena-agent | 
kontena-agent | D, [2017-03-02T09:50:05.191951 #1] DEBUG -- NetworkAdapters::WeaveExecutor: weaveexec ok: ["--local", "attach", "10.81.128.5/16", "--rewrite-hosts", "a746a42562999f5cbd9984e8aea6403f69f5dce783622f643e7514371b88f443"]
kontena-agent | 10.81.128.5
kontena-agent | 
kontena-agent | D, [2017-03-02T09:50:06.356523 #1] DEBUG -- NetworkAdapters::WeaveExecutor: weaveexec ok: ["--local", "attach", "10.81.128.28/16", "--rewrite-hosts", "ba04a7ced236046c8886fe852e5ffe12d977c3c3b480cc2dee547f0398129b37"]
kontena-agent | weave container is not present. Have you launched it?
kontena-agent | 10.81.128.28
kontena-agent | 
kontena-agent | D, [2017-03-02T09:50:06.923496 #1] DEBUG -- NetworkAdapters::WeaveExecutor: weaveexec ok: ["--local", "attach", "10.81.128.3/16", "--rewrite-hosts", "653e4574f253f572f9c584b279ec5d64dc602cac8f7e646ac8de606df7277316"]
kontena-agent | weave container is not present. Have you launched it?
kontena-agent | 10.81.128.3
kontena-agent | 
```

SpComb:
> @SpComb The alternative to this would be to change the `Weave` actor to re-notify `network_adapter:start` after upgrades.

Having the `Weave` actor notify after re-creating might actually work to avoid the race between `network_adapter:started` and `container:event` notifications?

jnummelin:
> Having the Weave actor notify after re-creating might actually work to avoid the race between network_adapter:started and container:event notifications?

Should work, and would simplify things a bit.

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -29,6 +29,18 @@ def on_weave_start(topic, data)
       self.start
     end
 
+    # Ensure that all service containers are attached and registered in Weave DNS.
+    #
+    # Only called after the network_adapter is running?
```
SpComb: Turns out this is not actually the case. It might be `already_started?`, but not `running?`, if weave failed/crashed during the `Weave#launch`...

SpComb:
>> Having the Weave actor notify after re-creating might actually work to avoid the race between network_adapter:started and container:event notifications?
> 
> Should work, and would simplify things a bit.

Implemented this alternative over in #2278

SpComb:
Closed by #2278

PR #2176, opened 2017-04-18
Make ssl cert path configurable

marcbosserhoff:
so that Excon can load the certs from global cert repo.
The default is set for ubuntu here. After inserting this
line i could connect to my master with letencrypt certs
on the haproxy, as i inserted the ca to ubuntu global
cert repos.

### DIFF
--- a/cli/lib/kontena/client.rb
+++ b/cli/lib/kontena/client.rb
@@ -55,6 +55,7 @@ module Kontena
         connect_timeout: ENV["EXCON_CONNECT_TIMEOUT"] ? ENV["EXCON_CONNECT_TIMEOUT"].to_i : 5,
         read_timeout:    ENV["EXCON_READ_TIMEOUT"]    ? ENV["EXCON_READ_TIMEOUT"].to_i    : 30,
         write_timeout:   ENV["EXCON_WRITE_TIMEOUT"]   ? ENV["EXCON_WRITE_TIMEOUT"].to_i   : 5,
+        ssl_ca_path:     ENV["EXCON_SSL_CA_PATH"]     ? ENV["EXCON_SSL_CA_PATH"] : "/etc/ssl/certs",
         ssl_verify_peer: ignore_ssl_errors? ? false : true
       }
       if ENV["DEBUG"]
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/client.rb
```diff
@@ -55,6 +55,7 @@ def initialize(api_url, token = nil, options = {})
         connect_timeout: ENV["EXCON_CONNECT_TIMEOUT"] ? ENV["EXCON_CONNECT_TIMEOUT"].to_i : 5,
         read_timeout:    ENV["EXCON_READ_TIMEOUT"]    ? ENV["EXCON_READ_TIMEOUT"].to_i    : 30,
         write_timeout:   ENV["EXCON_WRITE_TIMEOUT"]   ? ENV["EXCON_WRITE_TIMEOUT"].to_i   : 5,
+        ssl_ca_path:     ENV["EXCON_SSL_CA_PATH"]     ? ENV["EXCON_SSL_CA_PATH"] : "/etc/ssl/certs",
```
kke: I'm not sure it's a good idea to default to `/etc/ssl/certs` without the ENV.

Excon will already try using `ENV['SSL_CERT_FILE']` or `ENV['SSL_CERT_DIR']` if this option is not set.

https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb#L34

marcbosserhoff: I see, do you have an idea how to implement this better? in my case kontena cli wont load /etc/ssl/certs by default from this path in ubuntu... maybe i can pass some parameter so that excon can pick it up?

kke: Did you try with `SSL_CERT_DIR=/etc/ssl/certs`?

marcbosserhoff: Nice that worked :) Didnt find anything about this in the documentation or the slack channel... so you can drop the pull request. Maybe you can add this "as tip" to the protip when displaying the error message ;)

kke: Maybe you can add that to https://github.com/kontena/kontena/blob/master/cli/lib/kontena/command.rb#L213 in this PR

SpComb:
Replaced by #2201

Issue #2154, opened 2017-04-13
Agent starts outdated container instead of re-creating it


SpComb:
If a `ServicePod` is out of date (the service `updated_at` is newer than the container `created_at`), but stopped, the agent `ServicePodWorker` will only start the outdated container, and not re-create it, because of the order of checks:

```ruby
      elsif service_container && service_pod.running? && !service_container.running?
        info "starting #{service_pod.name}"
        ensure_started
      elsif service_pod.running? && (service_container && service_container_outdated?(service_container, service_pod))
        info "re-creating #{service_pod.name}"
        ensure_running
```

Then 20s later, it will have picked up the Docker container start event, notice that the container is running and out of date, and re-create it.

```
D, [2017-04-13T09:03:13.500997 #1] DEBUG -- Kontena::RpcServer: rpc notification: Kontena::Rpc::ServicePodsApi#notify_update ["start"] D, [2017-04-13T09:03:13.575368 #1] DEBUG -- Kontena::RpcClient: waited 0.1s of 30.0s until: request /node_service_pods/list has response wth id=747628547 yielded Array D, [2017-04-13T09:03:13.627704 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-1: running I, [2017-04-13T09:03:13.640265 #1]  INFO -- Kontena::Workers::ServicePodWorker: starting redis-1 I, [2017-04-13T09:03:13.655237 #1]  INFO -- Kontena::ServicePods::Starter: starting service: /redis-1

[...]

PR #2177, opened 2017-04-19
Check that image is up-to-date in ServicePodWorker

### LINKED ISSUE #2171: Service with newer image is not deployed without force
Kontena Agent is not checking is image updated when doing deploy, this results to situation that service which has only image updated (no changes in configuration) won't be deployed without `--force` option.

jakolehm:
Fixes #2171

### DIFF
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -81,12 +81,12 @@ module Kontena::Workers
       if service_pod.running? && service_container.nil?
         info "creating #{service_pod.name}"
         ensure_running
-      elsif service_container && service_pod.running? && !service_container.running?
-        info "starting #{service_pod.name}"
-        ensure_started
       elsif service_pod.running? && (service_container && service_container_outdated?(service_container))
         info "re-creating #{service_pod.name}"
         ensure_running
+      elsif service_container && service_pod.running? && !service_container.running?
+        info "starting #{service_pod.name}"
+        ensure_started
       elsif service_pod.stopped? && (service_container && service_container.running?)
         info "stopping #{service_pod.name}"
         ensure_stopped
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -157,8 +157,9 @@ module Kontena::Workers
     def service_container_outdated?(service_container, service_pod)
       creator = Kontena::ServicePods::Creator.new(service_pod)
       creator.container_outdated?(service_container) ||
-        creator.labels_outdated?(service_pod.labels, service_container) ||
-          creator.recreate_service_container?(service_container)
+          creator.labels_outdated?(service_pod.labels, service_container) ||
+          creator.recreate_service_container?(service_container) ||
+          creator.image_outdated?(service_pod.image_name, service_container)
     end
 
     # @param [Kontena::Models::ServicePod] service_pod
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -156,10 +156,25 @@ module Kontena::Workers
     # @param [Kontena::Models::ServicePod] service_pod
     def service_container_outdated?(service_container, service_pod)
       creator = Kontena::ServicePods::Creator.new(service_pod)
-      creator.container_outdated?(service_container) ||
+      outdated = creator.container_outdated?(service_container) ||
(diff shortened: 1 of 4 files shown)


agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -157,8 +157,9 @@ def ensure_terminated
     def service_container_outdated?(service_container, service_pod)
       creator = Kontena::ServicePods::Creator.new(service_pod)
       creator.container_outdated?(service_container) ||
-        creator.labels_outdated?(service_pod.labels, service_container) ||
-          creator.recreate_service_container?(service_container)
+          creator.labels_outdated?(service_pod.labels, service_container) ||
+          creator.recreate_service_container?(service_container) ||
+          creator.image_outdated?(service_pod.image_name, service_container)
```
SpComb: Compared to `Kontena::ServicePods::Creator#service_uptodate?`, this is still missing the `service_container.config['Image'] != service_pod.image_name`. Would that also be needed here?

I'm guessing it's about something like detecting changes to `config['Image']` that happen when new images are pulled and the image tags update to point to new images?

SpComb:
If I understood #2171 correctly, that would be a service with something like `image: foo:latest`, and then `kontena service deploy` should actually do a `docker pull foo:latest` and re-create the service container if an updated image has been pushed for the `latest` tag.

I don't see that happening here? It seems like the `ServicePodWorker` would need to `ensure_image` to actually pull the latest image for that tag before checking if the container image is out of date.

In terms of the smallest possible change, that might imply dropping the `service_container_outdated?` check and just unconditionally calling `ensure_running` to invoke the existing `Kontena::ServicePods::Creator` logic whenever the `deploy_rev` changes... Longer-term, I think that the log should be moved into some kind of `ImageManager` that deals with them....

BTW also fix #2154 while you're at it?

jakolehm:
@SpComb good point, added image pull before comparing.

jakolehm:
Refactored comparison logic to `ServicePodWorker` and wrote some additional tests. Also noticed that comparing timestamps is not very reliable because image might actually change so that timestamp is still "older" than what container has => changed image comparison to check id's.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -84,7 +84,7 @@ def ensure_desired_state
       elsif service_container && service_pod.running? && !service_container.running?
         info "starting #{service_pod.name}"
         ensure_started
-      elsif service_pod.running? && (service_container && service_container_outdated?(service_container, service_pod))
+      elsif service_pod.running? && (service_container && service_container_outdated?(service_container))
```
SpComb: Might be worth fixing #2154 too: the agent can hit the earlier `!service_container.running?` -> `ensure_started` case even if the container is outdated.

Plus, right now it looks like the `recreate_service_container?` check is broken and will never be hit, because the agent will always just start the `!running?` container instead of re-creating it?

jakolehm: done

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -153,12 +153,62 @@ def ensure_terminated
...
+      outdated = container_outdated?(service_container) ||
+          labels_outdated?(service_container) ||
+          recreate_service_container?(service_container)
+      return true if outdated
+
+      image_puller.ensure_image(
+        service_pod.image_name, service_pod.deploy_rev, service_pod.image_credentials
+      )
+
+      if image_outdated?(service_container)
```
SpComb: `if wtf then true else false end` :sunglasses:

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -153,12 +153,62 @@ def ensure_terminated
...
-      creator.container_outdated?(service_container) ||
-        creator.labels_outdated?(service_pod.labels, service_container) ||
-          creator.recreate_service_container?(service_container)
+    def service_container_outdated?(service_container)
+      outdated = container_outdated?(service_container) ||
+          labels_outdated?(service_container) ||
+          recreate_service_container?(service_container)
+      return true if outdated
+
+      image_puller.ensure_image(
```
SpComb: Now that the `ServicePodWorker` handles the image re-pull on update checks, the `ServicePods::Creator` probably doesn't need to pull it twice anymore. It does still need to pull it, because the `config_container` -> `Weave` stuff needs access to the image before creating the container. Perhaps the `ServicePods::Creator#create_container` -> `ensure_image` call could be removed?

PR #2179, opened 2017-04-19
kontena.yml reference improvements

jakolehm:


docs/references/kontena-yml.md
```diff
@@ -419,25 +424,29 @@ user: app_user
...
+
+##### Data volume:
 
 ```
 volumes:
- - /data/mysql:/var/lib/mysql
+  - /var/lib/mysql
 ```
 
-Named volumes are supported in the service declaration, but not in the top-level `volumes` key. When defining named volume in the service declaration the default driver configured by the Docker Engine will be used (in most cases, this is the local driver). If volume does not exist it will be created.
```
SpComb: This same sentence is also in `docs/references/docker-compose-support.md`.

docs/references/kontena-yml.md
```diff
@@ -419,25 +424,29 @@ user: app_user
...
+Data volume is created on the fly. Volume is persistent only if service is marked as `stateful: true`.
```
SpComb: Anonymous data volumes are created on the fly. The volume is only persistent if the service is marked as `stateful: true`.

docs/references/kontena-yml.md
```diff
@@ -74,7 +79,7 @@ instances: 1
 
 #### stateful
 
-Mark service as stateful (default: false). Kontena will create and automatically mount a data volume container for the service. This option also instructs the scheduler to bind the service instance to the scheduled host so that the volume can be mapped when the service is updated.
+Mark service as stateful (default: false). Kontena will create and automatically mount a data volume container for the service. This option also instructs the scheduler to not make any automatic scheduling decisions after initial deployment is done and bind the service instance to the scheduled host so that the volume can be mapped when the service is updated.
```
SpComb: This option binds each service instance to the host node that is first scheduled on to, so that the same volumes can be mapped whenever the service is updated. This option also instructs the scheduler to not make any automatic scheduling decisions for the service after initial deployment is done.

docs/references/kontena-yml.md
```diff
@@ -74,7 +79,7 @@ instances: 1
 
 #### stateful
 
-Mark service as stateful (default: false). Kontena will create and automatically mount a data volume container for the service. This option also instructs the scheduler to bind the service instance to the scheduled host so that the volume can be mapped when the service is updated.
+Mark service as stateful (default: false). Kontena will create and automatically mount a data volume container for the service. This option binds each service instance to the host node that is first scheduled on to, so that the same volumes can be mapped whenever the service is updated. This option also instructs the scheduler to not make any automatic scheduling decisions for the service after initial deployment is done.
```
SpComb: This reminds me that the `using-kontena/nodes.md` `Decomissioning nodes` should have some mention of how `kontena node rm` will cause stateful service instances to be re-deployed to new nodes and lose their state.

PR #1300, opened 2016-11-09
Use ipam and weave as standalone without docker plugins

jnummelin:
This PR makes networking use the `weave attach`model instead of Docker network and IPAM plugins as the Docker networking stuff proved to be highly unstable and unusable.

Very much WIP still, mainly a starting skeleton to see if and how the ipam integration could work. I've tried to make all ipam related handling within the weave adapter so it's "confined" only in one place.

Still lots of things on the todo list for this:
- [x] Make the ipam understand the non-docker managed ips/networks in the cleanup
~~- [ ] NetworkMigrator should reserve addresses of existing containers~~, let's handle this as separate PR
- [x] update and fix Tests
- [x] code cleanup
- [x] ipam client error and response handling
- [x] master to understand the old style ip's, again

@jakolehm @SpComb  Take a look on the general idea. Don't focus on code details yet. :)

PR #955, opened 2016-09-02
Move ipam functionality into docker plugin

jnummelin:
This PR removes all IPAM related functionality from master and agent. In future IPAM is fully handled by custom Docker plugin.

WIP flag as probably this requires still a bit of fine-tuning.

Things todo:
- [x] Add tests for new stuff
- [x] Weavewait "migration" into no-op binary

**Note:** Handling of multiple networks will be another PR in the near future. :)

PR #2181, opened 2017-04-19
Agent: Remove unused ContainerStarterWorker

SpComb:
This was introduced as part of the network plugins (#955), but disabled when reverting back to the weaveexec model: #1300 https://github.com/kontena/kontena/commit/eb492ede4f3e39c3b2502a0b1f09af0c31666639#diff-e39179f2a45969bbee881391fcb03770L17

At this point it's just dead code, and and think the `ServicePodWorker` would do the same thing, but much better.


agent/lib/kontena/workers/container_starter_worker.rb
```diff
@@ -1,42 +0,0 @@
...
-module Kontena::Workers
-  class ContainerStarterWorker
-    include Celluloid
-    include Celluloid::Notifications
-    include Kontena::Logging
-    include Kontena::Helpers::WeaveHelper
-
-    def initialize
-      info 'initialized'
-      #subscribe('network:ready', :on_overlay_start)
```
SpComb: no-op'd

PR #2188, opened 2017-04-20
Do not add nice when calculating node CPU usage

nevalla:
This PR will remove `nice` value from node's CPU usage calculation. Although it's consuming(?) CPU it can give wrong impression about how busy the server is or how much free capacity there are.

### DIFF
--- a/server/app/models/host_node_stat.rb
+++ b/server/app/models/host_node_stat.rb
@@ -179,8 +179,7 @@ class HostNodeStat
           '$avg': {
             '$add': [
               { '$ifNull': ['$cpu.user', 0] },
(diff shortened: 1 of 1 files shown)


SpComb:
Turns out that the agent node stats code is just buggy for the `nice` diff: https://github.com/kontena/kontena/blob/master/agent/lib/kontena/workers/node_stats_worker.rb#L240

```ruby
 nice_ticks = current_cpu.nice = prev_cpu.nice
```

spot the bug vs

```ruby
 idle_ticks = current_cpu.idle - prev_cpu.idle
```

jakolehm:
Proper fix in #2190

PR #2189, opened 2017-04-20
Do not send un-managed volume infos to master

jnummelin:
Currently agent sends all volume infos to master and master tries to match those to existing volume instances. It creates two mongo queries per volume plus pub-sub "overhead".

Fixed agent to send only info on volumes where the proper `io.kontena.volume_instance.id` label is in place.

### DIFF
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -94,7 +94,12 @@ module Kontena::Workers::Volumes
         'volume_instance_id' => data.dig('Labels', 'io.kontena.volume_instance.id'),
         'volume_id' => data.dig('Labels', 'io.kontena.volume.id')
       }
-      rpc_client.async.request('/node_volumes/set_state', [node.id, volume])
+      # Only send "managed" volumes to server
+      if volume['volume_instance_id']
+        rpc_client.async.request('/node_volumes/set_state', [node.id, volume])
+      else
+        debug "Skip sending un-managed volume: #{volume['name']}"
+      end
     end
 
     # Checks if given volume exists with the expected driver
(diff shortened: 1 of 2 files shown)


SpComb:
Could the server's `Rpc::NodeVolumeHandler#set_state` now also be improved to fail and log a warning if the volume_instance was not found?

jnummelin:
It could fail, but that actually makes me think why do we even send then the info at all? I mean if server does not find the `VolumeInstance` and it doesn't create it and instead fails with a warning, what's the point to even send those volume infos?

SpComb:
Even though it is the server that decides which `VolumeInstance`s to create right now, and the `/node_volumes/set_state` RPC doesn't actually do anything at all, I think there are two reasons to keep the `/node_volumes/set_state` RPC:

* Reporting the actual state of the volume back to the server, which could update the `VolumeInstance`
* Discovery of Docker volumes that show up via the volume plugins on other nodes, after the first volume instance is scheduled, and the volume is created

    This would require other changes to the volume management, though, since the volume labels do not propagate.

SpComb:
This is an important fix for 1.2, because we've seen cases where host nodes might have several thousand orphaned anonymous Docker volumes, and sync'ing those to the master on startup causes RPC issues.

jnummelin:
> This is an important fix for 1.2, because we've seen cases where host nodes might have several thousand orphaned anonymous Docker volumes, and sync'ing those to the master on startup causes RPC issues.

Makes me think whether there should be a worker on agent that does `docker prune` for the volumes periodically? Not part of this PR for sure

SpComb:
>> This is an important fix for 1.2, because we've seen cases where host nodes might have several thousand orphaned anonymous Docker volumes, and sync'ing those to the master on startup causes RPC issues.
> 
> Makes me think whether there should be a worker on agent that does docker prune for the volumes periodically? Not part of this PR for sure

AFAICT the agent has always done `Docker::Container.remove(v: true)`, so I'm not entirely sure why those dangling volumes have accumulated.

PR #2190, opened 2017-04-20
Fix nice cpu

daniel-bytes:
Fixing fat-finger typo in the `nice` cpu calculation in `node_stats_worker.rb`.

Also, using simpler numbers in the corresponding spec (including something greater than zero for `nice`, the zero value is what caused this bug to go unnoticed).  

Finally, including big comment in spec showing the math because.. math.


SpComb:
I think the reason this passed unnoticed is that the default CoreOS install just doesn't run any processes at nice levels > 0, so the cpu nice ticks will tend to just be zero.

PR #2201, opened 2017-04-22
SSL: Give hint about setting certs path

marcbosserhoff:
if kontena cannot find system ca bundle, when its at another location
based on the underlying linux system standards


marcbosserhoff:
Build failed in "server" section, but i only touched the cli... so trunk seems to be broken anyways ;)

PR #2205, opened 2017-04-24
Add rb-readline to CLI installer

### LINKED ISSUE #2194: OSX cli package (omnibus) is missing readline
Without readline kontena shell plugin cannot be loaded.

Workaround:
```
$ cd /opt/kontena/embedded/bin/
$ sudo ./gem install rb-readline
```

kke:
Installs rb-readline into the embedded ruby to support running kosh.

Fixes #2194

### DIFF
--- a/cli/omnibus/config/software/kontena.rb
+++ b/cli/omnibus/config/software/kontena.rb
@@ -5,7 +5,6 @@ dependency "ruby"
 dependency "rubygems"
 dependency "libxml2"
 dependency "libxslt"
-
 whitelist_file "./wrappers/sh/kontena"
 build do
   gem "install rb-readline -v 0.5.4 --no-ri --no-doc"
--- a/cli/omnibus/config/software/kontena.rb
+++ b/cli/omnibus/config/software/kontena.rb
@@ -5,8 +5,10 @@ dependency "ruby"
 dependency "rubygems"
 dependency "libxml2"
 dependency "libxslt"
+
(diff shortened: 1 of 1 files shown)


cli/omnibus/config/software/kontena.rb
```diff
@@ -7,6 +7,7 @@
 dependency "libxslt"
 whitelist_file "./wrappers/sh/kontena"
 build do
+  gem "install rb-readline -v 0.5.4 --no-ri --no-doc"
```
SpComb: Ideally this should be a dependency of the `kontena-plugin-shell` gem, but apparently conditional dependencies for different platforms (like omnibus not having a working stdlib readline?) are difficult.

kke: Also the omnibus thing should be able to produce windows installers if you run it on windows. Rb-readline README says that one of the reasons it was written is because readline is nearly impossible to build on windows. (kosh seems to work fine on win10 and rubyinstaller installed ruby, so not sure if relevant)

kke:
It seemed to work on my mac. @nevalla @jnummelin @jakolehm can verify.

Binary is in dropbox temp/

jnummelin:
Works on my mac at least

kke:
No change in behavior with windows when rb-readline added. Must be something else that is causing it to be weird.

SpComb:
Is this `status/merge`?

PR #2207, opened 2017-04-24
Refactor agent WeaveHelper to avoid unnecessary sync actor calls

SpComb:
Avoid unnecessary remote actor calls like `Actor[:network_adapter].adapter_image?(event.from)`, when the same logic can easily be implemented as a local helper method call. This avoids side-effects like having the `EventWorker` actor crash with `DeadActorError`s if the `Weave` actor crashes.

Also moves the Weave API client stuff out of the `WeaveHelper` into the classes where they are actually used.


SpComb:
@jnummelin PTAL

PR #2209, opened 2017-04-24
Exit with non-zero exit code when commands run other commands that fail 

### LINKED ISSUE #1637: Stack upgrade --deploy return code should not be zero

kke:
Fixes #1637
Replaces #1702

This PR makes `Kontena.run` re-raise the SystemExit if status is non-zero unless called with `returning: :status` in which case it will return the exit code to caller.

```console
$ kontena stack upgrade redis redis.yml || echo "dang"
[error] Foofoo
dang
```

### DIFF
--- a/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
@@ -57,7 +57,7 @@ module Kontena
         ]
         Retriable.retriable do
           ENV["DEBUG"] && $stderr.puts("Running: #{cmd}")
-          Kontena.run(cmd)
+          raise "retry" unless Kontena.run(cmd, returning: :status).zero?
         end
       end
     end
--- a/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
@@ -57,7 +57,7 @@ module Kontena
         ]
         Retriable.retriable do
           ENV["DEBUG"] && $stderr.puts("Running: #{cmd}")
-          raise "retry" unless Kontena.run?(cmd)
+          Kontena.run!(cmd)
         end
       end
     end
--- a/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
@@ -49,7 +49,12 @@ module Kontena
         end
 
         require 'shellwords'
-        cmd = "master login --no-login-info --skip-grid-auto-select --verbose --name #{command.result[:name].shellescape} --code #{command.result[:code].shellescape} #{new_master.url.shellescape}"
+        cmd = [
+          'master', 'login', '--no-login-info' ,'--skip-grid-auto-select', '--verbose',
+          '--name', command.result[:name],
+          '--code', command.result[:code],
+          new_master.url
+        ]
         Retriable.retriable do
           ENV["DEBUG"] && $stderr.puts("Running: #{cmd}")
           Kontena.run(cmd)
--- a/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
@@ -57,7 +57,7 @@ module Kontena
         ]
         Retriable.retriable do
           ENV["DEBUG"] && $stderr.puts("Running: #{cmd}")
-          raise "retry" unless Kontena.run(cmd, returning: :status).zero?
+          raise "retry" unless Kontena.run?(cmd)
         end
       end
     end
--- a/cli/lib/kontena/callbacks/master/deploy/60_configure_auth_provider_after_deploy.rb
+++ b/cli/lib/kontena/callbacks/master/deploy/60_configure_auth_provider_after_deploy.rb
@@ -15,21 +15,16 @@ module Kontena
         args
       end
 
-      def configure_auth_provider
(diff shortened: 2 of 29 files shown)


cli/lib/kontena_cli.rb
```diff
@@ -6,7 +6,7 @@ module Kontena
   #
   # @param [String,Array<String>] command_line
   # @return [Fixnum] exit_code
```
SpComb: The `@param`, `@raise` and `@return` docs need updating.

Seems that it returns the `Kontena::Command.run` per default in the success case, which is either whatever `execute` returns, or `true` if it does `exit(0)`.

cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
```diff
@@ -49,10 +49,15 @@ def after
...
+        cmd = [
+          'master', 'login', '--no-login-info' ,'--skip-grid-auto-select', '--verbose',
+          '--name', command.result[:name],
+          '--code', command.result[:code],
+          new_master.url
+        ]
         Retriable.retriable do
           ENV["DEBUG"] && $stderr.puts("Running: #{cmd}")
-          Kontena.run(cmd)
+          raise "retry" unless Kontena.run(cmd, returning: :status).zero?
```
SpComb: Most of these callers seem to be of the form `fail unless Kontena.run(..., :returning => :status).zero?`. Would the default `Kontena.run(cmd)` case behave identically, given that it will raise an exception, and thus retry?

cli/lib/kontena/callbacks/master/deploy/60_configure_auth_provider_after_deploy.rb
```diff
@@ -8,15 +8,15 @@ class ConfigureAuthProviderAfterDeploy < Kontena::Callback
...
         args << "--cloud-master-id #{command.cloud_master_id}" if command.cloud_master_id
         args << "--provider #{command.result[:provider]}" if command.result[:provider]
         args << "--version #{command.result[:version]}" if command.result[:version]
-        args.join(' ')
+        args
       end
 
       def configure_auth_provider
-        Kontena.run("master init-cloud #{init_cloud_args}")
+        Kontena.run(['master', 'init-cloud'] + init_cloud_args, returning: :status)
```
SpComb: Where is the status return value going in these cases?

cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb
SpComb: Not equivalent because `Kontena.run` will re-raise `SystemExit` in error cases, which is not a `StandardError`, and would thus likely behave differently than a `RuntimeError` from `raise "retry"`.

cli/lib/kontena_cli.rb
```diff
@@ -1,32 +1,41 @@
...
 
+  # Run a kontena command and return true if the command did not raise or exit with a non-zero exit code. Raises nothing.
+  # @param [String,Array<String>] command_line
+  # @return [TrueClass,FalseClass] success
+  def self.run?(*cmdline)
+    run(*cmdline)
+    true
+  rescue SystemExit => ex
+    ex.status.zero?
+  rescue ScriptError, NoMemoryError, SystemStackError, StandardError => ex
```
SpComb: Given that a lot of the usage is like `Retriable.retriable do raise "retry" unless Kontena.run?(cmd) end`.... does it make sense to retry on these errors? AFAICT these errors were not being retried before.

cli/lib/kontena/cli/master/user/invite_command.rb
```diff
@@ -39,11 +39,11 @@ def execute
             puts "  * command: kontena master join #{current_master.url} #{response['invite_code']}"
           end
           roles.each do |role|
-            Kontena.run(["master", "user", "role", "add", role, email])
+            raise "Failed to add role" unless Kontena.run?(["master", "user", "role", "add", role, email])
```
SpComb: I wonder if this kind of thing would be easier if `Kontena.run` just didn't propagate `SystemExit`, re-raising it as something that's a `StandardError`?

cli/lib/kontena_cli.rb
```diff
@@ -1,32 +1,41 @@
...
-    ENV["DEBUG"] && $stderr.puts("Command raised #{ex} with message: #{ex.message}\n#{ex.backtrace.join("\n  ")}")
-    returning == :status ? 1 : nil
+    ENV["DEBUG"] && $stderr.puts("Command raised #{ex.class.name} with message: #{ex.message}\n#{ex.backtrace.join("\n  ")}")
+    raise ex
   end
 
+  # Run a kontena command and return true if the command did not raise or exit with a non-zero exit code. Raises nothing.
+  # @param [String,Array<String>] command_line
+  # @return [TrueClass,FalseClass] success
+  def self.run?(*cmdline)
```
SpComb: Just nitpicking on naming now, but having `run?` as a variant of `run` that doesn't raise exceptions, but still has side-effects, doesn't seem right. Shouldn't it be `run!` that raises, and then a `run` wrapper that rescues the `SystemExit` to return `false`? Most callers would then call `run!`.

Although, if I understand the callers correctly, then it might also work to just have a single `run` method that re-raised `SystemExit` as a `StandardError`?

jnummelin: I fully agree on the naming, "banged" versions are expected to raise.

PR #2211, opened 2017-04-24
Fix GridService#deploy_pending? to consider all unfinished deploys as pending

### LINKED ISSUE #2208: Server: GridServiceHealthMonitorJob can pile up deployments
Given a failing health check, `GridServiceHealthMonitorJob` will just blindly create new deployments if the service overall health drops too much. In case of mis-configured (or anyway heavily failing) health check this can cause service deployments to pile up heavily with these kind of log entries
```
I, [2017-04-24T11:48:10.369721 #11]  INFO -- GridServiceSchedulerWorker: delaying foo/bar/some-service deploy because there is another deploy in progress
```
These will probably also mean timeouts with distributed locks:
```
W, [2017-04-24T11:52:45.536157 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:45.763583 #9]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:45.794109 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:45.805272 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:46.810364 #9]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:46.812494 #9]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:46.845791 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:56.563669 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:56.793793 #9]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:56.815847 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
W, [2017-04-24T11:52:56.844696 #11]  WARN -- Module: timeout after waiting 10.0s of 10.0s until: lock check_deploy_queue:580ddcc376f5980007001319 is available
```

`GridServiceHealthMonitorJob` should check if there is already a pending deployment for the given service and only if not create the deployment.

jnummelin:
fixes #2208 

There was a racey condition when the GridServiceSchedulerWorker check the deploy "queue", it marks deploys with some timestamp for started_at when it picks them up from the queue and only after a 30 "wait" marks them back at not started if there were previous ongoing deploys. This leaves a window where the previous logic (`started_at != nil`) would allow to create multiple deploys from e.g. failing health check that would pile up.

### DIFF
--- a/server/app/models/grid_service.rb
+++ b/server/app/models/grid_service.rb
@@ -147,7 +147,7 @@ class GridService
   end
 
   def deploy_pending?
-    self.grid_service_deploys.where(started_at: nil).count > 0
+    self.grid_service_deploys.where(finished_at: nil).count > 0
   end
 
   # @return [Boolean]
--- a/server/app/models/grid_service.rb
+++ b/server/app/models/grid_service.rb
@@ -146,8 +146,17 @@ class GridService
     self.stack.exposed_service?(self)
   end
 
+  # Are there any deploys in queue for this service
+  # @return [Boolean]
   def deploy_pending?
-    self.grid_service_deploys.where(finished_at: nil).count > 0
+    self.grid_service_deploys.where(started_at: nil).count > 0
+  end
+
+  # Are there any deploys in queue or in progress for this service
+  # Ignores any possibly stale deploys that have been started but never finished
+  # @return [Boolean]
+  def deploy_outstanding?
+    self.grid_service_deploys.any_of({:started_at => nil, :finished_at => nil}, {:started_at.gt => 30.minutes.ago , finished_at: nil}).count > 0
   end
 
   # @return [Boolean]
--- a/server/spec/models/grid_service_spec.rb
+++ b/server/spec/models/grid_service_spec.rb
@@ -305,4 +305,20 @@ describe GridService do
       expect(subject.affinity).to eq(['label!=type=ssd'])
     end
   end
+
+  describe '#deploy_pending?' do
+    it 'returns false when no deployments' do
+      expect(subject.deploy_pending?).to be_falsey
+    end
+
+    it 'returns true when queued deployments' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: nil, finished_at: nil)
+      expect(subject.deploy_pending?).to be_truthy
+    end
+
+    it 'returns true when un-finished deployments' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: Time.now, finished_at: nil)
+      expect(subject.deploy_pending?).to be_truthy
+    end
+  end
 end
--- a/server/spec/models/grid_service_spec.rb
+++ b/server/spec/models/grid_service_spec.rb
@@ -306,19 +306,24 @@ describe GridService do
     end
   end
 
-  describe '#deploy_pending?' do
+  describe '#deploy_outstanding?' do
     it 'returns false when no deployments' do
-      expect(subject.deploy_pending?).to be_falsey
+      expect(subject.deploy_outstanding?).to be_falsey
     end
 
     it 'returns true when queued deployments' do
       GridServiceDeploy.create!(grid_service: subject, started_at: nil, finished_at: nil)
-      expect(subject.deploy_pending?).to be_truthy
+      expect(subject.deploy_outstanding?).to be_truthy
     end
 
     it 'returns true when un-finished deployments' do
-      GridServiceDeploy.create!(grid_service: subject, started_at: Time.now, finished_at: nil)
-      expect(subject.deploy_pending?).to be_truthy
+      GridServiceDeploy.create!(grid_service: subject, started_at: 10.minutes.ago, finished_at: nil)
+      expect(subject.deploy_outstanding?).to be_truthy
+    end
+
+    it 'returns true when un-finished stale deployments' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: 32.minutes.ago, finished_at: nil)
+      expect(subject.deploy_outstanding?).to be_falsey
     end
   end
 end
--- a/server/spec/models/grid_service_spec.rb
+++ b/server/spec/models/grid_service_spec.rb
@@ -321,8 +321,18 @@ describe GridService do
       expect(subject.deploy_outstanding?).to be_truthy
     end
 
-    it 'returns true when un-finished stale deployments' do
+    it 'returns false when un-finished stale deployments' do
       GridServiceDeploy.create!(grid_service: subject, started_at: 32.minutes.ago, finished_at: nil)
+      expect(subject.deploy_outstanding?).to be_falsey
+    end
+
+    it 'returns false when finished deployments' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: 32.minutes.ago, finished_at: 26.minutes.ago)
+      expect(subject.deploy_outstanding?).to be_falsey
+    end
+
+    it 'returns false when finished deployments with no started_at' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: nil, finished_at: 26.minutes.ago)
       expect(subject.deploy_outstanding?).to be_falsey
     end
   end
--- a/server/app/jobs/grid_service_health_monitor_job.rb
+++ b/server/app/jobs/grid_service_health_monitor_job.rb
@@ -30,7 +30,7 @@ class GridServiceHealthMonitorJob
     min_health = service.deploy_opts.min_health || 0.8
     expected_health = 1 - min_health
     if health_percent < expected_health
-      service.running? && !service.deploy_pending?
+      service.running? && !service.deploy_outstanding?
     else
       false
     end
--- a/server/spec/jobs/grid_service_health_monitor_job_spec.rb
+++ b/server/spec/jobs/grid_service_health_monitor_job_spec.rb
@@ -49,7 +49,7 @@ describe GridServiceHealthMonitorJob, celluloid: true do
           health_status: {healthy: 1, total: 6},
           deploy_opts: double({min_health: 0.8}),
           running?: true,
-          deploy_pending?: false
+          deploy_outstanding?: false
         })
       expect(subject.deploy_needed?(service)).to be_truthy
     end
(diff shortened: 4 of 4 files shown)


SpComb:
The `GridService` has two methods that are almost but not entirely unlike: `deploy_pending?` and `deploying?`. It seems that `deploying?` only considers something that is `started_at` but not yet `finished_at`? They should have some kind of comment explaining this difference, because it's not immediately apparent.

Anyways, the `GridServiceHealthMonitorJob` seems to be the only thing that calls `GridService.deploy_pending?`, so this change should be fine?

SpComb:
Based on some testing, this seems to be even more broken: multiple overlapping pending deploys for a service block the `GridServiceSchedulerWorker` from actually deploying the service, because each attempt to deploy the service after the 30s interval will always find some other `GridServiceDeploy` that has `:started_at.ne => nil, :finished_at => nil`... after the first deploy, the service just gets stuck, and doesn't get deployed anymore...

SpComb:
The `service.deploy unless service.deploy_pending?` logic is still inherently racy, but this is still much better than before.

I managed to cause duplicate deploys by scrolling the server console and blocking the logging output, I think... the logging timestamps somehow went out of order :see_no_evil: 

```
I, [2017-04-24T18:23:26.275478 #21141]  INFO -- GridServiceDeployer: service development/null/redis-unhealthy has been deployed
D, [2017-04-24T18:23:28.920088 #21153] DEBUG -- GridServiceHealthMonitorJob: service=redis-unhealthy health=0.0 deploy_pending=false
I, [2017-04-24T18:23:28.923919 #21153]  INFO -- GridServiceHealthMonitorJob: service health too low, triggering full deploy for development/null/redis-unhealthy
I, [2017-04-24T18:23:29.128509 #21153]  INFO -- GridServiceDeployer: starting to deploy development/null/redis-unhealthy
I, [2017-04-24T18:23:29.200267 #21153]  INFO -- GridServiceDeployer: throttling service instance development/null/redis-unhealthy deploy because of min_health limit (1 instances in-progress)
I, [2017-04-24T18:23:29.201551 #21153]  INFO -- GridServiceInstanceDeployer: Deploying service instance development/null/redis-unhealthy-1 to node core-01 at 2017-04-24 15:23:29 UTC...
D, [2017-04-24T18:22:25.834823 #21145] DEBUG -- GridServiceHealthMonitorJob: service=redis-unhealthy health=0.0 deploy_pending=false
I, [2017-04-24T18:22:25.848619 #21145]  INFO -- GridServiceHealthMonitorJob: service health too low, triggering full deploy for development/null/redis-unhealthy
```

```
D, [2017-04-24T18:23:29.761755 #21153] DEBUG -- GridServiceInstanceDeployer: waiting 1.0s of 300.0s until: service development/null/redis-unhealthy-1 is running on node development/core-01 at 2017-04-24 15:23:29 UTC
D, [2017-04-24T18:23:30.262686 #21153] DEBUG -- GridServiceInstanceDeployer: waiting 1.5s of 300.0s until: service development/null/redis-unhealthy-1 is running on node development/core-01 at 2017-04-24 15:23:29 UTC
D, [2017-04-24T18:23:31.264781 #21153] DEBUG -- GridServiceInstanceDeployer: waiting 2.5s of 300.0s until: service development/null/redis-unhealthy-1 is running on node development/core-01 at 2017-04-24 15:23:29 UTC
D, [2017-04-24T18:23:32.297422 #21153] DEBUG -- GridServiceHealthMonitorJob: service=redis-unhealthy health=0.0 deploy_pending=true
```

SpComb:
OTOH I'm not even sure what the point of the `GridServiceHealthMonitorJob` is... deploying an unhealthy service doesn't actually do anything, because the service instances are scheduled onto host nodes and running.... re-deploying isn't going to reschedule them onto different nodes, or re-create them?

server/app/models/grid_service.rb
```diff
@@ -147,7 +147,7 @@ def stack_exposed?
   end
 
   def deploy_pending?
-    self.grid_service_deploys.where(started_at: nil).count > 0
+    self.grid_service_deploys.where(finished_at: nil).count > 0
```
SpComb: This may run into problems with stale `GridServiceDeploy`s where a `GridServiceSchedulerWorker` has set `started_at` (causing other workers to ignore it), but the worker fails to either perform the deploy (such as #2213) or un-set `started_at` (like if the server crashes during the blocked-deploy operation).

AFAIK this means that there could be any number of really old stale `GridServiceDeploy`s that match this query, because AFAIK we don't clean out old/stale `GridServiceDeploy`s?

jakolehm: If deploy has started but not finished is it really considered as pending or in-progress? Original idea behind `#deploy_pending?` was to check if service had deployment in a queue (but not yet started).

server/app/models/grid_service.rb
```diff
@@ -147,7 +147,7 @@ def stack_exposed?
   end
 
   def deploy_pending?
```
jakolehm: We should rename this method if we change the behaviour.

server/app/models/grid_service.rb
SpComb: > AFAIK this means that there could be any number of really old stale GridServiceDeploys that match this query, because AFAIK we don't clean out old/stale GridServiceDeploys?

These kind of stale deploys do seem to exist, and they don't appear to get cleaned out: `puts GridServiceDeploy.where(:finished_at => nil).asc(:created_at).map{|sd| "#{sd.created_at} #{sd.grid_service ? sd.grid_service.to_path : '<destroyed>'} is #{sd.deploy_state} with started_at=#{sd.started_at} finished_at=#{sd.finished_at}" }.join("\n")`
```
2016-11-22 09:18:20 UTC <destroyed> is created with started_at=2016-11-22T09:18:20+00:00 finished_at= 2017-02-24 13:37:58 UTC development/ingress-lb/lb is ongoing with started_at=2017-02-24T13:37:58+00:00 finished_at= 2017-03-03 09:19:53 UTC development/ingress-lb/lb is ongoing with started_at=2017-03-03T09:19:54+00:00 finished_at= 2017-04-03 07:23:40 UTC development/ingress-lb/lb is created with started_at=2017-04-03T07:23:40+00:00 finished_at= 2017-04-06 06:47:24 UTC development/ingress-lb/lb is created with started_at=2017-04-06T06:47:24+00:00 finished_at= 2017-04-11 07:56:23 UTC development/ingress-lb/lb is created with started_at=2017-04-11T07:56:24+00:00 finished_at= 2017-04-24 16:05:56 UTC development/null/redis-test is created with started_at=2017-04-24T16:05:56+00:00 finished_at= 2017-04-24 16:06:21 UTC development/null/redis-test is created with started_at=2017-04-24T16:10:59+00:00 finished_at=
```

server/app/models/grid_service.rb
jnummelin: Roger. I took the interpretation of `pending`too loosely. :)

How about `deploy_outstanding?`? At least in Cambridge dictionary it's defined as:
> Not yet solved or done

SpComb: In #2221, I just called this `deploying?`, and renamed the existing `deploying?` method to `deploy_running?`.

SpComb: To fix this for 1.2.1 before #2221 is ready, probably better to just leave `deploying?` as-is, and have the `GridServiceHealthMonitorJob` call something like `deploy_outstanding?`.

However, it still needs to ignore all of the ancient deploys that were marked as `started_at` but never actually run.

jnummelin:
"Reverted" the `deploy_pending?` logic and added `deploy_outstanding?` method which also ignores any ancient deploys that have not finished at all.

server/spec/models/grid_service_spec.rb
```diff
@@ -305,4 +305,25 @@
...
+      GridServiceDeploy.create!(grid_service: subject, started_at: nil, finished_at: nil)
+      expect(subject.deploy_outstanding?).to be_truthy
+    end
+
+    it 'returns true when un-finished deployments' do
+      GridServiceDeploy.create!(grid_service: subject, started_at: 10.minutes.ago, finished_at: nil)
+      expect(subject.deploy_outstanding?).to be_truthy
+    end
+
+    it 'returns true when un-finished stale deployments' do
```
SpComb: `returns false`

server/spec/models/grid_service_spec.rb
```diff
@@ -305,4 +305,25 @@
       expect(subject.affinity).to eq(['label!=type=ssd'])
     end
   end
+
+  describe '#deploy_outstanding?' do
```
SpComb: Also needs a spec for `returns false when deployments have finished`, i.e. both `started_at` and `finished_at`.

server/spec/models/grid_service_spec.rb
jnummelin: copy-pasta strikes again...

server/spec/models/grid_service_spec.rb
jnummelin: done

server/app/models/grid_service.rb
```diff
@@ -146,10 +146,19 @@ def stack_exposed?
     self.stack.exposed_service?(self)
   end
 
+  # Are there any deploys in queue for this service
+  # @return [Boolean]
   def deploy_pending?
```
jakolehm: We should remove this method if it's not used anymore?

jnummelin: it is used in some existing (prior to this PR)  specs

PR #2220, opened 2017-04-25
Fix validation of service volume mappings with permission flags

### LINKED ISSUE #2219: Service volume mounts with :ro breaks validation in 1.2.0
```yaml
services:
  foo:
    volumes:
     - /var/run/docker.sock:/var/run/docker.sock:ro
     - /proc:/hostfs/proc:ro
     - /sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro
     - /:/hostfs:ro
...
[errror]
volumes:
         ro: mount point defined 4 times
```

kke:
Fixes #2219

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
@@ -33,7 +33,7 @@ module Kontena::Cli::Stacks
 
       # bar -> bar
       # foo:bar -> bar
-      # foo:bar:baz -> bar
+      # foo:bar:ro -> bar
       def mount_point_from_volume_mapping(string)
         string.gsub(/:(?:ro|rw)\z/, '').split(':').last
       end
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
```diff
@@ -31,6 +31,13 @@ def initialize
         Validations::CustomValidators.load
       end
 
+      # bar -> bar
+      # foo:bar -> bar
+      # foo:bar:ro -> bar
+      def mount_point_from_volume_mapping(string)
+        string.gsub(/:(?:ro|rw)\z/, '').split(':').last
```
SpComb: See the server's [`parse_volume`](https://github.com/kontena/kontena/blob/v1.2.0/server/app/helpers/volumes_helpers.rb#L3) method for how volumes should be parsed.

There can also be other flags than just `ro` and `rw`, but they are only valid for named and bind volumes, not anonymous volumes, and thus they are always the third `:`-separated field.

* `/var/lib/foo` this form of anonymous volumes can not have flags; the mount point is the first field
* `/var/lib/foo:/var/lib/foo` this form of bind mounts supports optional flags; the mount point is the second field
* `/var/lib/foo:/var/lib/foo:ro` with optional flags
* `foo:/var/lib/foo` same thing for named volumes

PR #2224, opened 2017-04-26
Version 1.2.1.dev

SpComb:
Bump git master from 1.2.0.dev -> 1.2.1.dev, which is newer than the current 1.2.0 release in the `v1.2` release branch.


kke:
There's still no revision/commit in master that would point to 1.2.0.

Brew formulas usually do something like:
```
url "https://github.com/zsh-users/zsh-syntax-highlighting.git", 
  :revision => "15d4587514a3beaa13972093e335bf685b6726a9"
```

I can do `:tag => 'v1.2.0, :revision => 'abcdefg'', but it's unusual.

SpComb:
Well, there's no revision/commit in master that would point to 1.1.5 either... All of the 1.2.x tags are in the `v1.2` branch now. Dunno if it's the right way to do things.

Issue #1723, opened 2017-01-25
SSH key management for master & nodes


japsu:
At the moment it is needlessly complicated to add and remove SSH keys that have console access to the master and nodes.

I propose adding Kontena CLI commands for managing SSH keys in the master and in nodes:

```
kontena master ssh-keys add id_rsa.pub
kontena master ssh-keys add - < id_rsa.pub
kontena master ssh-keys list
kontena master ssh-keys remove japsu@myrmidon-20171125 # from key comment field

kontena grid ssh-keys add id_rsa.pub
kontena grid ssh-keys add - < id_rsa.pub
kontena grid ssh-keys add --all-keys-from-master
kontena grid ssh-keys list
kontena grid ssh-keys remove japsu@myrmidon-20171125
kontena grid ssh-keys remove --all-keys-not-in-master
```

We could even add support for shared keys that are stored in the vault and which are used automatically by `kontena node ssh` and `kontena master ssh`.

kke:
The ssh keys could also be used to authenticate API communication. (JWT's digested using ssh keys).

There could also be a module for sshd that would read the authorized keys from mongo/master api instead of trying to keep .ssh/authorized_keys up to date.

SpComb:
> There could also be a module for sshd that would read the authorized keys from mongo/master api instead of trying to keep .ssh/authorized_keys up to date.

You don't want SSH logins to the master to fail when you're trying to SSH in to debug a broken mongo/master API :)

It might make sense on the nodes, but Integrating sshd `AuthorizedKeysCommand` + kontena-agent via Docker on the nodes sounds painful.

SpComb:
This functionality overlaps with the cloud-config stuff used by the plugins to manage SSH keys on CoreOS nodes. AFAIK CoreOS may also overwrite the `~core/.ssh/authorized_keys` file at boot, using the SSH keys in the cloud metadata? This would mean that anything that the agent itself writes to `~core/.ssh/authorized_keys` will get reverted (there are workarounds, like IIRC `~core/.ssh/authorized_keys.d/`.

The ideal thing would really be some mechanism to update and re-apply the cloud-config on existing CoreOS nodes, instead of starting to re-implement overlapping cloud-config/Puppet/etc functionality within the kontena-agent.

OFC the CoreOS `cloud-config` stuff doesn't really apply to e.g. Ubuntu nodes...

SpComb:
[...]

Issue #912, opened 2016-08-11
Updating master SSL cert


adrianchifor:
Once I've created a master server (DigitalOcean CoreOS) with an SSL cert (Let's Encrypt), how can I update the certificate on the server when the initial one expires?

mudspot:
I have met this same issue.

I tried replacing /etc/kontena-server.pem

But after rebooting, the original Kontena-server.pem got restored.

My Kontena master was created on AWS with the "kontena master create " command

SpComb:
The `/etc/kontena-server.pem` is provisioned via the user-data cloud-config. The cloud-config gets re-applied at every boot, so it will restore the original cert.

You either need to update the cloud-config in the AWS EC2 instance user-data, or write it out as a new file (`/etc/kontena-server-2.pem` or whatever), and use the `/etc/kontena-server.local.env` to point `SSL_CERT=/etc/kontena-server-2.pem` at that non-cloudconfig'd file.

mudspot:
Works perfect!!

I used the second method.
I put it into `/etc/kontena-server.custom.env` (where I have also defined my KONTENA_VERSION

jnummelin:
> Works perfect!!
closing as solved. :)

PR #2228, opened 2017-04-26
docs/digitalocean: mention caveat about immutable cloud-config.yml

japsu:
Add the following caveat to `docs/getting-started/installing/digitalocean.md`:

The Kontena DigitalOcean plugin creates CoreOS virtual machines for the master
and nodes. CoreOS machines are configured using a configuration file called
`cloud-config.yml`. On DigitalOcean, this configuration file is stored in
instance metadata.

DigitalOcean provides no way to update instance metadata, including
`cloud-config.yml`, after the instance has been created. This means that if
you want to, for example, update SSH keys or set `sysctl`s in a persistent
fashion, you will need to create new nodes and decommission old ones.

### DIFF
--- a/docs/getting-started/installing/digitalocean.md
+++ b/docs/getting-started/installing/digitalocean.md
@@ -10,6 +10,14 @@ title: DigitalOcean
 - [Installing Kontena Nodes](digitalocean.md#installing-kontena-nodes)
 - [DigitalOcean Plugin Command Reference](digitalocean.md#digitalocean-plugin-command-reference)
 
+## Caveats
+
+### Cannot update `cloud-config.yml` after node creation
+
+The Kontena DigitalOcean plugin creates CoreOS virtual machines for the master and nodes. CoreOS machines are configured using a configuration file called `cloud-config.yml`. On DigitalOcean, this configuration file is stored in [instance metadata](https://www.digitalocean.com/community/tutorials/an-introduction-to-droplet-metadata).
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/installing/digitalocean.md
```diff
@@ -10,6 +10,14 @@ title: DigitalOcean
 - [Installing Kontena Nodes](digitalocean.md#installing-kontena-nodes)
 - [DigitalOcean Plugin Command Reference](digitalocean.md#digitalocean-plugin-command-reference)
 
+## Caveats
+
+### Cannot update `cloud-config.yml` after node creation
```
SpComb: The title and phrasing here is a bit odd in terms of the remainder of the documentation, because the cloud-config is mostly just an internal detail of how the DigitalOcean plugin is implemented, so the relationship to the plugin parameters documented here is a bit vauge and unactionable. Instead, it should probably mention the specific `kontena digitalocean node create` parameters that are problematic, and how to reconfigure those later after provisioning... Those parameters and some related GitHub issues would be:

* `kontena digitalocean (master|node) create --ssh-key` #1723 (changes to `~core/.ssh/authorized_keys` get reverted at boot)
* `kontena digitalocean (master|node) create --version` #2289 (agent reboots with the wrong version after master version upgrades)
* `kontena digitalocean master create --ssl-cert` #912 (need to use the non-cloudconfig'd `/etc/kontena-server.local.env` to update)
* `kontena digitalocean master create --mongodb-uri` (dunno about this one, it's written directly into the systemd unit, would need a drop-in to override)

SpComb:
The `docs` have been moved to https://github.com/kontena/docs, any docs PRs need to be re-submitted there.

PR #2240, opened 2017-05-02
Fix registry create command

jakubknejzlik:
When running 1.2.1 version I'm getting 

```
uninitialized constant Kontena::Cli::Registry::CreateCommand::SecureRandom
``` 

Adding `require 'securerandom'` fixed it in my case.

### DIFF
--- a/cli/lib/kontena/cli/registry/create_command.rb
+++ b/cli/lib/kontena/cli/registry/create_command.rb
@@ -1,4 +1,5 @@
 require_relative '../stacks/stacks_helper'
+require 'securerandom'
 
 module Kontena::Cli::Registry
   class CreateCommand < Kontena::Command


SpComb:
Broken by #1093, this was required by `cli/lib/kontena/cli/apps/init_command.rb` before.

Looking for other usages of `SecureRandom`, it looks like `cli/lib/kontena/callbacks/master/deploy/50_authenticate_after_deploy.rb` may also be broken for the case of duplicate master names.

kke:
As a workaround until 1.2.2 is released you can add the line:

```ruby
require "securerandom"
```

to the beginning of the broken command file,  which you can find with:

```console
gem contents kontena | grep registry/create_command
```

or

```console
kontena whoami --bash-completion-path | sed 's/scripts.*/cli\/registry\/create_command.rb/'
```

And to do this all automatically in bash:

```console
ruby -i -pe 'sub(/require_relative/, "require \"securerandom\"\nrequire_relative")' `kontena whoami --bash-completion-path | sed 's/scripts.*/cli\/registry\/create_command.rb/'`
```

PR #2242, opened 2017-05-02
Add v4auth to create registry with s3 driver

jakubknejzlik:
Since registry 2.6.0 auth v4 is required. This could be mentioned in docs (also default value could be changed in cli).

### DIFF
--- a/docs/using-kontena/image-registry.md
+++ b/docs/using-kontena/image-registry.md
@@ -44,7 +44,7 @@ $ kontena vault write REGISTRY_STORAGE_S3_SECRETKEY <secret_key>
 Create registry service:
 
 ```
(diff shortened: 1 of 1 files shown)


SpComb:
ping @jakolehm @jnummelin can you test the S3 registry? Should we update the `--s3-v4auth` default?

jnummelin:
Based on [AWS docs](http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) I think it would be safe to change the default:
> Amazon S3 supports Signature Version 4, a protocol for authenticating inbound API requests to AWS services, in all AWS regions

SpComb:
Based on #2606 I suspect we shouldn't just document this as `--s3-v4auth`, but change that to be the default instead?

PR #2243, opened 2017-05-02
Move inclusion of Kontena::Cli::Common to Kontena::Command

kke:
`Kontena::Cli::Common` is included in nearly all of the commands.

..so it can be included in `Kontena::Command` from which all of the commands are inherited from.

This opens up some optimization and refactoring opportunities to further eradicate the Kontena::Cli::Common anti-pattern in the future.


SpComb:
Seems like an ok refactoring for 1.4, just needs merge... and testing to find the one or two subtly broken commands that seem to happen with these CLI refactorings :/

PR #2245, opened 2017-05-03
Fix random strategy behaviour

### LINKED ISSUE #2244: Random strategy moves services constantly
Regression in 1.2.x

jakolehm:
Fixes #2244 

Includes also cleanup of scheduler internals.

### DIFF
--- a/server/app/services/scheduler/strategy/random.rb
+++ b/server/app/services/scheduler/strategy/random.rb
@@ -1,34 +1,58 @@
+require_relative 'common'
+
 module Scheduler
   module Strategy
     class Random
-
-      # @param [Integer] node_count
-      # @param [Integer] instance_count
-      def instance_count(node_count, instance_count)
-        instance_count.to_i
-      end
+      include Scheduler::Strategy::Common
 
       # @return [ActiveSupport::Duration]
       def host_grace_period
         30.seconds
       end
 
-      ##
       # @param [GridService] grid_service
       # @param [Integer] instance_number
       # @param [Array<HostNode>] nodes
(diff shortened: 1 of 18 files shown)


server/app/services/scheduler/strategy/random.rb
```diff
@@ -1,34 +1,58 @@
...
+      # @return [Scheduler::Node]
+      def find_stateless_node(grid_service, instance_number, nodes)
+        if deployment?
+          nodes.sample
+        else 
           prev_instance = grid_service.grid_service_instances.find_by(
             grid_service: grid_service, instance_number: instance_number
           )
           if prev_instance
-            prev_instance.host_node
```
SpComb: This also fixes #2254

jakolehm:
Going to open separate pr for 1.2.2.

SpComb:
Needs merge after #2256 

I think this PR is still useful for the `GridServiceDeployWorker` and `GridServiceDeployer` -> `GridServiceScheduler` refactoring on its own, but I don't really understand what we need the automatic/manual deployment modes and scheduler/deployment scheduling modes for, since we fixed the only usage of those in #2256

jnummelin:
Do we really need this for 1.3 as the behaviour if already fixed in 1.2.2?

SpComb:
I don't think this belongs under the 1.3.X milestone... tentatively bumping to 1.4 for the refactoring, but this PR seems out of date, so perhaps just close it...?

SpComb:
Closing as obsolete. Wouldn't mind if we picked up the refactoring bits as a new PR.

PR #2247, opened 2017-05-03
Add require for SecureRandom in after deploy hook

kke:
Just like #2240


kke:
Won't hurt, better safe than sorry.

PR #2249, opened 2017-05-03
Enable clearing of statsd, log forwarder and default affinity in "kontena grid update"

### LINKED ISSUE #2230: grid statsd server cannot be cleared once set
Once a `--statsd-server` is set for a grid, it can be changed, but there is no way to un-set it, and stop the agent from exporting stats.

The CLI needs to support something like `kontena grid update --no-statsd`. The server `PUT /v1/grids/:name` API needs to support something like `{"stats": {}}`.

kke:
Fixes #2230 by always sending a full update of the grid object by modifying the response of `client.get("grid/#{name}")` (which is a more RESTy approach to PUT requests).

Trickery performed to detect if the request came from a previous version of CLI that does not send a full update (could be done in the controller by snooping on the useragent).

### DIFF
--- a/cli/lib/kontena/cli/grids/update_command.rb
+++ b/cli/lib/kontena/cli/grids/update_command.rb
@@ -7,36 +7,53 @@ module Kontena::Cli::Grids
 
     parameter "NAME", "Grid name"
     option "--statsd-server", "STATSD_SERVER", "Statsd server address (host:port)"
-    option "--default-affinity", "[AFFINITY]", "Default affinity rule for the grid", multivalued: true
+    option "--disable-statsd-server", :flag, "Disable statsd exporting", default: false
+    option "--default-affinity", "AFFINITY", "Default affinity rule for the grid", multivalued: true
+    option "--disable-default-affinity", :flag, "Disable default affinity rule", default: false
     option "--log-forwarder", "LOG_FORWARDER", "Set grid wide log forwarder"
-    option "--log-opt", "[LOG_OPT]", "Set log options (key=value)", multivalued: true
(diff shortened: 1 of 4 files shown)


server/app/mutations/grids/update.rb
```diff
@@ -54,24 +67,31 @@ def validate
...
-      end
-      if self.default_affinity
         attributes[:default_affinity] = self.default_affinity
+      else # assume this is pre 1.2.2 client that didn't send a full update
+        attributes[:stats] = self.stats if self.stats
+        attributes[:default_affinity] = self.default_affinity if self.default_affinity
       end
 
       if self.logs
         if self.logs[:forwarder] == 'none'
```
SpComb: After changing the `forwarder` to be an `optional` field, the `self.logs[:forwarder]` may now be `nil`. As is, this would result in `attributes[:grid_logs_opts] = GridLogsOpts.new(forwarder: nil, opts: nil)`, which is weird.

IMO better to just accept `self.logs[:forwarder].nil? || self.logs[:forwarder] == 'none'` -> `attributes[:grid_logs_opts] = nil`.

cli/lib/kontena/cli/grids/update_command.rb
```diff
@@ -7,36 +7,53 @@ class UpdateCommand < Kontena::Command
...
+        payload['stats'] = {
+          'statsd' => {
+            'server' => server,
+            'port' => port || 8125
           }
         }
       end
 
+      if disable_statsd_server?
+        payload['stats'] = { 'statsd' => nil }
```
SpComb: This is equivalent to just `payload['stats'] = {}`.

PR #2250, opened 2017-05-03
Restore "kontena login" and "logout" commands

kke:
From community Slack:

image

This PR restores the good old "kontena login", but with a twist:

`kontena login master` is exactly the same as `kontena master login`
`kontena login cloud` is exactly the same as `kontena cloud login`

And therefore:
`kontena login` gives you the subcommand list:

```
Usage:
    kontena login [OPTIONS] SUBCOMMAND [ARG] ...

Parameters:
    SUBCOMMAND                    subcommand
    [ARG] ...                     subcommand arguments

Subcommands:
    master                        Login to a Kontena Master
    cloud                         Login to a Kontena Cloud account
```

Same goes with `kontena logout`.

### DIFF
--- a/cli/lib/kontena/cli/login_command.rb
+++ b/cli/lib/kontena/cli/login_command.rb
@@ -1,0 +1,6 @@
+module Kontena::Cli
+  class LoginCommand < Kontena::Command
+    subcommand "master", "Login to a Kontena Master", load_subcommand('master/login_command')
+    subcommand "cloud", "Login to a Kontena Cloud account", load_subcommand('cloud/login_command')
+  end
+end
--- a/cli/lib/kontena/cli/logout_command.rb
+++ b/cli/lib/kontena/cli/logout_command.rb
@@ -1,10 +1,4 @@
 class Kontena::Cli::LogoutCommand < Kontena::Command
-  include Kontena::Cli::Common
-
-  banner "Command removed, use 'kontena master logout' to log out of the Kontena Master"
-  banner "or 'kontena cloud logout' to log out of the Kontena Cloud", false
-
-  def execute
-    exit_with_error("Command removed. Use #{"kontena master logout".colorize(:yellow)} to log out of the Kontena Master")
-  end
+  subcommand "master", "Logout from Kontena Masters", load_subcommand('master/logout_command')
+  subcommand "cloud", "Logout from Kontena Cloud account", load_subcommand('cloud/logout_command')
 end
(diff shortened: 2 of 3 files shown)


SpComb:
Should `kontena logout` behave the same, then?

```
$ kontena logout
 [error] Command removed. Use kontena master logout to log out of the Kontena Master
$ kontena logout -h
Usage:
    kontena logout [OPTIONS]

  Command removed, use 'kontena master logout' to log out of the Kontena Master
  or 'kontena cloud logout' to log out of the Kontena Cloud

Options:
    -D, --debug                   Enable debug (default: $DEBUG)
```

kke:
It does now.

SpComb:
LGTM in terms of implementation, but not sure about the design.

Having both `kontena cloud login` and `kontena login cloud` seems more confusing than just killing the deprecated `kontena login` command entirely.

Alternatively, a single `kontena login http://.../` command that does both a cloud login + master login...?

Same for `kontena logout` - hvae it logout of both master and cloud?

jnummelin:
> Same for kontena logout - have it logout of both master and cloud?

Are people always wanting to logout from both master and cloud? I think there are cases where one wants to logout only from master but not from cloud.

SpComb:
> Are people always wanting to logout from both master and cloud? I think there are cases where one wants to logout only from master but not from cloud.

Sure, that's what the current `kontena master logout` does.

jnummelin:
> Sure, that's what the current kontena master logout does.

Yes, but the point was that if `kontena logout` does both of those tasks, is it more or less confusing for users? :)

TBH I don't know which one is better way, they both have pros and cons. ¯\_(ツ)_/¯

jnummelin:
@jakolehm WDYT, is the proposed changes to have `kontena login master == kontena master login` and `kontena login cloud == kontena cloud login` ok, or source of confusion?

kke:
Pehaps the main command help should just have some kind of "getting started" banner.

image

?

jnummelin:
IMO banner approach is (also) good, it pretty much achieves the desired results without adding duplication in commands.

jakolehm:
> Having both kontena master logout and kontena logout master seems more confusing than just killing the deprecated kontena logout command entirely.

I agree, it's confusing.

jnummelin:
So having multiple commands do the same things is now deemed as confusing, for a good reason. 

Maybe the best alternative to "solve" this issue is the banner approach.

kke:
Replaced with #2385

Issue #2231, opened 2017-04-27
Agent: Actor crash-and-restart loop will eat lot of resources


jnummelin:
Given for example a mis-configuration of statsd server, the `Kontena::Workers::StatsWorker` will enter a really fast crash-and-restart loop:
```
I, [2017-04-27T13:10:24.050432 #1]  INFO -- Kontena::Workers::StatsWorker: cadvisor is running, starting stats loop
I, [2017-04-27T13:10:24.066581 #1]  INFO -- Kontena::Workers::StatsWorker: exporting stats via statsd to udp://none:8125
I, [2017-04-27T13:10:24.118177 #1]  INFO -- Kontena::Workers::StatsWorker: initialized
E, [2017-04-27T13:10:24.090364 #1] ERROR -- : Actor crashed!
SocketError: getaddrinfo: Name does not resolve
    /usr/lib/ruby/gems/2.3.0/gems/statsd-ruby-1.3.0/lib/statsd.rb:430:in `ip'
    /usr/lib/ruby/gems/2.3.0/gems/statsd-ruby-1.3.0/lib/statsd.rb:430:in `block in connect'
    /usr/lib/ruby/gems/2.3.0/gems/statsd-ruby-1.3.0/lib/statsd.rb:419:in `synchronize'
    /usr/lib/ruby/gems/2.3.0/gems/statsd-ruby-1.3.0/lib/statsd.rb:419:in `connect'
    /usr/lib/ruby/gems/2.3.0/gems/statsd-ruby-1.3.0/lib/statsd.rb:274:in `initialize'
    /app/lib/kontena/workers/stats_worker.rb:27:in `new'
    /app/lib/kontena/workers/stats_worker.rb:27:in `configure_statsd'
    /app/lib/kontena/workers/stats_worker.rb:37:in `block in start'
    /app/lib/kontena/observer.rb:48:in `call'
    /app/lib/kontena/observer.rb:57:in `observed'
    /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
    /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
    /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/async.rb:7:in `dispatch'
[...]

PR #2251, opened 2017-05-03
Add --no-statsd-server and --no-default-affinity to "kontena grid update"

### LINKED ISSUE #2230: grid statsd server cannot be cleared once set

### LINKED ISSUE #2252: CLI: kontena grid update clears out default-affinity
If a grid has a default affinity set:

```
development:
  uri: ws://localhost:9292/
  initial_size: 1
  default_affinity: 
    - label!=special
  subnet: 10.81.0.0/16
  supernet: 10.80.0.0/12
  stats:
    nodes: 2 of 2
    cpus: 2
    load: 0.41 0.4 0.36
    memory: 0.94 of 1.95 GB
    filesystem: 8.11 of 31.14 GB
    users: 1
    services: 6
    containers: 98
```

Then a `kontena grid update` will wipe out that default affinity filter when updating any other parameter:

`DEBUG=api bundle exec bin/kontena grid update --log-forwarder fluentd --log-opt fluentd-address=127.0.0.1 development`
```
...
[API Client Request]: PUT http://localhost:9292/v1/grids/development | Headers: {Accept: application/json, Content-Type: application/json, Authorization: Bearer}  | Body: 
{"logs":{"forwarder":"fluentd","opts":{"fluentd-address":"127.0.0.1"}},"default_affinity":[]}
[API Client Response]: Headers: {Content-Type: application/json}  | Status: 200 | Body: 
{"id":"development","name":"development","token":"tB...0A==","initial_size":1,"stats":{"statsd":null},"default_affinity":[],"trusted_subnets":[],"node_count":2,"service_count":6,"container_count":98,"user_count":1,"subnet":"10.81.0.0/16","supernet":"10.80.0.0/12","logs":{"forwarder":"fluentd","opts":{"fluentd-address":"127.0.0.1"}}}
```

The default affinity is now gone from `kontena grid show`:
```
development:
  uri: ws://localhost:9292/
  initial_size: 1
  default_affinity: 
  subnet: 10.81.0.0/16
  supernet: 10.80.0.0/12
  stats:
    nodes: 2 of 2
    cpus: 2
    load: 0.41 0.4 0.36
    memory: 0.94 of 1.95 GB
    filesystem: 8.11 of 31.14 GB
    users: 1
    services: 6
    containers: 98
logs:
  forwarder: fluentd
  fluentd-address: 127.0.0.1
```

This is a bug in the CLI, where the `cli/grids/update_command.rb` conditional will always be true for the default `[]` value of the multi-valued `--default-affinity` parameter:

```ruby
      if default_affinity_list
        payload[:default_affinity] = default_affinity_list
      end
```

kke:
Fixes #2230 by adding `--no-statsd-server` to `kontena grid update`.
Also:
Fixes #2252 (default_affinity was always cleared when you ran `kontena grid update` without `--default-affinity` option.

### DIFF
--- a/cli/lib/kontena/cli/grids/update_command.rb
+++ b/cli/lib/kontena/cli/grids/update_command.rb
@@ -9,7 +9,7 @@ module Kontena::Cli::Grids
     option "--statsd-server", "STATSD_SERVER", "Statsd server address (host:port)"
     option "--no-statsd-server", :flag, "Unset statsd server setting"
     option "--default-affinity", "[AFFINITY]", "Default affinity rule for the grid", multivalued: true
-    option "--log-forwarder", "LOG_FORWARDER", "Set grid wide log forwarder"
+    option "--log-forwarder", "LOG_FORWARDER", "Set grid wide log forwarder (set to 'none' to disable)"
     option "--log-opt", "[LOG_OPT]", "Set log options (key=value)", multivalued: true
 
     def execute
@@ -28,7 +28,7 @@ module Kontena::Cli::Grids
       end
 
       if no_statsd_server?
-        payload[:stats] = { statsd: { server: 'none', port: 0 } }
+        payload[:stats] = { statsd:  nil }
       end
 
       if log_forwarder
@@ -38,9 +38,10 @@ module Kontena::Cli::Grids
         }
       end
 
-      if default_affinity_list
+      unless default_affinity_list.empty?
         payload[:default_affinity] = default_affinity_list
       end
+
       client(token).put("grids/#{name}", payload)
     end
 
--- a/cli/lib/kontena/cli/grids/update_command.rb
+++ b/cli/lib/kontena/cli/grids/update_command.rb
@@ -7,6 +7,7 @@ module Kontena::Cli::Grids
 
     parameter "NAME", "Grid name"
     option "--statsd-server", "STATSD_SERVER", "Statsd server address (host:port)"
+    option "--no-statsd-server", :flag, "Unset statsd server setting"
     option "--default-affinity", "[AFFINITY]", "Default affinity rule for the grid", multivalued: true
     option "--log-forwarder", "LOG_FORWARDER", "Set grid wide log forwarder"
     option "--log-opt", "[LOG_OPT]", "Set log options (key=value)", multivalued: true
@@ -24,6 +25,10 @@ module Kontena::Cli::Grids
             port: port || 8125
           }
         }
+      end
+
+      if no_statsd_server?
+        payload[:stats] = { statsd: { server: 'none', port: 0 } }
       end
 
       if log_forwarder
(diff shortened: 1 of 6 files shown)


SpComb:
This looks appropriate for 1.2.x, provided that nobody has a statsd server called `none` :)

Compared to the `logs.forwarder` enum with a strictly limited set of values, the `statsd.server` is just an arbitrary DNS hostname... perhaps it would be better to use something like `{"stats": {}}`, or `{"stats": {"statsd": {}}` (equivalent to `{"stats": {"statsd": {"server": null}}`) instead, to avoid special-casing the `none` hostname?

kke:
That forces to change the mutation validation optional or some strange lambda thing.

SpComb:
> That forces to change the mutation validation to optional or some strange lambda thing.

That should be okay in terms of 1.2.x compat, changing the `stats.statsd` field to be optional should not break existing 1.2.2 CLI -> 1.2.1 server cases?

It would means that 1.2.2 CLI `kontena grid update --no-statsd` would fail with a 1.2.1 master, but that's better than how the `{"stats": {"statsd": {"server": "none"}}}` would behave with a 1.2.1 server: using an unresolvable statsd server puts the agent stats workers into a crash loop that consumes resources:  #2231

kke:
Noted that currently `grid update` always clears default affinity because `if default_affinity_list` is never nil.

SpComb:
> Noted that currently grid update always clears default affinity because if default_affinity_list is never nil.

Yeah, I just noticed that while looking at the `DEBUG=api` requests... looks like a bug too.

SpComb:
Also fixes  #2252

cli/lib/kontena/cli/grids/update_command.rb
```diff
@@ -26,16 +27,21 @@ def execute
...
+
       if log_forwarder
         payload[:logs] = {
           forwarder: log_forwarder,
           opts: parse_log_opts
         }
       end
 
-      if default_affinity_list
+      unless default_affinity_list.empty?
```
SpComb: Perhaps there should also be a `--no-default-affinity` to clear out the list?

server/app/mutations/grids/update.rb
```diff
@@ -8,9 +8,9 @@ class Update < Mutations::Command
 
     optional do
       hash :stats do
-        required do
+        optional do
           hash :statsd do
-            required do
+            optional do
```
SpComb: This can remain as `required`, the `stats` -> `optional` is enough to accept`{"stats":{"statsd":null}}`.

PR #2255, opened 2017-05-03
Sign OSX pkg properly

jakolehm:


SpComb:
> Does it need some keyfile somewhere or can it be that simple?

Probably only available on our jenkins OSX build machine.

jakolehm:
It needs certificate in keychain (it's installed to build server).

SpComb:
Need to test this as a 1.2.2 pre-release.

Issue #1653, opened 2017-01-13
Agent statsd metrics exporter does not work after dns change


jakolehm:

SpComb:
Do you mean that the agent's `Statsd` client continues sending the stats to the old IP address, after the configured `kontena grid update --statsd-server=statds.foo` DNS name is updated to resolve to a different IP address?

jakolehm:
Yes, currently user needs to issue `grid update` again so that agent swaps the statsd client to a new one.

SpComb:
This sounds like the typical problem of DNS service discovery not working for most UDP-based applications.

It looks like statsd-ruby also supports a `Statsd.new(protocol: :tcp)`, which might act better in this situation? If the old statsd server goes away, the TCP connection will die, and it will reconnect and re-resolve the configured host: https://github.com/reinh/statsd/blob/v1.3.0/lib/statsd.rb#L428

jakolehm:
Yes, but tcp is not supported in every statsd server (for example telegraf).

PR #2262, opened 2017-05-04
Don't crash StatsWorker actor if statsd config fails

jnummelin:
per #2231 fast-crashing actor will start to leak a lot of memory and burns up the CPU.

This PR fixes the case where `StatsWorker` actor gets into crash-restart loop with non-working config.

The idea is that the statsd is tried configure again when the node info is next updated/observed.

### DIFF
--- a/agent/lib/kontena/workers/stats_worker.rb
+++ b/agent/lib/kontena/workers/stats_worker.rb
@@ -30,6 +30,9 @@ module Kontena::Workers
       else
         @statsd = nil
       end
+    rescue => error
+      warn "statsd configuration failed: #{error.message}"
+      warn error.backtrace.join('\n') if error.backtrace
     end
 
     def start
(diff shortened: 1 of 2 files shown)


agent/lib/kontena/workers/stats_worker.rb
```diff
@@ -30,6 +30,9 @@ def configure_statsd(node)
       else
         @statsd = nil
       end
+    rescue => error
+      warn "statsd configuration failed: #{error.message}"
+      warn error.backtrace.join('\n') if error.backtrace
```
SpComb: `.join("\n")`

jakolehm: Just thinking should we retry if it's dns name resolution error?

jnummelin: To be able to figure that out, I think it would need to dive into the `SocketError` message details. Doing logic based on exception messages is always a bit of a rabbit hole... :D

SpComb: Something like that would also be needed to fix #1653, but it's not really relevant for this change to workaround #2231

SpComb: If my original comment was missed: the `'\n'` prints the backtrace on a single line, this needs to be `"\n"`

PR #2180, opened 2017-04-19
Activate plugin gem and dependencies before loading the plugin

kke:
Currently the CLI using something like `load '/usr/foofoo/lib/kontena_cli_plugin.rb'` to load plugins.

This will not activate the plugins dependencies correctly. The plugins load paths will not be added to `$LOAD_PATH` and `/usr/foofoo/lib/kontena_cli_plugin.rb` will not be added to `$LOADED_FEATURES`.

This PR makes plugin_manager activate the plugin gem and its' dependencies before requiring (instead of loading) the `kontena_cli_plugin.rb`.

With `DEBUG=plugin` you get a bit more detailed report of what was loaded and added to load paths.

Also added "shell" to  the "you have not installed plugin X" notification thing.

Related reading: http://erik.hollensbe.org/2013/05/11/gem-activation-and-you/

PR #2263, opened 2017-05-04
CLI error and debug logging improvements

kke:
This PR enhances the debug/error logging of kontena cli.

* All exception backtraces will go to `~/.kontena/kontena.log` even when `ENV["DEBUG"]` is not set. You don't have  to run the command again to inspect the backtrace.
* All `ENV["DEBUG"] && $stderr.puts(...)` type of loggings have been converted to the block format: `debug { "message" }`  (the block will not be evaluated when log level is INFO and the message severity is DEBUG, saving some precious nanoseconds when not DEBUGging)
* The debug output contains milliseconds elapsed since previous log line, making it possible to do light performance profiling.
* The LOG_TARGET env can be used to control where the log goes to. This makes it possible to do something like `alias k="DEBUG=true LOG_TARGET=$HOME/.kontena/kontena.log kontena"` during development. LOG_TARGET defaults to STDERR when DEBUG is set and `~/.kontena/kontena.log` when DEBUG is unset.


cli/lib/kontena/cli/common.rb
```diff
@@ -11,11 +11,11 @@ module Cli
...
-        return @logger if @logger
-        @logger = Logger.new(ENV["DEBUG"] ? $stderr : $stdout)
-        @logger.level = ENV["DEBUG"].nil? ? Logger::INFO : Logger::DEBUG
-        @logger.progname = 'COMMON'
-        @logger
+        Kontena.logger
+      end
+
+      def debug(msg_or_progname = nil, &block)
+        block_given? ? Kontena.logger.debug(msg_or_progname || Kontena.logger.progname, &block) : Kontena.logger.debug(msg_or_progname)
```
SpComb: See how the [kontena-ipam `Logging`](https://github.com/kontena/kontena-ipam/blob/master/app/logging.rb#L33-L35) class wraps these:

```ruby
  def debug(message = nil, &block)
    logger.add(Logger::DEBUG, message, logging_prefix, &block)
  end
```

OTOH why not have everything just use `logger.debug { ... }` instead?

cli/lib/kontena/client.rb
```diff
@@ -44,8 +45,13 @@ def initialize(api_url, token = nil, options = {})
...
 
-      @logger = Logger.new(ENV["DEBUG"] ? $stderr : $stdout)
-      @logger.level = ENV["DEBUG"].nil? ? Logger::INFO : Logger::DEBUG
+      @logger = Logger.new($LOG_TARGET, 'weekly')
+      @logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+      @logger.formatter = proc do |severity, datetime, progname, msg|
+        timestamp = (1000 * (Time.now.to_f - ($PREVIOUS_LOG_TIME || $KONTENA_START_TIME))).to_i.to_s + "ms"
+        $PREVIOUS_LOG_TIME = Time.now.to_f
+        sprintf("%-6s", timestamp) + " #{progname}: #{msg}\n"
+      end
```
SpComb: Why not use the `Kontena.logger`, instead of copy-pasting this here?

cli/lib/kontena/client.rb
```diff
@@ -528,14 +535,16 @@ def ignore_ssl_errors?
     def handle_error_response(response)
       data = parse_response(response)
 
+      request_path = (last_request && last_request[:path] ? " (#{last_request[:path]})" : '')
```
SpComb: It looks like there is a `response.path` that you can use, avoiding the need for the `@last_request`.

cli/lib/kontena/client.rb
```diff
@@ -321,14 +327,15 @@ def request(http_method: :get, path:'/', body: nil, query: {}, headers: {}, resp
       }.merge(host_options)
 
       request_options.merge!(response_block: response_block) if response_block
+      @last_request = request_options
```
SpComb: I would prefer to get rid of the `@last_response`, and avoid adding a new `@last_request`...

cli/lib/kontena/command.rb
```diff
@@ -218,15 +218,19 @@ def run(arguments)
     end
   rescue Kontena::Errors::StandardError => ex
     raise ex if ENV['DEBUG']
-    abort(" [#{Kontena.pastel.red('error')}] #{ex.class.name} : #{ex.message}")
+    Kontena.logger.error("ERROR") { "#{ex.class.name} : #{ex.message}" }
```
SpComb: Fun fact: `Logger` can also take an `Exception` and print it out, including a backtrace and stuff:

```ruby
irb(main):007:0> begin
irb(main):008:1*   fail "something"
irb(main):009:1> rescue => exc
irb(main):010:1>   logger.error(exc)
irb(main):011:1> end
E, [2017-05-04T17:14:29.874143 #10387] ERROR -- : something (RuntimeError)
(irb):8:in `irb_binding'
...
```

Should we just be using that everywhere, instead of the hundreds of lines of hand-rolled exception-formatting that we have everywhere?

SpComb: Using `"ERROR"` as the `progname` looks weird, it repeats the `ERROR` twice, as the severity and the progname:

`logger.error("ERROR") { "foobar" }` -> `E, [2017-05-04T17:17:33.387675 #10387] ERROR -- ERROR: foobar`

cli/lib/kontena/debug_instrumentor.rb
```diff
@@ -52,14 +53,14 @@ def self.instrument(name, params = {}, &block)
         result << str
       end
 
-      if $stderr.tty?
+      if Kontena.logger.instance_variable_get(:@logdev).dev.tty?
```
SpComb: Ew :D Would be it better to just keep the `DebugInstrumentor` hardcoded to use $stderr?

cli/lib/kontena/cli/config.rb
```diff
@@ -26,20 +26,25 @@ def self.reset_instance
 
       def initialize
         super
-        @logger = Logger.new(ENV["DEBUG"] ? $stderr : $stdout)
-        @logger.level = ENV["DEBUG"].nil? ? Logger::INFO : Logger::DEBUG
+        @logger = Logger.new($LOG_TARGET, 'weekly')
```
SpComb: Why not use the `Kontena.logger` here, instead of duplicating this code?

cli/lib/kontena_cli.rb
```diff
@@ -87,6 +109,20 @@ def self.cli_root(*joinables)
...
+  def self.logger
+    return @logger if @logger
+    require 'logger'
+    @logger = Logger.new($LOG_TARGET, 'weekly')
+    @logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+    @logger.progname = 'CLI'
+    @logger.formatter = proc do |severity, datetime, progname, msg|
+      timestamp = (1000 * (Time.now.to_f - ($PREVIOUS_LOG_TIME || $KONTENA_START_TIME))).to_i.to_s + "ms"
+      $PREVIOUS_LOG_TIME = Time.now.to_f
+      sprintf("%-6s", timestamp) + " #{progname}: #{msg}\n"
```
SpComb: Timestamps for profiling are good, but I might prefer the default absolute timestamp formatting with µs resolution over only having relative ms timestamps for everything... They have more information (was this log from the last run, or something I tested yesterday?), but you it's still pretty easy to eyeball the relative durations:

```
D, [2017-05-04T14:25:10.029109 #2661] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2017-05-04T14:25:10.039161 #2661] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2017-05-04T14:25:10.043157 #2661] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2017-05-04T14:25:10.051214 #2661] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2017-05-04T14:25:10.068203 #2661] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
```

cli/lib/kontena_cli.rb
```diff
@@ -87,6 +109,20 @@ def self.cli_root(*joinables)
       File.join(Kontena.root, 'lib/kontena/cli', *joinables)
     end
   end
+
+  def self.logger
+    return @logger if @logger
+    require 'logger'
+    @logger = Logger.new($LOG_TARGET, 'weekly')
+    @logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+    @logger.progname = 'CLI'
```
SpComb: The agent/server have the nice touch of having this `def self.logger` being an `include` module, that uses the `self.class.name` as the progname. Would that also make sense for the CLI?

cli/lib/kontena/client.rb
kke: Getting rid of @last_response would require that `client.get(...)` etc would return a response object with `.success?`, `.data[]` etc, which would require changing everything that uses the client anywhere.

kke:
The rotation now holds only 1 log file of 64kb of size.

cli/lib/kontena/client.rb
kke: Mostly because of custom progname and because it was already duplicated.

cli/lib/kontena/command.rb
kke: All backtracelogging biz now done with `logger.error(ex)`. This only works with the default Logger::Formatter which is super noisy with PID's and what not.

cli/lib/kontena_cli.rb
kke: That would initialize a bunch of extra loggers writing to the same log device. Logger is only thread safe with one instance.

cli/lib/kontena/client.rb
```diff
@@ -44,9 +45,7 @@ def initialize(api_url, token = nil, options = {})
       uri = URI.parse(@api_url)
       @host = uri.host
 
-      @logger = Logger.new(ENV["DEBUG"] ? $stderr : $stdout)
-      @logger.level = ENV["DEBUG"].nil? ? Logger::INFO : Logger::DEBUG
-      @logger.progname = 'CLIENT'
+      @logger = Kontena.logger
```
SpComb: Would there be some nice way to set `@logger.progname = "CLIENT"` here, to avoid the progname repitition throughout the class?

It would need some method to return a logger, rather than just a singleton accessor... would it be worth making a mixin module for that?

cli/lib/kontena_cli.rb
```diff
@@ -1,3 +1,11 @@
+require 'logger'
+
+$KONTENA_START_TIME = Time.now.to_f
+at_exit do
+  Kontena.logger.debug { "Execution took #{(Time.now.to_f - $KONTENA_START_TIME).round(2)} seconds" }
```
SpComb: :+1: this is useful

cli/lib/kontena_cli.rb
```diff
@@ -87,6 +115,27 @@ def self.cli_root(*joinables)
...
+      @logger.formatter = ColorStrippingFormatter.new
+    end
+    @logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+    @logger.progname = 'CLI'
+    @logger
+  end
+
+  class ColorStrippingFormatter < Logger::Formatter
+    def msg2str(msg)
+      msg.gsub!(/\e+\[{1,2}[0-9;:?]+m/m, '') if msg.kind_of?(String)
```
SpComb: Mutating strings is probably best avoided... and this will probably fail with `ruby2.3 --enable-frozen-string-literal`?

```ruby
    msg = msg.gsub(....)
```

jnummelin:
> The rotation now holds only 1 log file of 64kb of size.
Feels a bit smallish. If the full API jsons are logged then this fills up pretty fast.

SpComb:
Needs merge for  #2180, but otherwise LGTM

cli/lib/kontena/command.rb
```diff
@@ -218,15 +218,17 @@ def run(arguments)
...
   rescue Errno::EPIPE
     # If user is piping the command outputs to some other command that might exit before CLI has outputted everything
     abort
   rescue Clamp::HelpWanted, Clamp::UsageError
     raise
   rescue => ex
     raise ex if ENV['DEBUG']
-    abort(" [#{Kontena.pastel.red('error')}] #{ex.class.name} : #{ex.message}\n         Rerun the command with environment DEBUG=true set to get the full exception")
+    Kontena.logger.error(ex)
+    abort(" [#{Kontena.pastel.red('error')}] #{ex.class.name} : #{ex.message}\n         See #{$LOG_TARGET} or run the command again with environment DEBUG=true set to see the full exception")
```
SpComb: Should be `Kontena.log_target`?

```
 [error] RuntimeError : orly
         See  or run the command again with environment DEBUG=true set to see the full exception
```

cli/lib/kontena_cli.rb
```diff
@@ -87,6 +115,26 @@ def self.cli_root(*joinables)
       File.join(Kontena.root, 'lib/kontena/cli', *joinables)
     end
   end
+
+  def self.logger
+    return @logger if @logger
+    if log_target.respond_to?(:tty?) && log_target.tty?
+      logger = Logger.new(log_target)
+      logger.datetime_format = "%M%S.%2N "
```
SpComb: This datetime format is rather odd, having just minutes and seconds reads funny. Also, the sub-second formatting could have at least three decimals to ms percision?

```
D, [4554.94 #873] DEBUG -- CONFIG: Current master: development
D, [4554.95 #873] DEBUG -- CONFIG: Current grid: development
```

kke:
It breaks `-D`

cli/lib/kontena_cli.rb
kke: The full timestamp is so long. I know or don't care what day and time it is when running CLI commands. The log file gets full timestamps.

The `DEBUG` duplicates `D`. The PID is 99% irrelevant and when it is relevant, you could see it in the log file.

SpComb: Sure, I agree that the timestamps for interactive `DEBUG=true` should be shorter than the default, and the PID isn't relevant here (the PID is relevant for e.g. the server logging, with multiple puma workers).

But `%M%S` isn't readable either. What is `4554.94`? Four and a half thousand seconds plus 940ms?

SpComb: I suggest something like `%T.%3N`. It's short and makes sense when you first see it.

```
irb(main):001:0> Time.now.strftime('%T.%3N')
=> "12:19:22.053"
```

jnummelin: That's how I would read it. So clearly needs some tuning still. I'd include also the hour, if a command runs just when the hour changes, the output would jump from `5959.90` -> `0000.10` or something alike. Just not readable/understandable IMO.

SpComb: Some kind of "seconds since process start" would also be fine, but I doubt there's an easy `logger.datetime_format` for that.

kke: It was not meant to be readable but informative for enlightened individuals :)
 
The latest commit restores the ms since last line display. It could also display ms since start, but I would guess every time you actually want to use that information it's the duration between events you are interested in, as you get the total execution time in the at_exit hook anyway.

jnummelin:
Some failing specs:
```
Failed examples:

rspec ./spec/kontena/kontena_cli_spec.rb:45 # Kontena#run! accepts a command line as a list of parameters
rspec ./spec/kontena/kontena_cli_spec.rb:50 # Kontena#run! accepts a command line as an array
rspec ./spec/kontena/kontena_cli_spec.rb:40 # Kontena#run! accepts a command line as string
rspec ./spec/kontena/kontena_cli_spec.rb:60 # Kontena#run! Re-raises the SystemExit when command exits with non-zero status
rspec ./spec/kontena/kontena_cli_spec.rb:55 # Kontena#run! Returns true if the command exits with SystemExit status 0
rspec ./spec/kontena/kontena_cli_spec.rb:83 # Kontena#run Returns true when nothing was raised
rspec ./spec/kontena/kontena_cli_spec.rb:78 # Kontena#run Returns true when status is 0
```

jnummelin:
The latest debug format seems to be ok IMO:
```
DEBUG=1 k node show foo
[D   40ms CLI] Kontena CLI 1.3.0.dev (ruby-2.3.3+x86_64-darwin15)
Activating plugin kontena-plugin-hello
Loading plugin kontena-plugin-hello
[D  100ms CLI] Running Kontena::MainCommand with ["node", "show", "foo"] -- callback matcher = 'nil'
[D    0ms CLI] Running Kontena::Cli::NodeCommand with ["show", "foo"] -- callback matcher = 'nil'
[D    1ms CLI] Running Kontena::Cli::Nodes::ShowCommand with ["foo"] -- callback matcher = 'nodes show'
[D    0ms CONFIG] Loading configuration from /Users/jussi/.kontena_client.json
[D    2ms CONFIG] Configuration loaded with 39 servers.
[D    0ms CONFIG] Current master: mac
[D    0ms CONFIG] Current grid: e2e
[D    0ms CLIENT] Excon opts: {:omit_default_port=>true, :connect_timeout=>5, :read_timeout=>30, :write_timeout=>5, :ssl_verify_peer=>true, :instrumentor=>Kontena::DebugInstrumentor}
[D    0ms CLIENT] [Request]: GET http://localhost/v1/nodes/e2e/foo | Headers: {Accept: application/json, Authorization: Bearer} 
[D   95ms CLIENT] [Error]: Expected([200, 201, 204]) <=> Actual(404 Not Found)

[D    0ms CLIENT] Request GET /v1/nodes/e2e/foo: 404 Not Found: {"error":"Not found"}
[D    0ms CLI] Execution took 0.24 seconds
[D    0ms CLI] Kontena::Errors::StandardError status 404
/Users/jussi/code/kontena2/kontena/cli/lib/kontena/client.rb:547:in `handle_error_response': Not found (/v1/nodes/e2e/foo) (Kontena::Errors::StandardError)
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/client.rb:353:in `rescue in request'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/client.rb:294:in `request'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/client.rb:207:in `get'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/cli/nodes/show_command.rb:14:in `execute'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/command.rb:201:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/command.rb:201:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
	from /Users/jussi/code/kontena2/kontena/cli/lib/kontena/command.rb:201:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/command.rb:132:in `run'
	from /Users/jussi/code/kontena2/kontena/cli/bin/kontena:17:in `<top (required)>'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/bin/kontena:22:in `load'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/bin/kontena:22:in `<main>'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/bin/ruby_executable_hooks:15:in `eval'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/bin/ruby_executable_hooks:15:in `<main>'
```

The exec times are not the whole truth as they are calculated between log entries. But it's still pretty close to reality and useful info.

kke:
> Activating plugin kontena-plugin-hello
> Loading plugin kontena-plugin-hello

Looks like that one outputs debug info differently. (plugin load times are interesting info)

kke:
>The exec times are not the whole truth as they are calculated between log entries. But it's still pretty close to reality and useful info.

Yes it's easy to get confused and think that the reported time was spent during that event and not between this and the previous log line.

```
[D    3ms CLI] Calculating a zillion pi decimals
[D    900ms CLI] Calculating 1+1
```

jnummelin:
@SpComb what do you think about the format?

kke:
```
[D    ^3ms CLI] Calculating a zillion pi decimals
[D  ^900ms CLI] Calculating 1+1
```

? :)

jnummelin:
Can we nail down the format, the code itself looks good.

The current format (ms since last log entry) is kinda handy to see the rough execution times, but as stated in above comments it can also be bit misleading.

/cc @jakolehm @SpComb

cli/lib/kontena/cli/log_formatters/compact.rb
```diff
@@ -0,0 +1,55 @@
...
+          when 'W' then Kontena.pastel.yellow('W')
+          when 'I' then Kontena.pastel.cyan('I')
+          when 'E', 'F' then Kontena.pastel.red('E')
+          else severity[0..0]
+          end
+        end
+
+        def colorize_time
+          elapsed = self.class.ms_since_last
+          str = sprintf("%6s", "#{elapsed}ms")
```
SpComb: > Can we nail down the format, the code itself looks good.
> 
> The current format (ms since last log entry) is kinda handy to see the rough execution times, but as stated in above comments it can also be bit misleading.

My suggestion would be something like `"%7.3f" % (Time.now.to_f - @first_log)`. Could still keep the coloring based on the `ms_since_last ` thresholds?

kke: 12 characters is a bit much. You could then just as well display "HH:MM:SS.%3N" which is 12 chars also.

SpComb: I meant 8 chars (the 7 includes the decimal point + 3 decimal places)

```
irb(main):001:0> "%7.3fs" % 0.0
=> "  0.000s"
```

kke:
image

jnummelin:
So this latest format shows time since the CLI launch and highlight the times if they go above the thresholds?

kke:
Yes. And omits the progname since it doesn't contain any relevant information value.

jnummelin:
@SpComb The latest format ok for you as well?

kke:
1 minute

jnummelin:
@kke `status/merge`?

PR #2265, opened 2017-05-05
Prevent WeaveWorker from doing duplicate subscription for container:event

### LINKED ISSUE #2225: Agent WeaveWorker starts seeing container events twice after weave restart event
Introduced by #1398 in Kontena 1.0, happens on `weave` container `restart` events: the `WeaveWorker` will re-call `start`, which will also re-subscribe to `container:event`, so the `WeaveWorker` starts getting multiple copies of each Docker event.

This can happen if weave crashes and gets restarted by Docker.

```
I, [2017-04-26T11:31:32.186142 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=067cc8d4aace21339d38f4bac43fff62b6aef7c55dee1b47fcfb164484a05772 at cidr=10.81.128.99/16 
I, [2017-04-26T11:31:32.198249 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=067cc8d4aace21339d38f4bac43fff62b6aef7c55dee1b47fcfb164484a05772 at cidr=10.81.128.99/16
```

Further restart events will also get handled multiple times...

```
I, [2017-04-26T11:25:17.126175 #1]  INFO -- Kontena::Workers::WeaveWorker: Restart after weave restart
I, [2017-04-26T11:25:17.127935 #1]  INFO -- Kontena::Workers::WeaveWorker: Restart after weave restart
D, [2017-04-26T11:25:17.824282 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.7s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:25:17.985824 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.9s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:25:18.411287 #1] DEBUG -- Kontena::NetworkAdapters::WeaveExecutor: weaveexec stream: ["--local", "ps"]
D, [2017-04-26T11:25:19.016518 #1] DEBUG -- Kontena::NetworkAdapters::WeaveExecutor: weaveexec stream: ["--local", "ps"]
```

In fact... they multiply exponentially? :bomb: 

```
D, [2017-04-26T11:42:46.313313 #1] DEBUG -- Kontena::Workers::WeaveWorker: skip start for container=/weave without overlay_cidr
D, [2017-04-26T11:42:46.340485 #1] DEBUG -- Kontena::Workers::WeaveWorker: skip start for container=/weave without overlay_cidr
D, [2017-04-26T11:42:46.416475 #1] DEBUG -- Kontena::Workers::WeaveWorker: skip start for container=/weave without overlay_cidr
D, [2017-04-26T11:42:46.489772 #1] DEBUG -- Kontena::Workers::WeaveWorker: skip start for container=/weave without overlay_cidr
D, [2017-04-26T11:42:47.101451 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.6s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:42:47.194944 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.7s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:42:47.263276 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.7s of 300.0s until: weave running yielded TrueClass
D, [2017-04-26T11:42:47.312263 #1] DEBUG -- Kontena::Workers::WeaveWorker: waited 0.8s of 300.0s until: weave running yielded TrueClass
```

jnummelin:
As WeaveWorker#start can be called multiple times for different cases (weave restart for example) it does subscribe to `container:event` each time. This leads to exponential handling of the events as seen in #2225.

fixes #2225

### DIFF
--- a/agent/lib/kontena/workers/weave_worker.rb
+++ b/agent/lib/kontena/workers/weave_worker.rb
@@ -16,8 +16,8 @@ module Kontena::Workers
 
       @migrate_containers = nil # initialized by #start
 
-      @subscribed_container_event = false
-
+      @started = false # to prevent handling of container events before migration scan
+      subscribe('container:event', :on_container_event)
       if network_adapter.already_started?
         self.start
       else
@@ -33,15 +33,13 @@ module Kontena::Workers
       @migrate_containers = network_adapter.get_containers
       debug "Scanned #{@migrate_containers.size} existing containers for potential migration: #{@migrate_containers}"
 
-      # ready to start handling container events
-      # #start can be called multiple times, subscribe only if not done already as we don't want duplicate events
-      subscribe('container:event', :on_container_event) unless @subscribed_container_event
-      @subscribed_container_event = true
+      @started = true
 
       info 'attaching network to existing containers'
       Docker::Container.all(all: false).each do |container|
         self.start_container(container)
       end
+      
     end
 
     def on_dns_add(topic, event)
@@ -52,6 +50,8 @@ module Kontena::Workers
     # @param [String] topic
     # @param [Docker::Event] event
     def on_container_event(topic, event)
+      return unless started?
+
       if event.status == 'start'
         container = Docker::Container.get(event.id) rescue nil
         if container
@@ -68,6 +68,10 @@ module Kontena::Workers
       elsif event.status == 'destroy'
         self.on_container_destroy(event)
       end
+    end
+
+    def started?
+      @started
     end
 
     # Ensure weave network for container
--- a/agent/lib/kontena/workers/weave_worker.rb
+++ b/agent/lib/kontena/workers/weave_worker.rb
@@ -16,6 +16,8 @@ module Kontena::Workers
 
       @migrate_containers = nil # initialized by #start
 
+      @subscribed_container_event = false
+
       if network_adapter.already_started?
         self.start
       else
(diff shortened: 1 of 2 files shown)


SpComb:
:ok_hand: for fixing this separately from #2173, but I'm uncomfortable making the already-complicated `initialize` `subscribe` -> `start` `maybe-subscribe` logic even more complicated like this.

If not doing the `start_all_containers` split in #2173, then I would suggest moving the `subscribe('container:event')` back into `initialize`, which seems to be the right place for that kind of only-ever-once thing. The reason it was moved to `start` was to prevent `start_container` from running without `@migrate_containers` being populated. Instead, have `start` set some `@started = true` flag, and then `on_container_event` can `return unless @started` to ignore events until ready. IMO this more explicit `initialize` `subscribe container:event` -> `start` `@started = true` -> `on_container_event` `if @started` state seems better than the chained/conditional subscribe logic that is getting out of hand here.

jnummelin:
Refactored based on the suggestions

agent/lib/kontena/workers/weave_worker.rb
```diff
@@ -48,6 +50,8 @@ def on_dns_add(topic, event)
     # @param [String] topic
     # @param [Docker::Event] event
     def on_container_event(topic, event)
+      return unless started?
```
SpComb: Worth a comment like `# cannot run start_container before start has populated @migrate_containers`

PR #2268, opened 2017-05-05
Fix GridServiceSerializer deploy_opts min_health typo

SpComb:
Randomly spotted in the cloud websocket events debug output. I think that's the only thing affected?


SpComb:
Ping @nevalla

Issue #823, opened 2016-06-29
Request timeouts on long-running service container exec commands


christofferh:
We run our master in Heroku which has a 30 second connection timeout, after that the connection is dropped. Running `kontena container exec` with a command that exceeds this time limit will error out in the following fashion:

```
$ kontena container exec my-api-1 sleep 40 && echo "Success!"
<!DOCTYPE html>
    <html>
    <head>
      <meta name="viewport" content="width=device-width, initial-scale=1">
      <style type="text/css">
        html, body, iframe { margin: 0; padding: 0; height: 100%; }
        iframe { display: block; width: 100%; border: none; }
      </style>
    <title>Application Error</title>
    </head>
    <body>
      <iframe src="//s3.amazonaws.com/heroku_pages/error.html">
        <p>Application Error</p>
      </iframe>
    </body>
    </html>
```

Could a solution be to return the request immediately and start tailing the container log (with the addition that exit codes needs to be handled properly)?

SpComb:
#2271 would use a CLI-server websocket connection for the `exec` commands, which would probably also help with request timeouts for long-running execs?

SpComb:
I'm guessing that this is no longer an issue with the websocket-based container execs in 1.3.

PR #2271, opened 2017-05-08
Service and container exec interactive tty support

### LINKED ISSUE #961: CLI: interactive TTY support
Currently `$ kontena container exec redis-1 passwd` or `$ kontena container exec railsapp rails console` will just hang.

There should be a `$ kontena container tty redis-1 passwd` command that runs `docker exec -ti` in the agent where `redis-1` is being deployed and streams the stdout and stdin to/from the CLI.

jakolehm:
Fixes #961

### DIFF
--- a/agent/lib/kontena/actors/container_exec.rb
+++ b/agent/lib/kontena/actors/container_exec.rb
@@ -16,7 +16,6 @@ module Kontena::Actors
       @container = container
       @read_pipe, @write_pipe = IO.pipe
       info "initialized (session #{@uuid})"
-      @mutex = Mutex.new
     end
 
     def input(input)
@@ -25,7 +24,7 @@ module Kontena::Actors
 
     def run(cmd)
       info "starting command: #{cmd}"
-      _, _, exit_code = @container.exec(cmd, @opts) do |stream, chunk|
+      _, _, exit_code = @container.exec(cmd) do |stream, chunk|
         rpc_client.notification('/container_exec/output', [@uuid, stream, chunk.force_encoding(Encoding::UTF_8)])
       end
     ensure 
--- a/agent/lib/kontena/actors/container_exec.rb
+++ b/agent/lib/kontena/actors/container_exec.rb
@@ -1,0 +1,52 @@
+require_relative '../helpers/rpc_helper'
+
+module Kontena::Actors
+  class ContainerExec
+    include Celluloid
+    include Kontena::Logging
+    include Kontena::Helpers::RpcHelper
+
+    exclusive :input
+
+    attr_reader :uuid
+
+    # @param [Docker::Container] container
+    def initialize(container)
+      @uuid = SecureRandom.uuid
+      @container = container
+      @read_pipe, @write_pipe = IO.pipe
+      info "initialized (session #{@uuid})"
+      @mutex = Mutex.new
+    end
+
+    def input(input)
+      @write_pipe.write(input)
+    end
+
+    def run(cmd)
+      info "starting command: #{cmd}"
+      _, _, exit_code = @container.exec(cmd, @opts) do |stream, chunk|
+        rpc_client.notification('/container_exec/output', [@uuid, stream, chunk.force_encoding(Encoding::UTF_8)])
+      end
+    ensure 
+      info "command finished: #{cmd} with code #{exit_code}"
+      rpc_client.notification('/container_exec/exit', [@uuid, exit_code])
+      self.terminate
+    end
+
+    def interactive(cmd)
+      info "starting interactive session: #{cmd}"
+      opts = {tty: true, stdin: @read_pipe}
+      exit_code = 0
+      defer {
+        _, _, exit_code = @container.exec(cmd, opts) do |data|
+          rpc_client.notification('/container_exec/output', [@uuid, 'stdout', data.force_encoding(Encoding::UTF_8)])
+        end
+      }
+    ensure 
+      info "interactive session finished: #{cmd} with code #{exit_code}"
+      rpc_client.notification('/container_exec/exit', [@uuid, exit_code])
+      self.terminate
+    end
+  end
+end
--- a/agent/lib/kontena/actors/container_exec.rb
+++ b/agent/lib/kontena/actors/container_exec.rb
@@ -18,10 +18,12 @@ module Kontena::Actors
       info "initialized (session #{@uuid})"
     end
 
+    # @param [String] input
     def input(input)
       @write_pipe.write(input)
     end
 
+    # @param [String] cmd
     def run(cmd)
       info "starting command: #{cmd}"
       _, _, exit_code = @container.exec(cmd) do |stream, chunk|
@@ -29,10 +31,10 @@ module Kontena::Actors
       end
     ensure 
       info "command finished: #{cmd} with code #{exit_code}"
-      rpc_client.notification('/container_exec/exit', [@uuid, exit_code])
-      self.terminate
+      shutdown(cmd, exit_code)
     end
 
+    # @param [String] cmd
     def interactive(cmd)
       info "starting interactive session: #{cmd}"
       opts = {tty: true, stdin: @read_pipe}
@@ -44,6 +46,12 @@ module Kontena::Actors
       }
     ensure 
       info "interactive session finished: #{cmd} with code #{exit_code}"
+      shutdown(cmd, exit_code)
+    end
+
+    # @param [String] cmd
+    # @param [Integer] exit_code
+    def shutdown(cmd, exit_code)
       rpc_client.notification('/container_exec/exit', [@uuid, exit_code])
       self.terminate
     end
--- a/agent/lib/kontena/rpc/docker_container_api.rb
+++ b/agent/lib/kontena/rpc/docker_container_api.rb
@@ -116,7 +116,7 @@ module Kontena
       # @param [String] container_id
       # @return [Hash]
       def create_exec(container_id)
-        container = Docker::Container.get(id)
+        container = Docker::Container.get(container_id)
         executor = Kontena::Actors::ContainerExec.new(container)
         actor_id = "container_exec_#{executor.uuid}"
         Celluloid::Actor[actor_id] = executor
--- a/agent/lib/kontena/rpc/docker_container_api.rb
+++ b/agent/lib/kontena/rpc/docker_container_api.rb
@@ -1,6 +1,6 @@
 require 'docker'
 require_relative '../helpers/iface_helper'
-
+require_relative '../actors/container_exec'
 module Kontena
   module Rpc
 
@@ -19,7 +19,7 @@ module Kontena
       def show(id)
         container = Docker::Container.get(id)
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -30,7 +30,7 @@ module Kontena
         container = Docker::Container.get(id)
         container.start
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -53,7 +53,7 @@ module Kontena
         container = Docker::Container.get(id)
         container.restart
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -64,7 +64,7 @@ module Kontena
         container = Docker::Container.get(id)
         container.pause
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -75,7 +75,7 @@ module Kontena
         container = Docker::Container.get(id)
         container.unpause
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -87,7 +87,7 @@ module Kontena
         container = Docker::Container.get(id)
         container.kill(opts)
         container.json
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
@@ -97,7 +97,7 @@ module Kontena
       def top(id)
         container = Docker::Container.get(id)
         container.top
-      rescue Docker::Error::NotFoundError => exc
+      rescue Docker::Error::NotFoundError
         raise RpcServer::Error.new(404, 'Not found')
       end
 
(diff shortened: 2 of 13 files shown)


jakolehm:
How to test:

```
$ k service create redis redis:3-alpine && k service deploy redis
$ k service exec --interactive redis sh
```

jnummelin:
There seems to be something failing in an interval:
```
/ # no implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integerno implicit conversion of String into Integer
```

agent/lib/kontena/actors/container_exec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    attr_reader :uuid
+
+    # @param [Docker::Container] container
+    def initialize(container)
+      @uuid = SecureRandom.uuid
+      @container = container
+      @read_pipe, @write_pipe = IO.pipe
+      info "initialized (session #{@uuid})"
+      @mutex = Mutex.new
```
SpComb: Unused?

cli/lib/kontena/cli/containers/exec_command.rb
```diff
@@ -7,15 +7,61 @@ class ExecCommand < Kontena::Command
...
+      end
+      stdin_thread = Thread.new {
+        STDIN.raw {
+          while char = STDIN.readpartial(1024)
+            ws.send(char)
+          end
+        }
+      }
+      stdin_thread.join
+    end
```
SpComb: Some helper module with `interactive_exec(container)` for service+container exec commands.

server/app/services/docker/streaming_executor.rb
```diff
@@ -0,0 +1,70 @@
...
+          @ws.send(JSON.dump({ stream: data['stream'], chunk: data['chunk'] }))
+        end
+      end
+    end
+
+    def register_tty_ws_events
+      init = true
+      @ws.on(:message) do |event|
+        if init == true
+          cmd = ['/bin/sh', '-c', event.data]
```
SpComb: Take the command to exec as a POST parameter, and just keep the websocket for the I/O streaming?

agent/lib/kontena/rpc/docker_container_api.rb
```diff
@@ -109,18 +109,72 @@ def top(id)
...
+        container = Docker::Container.get(id)
+        executor = Kontena::Actors::ContainerExec.new(container)
+        actor_id = "container_exec_#{executor.uuid}"
+        Celluloid::Actor[actor_id] = executor
+        { id: executor.uuid }
+      rescue Docker::Error::NotFoundError
+        raise RpcServer::Error.new(404, 'Container not found')
+      end
+
+      def run_exec(id, cmd)
```
SpComb: Why separate create + run/tty_exec RPCs, and not just a single RPC for both?

This has the slight weirdness of accepting multiple `run_exec/tty_exec` RPCs for the same `create_exec` ID...

agent/lib/kontena/rpc/docker_container_api.rb
```diff
@@ -109,18 +109,72 @@ def top(id)
...
+        if executor
+          executor.async.input(input)
+        end
+      end
+
+      def terminate_exec(id)
+        actor_id = "container_exec_#{id}"
+        executor = Celluloid::Actor[actor_id]
+        if executor
+          Celluloid::Actor.kill(executor) if executor.alive?
```
SpComb: Does this terminate the actual `Docker::Exec` and the exec process? I don't think `Celluloid::Actor.kill` is appropriate here, this should just be a `sync` call that waits for it to stop the exec?

agent/lib/kontena/rpc/docker_container_api.rb
```diff
@@ -109,18 +109,72 @@ def top(id)
...
+        { id: executor.uuid }
+      rescue Docker::Error::NotFoundError
+        raise RpcServer::Error.new(404, 'Container not found')
+      end
+
+      def run_exec(id, cmd)
+        actor_id = "container_exec_#{id}"
+        executor = Celluloid::Actor[actor_id]
+        if executor
+          executor.async.run(cmd)
```
SpComb: Why `async` and not `sync`.... this ignores exec errors.

server/app/services/docker/streaming_executor.rb
jakolehm: Not sure if it's a good idea.. I/O stream might miss something because there would be delay between POST and websocket connection?

agent/lib/kontena/rpc/docker_container_api.rb
jakolehm: Not sure, should we return exec errors using exec session id? If we remove async here it will block until cmd is completed -> we should then also call this api with request (not notify).

agent/lib/kontena/rpc/docker_container_api.rb
jakolehm: I'll refactor run/tty into one RPC endpoint.

agent/lib/kontena/rpc/docker_container_api.rb
jakolehm: It does not kill actual exec process, only actor. I don't know if we actually have any way to kill the process here?

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -12,14 +13,54 @@ class ExecCommand < Kontena::Command
...
+        elsif container['status'] != 'running'
+          exit_with_error "Service #{name} container #{container['name']} is not running, it is #{container['status']}"
+        else
+          interactive_exec(container)
+        end
+      else
+        if interactive?
+          interactive_exec(running_containers.first)
+        else 
+          exec_container(running_containers.first)
```
jakolehm: @SpComb should we change "normal" exec to use websocket (because exec through http call has many problems)?

server/app/services/docker/streaming_executor.rb
SpComb: You can still do the actual exec RPC on the `ws.open`, but I mean take the command from the POST parameters, not from the first websocket frame. Simpler and cleaner than special-casing the first frame?

cli/lib/kontena/cli/services/exec_command.rb
SpComb: yes, using websockets for long-running execs would probably work to fix #823

server/app/services/docker/streaming_executor.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def create_session
+      @exec_session = @client.request('/containers/create_exec', @container.container_id)
+    end
+
+    def subscribe_to_session
+      @subscription = MongoPubsub.subscribe("container_exec:#{@exec_session['id']}") do |data|
+        if data.has_key?('exit')
+          @ws.send(JSON.dump({ exit: data['exit'] }))
+          @ws.close
```
SpComb: This could also use the close `code` (1000 vs something else) with a JSON payload for the exit event?

cli/lib/kontena/cli/helpers/exec_helper.rb
```diff
@@ -0,0 +1,32 @@
+module Kontena::Cli::Helpers
+  module ExecHelper
+
+    # @param [WebSocket::Client::Simple] ws 
+    # @return [Thread]
+    def stream_stdin_to_ws(ws)
```
SpComb: If we want complete support for interactive TTY execs, then we may also need to consider support for things like `SIGTTYWINCH` to send Docker API `POST exec/.../resize?h=40&w=80`?

jakolehm: Yes, but that can be added later.

agent/lib/kontena/rpc/docker_container_api.rb
jnummelin: `container.exec` returns a `Docker::Exec` instance but it does not seem to have any capabilities to kill the process.

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -69,16 +46,114 @@ def execute
...
         return ret
       elsif instance
-        if !(container = service_containers.find{|container| container['instance_number'] == instance})
+        if !(container = service_containers.find{|c| c['instance_number'] == instance})
           exit_with_error "Service #{name} does not have container instance #{instance}"
         elsif container['status'] != 'running'
           exit_with_error "Service #{name} container #{container['name']} is not running, it is #{container['status']}"
         else
-          exec_container(container)
+          interactive_exec(container)
```
jakolehm: Fixed.

agent/lib/kontena/actors/container_exec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    exclusive :input
+
+    attr_reader :uuid
+
+    # @param [Docker::Container] container
+    def initialize(container)
+      @uuid = SecureRandom.uuid
+      @container = container
+      @read_pipe, @write_pipe = IO.pipe
```
jnummelin: Not sure if any other naming would make it more clear to which "direction" each of these are meant for. Now, at least for me, it always takes a couple of readings to figure out. :)

cli/lib/kontena/cli/helpers/exec_helper.rb
jnummelin: +1 for adding `SIGTTYWINCH` etc. later on

agent/lib/kontena/rpc/docker_container_api.rb
SpComb: Yes, it looks like the Docker API doesn't have anything to kill an exec, which is surprising.

If you SIGINT a `docker exec -i ... sleep 300` via Ctrl-c, then it just kills the CLI process, and leaves the exec process (`sleep 300`) running inside the container.

With `docker exec -it` with raw tty mode, a ^C results in "\x03" on the TTY, which the CLI passes through to the API, and gets delivered to the pseudotty within the container, signaling the process.

SpComb: Still think the `Celluloid::Actor.kiill` is a bit drastic, that's an interrupting `Thread.kill` which should only be used when absolutely necessary, not under normal operation.

Just call some normal method that does `self.terminate` instead?

cli/lib/kontena/cli/helpers/exec_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    # @param [Websocket::Frame::Incoming] msg
+    def handle_message(msg)
+      data = parse_message(msg)
+      if data.is_a?(Hash)
+        if data.has_key?('exit')
+          exit data['exit'].to_i
+        else
+          $stdout << data['chunk']
```
SpComb: The `data['stream']` stdout vs stderr?

cli/lib/kontena/cli/helpers/exec_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+      end
+    rescue => exc
+      $stderr << "#{exc.class.name}: #{exc.message}"
```
SpComb: `logge.error exc`

agent/lib/kontena/rpc/docker_container_api.rb
```diff
@@ -109,18 +109,69 @@ def top(id)
...
+
+        {}
+      end
+
+      def tty_input(id, input)
+        actor_id = "container_exec_#{id}"
+        executor = Celluloid::Actor[actor_id]
+        if executor
+          executor.async.input(input)
+        end
```
SpComb: `raise RpcServer::Error.new` 404

server/app/services/docker/streaming_executor.rb
```diff
@@ -0,0 +1,106 @@
...
+      @ws.on(:message) do |event|
+        begin
+          data = JSON.parse(event.data)
+          if data.has_key?('cmd')
+            if shell
+              cmd = ['/bin/sh', '-c', data['cmd'].join(' ')]
+            else 
+              cmd = data['cmd']
+            end
+            @client.notify('/containers/run_exec', @exec_session['id'], cmd, false)
```
SpComb: This will re-notify the `run_exec` if there are more websocket messges unless the client does pass `?interactive=true`

The further execs are likely to fail, but still result in a `/contanier_exec/exit` RPC from the agent?

server/app/services/docker/streaming_executor.rb
SpComb: I take this back, the websocket connection must be a GET request and can't have any POST request body parameters.

It might still be possible to take the command as query parameter or header, but reading the command from the first websocket frame also makes sense.

SpComb:
If I have the kontena master added with a URL like `http://localhost:9292/` with a trailing slash, then the `kontena service exec redis ls` command fails and gets stuck without actually running the exec or exiting. Server logs:

```
127.0.0.1 - - [30/May/2017:13:10:53 +0300] "GET //v1/containers/development/core-02/redis-1/exec HTTP/1.1" 404 - 0.0937
2017-05-30 13:10:53 +0300: HTTP parse error, malformed request (): #<Puma::HttpParserError: Invalid HTTP format, parsing fails.>
2017-05-30 13:10:53 +0300: ENV: {"rack.version"=>[1, 3], "rack.errors"=>#<IO:<STDERR>>, "rack.multithread"=>true, "rack.multiprocess"=>false, "rack.run_once"=>false, "SCRIPT_NAME"=>"", "QUERY_STRING"=>"", "SERVER_PROTOCOL"=>"HTTP/1.1", "SERVER_SOFTWARE"=>"puma 3.6.0 Sleepy Sunday Serenity", "GATEWAY_INTERFACE"=>"CGI/1.2"}
```

With some wireshark dump, looks like the server fails the initial websocket request on the leading `//...` double-slash, and then the initial websocket frame is parsed as a pipeline request, so the server responds with HTTP 400, which seems to get ignored by the `WebSocket::Client::Simple`... there is no `on :close` or `on :error` event :(

```
GET //v1/containers/development/core-02/redis-1/exec HTTP/1.1
Upgrade: websocket
Connection: Upgrade
Host: localhost:9292
Authorization: Bearer 87e3736531ae619025235a9d36aed0ea70a96f4f1855601b68799433e72f30b7
Accept: application/json
Sec-WebSocket-Version: 13
Sec-WebSocket-Key: d89iVer+ziZIdyqnO9dOvA==

HTTP/1.1 404 Not Found
Content-Type: text/html
X-Kontena-Version: 1.3.0.pre2
Content-Length: 0

...V.@.t.-.t...:.b.+HTTP/1.1 400 Bad Request
```

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -69,16 +47,116 @@ def execute
...
+        exit 1
+      end
+      
+      stream_stdin_to_ws(ws).join
+    end
+
+    # @param [Hash] container
+    # @return [String]
+    def ws_url(container)
+      "#{require_current_master.url.sub('http', 'ws')}/v1/containers/#{container['id']}/exec"
```
SpComb: Fails if the current master URL has a trailing `/`, the resulting `GET //v1/...` results in HTTP 404

agent/lib/kontena/actors/container_exec.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    # @param [String] cmd
+    def interactive(cmd)
+      info "starting interactive session: #{cmd}"
+      opts = {tty: true, stdin: @read_pipe}
+      exit_code = 0
+      defer {
+        _, _, exit_code = @container.exec(cmd, opts) do |data|
+          rpc_client.notification('/container_exec/output', [@uuid, 'stdout', data.force_encoding(Encoding::UTF_8)])
```
jnummelin: At least the usual tty control codes seem to work, tested:
- redis-cli, with the fancy `intellisense`
- lolcat, heavy use of color codes
- tty flasher:
```
flasher () { while true; do printf \\e[?5h; sleep 0.1; printf \\e[?5l; read -s -n1 -t1 && break; done; }
```

jnummelin:
There are some opportunities to improve/refactor, but this is a good starting point.

One item that needs some attention is the fact that the CLI side WS client does not play too well with un-expected responses such as 404. That just leaves the client hanging.

cli/lib/kontena/cli/containers/exec_command.rb
```diff
@@ -1,21 +1,43 @@
...
+      base = self
+      ws = connect(ws_url("#{current_grid}/#{container_id}"), token)
+      ws.on :message do |msg|
+        base.handle_message(msg)
+      end
+      ws.on :open do
+        ws.send(cmd)
+      end
+      ws.on :error do |e|
+        exit_with_error('fuck')
```
SpComb: :thinking:

cli/lib/kontena/cli/services/exec_command.rb
```diff
@@ -69,16 +47,98 @@ def execute
...
+    # @param [Hash] container
+    # @return [Boolean]
+    def normal_exec(container)
+      base = self
+      cmd = JSON.dump({ cmd: cmd_list })
+      exit_status = nil
+      token = require_token
+      url = ws_url(container['id'])
+      url << '?shell=true' if shell?
+      ws = connect(url, token)
```
SpComb: Missing `ws.on :error` handler to abort the CLI.

If any of the `on :open` etc handlers raises an error, then the CLI just dies silently, because the `Websocket::Client::Simple` thread dies, and so the main CLI thread just gets stuck waiting for it.

SpComb:
> the server responds with HTTP 400, which seems to get ignored by the `WebSocket::Client::Simple`... there is no `on :open`, `on :close` or `on :error` event :(

Actually, for a 404 response, it does call `on :open`, but not `on :error`. 

This is because [`Websocket::Client::Simple` calls `Websocket::Handshake::Client#finished?`](https://github.com/shokai/websocket-client-simple/blob/v0.3.0/lib/websocket-client-simple/client.rb#L51), but [that is also true on errors](https://github.com/imanel/websocket-ruby/blob/v1.2.3/lib/websocket/handshake/base.rb#L42-L44). It [needs to also call `valid?`](https://github.com/imanel/websocket-ruby/blob/v1.2.3/lib/websocket/handshake/client.rb#L28-L32), and emit an error instead of open in that case...

I don't see an upstream issue for this...

PR #2272, opened 2017-05-08
Speed up --bash-completion-path for faster shell loading

### LINKED ISSUE #2270: Whoami command (for bash completions) is very slow
Hi,

After enabling Bash completions for Kontena CLI, I noticed it takes a long time to load my bash profile and initialize new terminals. I traced this to the `kontena whoami` command which takes 1.2 seconds to run. Is there any way we can optimize this?

I figured out I can get the path using this command (although I'm not sure it is correct or robust):

```
gem contents kontena-cli | grep /lib/kontena/scripts/init | head -n 1
```

This still takes 0.5 seconds to run, but that's better.

It seems silly that such a simple command should be so expensive. Any ideas?

kke:
Fixes #2270 

This PR moves the handling of `--bash-completion-path` to `bin/kontena` to avoid loading so much files for nothing.

Before:

```console
$ time for i in {1..10}; do kontena whoami --bash-completion-path; done
real	0m6.766s
user	0m5.279s
sys	0m1.185s
```

After:

```console
$ time for i in {1..10}; do kontena whoami --bash-completion-path; done
real	0m1.109s
user	0m0.665s
sys	0m0.207s
```

### DIFF
--- a/cli/bin/kontena
+++ b/cli/bin/kontena
@@ -11,6 +11,8 @@ STDOUT.sync = true
 if ARGV[0] == 'complete'
   ARGV.delete_at(0)
   require 'kontena/scripts/completer'
+elsif ARGV.last == '--bash-completion-path'
+  puts File.realpath(File.join(__dir__, '../lib/kontena/scripts/init'))
 else
   require 'kontena_cli'
   Kontena::PluginManager.instance.init
--- a/cli/lib/kontena/cli/whoami_command.rb
+++ b/cli/lib/kontena/cli/whoami_command.rb
@@ -5,6 +5,7 @@ class Kontena::Cli::WhoamiCommand < Kontena::Command
 
   def execute
     if self.token?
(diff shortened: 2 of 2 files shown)


SpComb:
Also update the `cli/README.md` documentation to just use `kontena --bash-completion-path`? Still need to keep the `kontena whoami --bash-completion-path` for backwards-compatibility.

kke:
There will be a complete change to the completion stuff in the near future anyway, at which point the `--bash-completion-path` should give a deprecation warning.

kke:
Fixed in #2335

kke:
Actually this is still valid.

SpComb:
Should the docs be updated to use `NO_PLUGINS=true kontena whoami --bash-completion-path` instead?

kke:
There's a PR's for all plugins to load faster (by using the subcommand loader)

kke:
Outdated

PR #2221, opened 2017-04-25
Fix server deploy queuing

SpComb:
Fixes #2208 Server: GridServiceHealthMonitorJob can pile up deployments
Fixes #2212 Service deploys can get stuck if there are several queued deploys when a service finishes deploying
Fixes #2213 Queued deploys time out after 10s if another deploy is running, leaving them stuck
Fixes #2214 Stopping service does not abort running deploy
Fixes #2215 Server GridServiceDeployer can run for longer than 5-10 minutes, allowing further deploys to run simultaneously 

This introduces a new `queued_at` timestamp for `GridServiceDeploy`s. The `GridServiceSchedulerWorker` picks up `created` deploys that do not yet have any `queued_at` set, and atomically sets their `queued_at` timestamp. While holding the `check_deploy_queue` dlock, the worker then checks if the service is `deploy_running?` (has deploys with `started_at` set but not `finished_at`). If the there are no running deploys, then the worker sets `started_at` and releases the dlock. This prevents other deploys from being started while the worker proceeds to run the `GridServiceDeployer`, which sets the state to `ongoing`. Once the `GridServiceDeployer` is done, it sets `finished_at`, and the deploy is complete in either the `success` or `error` state.

In the case of multiple pending deploys for the same service, the pending deploys simply remain queued up with their `queued_at` timestamp set, but not having any `started_at` timestamp set. These queued deploys will get picked up again once their `queued_at` timestamp is more than 30s in the past, if they are still not `started_at`. This also works in the case of dlock timeouts or server crashes, allowing the queued deploy to be retried. The 30s re-queueing interval must be long enough to ensure that any failed attempts at acquiring the dlock have timed out.

The `ServiceDeployQueuedAt` migration will set the `queued_at` timestamp for any existing deploys, causing them to be ignored by the new `GridServiceSchedulerWorker`, provided that they have a `started_at` set. This takes care of existing deploys that have been left in the `created` state with `started_at` set, which can currently be caused by dlock timeouts (see #2213) or the earlier queueing logic if the server was stopped before the `after(30) { ... }` task was run. These should be much less frequent with the changes in this PR, since any `queued_at` deploys will be picked up again after 30s. The migration should also behave correctly in the case of any pending deployments when the server was shut down: they will be run 30s after the migration happens, because they have not yet been marked as `started_at` :)

[...]

PR #2195, opened 2017-04-21
[WIP] Fancy deploy progress monitoring

kke:
Seems to work quite ok but perhaps a bit too complex for now.

Would fix #2114

PR #2281, opened 2017-05-08
Fix handling of aborted service deploys

SpComb:
Fixes the handling of aborted services deploys that remained TODO in #2221

* Fix the `GridServiceDeployer` deploy abort error handling

    The error message was broken: `undefined method' 'deploy_state' for #<GridServiceDeployer:0x00000004cb1ca0>`

    The spec was also broken, and was actually failing with `Celluloid::NotActive`.

* Fix the `GridServiceDeploy#abort!` to set the `GridServiceDeploy#deploy_state` to `error`
    
    This fixes the `StackDeployWorker` and CLI stack/service deploy helpers to correctly handle aborted deploys as un-succesful cases.

* Fix the CLI `kontena stack deploy` helpers to wait for `finished_at`, instead of the deploy state

    The `kontena stack deploy` command would previously hang on an aborted service deploy, because it expects a `success` or `error` state instead of looking at the `finished_at`. Although this change to set the `error` state fixes this, it's better and more reliable to just wait for `finished_at`. It also matches what the `kontena service deploy` does.

### DIFF
--- a/cli/lib/kontena/cli/stacks/stacks_helper.rb
+++ b/cli/lib/kontena/cli/stacks/stacks_helper.rb
@@ -37,12 +37,12 @@ module Kontena::Cli::Stacks
           service_deploy = deployment['service_deploys'].find{ |s| s['state'] == 'ongoing' }
           if service_deploy
             tracked_services << service_deploy['id']
-            wait_for_service_deploy(service_deploy, states)
+            wait_for_service_deploy(service_deploy)
           end
           if states.include?(deployment['state'])
             deployed = true
             deployment['service_deploys'].select{ |s| !tracked_services.include?(s['id']) }.each do |s|
-              wait_for_service_deploy(s, states)
+              wait_for_service_deploy(s)
             end
           else
             sleep 1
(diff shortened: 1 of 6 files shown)


SpComb:
Changed this PR to just have `GridServiceDeploy#abort!` set the `error` state, instead of the more complicated approach of changing all users of the grid service deploy state to check for `!success?` after `finished_at` is set.

cli/lib/kontena/cli/stacks/stacks_helper.rb
```diff
@@ -69,19 +69,15 @@ def wait_for_deploy_to_finish(deployment, timeout = 600)
...
-          r = client.get("services/#{service_deploy['service_id']}/deploys/#{service_deploy['id']}")
-          if states.include?(r['state'])
-            service_deployed = true
-            spin.fail if r['state'] == 'error'
-          else
-            sleep 1
-          end
+        until service_deploy['finished_at']
+          sleep 1
+          service_deploy = client.get("services/#{service_deploy['service_id']}/deploys/#{service_deploy['id']}")
```
SpComb: Now using the same logic as in the CLI [`ServicesHelper#wait_for_deploy_to_finish`](https://github.com/kontena/kontena/blob/6db8a3f89da6cca391d1c331d5ad4c3ae0e3ab79/cli/lib/kontena/cli/services/services_helper.rb#L290-L293). The two really need refactoring to share the same code, but that's over in #2195

PR #2280, opened 2017-05-08
Abort deploy gracefully in exceptional cases

jnummelin:
Fixes `GridServiceSchedulerWorker#check_deploy_queue` to "gracefully" abort deploys in case of unhandled errors.

### TODO
- Should there be a specific check for removed service as happened in #2275? It would be probably nicer in logs instead of:
```
NoMethodError: undefined method `deploy_running?' for nil:NilClass
...
```

fixes #2275

PR #2284, opened 2017-05-09
Do not run workers in server tests or utilities

SpComb:
The server `GridServiceSchedulerWorker`, `StackDeployWorker` and `StaackRemoveWorker` actors were running during the server specs. This caused flaky specs when creating `GridServiceDeploy` objects in specs, because they might get modified by the `GridServiceSchedulerWorker` between the create and the expect:

```
  1) GridServiceSchedulerWorker#check_deploy_queue fails if fetch_deploy_item returns a non-pending deploy
     Failure/Error: expect{subject.check_deploy_queue}.to raise_error(RuntimeError, "deploy not pending")
       expected RuntimeError with "deploy not pending" but nothing was raised
     # ./spec/jobs/grid_service_scheduler_worker_spec.rb:49:in `block (3 levels) in <top (required)>'
```

```
  2) GridServices::Scale#run sets state to deploy_pending
     Failure/Error: expect {
       expected result to have changed from false to true, but did not change

     # ./spec/mutations/grid_services/scale_spec.rb:35:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (2 levels) in <top (required)>'
```

This extends the same `ENV['RACK_ENV'] == 'test'` logic used to not run the server jobs to also prevent running the workers. The `Stacks::Remove` and `Stacks::Deploy` specs already mocked out the corresponding workers, so the specs still pass without those workers running.

### DIFF
--- a/server/app/boot_jobs.rb
+++ b/server/app/boot_jobs.rb
@@ -5,6 +5,5 @@ require_relative 'services/mongodb/migrator'
 unless ENV['RACK_ENV'] == 'test' || ENV['NO_MONGO_PUBSUB']
   MongoPubsub.start!(PubsubChannel)
   JobSupervisor.run!
+  WorkerSupervisor.run!
 end
-
-WorkerSupervisor.run!


SpComb:
Per the `NO_MONGO_PUBSUB` flag it seems like this also applies to the `rake reset_admin` and `bin/kontena-console` processes: they would have run the `GridServiceSchedulerWorker` actors before?

jnummelin:
oh, that could've caused some odd effects. Luckily the `reset_admin` task is quite short living, not sure if many are using the `bin/console` command.

SpComb:
> oh, that could've caused some odd effects. Luckily the reset_admin task is quite short living, not sure if many are using the bin/console command.

Those and the specs should really be fixed to just require `app/boot.rb` without requiring `app/boot_jobs.rb`

server/app/boot_jobs.rb
```diff
@@ -5,6 +5,5 @@
 unless ENV['RACK_ENV'] == 'test' || ENV['NO_MONGO_PUBSUB']
```
jakolehm: Just wondering why `NO_MONGO_PUBSUB` skips also supervisors...

SpComb:
This seems to also hit the `GridServiceDeployer` specs hard:

```
  1) GridServiceDeployer for a grid with host nodes #deploy for multiple concurrent instance deploys fails the service deploy if one of the concurrent instance deploys fail

     Failure/Error: expect{

       expected result to have changed to 2, but is now 1

     # ./spec/services/grid_service_deployer_spec.rb:162:in `block (5 levels) in <top (required)>'
```

```
  2) GridServiceDeployer for a grid with host nodes #deploy for multiple concurrent instance deploys fails the service deploy if aborted

     Failure/Error: expect(grid_service_deploy.reason).to eq "halting deploy of test-grid/null/redis, deploy was aborted: testing"

       

       expected: "halting deploy of test-grid/null/redis, deploy was aborted: testing"

            got: "halting deploy of test-grid/null/redis, deploy was aborted: service deploy aborted: deploy not pending"

       

       (compared using ==)

     # ./spec/services/grid_service_deployer_spec.rb:193:in `block (5 levels) in <top (required)>'
```

Not sure about the first one, but the second one is definitely a result of this + #2280

server/app/boot_jobs.rb
SpComb: It's a badly named hack, but it seems to only be used by the rake and `bin/kontena-console` utilities to avoid running background tasks/pubsub stuff.

```
./app/boot_jobs.rb:unless ENV['RACK_ENV'] == 'test' || ENV['NO_MONGO_PUBSUB']
./lib/tasks/reset_admin.rake:    ENV['NO_MONGO_PUBSUB'] = 'true'
./bin/kontena-console:ENV['NO_MONGO_PUBSUB'] = 'true'
```

SpComb: Turns out that the `bin/kontena-console` and `rake kontena:reset_admin` processes do not even load `app/boot_jobs.rb`, so AFAIK the `NO_MONGO_PUBSUB` env does not actually have any effect on anything.

OTOH it does mean that if something else is changed and those processes do end up requiring `app/boot_jobs.rb`, then they would still be protected from launching the supervisors/workers. IMO for this purpose it would actually be better to `fail if ENV['NO_MONGO_PUBSUB']` instead?

SpComb:
Tested that `rake kontena:reset_admin` and `bin/kontena-console` still work with this change to not start the workers when `NO_MONGO_PUBSUB` is defined. They don't currently load the `app/boot_jobs.rb` file, so it's a no-op.

Issue #1979, opened 2017-03-15
No such image: weaveworks/weaveexec:1.8.2


panuhorsmalahti:
I've had this error a few times during a deployment:

node1:
```
I, [2017-03-15T10:02:22.507868 #1]  INFO -- Kontena::NetworkAdapters::Weave: Attach container=e5a9ddc0e26ec87159d6ec95ba3fb55874fcbd1f9a2e9475a43aaf7df5c5f729 at cidr=10.81.128.23/16
E, [2017-03-15T10:02:22.515015 #1] ERROR -- : Actor crashed!
Docker::Error::NotFoundError: No such image: weaveworks/weaveexec:1.8.2

	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:46:in `rescue in request'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:38:in `request'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
	/usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/container.rb:323:in `create'
	/app/lib/kontena/network_adapters/weave_executor.rb:22:in `execute'
	/app/lib/kontena/network_adapters/weave_executor.rb:115:in `attach'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
[...]

PR #2287, opened 2017-05-09
Better error handling for ServicePodWorker#image_outdated?

jnummelin:
In case of `Docker::Image.get` raised some error the `image_outdated?` would just return true with no indication why it failed. `rescue nil` strikes-again. :)

With the fix any un-expected errors are just let through and eventually seen on cli and event logs:
```
$ k service deploy nginx
 [fail] Deploying service nginx      
 [error] Kontena::Errors::StandardErrorArray : halting deploy of e2e/null/nginx, one or more instances failed:
	⊗ Failed to deploy instance e2e/null/nginx-1 to node moby: GridServiceInstanceDeployer::ServiceError: RuntimeError: BoomBoom
```
```
$ k service events nginx
TIME                      TYPE                 MESSAGE
2017-05-09T12:29:51.492Z  deploy               service e2e/null/nginx deploy started
2017-05-09T12:29:54.682Z  deploy               Failed to deploy service instance e2e/null/nginx-1 to node moby: GridServiceInstanceDeployer::ServiceError: RuntimeError: BoomBoom
2017-05-09T12:29:54.688Z  deploy               deploy of e2e/null/nginx errored: halting deploy of e2e/null/nginx, one or more instances failed
```

### DIFF
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -195,10 +195,13 @@ module Kontena::Workers
     # @param [Docker::Container] service_container
     # @return [Boolean]
     def image_outdated?(service_container)
-      image = Docker::Image.get(service_pod.image_name) rescue nil
-      return true unless image
-      return true if image.id != service_container.info['Image']
-
+      begin
+        image = Docker::Image.get(service_pod.image_name)
+        return true if image.id != service_container.info['Image']
+      rescue Docker::Error::NotFoundError
+        info "failed to get image: #{service_pod.image_name}"
+        return true
+      end
       false
     end
 
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -197,7 +197,10 @@ module Kontena::Workers
     def image_outdated?(service_container)
       begin
         image = Docker::Image.get(service_pod.image_name)
-        return true if image.id != service_container.info['Image']
+        if image.id != service_container.info['Image']
+          info "containers image(#{service_container.info['Image']} does not match images id(#{image.id})}"
+          return true
+        end
       rescue Docker::Error::NotFoundError
         info "failed to get image: #{service_pod.image_name}"
         return true
--- a/agent/spec/lib/kontena/workers/service_pod_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/service_pod_worker_spec.rb
@@ -291,6 +291,20 @@ describe Kontena::Workers::ServicePodWorker do
       allow(Docker::Image).to receive(:get).with(service_pod.image_name).and_return(image)
       expect(subject.image_outdated?(service_container)).to be_falsey
     end
+
+    it 'returns true if Docker::Error::NotFoundError raised' do
+      service_container = double(:service_container)
+      allow(Docker::Image).to receive(:get).with(service_pod.image_name).and_raise(Docker::Error::NotFoundError)
+      expect(subject.image_outdated?(service_container)).to be_truthy
+    end
+
+    it 'lets un-expected errors fall through'  do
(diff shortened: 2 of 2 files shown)


agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -195,10 +195,13 @@ def container_outdated?(service_container)
     # @param [Docker::Container] service_container
     # @return [Boolean]
     def image_outdated?(service_container)
-      image = Docker::Image.get(service_pod.image_name) rescue nil
-      return true unless image
-      return true if image.id != service_container.info['Image']
-
+      begin
+        image = Docker::Image.get(service_pod.image_name)
+        return true if image.id != service_container.info['Image']
```
SpComb: We could also start logging INFO-level messages when these conditions are true.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -195,10 +195,16 @@ def container_outdated?(service_container)
...
-      return true if image.id != service_container.info['Image']
-
+      begin
+        image = Docker::Image.get(service_pod.image_name)
+        if image.id != service_container.info['Image']
+          info "containers image(#{service_container.info['Image']} does not match images id(#{image.id})}"
+          return true
+        end
+      rescue Docker::Error::NotFoundError
+        info "failed to get image: #{service_pod.image_name}"
```
jnummelin: Not directly on this code path, but there seems to be cases when docker gives `No such image` although image listing shows them in https://github.com/kontena/kontena/issues/1979

SpComb: > Not directly on this code path, but there seems to be cases when docker gives No such image although image listing shows them in #1979

That's from `Docker::Container.create({'Image' => ...})` with some odd `docker.io/...` prefix in the `docker images` output...

The `ImagePullWorker#ensure_image` will already run `Docker::Image.get(...)` and abort with an error if that fails in any way.

SpComb: I think the best way to avoid this ambiguity would be for the `ImagePullWorker` to return the `Docker::Image`, and just use that here in this check. No need to think about errors here.

The `ImagePullWorker` shouldn't even need to call `Docker::Image.get`, because the `Docker::Image.create` will itself run `get` and return that.

SpComb: I'm not saying that this `rescue` code here is broken... but it feels impossible for it to ever happen, so I'm not sure what the agent really should do if it does happen. Better to just let it fail, unless we know how this could ever happen and how the agent should deal with it in that scenario. This `return true` assumes that the `ServicePods::Creator` -> `Docker::Container.create` with that image might still succeed.

Using the `Docker::Image` returned by the `ImagePullWorker` would avoid these "impossible" errors.

SpComb: Given, the `ImagePullWorker#ensure_image` will not touch the Docker API if the image is cached... but the `ServicePodWorker` only calls `ImagePullWorker#ensure_image` if the deploy_rev has changed, so it seems guaranteed that there will have been two succesfull calls to `Docker::Image.get` immediately prior to this third call. I seems bizarre if this one would then fail.

So this change to limit the `rescue` => `return true` behavior to the `Docker::Error::NotFound` case is definitely an improvement over the existing `rescue nil', but it doesn't make any sense anymore in the larger picture of how the `ServicePodWorker` and `ImagePullWorker` currently interact?

jnummelin: Yes, I think you're right. I'll close this one as the possible refactoring how `ServicePodWorker` and `ImagePullWorker`  interact is a bit of different thing.

jnummelin:
Closing as `not-relevant`, see the discussion for reasonings.

PR #2304, opened 2017-05-11
Fix Agent ServicePodWorker#container_outdated? to fail if service updated_at is in the future

### LINKED ISSUE #2298: Agent keeps re-creating service container if master node clock is ahead of host node clock
The when deploying an updated service with changes, or using `kontena service deploy --force`, the server will update the `GridService.updated_at` timestamp in mongo with its own `Time.now`. When applying the deploy and determining if the container needs to be re-created, the agent [`ServicePodWorker#container_outdated?`](https://github.com/kontena/kontena/blob/fd4771a393e63ea171ea8690a324206483597095/agent/lib/kontena/workers/service_pod_worker.rb#L191) compares the `updated_at` timestamp provided by the server with the `Created` timestamp provided by the Docker API.

If the master and host node clocks are out of sync such that the service `updated_at` timestamp set by the server is in the future, this leads to a loop where the agent `ServicePodWorker` considers the newly created container as being out of date, and thus keeps re-creating it. Until the agent's clock eventually catches up with the server's clock, the newly created container will always have a `Created` timestamp that is earlier than the service's `updated_at` timestamp.

SpComb:
Fixes #2298 by not having the agent re-create service containers in a loop if the server clock is ahead of the agent clock

* Fail the `ServicePodWorker#apply` if the service `updated_at` timestamp is more than `KONTENA_CLOCK_SKEW=1.0` seconds in the future

    This causes the service deploy to fail with an informative error instead when the service is updated, and the agent clock is in the past / master clock is in the future.

* Log INFO when the service container is out of date

## Testing

Using `timedatectl` to set the host node date into the past, and deploying with an updated timestamp:

```
$ kontena service deploy --force redis
 [fail] Deploying service redis      
 [error] Kontena::Errors::StandardErrorArray : halting deploy of development/null/redis, one or more instances failed:
	⊛ Deployed instance development/null/redis-1 to node core-01
	⊗ Failed to deploy instance development/null/redis-2 to node core-02: GridServiceInstanceDeployer::ServiceError: RuntimeError: service updated_at 2017-05-11T10:17:42+00:00 is in the future
```
```
D, [2017-05-11T10:16:58.677023 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-2: running
W, [2017-05-11T10:16:58.691241 #1]  WARN -- Kontena::Workers::ServicePodWorker: failed to sync redis-2 at 2017-05-11 10:18:00 UTC: service updated_at 2017-05-11T10:17:42+00:00 is in the future
W, [2017-05-11T10:16:58.691650 #1]  WARN -- Kontena::Workers::ServicePodWorker: service updated_at 2017-05-11T10:17:42+00:00 is in the future (RuntimeError)
/app/lib/kontena/workers/service_pod_worker.rb:194:in `container_outdated?'
/app/lib/kontena/workers/service_pod_worker.rb:172:in `service_container_outdated?'
/app/lib/kontena/workers/service_pod_worker.rb:98:in `ensure_desired_state'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
/app/lib/kontena/workers/service_pod_worker.rb:33:in `update'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/async.rb:7:in `dispatch'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
```

Once the host node clock catches up, then the deploy will still pass:

```
D, [2017-05-11T10:18:26.325570 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-2: running
I, [2017-05-11T10:18:26.352555 #1]  INFO -- Kontena::Workers::ServicePodWorker: service updated at 2017-05-11T10:17:42+00:00 after service container created at 2017-05-11T07:12:34+00:00
I, [2017-05-11T10:18:26.353040 #1]  INFO -- Kontena::Workers::ServicePodWorker: re-creating redis-2
```


SpComb:
A different approach would be to drop the `Created` vs `updated` timestamp comparison and use the `io.kontena.container.service_revision` label instead. That would avoid the clock skew issues.

SpComb:
Ideally the agent would still re-create the container on the initial `apply` when the `updated` timestamp changes on a `deploy_rev` bump, even if the `updated` timestamp is the future, but just fail with warnings on the later `apply` cycles. That would allow the `kontena service deploy` to succeed despite the clock skew, but it would make the `ServicePodWorker` state even more complicated than it already is.

SpComb:
Even with the #2144 check for clock skew when connecting, this is still relevant, as the clock might drift while the agent is connected. There should probably still be a little bit of acceptable skew, though.

jnummelin:
Do we still need this as the connection handshake already has the clock skew checks?

SpComb:
The clock skew can happen after the agent connects, so yes, the agent can still go into a recreate-container loop if the agent or server clock drifts while the websocket connection remains open.

SpComb:
Either this or #2371 should be included in 1.4.

@jakolehm PTAL, added a skew threshold?

jakolehm:
I would prefer this over #2371 (for 1.4).

PR #2306, opened 2017-05-12
Node scheduling availability

### LINKED ISSUE #1030: Node evacuate
It should be possible to "evacuate" a node for example before doing maintenance work (or before terminating a node). Evacuate could just disable node from accepting any new services and reschedule stateless services to different nodes .

jnummelin:
This PR implements node evacuate functionality.

When a node is evacuated, all stateless services on it will be re-scheduled for other nodes. Stateful services will be stopped if they are running.

Once a node is put in evacuate mode with `kontena node update --availability drain xyz`, no services will be deployed onto it. When a node is taken out of evacuate mode by `kontena node update --availability active xyz` it will again be available for service scheduling.

fixes #1030 

### Todo

- [x] API docs
- [x] Docs


server/app/mutations/host_nodes/evacuate.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def validate
+      add_error(:host_node, :already_evacuated, "Node #{self.host_node.name} is already evacuated") if self.host_node.evacuated?
+    end
+
+    def execute
+      self.host_node.set(:evacuated => true)
+
+      re_deploy_needed_services(self.host_node)
```
SpComb: Thinking... is this exactly equivalent to marking the node as offline, and running the `GridScheduler.reschedule`, or is are some differences?

jnummelin: `GridScheduler.reschedule` would run for all services. In evacuation case only some of the services need scheduling.

jnummelin:
Per the internal discussion I refactored this to be more generic in term of the possible "schedulability" states. Now there's more generic `availability` state for each node. This PR implements the evacuation by using state `drain` and a node can be put into active scheduling again by setting the state to `active`.

jnummelin:
@jakolehm @SpComb Any feedback on the design level highly appreciated.

cli/lib/kontena/cli/nodes/update_command.rb
```diff
@@ -5,14 +5,18 @@ class UpdateCommand < Kontena::Command
...
     def execute
       require_api_url
       require_current_grid
       token = require_token
 
       data = {}
-      data[:labels] = label_list if label_list
+      node = client.get("nodes/#{current_grid}/#{node_id}")
+      # Send the previous labels if none given
+      data[:labels] = label_list.empty? ? Array(node['labels']) : label_list
```
SpComb: It can just omit the `labels` from the POST if empty, it doesn't need to roundtrip them.... I think?

Similar fix in https://github.com/kontena/kontena/pull/2504#discussion_r124803890, needs an issue, I guess

docs/using-kontena/nodes.md
```diff
@@ -42,6 +42,36 @@ The re-scheduling of grid services will happen after the node offline grace peri
 - `ha`: 2 minutes
 - `random`: 30 seconds
 
+### Node availability
+
+Node availablity for scheduling services can be controlled with `kontena node availablity` command. The current supported states are: `drain`and `active`.
```
SpComb: Would `kontena node update --availability=active moby` also work? Not saying which is better.

docs/using-kontena/nodes.md
```diff
@@ -42,6 +42,36 @@ The re-scheduling of grid services will happen after the node offline grace peri
...
+
+Kontena scheduler can assign tasks into any node marked as `active` availability. `active` is also the default state for all nodes. To re-activate a node use `kontena node availability xyz active`
+
+#### Drain
+
+For a planned maintenance or node decommisioning it is a good idea to first drain the node so there are minimal disruptions on the services running. To drain the services from a node use `kontena node availability xyz drain`
+
+Node draining means that all the stateless services will be re-scheduled immediately out from the node. Any stateful services running on that node will be stopped.
+
+Once the maintenance is over and a node can be put back to work it can be put into active state using `kontena node availability xyz dactive` command. This marks the node back in a `active` state where it will get new deployments as Kontena scheduler will see this node back in the available nodes list.
```
SpComb: > dactive

docs/using-kontena/nodes.md
```diff
@@ -42,6 +42,36 @@ The re-scheduling of grid services will happen after the node offline grace peri
...
+... snip ...
+```
+
+#### Active
+
+Kontena scheduler can assign tasks into any node marked as `active` availability. `active` is also the default state for all nodes. To re-activate a node use `kontena node availability xyz active`
+
+#### Drain
+
+For a planned maintenance or node decommisioning it is a good idea to first drain the node so there are minimal disruptions on the services running. To drain the services from a node use `kontena node availability xyz drain`
```
SpComb: What about `kontena node drain xyz`? It's a verb.

server/app/routes/v1/nodes_api.rb
```diff
@@ -82,6 +81,21 @@ def load_grid_node(grid_name, node_id)
           end
         end
 
+        r.post do
+          # POST /v1/nodes/:grid/:node/availability
+          r.on 'availability' do
```
SpComb: Dead code?

server/app/mutations/host_nodes/update.rb
```diff
@@ -4,19 +4,81 @@ module HostNodes
...
 
-      self.host_node
+      # Update availability if given and trigger appropriate actions
+      if self.availability && self.host_node.availability != self.availability
+
+        self.host_node.set(:availability => self.availability)
+
+        case self.availability
+        when 'active'
+          start_stateful_services(self.host_node)
```
SpComb: What about this `desired_state: 'stopped' if stateful && host_node.drain?` logic into the `Rpc::ServicePodSerializer`, or perhaps the `GridServiceInstanceDeployer`?

Then this wouldn't need to `stop_stateful_services` / `start_stateful_services`, just trigger deploys for each service on the node. Benefits:

* Errors halfway through these operations here might leave some of the services started/stopped - a deploy would be needed to fix those (re-update would not work)
* This does not actually wait for the services to start/stop, and automation trying to drain a node probably wants to wait for the services to stop - the standard deploy tracking would work for that, if this returned an array of service deploy IDs to track.

server/app/mutations/host_nodes/update.rb
```diff
@@ -4,19 +4,81 @@ module HostNodes
...
     include Common
+    include Logging
 
     required do
       model :host_node
       array :labels, nils: true
     end
 
+    optional do
+      string :availability, in: ['active', 'drain']
```
SpComb: Wasn't there also some third state, which would just stop new instances from being scheduled to the node, but keep existing instances running?

docs/using-kontena/nodes.md
SpComb: See below... `kontena node drain` could also wait for the services to actually stop?

SpComb:
PR description is missing #1030

server/app/mutations/host_nodes/update.rb
jnummelin: I think it was `pause` on some other orchestrator. (could not find it though on Docker Swarm or Rancher docs, maybe it was Kube). Anyway that can be added as separate PR IMO.

docs/using-kontena/nodes.md
jnummelin: oh, I forgot to update the docs in https://github.com/kontena/kontena/pull/2306/commits/a48f420b56d3d66bf77aeb95986f280201af09a1, that changed the availability to use `kontena node update --availability`

docs/using-kontena/nodes.md
jnummelin: CLI would have to figure out which service instances should be left on the node. Plus the corner case of draining the last node in the grid. Maybe enhancement for another PR?

server/app/mutations/host_nodes/update.rb
jnummelin: From the API point-of-view it feels a bit weird if `v1/nodes/...` API would be returning deploy IDs as they are not really directly relevant resources.

docs/using-kontena/nodes.md
jnummelin: If we want to have some waiting or other logic bound to `draining` or or other modes it really needs to be separate command from generic node update. @jakolehm what say you?

server/app/routes/v1/nodes_api.rb
jnummelin: yes, as per the internal discussion functionality was incorporated into node update API and forgot to remove these

cli/lib/kontena/cli/nodes/update_command.rb
jnummelin: I guess I was trying to use the API in "pure" PUT mode instead of using PUT in PATCH fashion :)

server/app/mutations/host_nodes/update.rb
jnummelin: If we want to really track the `drain` triggered deploys, IMO it has to be separate API (as it was in the beginning of this PR) that returns the deploy IDs created.

server/app/services/scheduler/filter/availability.rb
```diff
@@ -0,0 +1,26 @@
...
+      ##
+      # @param [GridService] service
+      # @param [Integer] instance_number
+      # @param [Array<HostNode>] nodes
+      # @raise [Scheduler::Error]
+      def for_service(service, instance_number, nodes)
+
+        nodes = nodes.select { |n| 
+          case n.availability
+          when 'active'
```
jakolehm: Availability strings should be constants in `HostNode` class?

cli/lib/kontena/cli/node_command.rb
```diff
@@ -7,6 +7,8 @@ class Kontena::Cli::NodeCommand < Kontena::Command
   subcommand ["remove","rm"], "Remove node", load_subcommand('nodes/remove_command')
   subcommand "label", "Node label specific commands", load_subcommand('nodes/label_command')
   subcommand "health", "Check node health", load_subcommand('nodes/health_command')
+  subcommand "availability", "Updatea node availability status", load_subcommand('nodes/availability_command')
+  subcommand "unevacuate", "Un-evacuate node", load_subcommand('nodes/unevacuate_command')
```
SpComb: Looks like you reviewed an older version of the branch.

cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -38,6 +38,7 @@ def fields
         version: 'agent_version',
         status:  'status',
         initial: 'initial',
+        availability: 'availability',
```
SpComb: I think we should be careful not to add too many new columns to `node ls`... how about showing the status as `drained` (in yellow?) instead of `online` for these nodes? Offline nodes would still just be shown as offline.

cli/lib/kontena/cli/nodes/update_command.rb
```diff
@@ -11,13 +11,15 @@ class UpdateCommand < Kontena::Command
 
     option ["-l", "--label"], "LABEL", "Node label", multivalued: true
     option "--clear-labels", :flag, "Clear node labels"
+    option "--availability", "AVAILABILITY", "Node scheduling availability"
```
SpComb: This should have some convenient reference for the available options:

`option "--availability", "active|drain", "Node scheduling availability"`

server/app/mutations/host_nodes/update.rb
```diff
@@ -18,7 +23,63 @@ def execute
...
+      end
+
+      self.host_node.reload
+    end
+
+    # Re-deploy needed services, new deployments will filter out evacuated node
+    def re_deploy_needed_services(host_node)
+      services = find_services(host_node)
+      services.each do |service|
+        GridServiceDeploy.create!(grid_service: service)
```
jnummelin: yes, IMO should be harmless.

server/app/models/host_node.rb
```diff
@@ -31,6 +36,7 @@ class HostNode
   field :agent_version, type: String
   field :docker_version, type: String
   field :connected_at, type: Time
+  field :availability, type: String, default: Availability::ACTIVE
```
jnummelin: I avoided the enum since it created bit of a hassle in mongoid upgrade and instead went with enum-like class module thingy.

cli/lib/kontena/cli/nodes/list_command.rb
jnummelin: You mean something like this:
image
@jakolehm WDYT?

jakolehm: Looks good. Green icon (on the left) describes node healthiness, right?

server/app/mutations/host_nodes/update.rb
jakolehm: Would it make this too bad if we just trust the scheduler and not force deployments?

jnummelin: It just makes the drain operation not to start "immediately" as it will happen basically on next scheduling loop.

cli/lib/kontena/cli/nodes/update_command.rb
jnummelin: good idea

server/app/mutations/host_nodes/update.rb
jakolehm: I think it's fair tradeoff, and we can always improve that scheduling loop to notice these kind of changes faster.

jnummelin: Done

SpComb: This `re_deploy_needed_services` method + `find_services` is now dead code?

jnummelin: yes, thought I removed those but apparently didn't

server/app/mutations/host_nodes/update.rb
```diff
@@ -18,7 +23,47 @@ def execute
...
+      end
+
+      self.host_node.reload
+    end
+
+    def stop_stateful_services(host_node)
+      host_node.grid_service_instances.each do |instance|
+        if instance.grid_service.stateful?
+          info "setting desired state to stopped for instance #{instance.grid_service.to_path}-#{instance.instance_number}"
+          instance.set(desired_state: 'stopped')
```
SpComb: I would have said something about a `kontena service deploy` of the drained stateful service setting the `GridServiceInstance` `desired_state` back to `running`, but that's moot, because the drained state filters out the node and prevents the stateful service from being deployed:

```
$ kontena service deploy redis-stateful
 [fail] Deploying service redis-stateful      
 [error] 500 : Cannot find applicable node for service instance development/null/redis-stateful-2: Strategy Scheduler::Strategy::HighAvailability did not find any node:
	⊛ Deployed instance development/null/redis-stateful-1 to node core-01
```

PR #2241, opened 2017-05-02
Make stack environment variable validation fail if array item has no equals sign

kke:
Fixes #2238

PR #2058, opened 2017-04-04
Improve mutations to return multiple errors for a field

SpComb:
Fixes #2238 server-side validation of service array fields when creating/updating stacks

Many of the mutations validations use a fixed field even if they happen in a loop, which means that multiple invalid items for a service will only return a single validation error for the last invalid item, clobbering any validation errors for other preceding invalid items.

Converts the following mutations error fields into hashes:

* stack `volumes`

Converts the following mutations error fields into arrays, to allow each individual validation error to be returned:

* service `links`
* service `secrets`
* service `volumes`

Fixes the server stacks mutations to correctly handle array fields, where the resulting `Mutations::ErrorArray` currently breaks on `handle_service_outcome_errors` (see https://github.com/kontena/kontena/issues/2238#issuecomment-298569759)

```
ArgumentError (Invalid kind):
  /home/kontena/kontena/kontena/server/vendor/bundle/ruby/2.3.0/gems/mutations-0.7.2/lib/mutations/command.rb:113:in `add_error'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/common.rb:13:in `add_service_error'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/common.rb:21:in `block in handle_service_outcome_errors'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/common.rb:20:in `each'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/common.rb:20:in `handle_service_outcome_errors'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/create.rb:35:in `block in validate_services'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/create.rb:29:in `each'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/create.rb:29:in `validate_services'
  /home/kontena/kontena/kontena/server/app/mutations/stacks/create.rb:25:in `validate'
```

Monkey-patches `Mutations::Command#add_error` to support adding structured `Mutations::ErrorHash` and `Mutations::ErrorArray` errors. Also adds a new `validate_each` that builds a `Mutations::ErrorArray`, similar to the `Mutations::ArrayFilter` validation.

## Example
```yaml
stack: test/test
services:
  foo:
    image: redis
    depends_on:
      - asdf1
      - asdf2

  bar:
    image: redis
    depends_on:
      - asdf/foo
      - asdf3
```
#### `kontena stack install`
```
 [fail] Creating stack test      
[...]

PR #2315, opened 2017-05-15
Validate service envs on server, simplify syntax for parsing/validation

SpComb:
The regexps used to parse/validate service environment variables in #2241 included support for escapes like `FOO\=BAR=bar`, having kontena parse them as `{'FOO\\=BAR' => 'bar'}`. However, I don't think this kind of syntax to escape `=` chars in environment variable names makes a whole lot of sense, because I doubt that other applications are going to parse them the same way?

Remove support for special handling of such escapes, restoring the previous `split('=', 2)` implementation, because any attempt to use them is unlikely to end behave in any sane way. So far, I've only sen ruby interpret `ENV['FOO\\=BAR']`.

This PR also uses the same regexp to validate the service envs on the server side. However, the validation failures will result in HTTP 500 errors for the stacks mutations, pending improved support for handling complex error outcomes in #2058.

#### `kontena.yml`
```yaml
services:
  foo:
    environment:
      - FOO\=BAR=QUUX
```

#### `docker inspect`
```
[
    {
        "Config": {
            "Env": [
                "FOO\\=BAR=QUUX",
            ],
        },
    }
]
```

### `bash`
```
$ echo "${FOO\\}"
bash: ${FOO\\}: bad substitution
$ echo "${FOO\\=BAR}"
bash: ${FOO\\=BAR}: bad substitution
```

### `python`
```
>>> os.environ
{..., 'FOO\\': 'BAR=QUUX', ...}
>>> os.environ['FOO\\']
'BAR=QUUX'
>>> os.environ['FOO\\=BAR']
Traceback (most recent call last):
  ...
KeyError: 'FOO\\=BAR'
```

### `irb`
```
irb(main):001:0> ENV
=> {..., "FOO\"=>"BAR=QUUX", ...}
irb(main):002:0> ENV['FOO\\']
=> "BAR=QUUX"
irb(main):003:0> ENV['FOO\\=BAR']
=> "QUUX"
irb(main):011:0> ENV['FOO=BAR']  
=> nil
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/validations.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validations.rb
@@ -43,7 +43,7 @@ module Kontena::Cli::Stacks::YAML
               )
             end
           elsif value.is_a?(Array)
-            value.all? { |v| v.kind_of?(String) && v =~ /\A\S+=/ }
+            value.all? { |v| v.kind_of?(String) && v =~ /\A[^=]+=/ }
           else
             false
           end
--- a/cli/lib/kontena/cli/stacks/yaml/validations.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validations.rb
@@ -43,7 +43,7 @@ module Kontena::Cli::Stacks::YAML
               )
             end
           elsif value.is_a?(Array)
-            value.all? { |v| v.kind_of?(String) && v =~ /^.+=.*/ }
+            value.all? { |v| v.kind_of?(String) && v =~ /^([^=]+)=(.*)/ }
           else
             false
           end
--- a/cli/lib/kontena/cli/stacks/yaml/validations.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validations.rb
@@ -43,7 +43,7 @@ module Kontena::Cli::Stacks::YAML
               )
             end
           elsif value.is_a?(Array)
-            value.all? { |v| v.kind_of?(String) && v =~ /^([^=]+)=(.*)/ }
+            value.all? { |v| v.kind_of?(String) && v =~ /\A\S+=/ }
           else
             false
           end
--- a/cli/lib/kontena/cli/stacks/yaml/validations.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validations.rb
@@ -43,7 +43,7 @@ module Kontena::Cli::Stacks::YAML
               )
             end
           elsif value.is_a?(Array)
-            value.all? { |v| v.kind_of?(String) && v =~ /\A\S+(?<!\\)=.*/ }
+            value.all? { |v| v.kind_of?(String) && v =~ /^.+=.*/ }
           else
             false
           end
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -172,7 +172,7 @@ module GridServices
           end
           string :entrypoint
           array :env do
-            string matches: /\A\S+=/
+            string matches: /\A[^=]+=/
           end
           array :secrets do
             hash do
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -172,7 +172,7 @@ module GridServices
           end
           string :entrypoint
           array :env do
-            string matches: /^([^=]+)=(.*)/
+            string matches: /\A\S+=/
           end
           array :secrets do
             hash do
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -172,7 +172,7 @@ module GridServices
           end
           string :entrypoint
           array :env do
-            string
+            string matches: /^([^=]+)=(.*)/
           end
           array :secrets do
             hash do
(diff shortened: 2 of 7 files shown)


cli/lib/kontena/cli/stacks/yaml/validations.rb
```diff
@@ -43,7 +43,7 @@ def common_validations
               )
             end
           elsif value.is_a?(Array)
-            value.all? { |v| v.kind_of?(String) && v =~ /\A\S+(?<!\\)=.*/ }
+            value.all? { |v| v.kind_of?(String) && v =~ /^([^=]+)=(.*)/ }
```
SpComb: What's the best regexp to validate a sane env, without rejecting crazy edge cases that people might actually be using?

kke: I don't know but `^` is not the same as `\A`.

server/spec/mutations/stacks/create_spec.rb
```diff
@@ -300,6 +300,29 @@
       expect(outcome.errors.message).to eq 'services' => { 'foo' => {'name' => "Create failed"}, 'bar' => { 'links' => "Service soome-stack/foo does not exist"}}
     end
 
+    pending 'reports service error array outcomes' do
```
SpComb: This is pending the fixes in #2058 for handling `Mutations::ErrorArray` in `handle_service_outcome_errors`:

```
  1) Stacks::Create#run reports service error array outcomes
     Failure/Error: outcome = described_class.new(
     ArgumentError:
       Invalid kind
     # ./vendor/bundle/ruby/2.3.0/gems/mutations-0.7.2/lib/mutations/command.rb:113:in `add_error'
     # ./app/mutations/stacks/common.rb:13:in `add_service_error'
     # ./app/mutations/stacks/common.rb:21:in `block in handle_service_outcome_errors'
     # ./app/mutations/stacks/common.rb:20:in `each'
     # ./app/mutations/stacks/common.rb:20:in `handle_service_outcome_errors'
     # ./app/mutations/stacks/create.rb:35:in `block in validate_services'
     # ./app/mutations/stacks/create.rb:29:in `each'
     # ./app/mutations/stacks/create.rb:29:in `validate_services'
     # ./app/mutations/stacks/create.rb:25:in `validate'
     # ./vendor/bundle/ruby/2.3.0/gems/mutations-0.7.2/lib/mutations/command.rb:67:in `initialize'
     # ./spec/mutations/stacks/create_spec.rb:312:in `new'
     # ./spec/mutations/stacks/create_spec.rb:312:in `block (4 levels) in <top (required)>'
     # ./spec/mutations/stacks/create_spec.rb:311:in `block (3 levels) in <top (required)>'
```

server/app/mutations/grid_services/common.rb
```diff
@@ -172,7 +172,7 @@ def common_validations
           end
           string :entrypoint
           array :env do
-            string
+            string matches: /^([^=]+)=(.*)/
```
kke: `^` matches a character after any newline in the string, not the beginning of the string.

cli/lib/kontena/cli/stacks/yaml/validations.rb
SpComb: Right, it gets a little crazy if there's newlines in the env string..

```ruby
irb(main):005:0> /^([^=]+)=(.*)/.match "ASDF\nFOO=bar\nquux"
=> #<MatchData "ASDF\nFOO=bar" 1:"ASDF\nFOO" 2:"bar">
irb(main):006:0> /\A([^=]+)=(.*)/.match "ASDF\nFOO=bar\nquux"
=> #<MatchData "ASDF\nFOO=bar" 1:"ASDF\nFOO" 2:"bar">
```

What about `/\A\S+=(.*)\Z/`?

```
irb(main):017:0> /\A\S+=(.*)\Z/.match "ASDF\nFOO=bar\nQUUX"
=> nil
irb(main):018:0> /\A\S+=(.*)\Z/.match "FOO=bar\nQUUX"
=> nil
irb(main):019:0> /\A\S+=(.*)\Z/.match "ASDF\nFOO=bar"
=> nil
irb(main):020:0> /\A\S+=(.*)\Z/.match "FOO=bar"
=> #<MatchData "FOO=bar" 1:"bar">
```

kke:
```
^ Matches the beginning of a line.

$ Matches the end of a line.

\A Matches the beginning of the string.

\z Matches the end of the string.

\Z Matches the end of the string unless the string ends with a "\n", in which case it matches just before the "\n".
```

So `\Z` would strip the trailing linefeed from the value.

SpComb: OTOH there's no reason to not accept multi-line env values, so I suggest `/\A\S+=/`

```yaml
stack: test/test
services:
  foo:
    environment:
      - |
          FOO=
          ASDF1
          ASDF2
```

```
            "Env": [
                "FOO=\nASDF1\nASDF2",
            ],
```

```
$ echo $FOO
ASDF1 ASDF2
```

SpComb:
The current `/\A[^=]+=/` just leaves the slightly crazy case of env vars with whitespace and newlines in them... but those seem to work, so I don't see why we should fail them in validation? :)

```yaml
stack: test/test
services:
  foo:
    environment:
      - |
          FOO
          BAR
          =
          ASDF1
          ASDF2
```

```
            "Env": [
                "FOO\nBAR\n=\nASDF1\nASDF2",
            ],
```

```
>>> import os
>>> os.environ
{..., 'FOO\nBAR\n': '\nASDF1\nASDF2', ...}
>>> os.environ['FOO\nBAR\n']
'\nASDF1\nASDF2'
```

```
irb(main):001:0> ENV
=> {..., "FOO
BAR
"=>"\nASDF1\nASDF2", ...}
irb(main):004:0> ENV["FOO\nBAR\n"]
=> "\nASDF1\nASDF2"
```

PR #2316, opened 2017-05-15
Update cli Dockerfile to use latest docker version

jakubknejzlik:


cli/Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM alpine:3.5
 MAINTAINER Kontena, Inc. <info@kontena.io>
 
-ADD https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 /usr/local/bin/docker
+ADD https://get.docker.com/builds/Linux/x86_64/docker-latest.tgz /tmp/docker-latest.tgz
+RUN tar --strip-components=1 -xvzf /tmp/docker-latest.tgz -C /usr/local/bin
+RUN rm /tmp/docker-latest.tgz
```
SpComb: This still leaves the `/tmp/docker-latest.tgz` in the image, because each command is a separate layer. Use something like this instead?

```
RUN curl -sL https://get.docker.com/builds/Linux/x86_64/docker-latest.tgz | tar -C /usr/local/bin --strip-components=1 -xzv
```

SpComb:
The problem with embedding the latest version of Docker into the CLI image is that it will also use the latest Docker API version by default, and thus fail when run with any older version of the Docker engine on the host machine.

AFAIK the version of Docker embedded in the image thus forms a minimum version requirement on the version of Docker that is running on the host machine. Shipping Docker 1.10  in the `kontena/cli` image should work with any version of Docker >= 1.10 running on the host machine.

jakubknejzlik:
I see...actually I was not sure if keeping the old version is intentional and this was actually kind of proposal/question :).

Anyway when using this image in our CI the docker login email argument is little bit annoying. We are using single image to building, pushing and deploying the image. Fortunately with gitlab-ci we can split the build phase (docker build/login/push) into separate step with different image (although it will not be possible to use `kontena stack build`).

SpComb:
> Anyway when using this image in our CI the docker login email argument is little bit annoying (in latest version it's deprecated)

Can you explain this a bit more? I don't understand what you mean. The `docker run kontena/cli kontena stack build` is somehow misbehaving with the Docker 1.10 CLI, but it's fixed in the latest version of Docker? What is it that has changed?

jakubknejzlik:
`kontena stack build` should be fine. The issue is with `docker login` and it's email attribute. In version 1.10 it's required (and command fails with weird error btw [info here](https://gitlab.com/gitlab-org/gitlab-ce/issues/23478)). In latest version the command warns about deprecation of email attribute:

```
docker login -e test@example.com
Flag --email has been deprecated, will be removed in 1.14.
```

SpComb:
Closing because the `kontena/cli` Docker image needs to be pinned to some specific Docker version for compatibility reasons, and not pull in a different version when rebuilt with/without `--no-cache`.

We can consider upgrading it from 1.10.3 to some newer version if needed to fix issues, though... a GitHub issue explaining how the `docker login --email` is broken in the current version, and what Docker version it is fixed in would be helpful.

PR #2325, opened 2017-05-16
Fix server mutations validations regexps

### LINKED ISSUE #2323: Server Volume::Create validation accepts invalid scope
The mutation valdiates the `scope` parameter using a regexp, but it does not anchor it, so the scope can contain arbitrary prefixes/suffixes:

```ruby
      string :scope, matches: /instance|stack|grid/
```

This will probably fail in some ugly way when actually creating volume instances.

This should use the `string :scope, in: ['instance', 'stack', 'grid']` validator instead.

#### `kontena volume create --driver local --scope foo-instance-bar test-invalid`
```
[EXPERIMENTAL] The `kontena volume` commands are still experimental in Kontena 1.3, and may change in future releases
 [done] Creating volume test-invalid      
```
#### `kontena volume ls`
```
NAME          SCOPE            DRIVER      CREATED AT              
test-invalid  foo-instance-bar local       2017-05-16T09:17:07.415Z
```

### LINKED ISSUE #2324: Server string validations do not reject newlines
The server validations that create new named objects (`GridServices::Create`, `Stacks::Create`, `Volumes::Create` etc)  typically use a regexp to validate the name:

```ruby
      string :name, matches: /^(?!-)(\w|-)+$/ # do not allow "-" as a first character
```

However, the [use of `^` as an anchor in a Ruby Regexp](https://ruby-doc.org/core-2.3.1/Regexp.html#class-Regexp-label-Anchors) does not match the beginning of the string, it matches the beginning of a line. This means that the name can contain a newline, and any arbitrary garbage, provided that least one "line" is a valid name. This allows names like:

* `kontena service create $'redis\ntest'`
* `kontena service create $'redis\n{}test--<>' redis`
* `kontena service create $'redis\n☃' redis`

#### `DEBUG=api kontena service create $'redis\ntest' redis`
```
[API Client Request]: POST http://localhost:9292/v1/grids/development/services | Headers: {Accept: application/json, Content-Type: application/json, Authorization: Bearer}  | Body: 
{"name":"redis\ntest","image":"redis:latest","stateful":false,"secrets":[],"privileged":false,"log_opts":{}}
[API Client Response]: Headers: {Content-Type: application/json}  | Status: 201 | Body: 
{"id":"development/null/redis\ntest","created_at":"2017-05-16T09:21:59.031Z","updated_at":"2017-05-16T09:21:59.031Z","image":"redis:latest","affinity":[],"name":"redis\ntest","stateful":false,"user":null,"instances":1,"cmd":null,"entrypoint":null,"net":"bridge","dns":"redis\ntest.development.kontena.local","ports":[],"env":[],"secrets":[],"memory":null,"memory_swap":null,"cpu_shares":null,"volumes":[],"volumes_from":[],"cap_add":[],"cap_drop":[],"state":"initialized","grid":{"id":"development"},"stack":{"id":"development/null","name":"null"},"links":[],"log_driver":null,"log_opts":{},"strategy":"ha","deploy_opts":{"_id":{"$oid":"591ac4b7de357847e2000002"},"interval":null,"min_health":0.8,"wait_for_port":null},"pid":null,"instance_counts":{"total":0,"running":0},"hooks":[],"revision":1,"stack_revision":null}
 [done] Creating redis
test service      
```

#### `kontena service ls`
```
⊝ redis
test                                                 0 / 1      no       initialized                                                   
```

SpComb:
Fixes #2323
Fixes #2324 for a large number of cases

The server mutations validation regexps would accept invalid inputs, because they were either missing anchors, or using the `^` and `$` anchors. Using `/^...$/` anchors in validation `matches` regexps is generally a bug, since ruby Regexp use a multi-line mode where such a regexp will match invalid multi-line inputs. Change the validations to use either `string :foo, in: ['...']`, or `string :foo, matches: /\A...\z/`.


server/app/mutations/volumes/create.rb
```diff
@@ -3,8 +3,8 @@ class Create < Mutations::Command
 
     required do
       model :grid, class: Grid
-      string :scope, matches: /instance|stack|grid/
```
SpComb: I think this is the only case of #2323, I tried to look for others?

SpComb:
The ideal case would be to fix the `Mutations::StringFilter` `matches` check to always do an fully-anchored match. I can't think of any case where you would want to use `string :foo, matches: /.../` without anchoring the regexp to match the entire string.

SpComb:
There's still a couple uses of `/^` left in the server source, might also be worth converting those to `/\A`:

```
app/models/host_node.rb:        if match = label.match(/^region=(.+)/)
app/models/grid_service.rb:        {:volumes_from => {:$regex => /^#{self.name}-%s/}},
app/models/grid_service.rb:        {:volumes_from => {:$regex => /^#{self.name}-\d+/}},
app/models/grid_service.rb:        {:net => {:$regex => /^container:#{self.name}-%s/}},
app/models/grid_service.rb:        {:net => {:$regex => /^container:#{self.name}-\d+/}}
app/models/grid_service.rb:      if self.affinity.any?{|a| a.match(/^service(!=|==).+/)}
app/models/grid_service.rb:      if self.affinity.any?{|a| a.match(/^container(!=|==).+/)}
app/models/grid_service.rb:    return true if self.net.to_s.match(/^container:.+/)
app/routes/v1/config_api.rb:              regex = /^#{Regexp.escape(params['filter'].gsub(/\*/, ''))}/
app/services/scheduler/filter/affinity.rb:        if match = affinity.match(/^(.+)(==|!=)(.+)/)
app/services/scheduler/filter/dependency.rb:        !service.net.to_s.match(/^container:.+/).nil?
```

No uses of `$/` left.

server/app/models/host_node.rb
```diff
@@ -110,19 +110,6 @@ def schedule_counter
...
-      @region = 'default'.freeze
-      self.labels.to_a.each do |label|
-        if match = label.match(/^region=(.+)/)
-          @region = match[1]
-        end
-      end
-    end
-    @region
-  end
-
```
SpComb: Leftover from #2084

SpComb: Split to #2326 because there were also other cleanups

PR #2326, opened 2017-05-16
Cleanup leftover HostNode methods

SpComb:
Leftovers from #1849, #2084 and #2141

### DIFF
--- a/server/app/models/host_node.rb
+++ b/server/app/models/host_node.rb
@@ -110,19 +110,6 @@ class HostNode
     @schedule_counter ||= 0
   end
 
-  # @return [String]
-  def region
-    if @region.nil?
-      @region = 'default'.freeze
-      self.labels.to_a.each do |label|
-        if match = label.match(/^region=(.+)/)
-          @region = match[1]
-        end
-      end
-    end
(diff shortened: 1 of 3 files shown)


server/app/models/stack.rb
```diff
@@ -45,8 +45,4 @@ def exposed_service?(grid_service)
     return false unless latest_rev
     latest_rev.expose.to_s == grid_service.name
   end
-
-  def self.default_stack
-    where(name: NULL_STACK).first
-  end
```
SpComb: Added by #1849 but never used. Also broken, because it ignores the `grid`.

server/app/models/host_node.rb
```diff
@@ -105,24 +105,6 @@ def rpc_client(timeout = 300)
     RpcClient.new(self.node_id, timeout)
   end
 
-  # @return [Integer]
-  def schedule_counter
-    @schedule_counter ||= 0
-  end
```
SpComb: Specs were failing because some of them were still passing `HostNode` objects to the scheduler, not `Scheduler::Node`s. They now failed after removing this method.

PR #2329, opened 2017-05-17
Fix master WWW-Authenticate header

### LINKED ISSUE #2267: Master WWW-Authenticate header is invalid
```console
$ curl -i -XGET http://192.168.66.100:8080/v1/grids
HTTP/1.1 403 Forbidden
Content-Type: application/json
WWW-Authenticate: Bearer realm="kontena_master",
WWW-Authenticate:     error="invalid_token",
WWW-Authenticate:     error_description="Invalid token"
Content-Length: 61

{"error":"invalid_token","error_description":"Invalid token"}%
```

The token_authentication.rb Rack middleware does this:

```ruby
header{ 'WWW-Authenticate' => message_parts.join(",\n    ") }
```

Which apparently gets multiplied by Rack instead of sending a proper multiline header.

Obviously something like `response.headers['WWW-Authenticate']` will usually only contain the last line when handling the response.

(actually it's not invalid, just unusual. the https://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-21#page-22 says it's allowed but clients do not always handle that).

kke:
Fixes #2267

Server was splitting the WWW-Authenticate header to multiple lines, which caused rack to send multiple WWW-Authenticate headers, making it kind of invalid or at least hard to parse.

### DIFF
--- a/server/app/middlewares/token_authentication.rb
+++ b/server/app/middlewares/token_authentication.rb
@@ -120,7 +120,7 @@ class TokenAuthentication
     message_parts = ['Bearer realm="kontena_master"']
     message_parts << "error=\"#{error}\"" if error
     message_parts << "error_description=\"#{error_description}\"" if error_description
-    { 'WWW-Authenticate' => message_parts.join(",\n    ") }
+    { 'WWW-Authenticate' => message_parts.join(", ") }
   end
 
   def authenticate_header(error = nil, error_description = nil)


jnummelin:
Copying comment from the linked issue, more appropriate to have the conversation in the PR.

From RFC2617, https://tools.ietf.org/html/rfc2617#section-1.2

> Note: User agents will need to take special care in parsing the WWW-
Authenticate or Proxy-Authenticate header field value if it contains
more than one challenge, or if more than one WWW-Authenticate header
field is provided, since the contents of a challenge may itself
contain a comma-separated list of authentication parameters.

So I think it's been perfectly valid even before. :)

Why do we want/need to put the error details in the headers anyway?

kke:
Because the oauth spec says it's supposed to be there: https://tools.ietf.org/html/rfc6750#page-7

Anyway, the current header in fact is not valid since it has both multiple occurences and commas:

```
> puts c.get(path: "/v1/grids").headers['WWW-Authenticate']
Bearer realm="kontena_master",, error="invalid_token",, error_description="Invalid token"
```

jnummelin:
Ah, didn't know oauth mandated that. E_TOOMANYRFCS :D

PR #2331, opened 2017-05-17
Implement remote cloud login

### LINKED ISSUE #1758: cli: kontena cloud login needs --remote option
Usecase: I want to use different browser what opens by default. By having --remote option I would be able to copy'n'paste the URL to my choice of browser.

nevalla:
This PR will enable remote cloud login. Remote Cloud login is trying to follow OAuth device flow as much as possible: https://tools.ietf.org/html/draft-ietf-oauth-device-flow-05 

Fixes #1758

### DIFF
--- a/cli/lib/kontena/cli/cloud/login_command.rb
+++ b/cli/lib/kontena/cli/cloud/login_command.rb
@@ -8,6 +8,7 @@ module Kontena::Cli::Cloud
     option ['-c', '--code'], '[CODE]', 'Use an authorization code'
     option ['-v', '--verbose'], :flag, 'Increase output verbosity'
     option ['-f', '--force'], :flag, 'Force reauthentication'
+    option ['-r', '--remote'], :flag, 'Remote login'
 
     def execute
       if self.code && self.force?
@@ -35,8 +36,11 @@ module Kontena::Cli::Cloud
           finish and return
         end
       end
-
-      web_flow
+      if remote?
+        remote_login
+      else
+        web_flow
+      end
       finish
     end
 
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -51,6 +55,37 @@ def finish
...
     end
 
+    def remote_login
+      client_id = kontena_account.client_id || Kontena::Client::CLIENT_ID
+      params = {
+        client_id: client_id
+      }
+      cloud_url = kontena_account.url
+      client = Kontena::Client.new(cloud_url, nil)
+      auth_request_response = client.post('/auth_requests', {}, params)
```
kke: i think it's supposed to be in the body (and content type `application/x-www-form-urlencoded`)

cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -51,6 +55,37 @@ def finish
...
+      puts
+
+      code_request_params = {
+        client_id: client_id,
+        device_code: auth_request_response['device_code']
+      }
+      code_response = nil
+      spinner "Waiting for authentication" do
+        until code_response do
+          code_response = client.post("/auth_requests/code", {}, code_request_params) rescue nil
```
kke: instead of rescuing with nil, you could add expectations like in master login:

```ruby
        response = client.request(
          http_method: :get,
          path: authentication_path(auth_params),
          expects: [501, 400, 302, 403],
          auth: false
        )

        if client.last_response.status == 302
          ...
        end
```

nevalla:
@SpComb valid for 1.2.3?

PR #2338, opened 2017-05-18
[testing testing] Only run travis tests for components that have changed

kke:


kke:
It seems to work now.

PR #2344, opened 2017-05-19
Support capturing STDOUT in Kontena.run

kke:
In addendum to #2209 

Captures command STDOUT output and returns it as an array of strings.

```ruby
result = Kontena.run(%w(node ls -q), capture: true)
result.each do |node_name|
  puts node_name
end
```


SpComb:
Kinda :-1: on this, all the downsides of bash pipelines without any of the benefits... things like "list each node" should be some helper module mixin, or we could start figuring out some kind of separate `kontena-api` module layer to use in the CLI.

Issue #2347, opened 2017-05-22
CLI: kontena master init-cloud broken in 1.3.0.dev


jnummelin:
```
$ kontena master init-cloud --force
 [error] ArgumentError : Can't load cloud/master/add_command or /kontena/cli/lib/kontena/cli/cloud/master/add_command
```

jnummelin:
fixed in #2350

PR #2350, opened 2017-05-22
Fixing subcommand loading bug when running init-cloud

daniel-bytes:
This fixes issue #2347 .  Unnecessary `require`-ing of commands.

### DIFF
--- a/cli/lib/kontena/cli/cloud/master/add_command.rb
+++ b/cli/lib/kontena/cli/cloud/master/add_command.rb
@@ -1,6 +1,3 @@
-require 'kontena/cli/cloud_command'
-require 'kontena/cli/cloud/master_command'
-
 module Kontena::Cli::Cloud::Master
   class AddCommand < Kontena::Command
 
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/cloud/master/add_command.rb
```diff
@@ -1,6 +1,3 @@
-require 'kontena/cli/cloud_command'
-require 'kontena/cli/cloud/master_command'
-
 module Kontena::Cli::Cloud::Master
```
kke: I suppose those requires were there to make sure there's `Kontena::Cli::Cloud::Master` module tree. It would usually be better to just nest them like `module Kontena; module Cli; module ........`.

cli/lib/kontena/cli/master/user_command.rb
```diff
@@ -1,5 +1,3 @@
-require 'kontena/cli/master/user_command'
```
kke: I suppose this was the culprit, seems like this file required itself.

PR #2342, opened 2017-05-19
Serialize all pubsub payloads to avoid issues with dotted fields

jnummelin:
fixes #2340 

Newer mongoid doesn't allow dotted fields in hashes. There are multiple RPC cases where the hash payload has dotted fields.

This PR adds generic payload serialization on pubsub messages to avoid any such issues.

PR #2353, opened 2017-05-22
Fixing key type in cloud websocket notifications

daniel-bytes:
Fixes issue where notification events coming from master nodes were not being published over websocket.

Since we are only publishing data that comes from mongo pubsub (whose data is serialized) I don't think we ever need to support symbols as keys.

### DIFF
--- a/server/app/services/cloud/websocket_client.rb
+++ b/server/app/services/cloud/websocket_client.rb
@@ -180,18 +180,20 @@ module Cloud
 
     # @param [Hash] msg
     def send_notification_message(msg)
-      invalidate_users_cache if msg[:type] == 'User' # clear cache if users are modified
+      invalidate_users_cache if msg['type'] == 'User' # clear cache if users are modified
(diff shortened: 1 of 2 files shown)


nevalla:
Messages are coming from [EventStream](https://github.com/kontena/kontena/blob/master/server/app/models/event_stream.rb) module where symbols are used as keys. Please could you verify that you are able to fetch values with string versions of keys or are further modifications needed?

SpComb:
Caused by #2342, where `MongoPubsub` now uses `MessagePack` to encode all pubsub payloads, converting hash symbol keys into string keys. Previously, symbols in pubsub payloads were preserved.

SpComb:
This would likely also be fixed by #2357

daniel-bytes:
#2357 is most likely a cleaner fix for this.  Should we can close this PR in favor of that one?

SpComb:
Should be fixed in #2357

PR #2363, opened 2017-05-23
Fix broken e2e link spec

jnummelin:
Got broken by #2058 which changed the error output format


jnummelin:
@SpComb PTAL, helps me captaining as I'm running e2e spec for almost all PRs I'm merging and now this gives a false negative :)

PR #2366, opened 2017-05-24
Load kontena-cli gemspec on start

kke:
This PR makes `lib/kontena_cli.rb` load `../kontena-cli.gemspec` and activate the spec.

This should prevent plugins from loading any earlier versions of CLI and prepending their paths to `$LOAD_PATH`. Especially when running in development from `bin/kontena`. Running the `gem install` version should already activate the CLI gem when loading.

Adding `puts Gem.loaded_specs['kontena-cli'].inspect` to some command's `execute` showed an earlier version of kontena-cli I have installed instead of the 1.3.0.dev I was running it from.

After the changes to `lib/kontena_cli.rb` in this PR, it showed the correct version.


cli/lib/kontena_cli.rb
```diff
@@ -1,3 +1,12 @@
+begin
+  unless Gem.loaded_specs.include?('kontena-cli')
+    load File.expand_path('../../kontena-cli.gemspec', __FILE__)
+    KontenaCliGemspec.activate
+    KontenaCliGemspec.activate_dependencies
+  end
+rescue LoadError
```
jnummelin: Why do we rescue if we don't do anything?

kke: Gems don't always include their own .gemspec file. We do by chance (because the file list is built using `git ls files` (which by the way adds a bunch of unnecessary files such as the omnibus dir, Dockerfile, Gemfile, etc and makes it impossible to `gem build` the cli unless you downloaded it using git))

Gem::Specification has `#to_ruby` which builds a .gemspec from scratch.

Gemspecs are installed to `$GEM_HOME/specifications/` generated using that `#to_ruby`, so it won't include any of the magic from the .gemspec in cli root.

Since loading it is not absolutely essential, escpecially not for a regular joe, it's better to just ignore the load error in case it happens.

SpComb:
I'm left wondering why we need to do this kind of low-level gemspec wrangling. Shouldn't this kind of `gem 'kontena-cli'` activation be standard procedure for all bin scripts corresponding to installed gems?

kke:
We are using a stanard bin wrapper:

```ruby
#!/usr/local/opt/ruby/bin/ruby
#
# This file was generated by RubyGems.
#
# The application 'kontena-cli' is installed as part of a gem, and
# this file is here to facilitate running it.
#

require 'rubygems'

version = ">= 0.a"

if ARGV.first
  str = ARGV.first
  str = str.dup.force_encoding("BINARY") if str.respond_to? :force_encoding
  if str =~ /\A_(.*)_\z/ and Gem::Version.correct?($1) then
    version = $1
    ARGV.shift
  end
end

if Gem.respond_to?(:activate_bin_path)
load Gem.activate_bin_path('kontena-cli', 'kontena', version)
else
gem "kontena-cli", version
load Gem.bin_path("kontena-cli", "kontena", version)
end
```

Documentation for activate_bin_path says:

> A side effect of this method is that it will activate the gem that contains the executable.

So, for installed kontena-cli this is no problemo, as stated in the description:

image

The problem this PR fixes should only occur when running `bin/kontena` from the `cli/bin` directory, because that does not activate the gem, it just does `require 'kontena_cli'` and does not even know that this thing is actually a gem.

SpComb:
I think we should maybe stop running `bin/kontena` directly instead of hacking in our own support for gem activation into the CLI itself.

kke:
> EDIT: for development, that means running bundle exec bin/kontena instead of just bin/kontena

Which is super difficult to combine with plugins.

SpComb:
IMO this PR is still a bad idea, we shouldn't need to do our own gem activation... OTOH it does mean that the only currently known way to dev with plugins is using `gem build kontena-cli.gemspec && gem install kontena-cli-*.dev.gem` to get the standard binwrapper.

PR #2370, opened 2017-05-26
Fix flaky vpn E2E specs

### LINKED ISSUE #2118: e2e specs vpn create flaky
No idea why, succeeds when run with `DEBUG_KOMMANDO=true`

```
  5) vpn create creates a vpn stack
     Failure/Error: expect(k.code).to eq(0)
     
       expected: 0
            got: 1
     
       (compared using ==)
     # ./spec/features/vpn/create_spec.rb:6:in `block (2 levels) in <top (required)>'
```

jnummelin:
E2E VPN specs were a bit flaky and the problem was seen on server logs:
```
StackDeployWorker: service e2e/vpn/server deploy failed: Cannot find applicable node for service instance e2e/vpn/server-1: Did not find any nodes with unused ports: [1194]
```

This is due to the "eventually consistent" nature of service/stack removal. Even though specs "cleaned" after running by doing `kontena stack rm --force vpn` it does not guarantee that actual VPN containers are really gone before the next spec starts.

Added generic helper to check container existence and to wait until container is gone.

Fixes #2118

### DIFF
--- a/test/spec/support/container_helper.rb
+++ b/test/spec/support/container_helper.rb
@@ -1,0 +1,19 @@
+module ContainerHelper
+
+  # Checks if a deployed container exists with given name
+  # @param [String] name of the container to check
+  def container_exist?(name)
+    k = run "kontena container ls"
+    k.out.include?(name)
+  end
+
+  # Waits until no container exists with the name
+  # @param [String] name of the container to check
+  def wait_until_container_gone(name)
+    loop do
+      return unless container_exist?(name)
+      sleep 1
+    end
+  end
+
+end
--- a/test/spec/support/container_helper.rb
+++ b/test/spec/support/container_helper.rb
@@ -3,7 +3,7 @@ module ContainerHelper
   # Checks if a deployed container exists with given name
   # @param [String] name of the container to check
   def container_exist?(name)
-    k = run "kontena container ls"
+    k = run "kontena container ls -q"
     fail "kontena container ls command failed: #{k.out}" if k.code != 0
     k.out.include?(name)
   end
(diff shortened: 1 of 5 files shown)


SpComb:
Fixes  #2118

test/spec/support/container_helper.rb
```diff
@@ -0,0 +1,19 @@
+module ContainerHelper
+
+  # Checks if a deployed container exists with given name
+  # @param [String] name of the container to check
+  def container_exist?(name)
+    k = run "kontena container ls"
+    k.out.include?(name)
```
SpComb: Should probably also check `k.code.zero?`?

jnummelin: Yes, would give false positives if `kontena container ls` errors. Maybe even just `fail` out if `k.code != 0`

test/spec/support/container_helper.rb
```diff
@@ -0,0 +1,20 @@
+module ContainerHelper
+
+  # Checks if a deployed container exists with given name
+  # @param [String] name of the container to check
+  def container_exist?(name)
+    k = run "kontena container ls"
```
SpComb: We just got `kontena container ls -q`, this would be a good use for it :)

Ideally this would also match on the `/vpn.server-1` suffix or something, to avoid false-positives with other similarly named containers, but that's not strictly necessary for this  PR.

jnummelin: good idea to use `-q`, pushed fix for that.

we know in this case at least that there's no similarly named container so should be safe.

PR #2377, opened 2017-05-29
Add tracking if service embedded docs changes

### LINKED ISSUE #2109: Stack upgrade / Service update will not re-deploy service on removal of embedded objects
Kontena 1.1 had an issue where every `kontena service update` or `kontena stack deploy` for a service with links to other services would consider the service as changed, and cause the service containers to be re-deployed. This happens because the embdded `GridServiceLink` documents were being re-created with a new mongo ObjectID on every update: #2090

However, fixing the `GridServices::Update` to keep the existing embedded objects unchanged in #2097 introduces a second bug, whereby removing embedded objects from the service will not cause mongoid to consider the service as changed, and will not touch the `updated_at` timestamp on `save`. This means that the service containers will not be re-deployed if multi-valued service fields stored as embedded mongo documents are removed. This affects all of the following service properties:

* hooks 
* links
* secrets
* volumes

The update will be reflected in `kontena service show`, but the container will still be running with the old hooks/secrets/volumes. Use `kontena service deploy --force` to re-deploy the container, and remove the hooks/secrets/volumes from the actual container configuration.

jnummelin:
Removing embedded models (secrets, links, hooks, volumes) from `GridService` didn't bump up the `revision` or `created_at`and thus did not deploy the changes without `--force` flag.

fixes #2109

### DIFF
--- a/server/app/mutations/grid_services/update.rb
+++ b/server/app/mutations/grid_services/update.rb
@@ -75,34 +75,34 @@ module GridServices
       attributes[:health_check] = self.health_check if self.health_check
       attributes[:volumes_from] = self.volumes_from if self.volumes_from
 
-      embeds_dirrty = false
+      embeds_changed = false
 
       if self.links
         attributes[:grid_service_links] = build_grid_service_links(
           self.grid_service.grid_service_links.to_a,
           self.grid_service.grid, grid_service.stack, self.links
         )
-        embeds_dirrty = attributes[:grid_service_links] != self.grid_service.grid_service_links.to_a
+        embeds_changed ||= attributes[:grid_service_links] != self.grid_service.grid_service_links.to_a
       end
 
       if self.hooks
         attributes[:hooks] = self.build_grid_service_hooks(self.grid_service.hooks.to_a)
-        embeds_dirrty = attributes[:hooks] != self.grid_service.hooks.to_a
+        embeds_changed ||= attributes[:hooks] != self.grid_service.hooks.to_a
       end
 
       if self.secrets
         attributes[:secrets] = self.build_grid_service_secrets(self.grid_service.secrets.to_a)
-        embeds_dirrty = attributes[:secrets] != self.grid_service.secrets.to_a
+        embeds_changed ||= attributes[:secrets] != self.grid_service.secrets.to_a
       end
       if self.volumes
         attributes[:service_volumes] = self.build_service_volumes(self.grid_service.service_volumes.to_a,
           self.grid_service.grid, self.grid_service.stack
         )
-        embeds_dirrty = attributes[:service_volumes] != self.grid_service.service_volumes.to_a
+        embeds_changed ||= attributes[:service_volumes] != self.grid_service.service_volumes.to_a
       end
       grid_service.attributes = attributes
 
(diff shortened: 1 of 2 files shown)


server/app/mutations/grid_services/update.rb
```diff
@@ -75,28 +75,34 @@ def execute
       attributes[:health_check] = self.health_check if self.health_check
       attributes[:volumes_from] = self.volumes_from if self.volumes_from
 
+      embeds_dirrty = false
```
SpComb: `embeds_changed`?

server/app/mutations/grid_services/update.rb
```diff
@@ -75,28 +75,34 @@ def execute
...
+
       if self.links
         attributes[:grid_service_links] = build_grid_service_links(
           self.grid_service.grid_service_links.to_a,
           self.grid_service.grid, grid_service.stack, self.links
         )
+        embeds_dirrty = attributes[:grid_service_links] != self.grid_service.grid_service_links.to_a
```
SpComb: Needs to be something like `embeds_changed ||= ...`

server/app/mutations/grid_services/update.rb
```diff
@@ -75,28 +75,34 @@ def execute
...
           self.grid_service.grid, self.grid_service.stack
         )
+        embeds_dirrty = attributes[:service_volumes] != self.grid_service.service_volumes.to_a
       end
       grid_service.attributes = attributes
 
-      if grid_service.changed?
+      if grid_service.changed? || embeds_dirrty
         info "updating service #{grid_service.to_path} with changes: #{changed(grid_service)}"
         grid_service.revision += 1
```
SpComb: Yes, this should work, because the `revision` change should be enough to actually mark the model as `changed?`, so that the `updated_at` also gets changed.

server/app/mutations/grid_services/update.rb
jnummelin: oh, very true :) I was testing with only single type of changes :)

server/app/mutations/grid_services/update.rb
jnummelin: yep

server/app/mutations/grid_services/update.rb
```diff
@@ -75,28 +75,34 @@ def execute
...
         attributes[:grid_service_links] = build_grid_service_links(
           self.grid_service.grid_service_links.to_a,
           self.grid_service.grid, grid_service.stack, self.links
         )
+        embeds_changed ||= attributes[:grid_service_links] != self.grid_service.grid_service_links.to_a
       end
 
       if self.hooks
         attributes[:hooks] = self.build_grid_service_hooks(self.grid_service.hooks.to_a)
+        embeds_changed ||= attributes[:hooks] != self.grid_service.hooks.to_a
```
SpComb: This might run into some issues with the `hooks` input param being an unordered hash of hook types, so the order of the resulting hooks might change... but ATM there is only a single type of hook (`post_start`), so this is moot.

The `build_grid_service_hooks` needs fixing to `sort` the `hooks` keys to guarantee stable ordering across updates though, if we ever do add more hook types.

PR #2379, opened 2017-05-30
Revert any_key_to_continue implementation

### LINKED ISSUE #2378: Press any key to continue with timeout prompt stays running after keypress
image

Also if it times out while waiting for the auth to complete, it will raise and the auth will fail because cli exits.

Made an issue to tty-prompt : https://github.com/piotrmurach/tty-prompt/issues/56

kke:
Fixes #2378 by restoring the previous implementation while waiting for a fix to tty-prompt.


jnummelin:
Is there an upstream issue for this in tty-prompt?

PR #2388, opened 2017-05-31
Improve platform identification on the Kontena Cloud

nevalla:
This PR will send `uuid` field among other master information to Kontena Cloud to improve Platform identification. We can display this information on the Kontena Cloud and user can identify and verify Kontena Platform better.

### DIFF
--- a/server/app/services/auth_provider.rb
+++ b/server/app/services/auth_provider.rb
@@ -98,7 +98,7 @@ class AuthProvider
 
     body = {
       data: {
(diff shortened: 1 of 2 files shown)


jnummelin:
> Doesn't this deanonymize the anonymous stats?

Not really, this PR does not touch the telemetry data, it will remain anonymous.

And there's always the option of disabling the telemetry data sending if that is an issue for a user.

PR #2394, opened 2017-06-01
Fix master join command option forwarding to master login

jnummelin:
`kontena master join ...` command "forwarded" some of the options and their values as single string in the command array. That broke down `Kontena.run` since when the commands are executed through that Clamp expects to get options and their values as separate array items.

```
$ k master join https://somemaster code

ERROR: Unrecognised option '--join code'

```

This PR fixes that by putting the options and their values as separate items in the command array.

### DIFF
--- a/cli/lib/kontena/cli/master/join_command.rb
+++ b/cli/lib/kontena/cli/master/join_command.rb
@@ -9,7 +9,7 @@ module Kontena::Cli::Master
 
     def execute
       params = []
-      params += ["--join", self.invite_code.shellescape]
+      params += ["--join", self.invite_code]
       params << "--remote" if self.remote?
       params += ["--name", self.name.shellescape] if self.name
       params << "--verbose" if self.verbose?
--- a/cli/lib/kontena/cli/master/join_command.rb
+++ b/cli/lib/kontena/cli/master/join_command.rb
@@ -11,7 +11,7 @@ module Kontena::Cli::Master
       params = []
       params += ["--join", self.invite_code]
       params << "--remote" if self.remote?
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/master/join_command.rb
```diff
@@ -9,9 +9,9 @@ class JoinCommand < Kontena::Command
 
     def execute
       params = []
-      params << "--join #{self.invite_code.shellescape}"
+      params += ["--join", self.invite_code.shellescape]
```
kke: No need to shellescape anymore with the array syntax.

(it was needed to add quotes in case a string had spaces)

SpComb: I doubt the shellescape helps here, because there's no shell there to parse them in `Kontena.run`? Only matters if the invite_code/name contains shell metacharacters, I suspect those will get mangled and broken ATM.

(could also be `--join=#{...}`)

jnummelin:
@kke Removed the unnecessary `shellescape`, PTAL

cli/lib/kontena/cli/master/join_command.rb
```diff
@@ -9,9 +9,9 @@ class JoinCommand < Kontena::Command
 
     def execute
       params = []
-      params << "--join #{self.invite_code.shellescape}"
+      params += ["--join", self.invite_code]
       params << "--remote" if self.remote?
-      params << "--name #{self.name.shellescape}" if self.name
+      params += ["--name", self.name.shellescape] if self.name
```
kke: this  does not need shellescape either

PR #2397, opened 2017-06-01
Finetune header formatting on list commands

jnummelin:
The change to `tty-table` usage for most `kontena xyz ls` commands also changed the table headers to be always blue in colour. Quite many are using black backgrounds on their terminals, including Kontena cloud, and blue-on-black is just not too readable.

This PR "reverts" the formatting closer to original upcase'd format, only by bolding the headers. As there's no colours used, the font colour is selected automatically based on the terminal.

Also the health icons were moved into their own column. Now when slight padding is added, the icons drifted a bit so I moved them to the same column with the service/node/stack names.

After the changes applied the outputs look like:
image

image

On black(ish) background the bolded white table headers seems to slightly blur, but is it too bad? IMO they are still quite readable.

### DIFF
--- a/cli/lib/kontena/cli/nodes/list_command.rb
+++ b/cli/lib/kontena/cli/nodes/list_command.rb
@@ -34,7 +34,6 @@ module Kontena::Cli::Nodes
     def fields
       return ['name'] if quiet?
       {
-        ' ' =>   'health_icon',
         name:    'name',
         version: 'agent_version',
         status:  'status',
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -67,7 +66,7 @@ def node_data
         unless quiet?
           grid_health = grid_health(grid, grid_nodes)
           grid_nodes.each do |node|
-            node['health_icon'] = health_icon(node_health(node, grid_health))
+            node['name'] = health_icon(node_health(node, grid_health)) + " " + node['name']
```
kke: The "blank" health icon column was there to offset the headers so that it looks like:
```
  FOOFOO FOO
* foofoo
? foofoo
```

instead of

```
FOOFOO FOO
* foofoo
? foofoo
```

PR #2407, opened 2017-06-02
Add --use-defaults option to kontena stack upgrade command

### LINKED ISSUE #2406: Add a switch to stack upgrade to not prompt values that have defaults
Perhaps `--use-defaults`.

kke:
With this PR you don't have to hit enter if you want to use the variable values from master metadata.

Currently they are used as defaults and prompted, but with `--use-defaults` they will be used as the final values instead.


kke:
Outdated

PR #2409, opened 2017-06-02
Add --values option to kontena stack show command

### LINKED ISSUE #2408: Add a switch to stack show to dump variable values
`kontena stack show --variables foofoo` would dump a YAML to stdout that can be used as input for `kontena stack install --values-from`.

kke:
Fixes #2408

This PR adds `--values` and `--values-to` options to `kontena stack show` command.

Variable names and values are also displayed in normal `stack show` output.

```
$ kontena stack show --values somestack > somestack-vars.yml
$ kontena stack show --values-to foo.yml somestack
```

```
$ kontena stack show twemproxy-redis1
```

image

### DIFF
--- a/cli/lib/kontena/cli/stacks/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/show_command.rb
@@ -13,7 +13,7 @@ module Kontena::Cli::Stacks
     requires_current_master
     requires_current_master_token
 
-    option '--variables', :flag, 'Output the variable-value pairs as YAML'
+    option '--values', :flag, 'Output the variable-value pairs as YAML'
 
     def execute
       variables? ? show_variables : show_stack
--- a/cli/lib/kontena/cli/stacks/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/show_command.rb
@@ -13,16 +13,24 @@ module Kontena::Cli::Stacks
     requires_current_master
     requires_current_master_token
 
+    option '--variables', :flag, 'Output the variable-value pairs as YAML'
+
     def execute
-      show_stack(name)
+      variables? ? show_variables : show_stack
     end
 
-    def fetch_stack(name)
+    def fetch_stack
       client.get("stacks/#{current_grid}/#{name}")
     end
 
-    def show_stack(name)
-      stack = fetch_stack(name)
+    def show_variables
+      require 'yaml'
+      stack = fetch_stack
+      puts stack['variables'].to_yaml
+    end
+
+    def show_stack
+      stack = fetch_stack
 
       puts "#{stack['name']}:"
       puts "  created: #{stack['created_at']}"
--- a/cli/lib/kontena/cli/stacks/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/show_command.rb
@@ -19,18 +19,18 @@ module Kontena::Cli::Stacks
       variables? ? show_variables : show_stack
     end
 
-    def fetch_stack
+    def fetch_stack(name)
       client.get("stacks/#{current_grid}/#{name}")
     end
 
     def show_variables
       require 'yaml'
-      stack = fetch_stack
+      stack = fetch_stack(name)
       puts stack['variables'].to_yaml
     end
 
     def show_stack
-      stack = fetch_stack
+      stack = fetch_stack(name)
 
       puts "#{stack['name']}:"
       puts "  created: #{stack['created_at']}"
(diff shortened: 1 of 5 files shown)


kke:
breaks service instance  resolver, fixing.

jakolehm:
Maybe `--variables` should be `--values-to`, just to match the `--variables-from` option?

kke:
Rewrite after #2707

kke:
Showing variables in normal `stack show` would show stuff from vault.

SpComb:
> Showing variables in normal stack show would show stuff from vault.

I think those are filtered out from the `variables` hash stored on the server?

I don't mean parsing the stack YAML.

cli/lib/kontena/cli/stacks/show_command.rb
```diff
@@ -13,16 +14,36 @@ class ShowCommand < Kontena::Command
...
+    end
+
+    def show_variables
+      puts variable_yaml
     end
 
-    def fetch_stack(name)
-      client.get("stacks/#{current_grid}/#{name}")
+    def clean_variables
+      rejected_keys = Set.new(%w(GRID STACK PARENT_STACK PLATFORM))
```
SpComb: I don't think this is necessarily needed, these are already filtered out of the [`result['variables'] = variable_values(without_defaults: true, without_vault: true)`](https://github.com/kontena/kontena/blob/3435b98ee9cb2a7261af813e97415b40796d66b8/cli/lib/kontena/cli/stacks/yaml/reader.rb#L215) that gets stored on the server?

`DEBUG=api bundle exec bin/kontena stack show --values redis`
```json
{
   "variables" : {
      "version" : "3.0"
   },
}
```
```

Comment (kke) [reply cli/lib/kontena/cli/stacks/show_command.rb]: I had `PARENT_STACK` in there, maybe that doesn't get cleaned

Comment (SpComb) [reply cli/lib/kontena/cli/stacks/show_command.rb]: Right, it looks like the `PARENT_STACK` is still getting stored on the server for child stacks:

```json
   "variables" : {
      "peers" : null,
      "PARENT_STACK" : "deptest"
   },
```

IMO better to fix the `result['variables'] = variable_values(...)` to also filter that out, rather than having to filter it out everywhere else that the variables are used... that also includes:

* https://github.com/kontena/kontena/blob/050b80b74da613d3f640423e52b7dd0ed0489642/cli/lib/kontena/cli/stacks/install_command.rb#L49
* https://github.com/kontena/kontena/blob/050b80b74da613d3f640423e52b7dd0ed0489642/cli/lib/kontena/cli/stacks/upgrade_command.rb#L143

kke: Now filtered in reader.rb

SpComb: Does that mean that this `clean_variables` method is no longer needed, and can be removed?

kke: I think it can stay, if the stack has been installed before this release it will have those variables.

SpComb: You mean the `GRID`/`STACK` variables? I think those were already filtered out before sending to the server in 1.3? 

https://github.com/kontena/kontena/blob/3588a33dfc992a0fb29603002ac33bfbe047c673/cli/lib/kontena/cli/stacks/yaml/reader.rb#L146-L148

cli/lib/kontena/cli/stacks/show_command.rb
```diff
@@ -32,6 +53,10 @@ def show_stack(name)
       puts "  version: #{stack['version']}"
       puts "  revision: #{stack['revision']}"
       puts "  expose: #{stack['expose'] || '-'}"
+      puts "  variables:#{' -' if stack['variables'].nil? || stack['variables'].empty?}"
+      stack['variables'].each do |var, val|
```
SpComb: The "stacks" created by the `kontena registry`/`vpn` commands unfortunately seem to be return `{"variables": null}` from the API... that should maybe be fixed somehow, as it's breaking this `kontena stack show registry`:

```
NoMethodError: undefined method `each' for nil:NilClass
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/show_command.rb:57:in `show_stack'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/show_command.rb:21:in `execute'
```

PR #2420, opened 2017-06-05
Added requires for excon and json to plugin_manager

### LINKED ISSUE #2417: CLI: Plugin search broken
```
$ kontena plugin search
 [error] NoMethodError : undefined method `new' for Excon:Module
         See /Users/jussi/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception
```

### LINKED ISSUE #2418: CLI: Plugin install broken
```
$ kontena plugin install packet
    *   Upgrading plugin packet .. \NoMethodError : undefined method `new' for Excon:Module
 [fail] Upgrading plugin packet     
 [done] Running cleanup    
```

kke:
Fixes #2417 
Fixes #2418 

However I think there's still room for improvement.

### DIFF
--- a/cli/lib/kontena/plugin_manager.rb
+++ b/cli/lib/kontena/plugin_manager.rb
@@ -2,6 +2,14 @@ require 'singleton'
 
 module Kontena
   class PluginManager
+    autoload :RubygemsClient, 'kontena/plugin_manager/rubygems_client'
+    Gem.autoload :DependencyInstaller, 'rubygems/dependency_installer'
+    Gem.autoload :Requirement, 'rubygems/requirement'
+    Gem.autoload :Uninstaller, 'rubygems/uninstaller'
+    Gem.autoload :Commands, 'rubygems/command'
+    Gem.autoload :DefaultUserInteraction, 'rubygems/user_interaction'
+    Gem.autoload :StreamUI, 'rubygems/user_interaction'
+    Gem::Commands.autoload :CleanupCommand, 'rubygems/commands/cleanup_command'
 
     include Singleton
 
@@ -22,9 +30,6 @@ module Kontena
     # @param pre [Boolean] install a prerelease version if available
     # @param version [String] install a specific version
     def install_plugin(plugin_name, pre: false, version: nil)
-      require 'rubygems/dependency_installer'
-      require 'rubygems/requirement'
-
       cmd = Gem::DependencyInstaller.new(
         document: false,
         force: true,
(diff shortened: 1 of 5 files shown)


jnummelin:
Any chance to have some level of sanity checking specs for these? I think some E2E specs would do the trick.

kke:
There's an e2e spec already that attempts a plugin install. Maybe it should test reinstall/upgrade too, I think this problem was only when installing an installed plugin again.

jnummelin:
And also the search was broken

jnummelin:
Why do we need the gem spec changes to fix this?

cli/lib/kontena/plugin_manager.rb
```diff
@@ -56,6 +56,8 @@ def uninstall_plugin(plugin_name)
     # Search rubygems for kontena plugins
     # @param pattern [String] optional search pattern
     def search_plugins(pattern = nil)
+      require 'excon'
+      require 'json'
```
jnummelin: Have to say I find it weird to have `require` on method level. This require juggling starts to be a slippery slope now if we start to mangle all requires on method level "just-in-case" and to optimise cli load time.

cli/lib/kontena_cli.rb
```diff
@@ -1,3 +1,12 @@
+begin
+  unless Gem.loaded_specs.include?('kontena-cli')
+    load File.expand_path('../../kontena-cli.gemspec', __FILE__)
+    KontenaCliGemspec.activate
+    KontenaCliGemspec.activate_dependencies
+  end
+rescue LoadError
+end
+
```
jnummelin: Why we need this?

kke: Dang, another branch leak, it's from another pr.

jnummelin: Your branches are trying to make some hostile takeover by sneaking in everywhere. Or is it Skynet? :D

cli/lib/kontena/plugin_manager.rb
kke: Another option is autoload. 

Plugin manager is a special case because it's loaded for every command so if it requires outside of methods it's global and those two are quite slow to load. 

Of course the root cause is that these methods should be in classes of their own.

cli/lib/kontena_cli.rb
kke: Probably missed some "you have unstaged changes and have to stash them" errors when checking out master.

PR #2421, opened 2017-06-05
Render service links with qualified name

### LINKED ISSUE #2416: Service unlink broken
```
$ kontena service show todo-app/web 
test/todo-app/web:
  created: 2017-06-02T09:18:56.304Z
  updated: 2017-06-02T12:32:22.773Z
  stack: test/todo-app
  desired_state: running
  image: jnummelin/todo-example:latest
  revision: 2
  stack_revision: 1
  stateful: no
  scaling: 3
  strategy: ha
  deploy_opts:
    min_health: 0.8
    wait_for_port: 9292
  dns: web.todo-app.test.kontena.local
  cmd: bundle exec puma -p 9292 -e production
  env: 
    - MONGODB_URI=mongodb://mongo-1:27017,mongo-2:27017,mongo-3:27017/todo_production
    - KONTENA_LB_VIRTUAL_HOSTS=todo-demo.kontena.io
    - KONTENA_LB_INTERNAL_PORT=9292
  links: 
    - infra/lb-external
    - mongo
    - redis
  instances:
        …
$ kontena service unlink todo-app/web redis
 [fail] Unlinking todo-app/web from redis      
    links:
    - Service todo-app/lb-external does not exist
```

jnummelin:
Both `kontena service link` and `kontena service unlink` commands were slightly broken.

The commands first get the service details (including links) and then send back modified links. The problem is that server sends the `name` attribute of the link as plain service name without stack prefix. Naturally CLI just posts those back as is for existing and un-touched links. This breaks things like:
```
$ kontena service show todo-app/web 
test/todo-app/web:
  created: 2017-06-02T09:18:56.304Z
  updated: 2017-06-02T12:32:22.773Z
  stack: test/todo-app
  desired_state: running
  image: jnummelin/todo-example:latest
  revision: 2
  stack_revision: 1
  stateful: no
  scaling: 3
  strategy: ha
  deploy_opts:
    min_health: 0.8
    wait_for_port: 9292
  dns: web.todo-app.test.kontena.local
  cmd: bundle exec puma -p 9292 -e production
  env: 
    - MONGODB_URI=mongodb://mongo-1:27017,mongo-2:27017,mongo-3:27017/todo_production
    - KONTENA_LB_VIRTUAL_HOSTS=todo-demo.kontena.io
    - KONTENA_LB_INTERNAL_PORT=9292
  links: 
    - infra/lb-external
    - mongo
    - redis
  instances:
        …
$ kontena service unlink todo-app/web redis
 [fail] Unlinking todo-app/web from redis      
    links:
    - Service todo-app/lb-external does not exist
```

As the `infra/lb-external` link is posted back with name `lb-external` server assumes we're trying to link to another service within the same stack.

This PR changes the server to send name with the stack prefix which is actually same as the `alias` field. Or are there cases where they would be different?

This does slightly alter the API, but IMO to better direction as we mention in the API docs that the name should be prefixed with stack name:
> name: The linked service. Within stack use <service_name>, otherwise use <stack_name>/<service_name>

fixes #2416

### DIFF
--- a/server/app/views/v1/grid_services/_grid_service.json.jbuilder
+++ b/server/app/views/v1/grid_services/_grid_service.json.jbuilder
@@ -34,7 +34,7 @@ json.stack do
   json.name grid_service.stack.name
 end
 json.links grid_service.grid_service_links.map { |s|
-  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.alias } if s.linked_grid_service
+  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.linked_grid_service.qualified_name } if s.linked_grid_service
 }.compact
 json.log_driver grid_service.log_driver
 json.log_opts grid_service.log_opts
--- a/server/app/views/v1/grid_services/_grid_service.json.jbuilder
+++ b/server/app/views/v1/grid_services/_grid_service.json.jbuilder
@@ -34,7 +34,7 @@ json.stack do
   json.name grid_service.stack.name
 end
 json.links grid_service.grid_service_links.map { |s|
-  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.linked_grid_service.name } if s.linked_grid_service
+  { id: s.linked_grid_service.to_path, alias: s.alias, name: s.alias } if s.linked_grid_service
 }.compact
 json.log_driver grid_service.log_driver
 json.log_opts grid_service.log_opts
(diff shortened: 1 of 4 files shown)


jnummelin:
Somehow using the `alias` also for `name` just feels wrong, although it seem to work in all cases I figured to test. :/ Another option is to use `GridService#qualified_name` which return the "stackified" name.

jakolehm:
I think it would be safer to use `GridService#qualified_name`.

PR #2427, opened 2017-06-05
Remove extra character in on :close event args

kke:
Some typo has sneaked into `services/exec_command`.

Line 110:

From:
```ruby
ws.on :close do |e|s
```

To:
```ruby
ws.on :close do |e|
```


jnummelin:
rebase/merge needed :(

jnummelin:
Fixed in #2432

PR #2429, opened 2017-06-05
Fix Authentication API debug message

### LINKED ISSUE #2428: Authentication API raises error on debug mode
Authentication API outputs message in debug mode when authentication request is successful. However in debug message `user` is referenced which is in most cases nil and thus causing error

nevalla:
Fixes #2428

### DIFF
--- a/server/app/routes/oauth2/authenticate_api.rb
+++ b/server/app/routes/oauth2/authenticate_api.rb
@@ -62,7 +62,8 @@ module OAuth2Api
           user = current_user
         end
 
-        debug { "User authentication request for #{user.email} accepted, redirecting to auth provider" }
+        email =  !user.nil? ? user.email : 'anonymous'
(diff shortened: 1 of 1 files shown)


server/app/routes/oauth2/authenticate_api.rb
```diff
@@ -62,7 +62,8 @@ class AuthenticateApi < Roda
           user = current_user
         end
 
-        debug { "User authentication request for #{user.email} accepted, redirecting to auth provider" }
+        email =  !user.nil? ? user.email : 'anonymous'
```
kke: User has username which should not be nil I think.

nevalla: the issue is that `user` is nil

kke: Right.

PR #2430, opened 2017-06-06
Inject variable names into liquid as VARIABLES

kke:
Now you can do something like:

```yaml
{{ VARIABLES | join: ',' }}
```


matti:
Currently needs to do this:
```
# {% assign secrets = "REDIS_URL,MONGO_URL" | split: "," %}
# {% for secret in secrets %}
      - secret: ${STACK}_{{secret}}
        name: {{secret}}
        type: env
# {% endfor %}
```
With this could do:
```
# {% for secret in VARIABLES | join: ',' %}
      - secret: ${STACK}_{{secret}}
        name: {{secret}}
        type: env
# {% endfor %}
```

kke:
It might not  actually work because I don't believe there's a `self.send` equivalent in liquid.

kke:
Looks like it won't work like this. Converted into issue #2431

Issue #1771, opened 2017-02-02
Set node labels from kontena-agent container labels


matti:
Currently setting node labels is cumbersome:

```
$ cat > /etc/docker/daemon.json <<DOCKERCONFIG
{
  "labels": ["region=<name_here>"],
  "insecure-registries": ["10.81.0.0/16"]
}
DOCKERCONFIG
```

or cloud-init...

Let's be awesome and take node labels from `kontena-agent` ! wohoo!

SpComb:
Would the `KONTENA_NODE_LABELS` env in #2590 also be acceptable?

matti:
@spcomb yes, that’s better

PR #2454, opened 2017-06-09
Docs (nodes): Fix dockerd labels without a value

matti:
seems like dockerd requires labels in --label=[key=value]


matti:
related https://github.com/kontena/kontena/issues/1771

matti:
Isnt the problem with kontena labels that can be just keys? As dockerd doesnt allow that kind of labels (anymore?)

PR #2455, opened 2017-06-09
Allow server port to be set from PORT env

matti:
This allows the Kontena master to be run in environments like Heroku.

### DIFF
--- a/docs/references/environment-variables.md
+++ b/docs/references/environment-variables.md
@@ -17,6 +17,7 @@ toc_order: 2
 - `AUTH_API_URL`: specifies authentication server url (default: https://auth.kontena.io)
 - `WEBSOCKET_TIMEOUT`: websocket timeout in seconds (default: 5.0)
 - `KONTENA_CLOCK_SKEW`: acceptable clock skew for connecting agents (default: 1.0s)
+- `PORT`: server port (default: 9292)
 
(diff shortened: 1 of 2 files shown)


SpComb:
Seems a bit odd and likely to be heroku-specific. Very weird that they require dynamic ports even for Docker images.

Does the existing [`server/Procfile`](https://github.com/kontena/kontena/blob/master/server/Procfile) (which does use `$PORT`) not work on heroku?

matti:
It works, but not for docker deployment. deploying to heroku without docker is so 2016z

SpComb:
I assume this is the only change needed for the custom `mattipaksula/kontena-server` image used there?

matti:
@SpComb yes, without this change Heroku will kill the dyno:

```
2017-06-12T13:34:15.395614+00:00 app[web.1]: I, [2017-06-12T13:34:15.395302 #8]  INFO -- LeaderElectorJob: won election ♚
2017-06-12T13:34:15.555326+00:00 app[web.1]: I, [2017-06-12T13:34:15.555225 #8]  INFO -- Watchdog<WebsocketBackend>: watchdog start
2017-06-12T13:34:15.558057+00:00 app[web.1]: [5] - Worker 0 (pid: 8) booted, phase: 0
2017-06-12T13:35:09.684526+00:00 heroku[web.1]: Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch
2017-06-12T13:35:09.684526+00:00 heroku[web.1]: Stopping process with SIGKILL
2017-06-12T13:35:09.864457+00:00 heroku[web.1]: Process exited with status 137
2017-06-12T13:35:09.871075+00:00 heroku[web.1]: State changed from starting to crashed
```

PR #2456, opened 2017-06-11
Upgrade Alpine to 3.6

jakolehm:
Also upgrades Ruby from 2.3.3 to 2.4.1.

https://alpinelinux.org/posts/Alpine-3.6.0-released.html
https://blog.heroku.com/ruby-2-4-features-hashes-integers-rounding


SpComb:
Does travis also need changes to run the tests/builds with ruby 2.4?

SpComb:
This only upgrades the server. Should we also upgrade the agent to alpine 3.5 / ruby 2.4?

SpComb:
The upgrade to ruby 2.4 seems to change the behavior of `OpenSSL::Cipher#key=` as used by `SymmetricEncryption`: https://github.com/ruby/openssl/issues/116

Previous versions seem to have truncated the configured `VAULT_KEY` string to the key length required by the cipher, the new version raises `ArgumentError` instead if the key is the wrong length.

```
api_1      | ArgumentError: key must be 32 bytes
api_1      | 	/usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/cipher.rb:526:in `key='
api_1      | 	/usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/cipher.rb:526:in `binary_decrypt'
api_1      | 	/usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/symmetric_encryption.rb:121:in `decrypt'
api_1      | 	/app/app/models/configuration.rb:37:in `decrypt'
api_1      | 	/app/app/models/configuration.rb:41:in `block in decrypt_all'
```

The `aes-256-cbc` cipher that we use for `SymmetricEncryption` requires a key of exactly 32 bytes (256-bit). I suppose we need to truncate/pad the `VAULT_KEY` ourselves to match the existing pre-upgrade behavior...

SpComb:
> The aes-256-cbc cipher that we use for SymmetricEncryption requires a key of exactly 32 bytes (256-bit). I suppose we need to truncate/pad the VAULT_KEY ourselves to match the existing pre-upgrade behavior...

It's a bad idea in general to take some kind of arbitrary passphrase string as input and use it as an AES key, I would have expected a library like `SymmetricEncryption` to process the key through a PBKDF first. Unfortunately we can't change how the `VAULT_KEY` is configured without breaking decryption of all existing vault secrets :/

SpComb:
> The upgrade to ruby 2.4 seems to change the behavior of OpenSSL::Cipher#key= as used by SymmetricEncryption: ruby/openssl#116

This is mentioned in the [ruby 2.4.1 embedded `ext/openssl` release notes](https://github.com/ruby/ruby/blob/v2_4_1/ext/openssl/History.md#notable-changes) as a fixed issue: https://bugs.ruby-lang.org/issues/12561

> OpenSSL::Cipher#key= and #iv= reject too long inputs. They used to truncate silently. [Bug #12561]

SpComb:
PoC patch to fix this regression by doing the same kind of truncation as I assume used to happen in ruby 2.3, per https://github.com/ruby/ruby/commit/ce635262f53b760284d56bb1027baebaaec175d1

```patch
--- a/server/app/initializers/01_symmetric_encryption.rb
+++ b/server/app/initializers/01_symmetric_encryption.rb
@@ -3,7 +3,7 @@ require 'symmetric_encryption/extensions/mongoid/encrypted'
 
 if ENV['VAULT_KEY'] && ENV['VAULT_IV']
   SymmetricEncryption.cipher = SymmetricEncryption::Cipher.new(
-      key:         ENV['VAULT_KEY'],
+      key:         ENV['VAULT_KEY'][0..31],
       iv:          ENV['VAULT_IV'],
       cipher_name: 'aes-256-cbc'
   )
```

This fix is required to maintain compatibility with existing `VAULT_KEY` configuration when decrypting secrets...

If the same vault key is also used for some kind of HMAC signature verification, then this truncation might also break things... unclear if the `SymmetricEncryption` gem does that, but e.g. rails `ActiveSupport::MessageEncryptor` requires supplying separate truncated encrpytion/decryption keys and un-truncated signature verification keys for the same issue: https://github.com/rails/rails/issues/28401

SpComb:
Some deprecation warnings:

```
api_1      | /app/app/services/rpc/fixnum_helper.rb:18: warning: constant ::Bignum is deprecated
api_1      | /app/app/services/rpc/fixnum_helper.rb:20: warning: constant ::Bignum is deprecated
```

SpComb:
This also explodes on the `VAULT_IV`, which seems to be 128-bit:

```
api_1      | I, [2017-06-12T11:44:50.458140 #347]  INFO -- Mongodb::Migrator: migrating CreateInitialAdmin
api_1      | iv must be 16 bytes
api_1      | iv must be 16 bytes
api_1      | /usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/cipher.rb:479:in `iv='
api_1      | /usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/cipher.rb:479:in `binary_encrypt'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/cipher.rb:235:in `encrypt'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/symmetric-encryption-3.8.2/lib/symmetric_encryption/symmetric_encryption.rb:181:in `encrypt'
api_1      | /app/app/helpers/digest_helper.rb:26:in `salt'
api_1      | /app/app/helpers/digest_helper.rb:32:in `digest'
api_1      | /app/app/helpers/digest_helper.rb:15:in `digest'
api_1      | /app/app/models/access_token.rb:62:in `block in <class:AccessToken>'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:448:in `instance_exec'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:448:in `block in make_lambda'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:164:in `block in halting'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:504:in `block in call'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:504:in `each'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:504:in `call'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:92:in `__run_callbacks__'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:778:in `_run_save_callbacks'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:81:in `run_callbacks'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/interceptable.rb:138:in `run_callbacks'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:116:in `prepare_insert'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:23:in `insert'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:180:in `block in create!'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/threaded/lifecycle.rb:161:in `_creating'
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:175:in `create!'
api_1      | /app/db/migrations/14_create_initial_admin.rb:9:in `up'
```

SpComb:
Also some deprecation warnings from `mutations` 0.7.2, need to also bump that to `0.8.1` with ruby 2.4 support? https://github.com/cypriss/mutations/blob/master/CHANGELOG.md#081 

```
api_1      | /usr/lib/ruby/gems/2.4.0/gems/mutations-0.7.2/lib/mutations/string_filter.rb:24: warning: constant ::Fixnum is deprecated
```

SpComb:
Needs the following patch to pass the e2e tests:

```
--- a/server/app/initializers/01_symmetric_encryption.rb
+++ b/server/app/initializers/01_symmetric_encryption.rb
@@ -3,8 +3,8 @@ require 'symmetric_encryption/extensions/mongoid/encrypted'
 
 if ENV['VAULT_KEY'] && ENV['VAULT_IV']
   SymmetricEncryption.cipher = SymmetricEncryption::Cipher.new(
-      key:         ENV['VAULT_KEY'],
-      iv:          ENV['VAULT_IV'],
+      key:         ENV['VAULT_KEY'][0...32],
+      iv:          ENV['VAULT_IV'][0...16],
       cipher_name: 'aes-256-cbc'
   )
 end
```

I suspect this would also explode with the 48-byte hex keys used by plugin provisioned masters.

jakolehm:
@SpComb just thinking if we need `Rpc::FixnumHelper` anymore (it has deprecated Bignum)?

SpComb:
Looks like travis is still running the server specs with ruby 2.3.1, it should also run them with ruby 2.4.1 to match the docker image.

SpComb:
> @SpComb just thinking if we need Rpc::FixnumHelper anymore (it has deprecated Bignum)?

Hmm... is that because mongoid/mongo is unable to store large integers? Not familiar with what the restrictions on that are... might still need something for that.

I suspect the `Rpc::FixnumHelper` is broken as-is... it looks like in ruby 2.4.1, `<integer>.is_a? Bignum` will always return true, so it will always convert all Integers to Float?

```
irb(main):012:0> (2 + 1).is_a? Bignum
(irb):12: warning: constant ::Bignum is deprecated
=> true
irb(main):013:0> (2**32 + 1).is_a? Bignum
(irb):13: warning: constant ::Bignum is deprecated
=> true
irb(main):014:0> (2**64 + 1).is_a? Bignum
(irb):14: warning: constant ::Bignum is deprecated
=> true
```

compared to ruby 2.3.1:

```
irb(main):008:0> (2**1 + 1).is_a? Bignum
=> false
irb(main):009:0> (2**32 + 1).is_a? Bignum
=> false
irb(main):010:0> (2**64 + 1).is_a? Bignum
=> true
```

SpComb:
Hmm... without the `fixnums_to_float`, how does mongo handle large integers in container and host-node stats?

SpComb:
Do the `HostNodeStat` and `ContainerStat` models need handle large unsigned 64-bit integers?

Testing with ruby 2.4.0:

```
irb(main):006:0> (2**63 - 1).to_bson
=> #<BSON::ByteBuffer:0x005593d9e073e8>
irb(main):007:0> (2**63).to_bson    
RangeError: 9223372036854775808 is not a valid 8 byte integer value.
	from /usr/lib/ruby/gems/2.4.0/gems/bson-4.2.1/lib/bson/integer.rb:172:in `out_of_range!'
	from /usr/lib/ruby/gems/2.4.0/gems/bson-4.2.1/lib/bson/integer.rb:112:in `to_bson'
	from (irb):7
	from /usr/bin/irb:11:in `<main>'
```

SpComb:
I'm tempted to merge this as-is, dropping the `Bignum` => `Float` conversion, and then re-deal with that later as a separate issue.

@daniel-bytes WDYT about the Ruby 2.4 Integer change effects on the `HostNodeStat` / `ContainerStat`? Have you seen large integers >= 2^63 that currently rely on the `Bignum` => `Float`  conversion?

SpComb:
Filed #2520 to think about the unsigned 64-bit integers thing.

PR #2457, opened 2017-06-11
Upgrade rack & roda to latest version

jakolehm:


SpComb:
Re-ran e2e specs with fixes, and passes apart from  #2522

PR #2461, opened 2017-06-12
Increase short RpcClient timeouts from 2s to 5s

### LINKED ISSUE #2460: Increase Rpc default timeout
The default timeout used in server-agent RPC is 2secs. In cases where there are e.g. multiple deployments ongoing both agent and server has their hands full of work. This makes the RPC channel hit timeouts and thus may cause failed instance deploys.
```
Failed to deploy instance xyz-1 to node foo-node: RpcClient::TimeoutError: Connection timeout (2.0s)
```

jnummelin:
In some cases the 2 seconds timeout used is just too short. This PR increases the cases where 2secs timeout was used to use 5 secs timeout.

Fixes #2460

### DIFF
--- a/server/app/services/grid_service_instance_deployer.rb
+++ b/server/app/services/grid_service_instance_deployer.rb
@@ -133,7 +133,7 @@ class GridServiceInstanceDeployer
 
   # @param [HostNode] node
   # @raise [RpcClient::TimeoutError]
-  def notify_node(node, timeout: 2.0)
+  def notify_node(node, timeout: 5.0)
     rpc_client = RpcClient.new(node.node_id, timeout)
     rpc_client.request('/service_pods/notify_update', [])
   end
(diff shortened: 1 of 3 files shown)


server/app/services/grid_service_instance_deployer.rb
```diff
@@ -133,7 +133,7 @@ def get_service_instance
 
   # @param [HostNode] node
   # @raise [RpcClient::TimeoutError]
-  def notify_node(node, timeout: 2.0)
+  def notify_node(node, timeout: 5.0)
     rpc_client = RpcClient.new(node.node_id, timeout)
     rpc_client.request('/service_pods/notify_update', [])
```
SpComb: What if this was `rpc_client.notify` instead of `request`? No need for a timeout in that case, and the agent RPC handler doesn't return or raise anything useful.

SpComb:
Could the RPC timeouts be a bit saner if we changed the default `RpcClient#initialize` timeout to something like 10s, and just used that for the majority of calls? Some of the RPC requests that actually do long operations in the agent RPC handlers could then use a 300s timeout.

PR #2462, opened 2017-06-12
Fix plugin installation by removing SafeYAML dependency

kke:
Ruby 2.0, which had no YAML.safe_load is no longer supported so we can drop the SafeYAML gem which was there for compatibility with older rubies.

Before:
```
$ kontena plugin install upcloud
    *   Installing plugin upcloud .. -Called 'load' without the :safe option -- defaulting to safe mode.
You can avoid this warning in the future by setting the SafeYAML::OPTIONS[:default_mode] option (to :safe or :unsafe).
Gem::InstallError : invalid gem: package is corrupt, exception while verifying: YAML data doesn't evaluate to gem specification (Gem::Exception) in /root/.kontena/gems/2.3.1/cache/kontena-plugin-upcloud-0.2.5.gem
 [fail] Installing plugin upcloud
```

After:
```
$ kontena plugin install upcloud
 [done] Installing plugin upcloud
Installed plugin upcloud version 0.2.5
```

For me, only seems to happen on Ruby < 2.4.0. Also for some reason, if I had some random perhaps invalid `kontena-*.gem` files in my current directory, I got:

```
Gem::Package::FormatError : package content (data.tar.gz) is missing in kontena-cli-1.3.0.pre2.gem
```


cli/lib/kontena/cli/plugins/uninstall_command.rb
```diff
@@ -6,10 +6,8 @@ class UninstallCommand < Kontena::Command
     include Kontena::Cli::Common
 
     parameter 'NAME', 'Plugin name'
-    option "--force", :flag, "Force remove", default: false, attribute_name: :forced
```
jnummelin: Let's not remove options in patch releases

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -94,13 +92,13 @@ def stack_from_reader(reader)
       stack
     end
 
-    def stack_from_yaml(filename, name: nil, values: nil, defaults: nil)
-      reader = reader_from_yaml(filename, name: name, values: values, defaults: defaults)
+    def stack_from_yaml(filename, name: nil, values: nil, defaults: nil, action: nil)
```
jnummelin: What is the `action` option?

kke:
`kontena plugin search/install/uninstall` + upgrade (rubygems didn't like SafeYAML, that's why there was `without_safe { ..... }` blocks.
`kontena stack validate` (uses YAML.safe_load)
`kontena vault export` (uses `{}.to_yaml`)

:white_check_mark:  2.2.3
:white_check_mark:  2.3.3
:white_check_mark:  2.4.1

cli/lib/kontena/cli/plugins/uninstall_command.rb
```diff
@@ -6,9 +6,11 @@ class UninstallCommand < Kontena::Command
     include Kontena::Cli::Common
 
     parameter 'NAME', 'Plugin name'
+
     option "--force", :flag, "Force remove", default: false, attribute_name: :forced
 
     def execute
+      exit_with_error "Plugin #{pastel.cyan(name)} is not installed" unless Kontena::PluginManager.instance.installed(name) && !ENV['NO_PLUGINS']
```
SpComb: Does this strictly belong to this PR? Anyways, the `Kontena::PluginManager#uninstall_plugin` will already do the same check and raise an almost identical error, so this is redundant.

kke: It's there to avoid the confirm dialog when the plugin isn't installed in the first place.

SpComb: Makes sense. The triple-negation on the `unless !ENV['NO_PLUGINS']` is just super confusing, but I trust this to do the right thing based on testing without trying to understand it :P

SpComb: BTW as a separate issue, it wouldn't hurt to drop the `Kontena::PluginManager#uninstall_plugin` and make this exit overridable with `--force`, because a broken plugin with `LoadError`'s won't be `installed`, and thus impossible to `kontena plugin uninstall` it ATM?

PR #2471, opened 2017-06-13
Fix plugin install exit code when install fails

### LINKED ISSUE #2470: CLI plugin install command exits with zero on errors
From #2469, separate bug:

```
/kontena/test # kontena plugin install aws && echo yay it worked
    *   Installing plugin aws .. /Called 'load' without the :safe option -- defaulting to safe mode.
You can avoid this warning in the future by setting the SafeYAML::OPTIONS[:default_mode] option (to :safe or :unsafe).
/Gem::InstallError : invalid gem: package is corrupt, exception while verifying: YAML data doesn't evaluate to gem specification (Gem::Exception) in /root/.kontena/gems/2.3.3/cache/kontena-cli-1.3.0.gem
 [fail] Installing plugin aws     
yay it worked
```

kke:
Fixes #2470

### DIFF
--- a/cli/lib/kontena/cli/plugins/install_command.rb
+++ b/cli/lib/kontena/cli/plugins/install_command.rb
@@ -14,34 +14,18 @@ module Kontena::Cli::Plugins
       installed_version = Kontena::PluginManager.instance.installed(name)
 
       if installed_version
-        installed = spinner "Upgrading plugin #{name.colorize(:cyan)}" do |spin|
-          begin
-            Kontena::PluginManager.instance.upgrade_plugin(name, pre: pre?)
-          rescue => ex
-            $stderr.puts pastel.red("#{ex.class.name} : #{ex.message}")
-            logger.error(ex)
-            spin.fail
-            ex
-          end
+        installed = spinner "Upgrading plugin #{name.colorize(:cyan)}" do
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/cli/plugins/install_command.rb
```diff
@@ -34,11 +35,14 @@ def execute
...
             $stderr.puts pastel.red("#{ex.class.name} : #{ex.message}")
             logger.error(ex)
-            spin.fail!
+            spin.fail
+            ex
           end
         end
       end
 
+      exit_with_error "Install failed: #{installed.message}" if installed.respond_to?(:message)
```
SpComb: Why not just let the exception raise itself instead?

kke: It does that now

SpComb:
Unable to review with the #2462 changes merged in.

I think the best way to do chained PRs on github would be to set the #2462 branch as the base for this PR, then github will show the diff against that PR. Once the base PR is merged, fix the base branch for this PR back to master.

kke:
Cleaned out the other PR

kke:
The rubygems commands/functions/whatever do write to STDOUT during spinning and mess out the output, but that's another issue for the future.

PR #2483, opened 2017-06-15
Fix usage of node IDs in CLI, server API JSON + docs

SpComb:
The documented node JSON was incorrect:
* `id` was documented as being the `:grid/:name` path, whereas it is actually the Docker ID
* `node_id` was documented, but did not exist
* `initial_member` was documented as `initial_node`
* `*_at` timestamp fields were empty, and missing `connected_at` (which was formatted incorrectly before #2529)

## API changes

Affects `/v1/grids/:grid/nodes` and `/v1/nodes/:grid/:node` 

* Add a new `node_id` field as documented
* Change the `id` field to use the intended `node.to_path` syntax, as documented

## CLI fixes

* Fix `kontena node show` to use the `node['node_id']` instead of `node['id']`
* Fix `kontena node labels remove` to not use `/v1/nodes/#{current_grid}/#{node['id']}` - this will fail with a HTTP 404 against the new server
* Fix CLI commands to use `/v1/nodes/#{node['id']}` when doing response -> request:
  * `kontena etcd health`
  * `kontena node labels add`
  * `kontena node labels remove`
  * `kontena node remove`
* Fix CLI `NODE` argument to be the `Node name` instead of `Node id` (trying to use the `:grid/:name` ID as a CLI argument will not work)


server/docs/source/index.html.md
```diff
@@ -286,15 +286,15 @@ follow | Stream logs
...
+	"id": "RQKP:Y32W:SB4H:7TNG:5BKC:R6ZO:5B25:C2AV:3Z3Q:SVPX:A76C:WPBX",
 	"name": "misty-sun-87",
 	"connected": "true",
-	"created_at": "",
-	"updated_at": "",
-	"last_seen_at": "",
+	"created_at": "2017-06-14T12:33:05.139Z",
+	"updated_at": "2017-06-14T13:37:36.968Z",
+	"last_seen_at": "2017-06-14T13:38:03.785Z",
+	"connected_at": "2017-06-14T12:33:05.084+00:00",
```
SpComb: Looks like a bug, should probably use the same `Z` UTC timezone suffis as the other timestamps... seems to be some odd `Time` vs `DateTime` difference.. :cry:

server/docs/source/index.html.md
```diff
@@ -286,15 +286,15 @@ follow | Stream logs
 
 ```json
 {
-	"id": "mygrid/misty-sun-87",
-	"node_id": "RQKP:Y32W:SB4H:7TNG:5BKC:R6ZO:5B25:C2AV:3Z3Q:SVPX:A76C:WPBX",
+	"id": "RQKP:Y32W:SB4H:7TNG:5BKC:R6ZO:5B25:C2AV:3Z3Q:SVPX:A76C:WPBX",
```
SpComb: This is really an API bug, it should be like it is in the documentation, but unfortunately it isn't. 

The `/v1/nodes/RQKP:Y32W:SB4H:...` API exists, but it it's deprecated and odd: it only supports GET/PUT, and it requires the `Kontena-Grid-Token` header.

server/docs/source/index.html.md
```diff
@@ -286,15 +286,15 @@ follow | Stream logs
 
 ```json
 {
-	"id": "mygrid/misty-sun-87",
-	"node_id": "RQKP:Y32W:SB4H:7TNG:5BKC:R6ZO:5B25:C2AV:3Z3Q:SVPX:A76C:WPBX",
```
SpComb: Also considering adding a new `node_id ` field to the API, which just contains the same value as the current `id` field. It would be for future-compatibility: clients could start using the `node_id` field, and it would be easier to change the `id` field later.

SpComb:
Alternatively to fixing the docs, we could also consider changing the API to match the docs... i.e. add a new `node_id` field with the current value, and change the existing `id` field to use `HostNode.to_path`...

I suspect most API users (including the CLI) are not going to be using the `node['id']` for the API request URLs, they mostly use `/v1/nodes/#{current_grid}/#{node['name']}`, which will continue to work.

Some of the plugins look like they use `nods/#{node['id']}` and include the `Kontena-Grid-Token` header, but those would still continue to work?

server/docs/source/index.html.md
jnummelin: Yeah, should definitely use the same format.

jnummelin:
> Some of the plugins look like they use nods/#{node['id']} and include the Kontena-Grid-Token header, but those would still continue to work?

I don't think so. The deprecated `/nodes/...` api uses `grid.host_nodes.find_by(node_id: id)` to find the node (to update labels) which would fail if the `node['id']` would be `HostNode.to_path`

jnummelin:
I do think it would be better in long run to actually fix the APIs but before that we need to fix all the plugins not to use the deprecated `/nodes/` API

SpComb:
> Some of the plugins look like they use nods/#{node['id']} and include the Kontena-Grid-Token header, but those would still continue to work?
> 
> I don't think so. The deprecated /nodes/... api uses grid.host_nodes.find_by(node_id: id) to find the node (to update labels) which would fail if the node['id'] would be HostNode.to_path

If we changed the `id` field to use `to_path`, then a client using `"/v1/nodes/#{node['id']}"` interpolation would end up using an URL like `"/v1/nodes/gridname/nodename"` - that would transparently use the current API, not the deprecated one.

jnummelin:
> If we changed the id field to use to_path, then a client using "/v1/nodes/#{node['id']}" interpolation would end up using an URL like "/v1/nodes/gridname/nodename" - that would transparently use the current API, not the deprecated one.

Yes, but they use different authentication which I fear some of the plugins will crash on.

SpComb:
> Yes, but they use different authentication which I fear some of the plugins will crash on.

At least the DO plugin is just using the standard CLI `Kontena::Client`, so I assume those requests also have the oauth token.

So while I think changing the node API `id` field might actually work, not sure it's a good idea either... better to fix the CLI and plugins first. But that would also be easier if there was a "fixed" `id` field...

server/docs/source/index.html.md
SpComb: Separate PR for fixing those: #2529

SpComb:
Interestingly, the CLI `kontena node label remove` command uses the `node['id']`:

```ruby
      node = client.get("nodes/#{current_grid}/#{node_id}")
      data = ...
      client.put("nodes/#{current_grid}/#{node['id']}", data)
```

```
DEBUG [Request]: PUT http://localhost:9292/v1/nodes/development/XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5 | Headers: {Accept: application/json, Content-Type: application/json, Authorization: Bearer}  | Body: 
{"labels":["test"]}
DEBUG [Response]: Headers: {Content-Type: application/json, X-Kontena-Version: 1.4.0.dev}  | Status: 200 | Body: 
{"id":"XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5",...}
```

This example would break if we changed the API `id` field.

SpComb:
Made some changes to the API to add the `node_id` field, and deprecate the `id` field without actually changing it.

server/docs/source/index.html.md
```diff
@@ -295,17 +295,21 @@ follow | Stream logs
 
 ## Node
 
+The `id` field is deprecated, use the `node_id` field available in Kontena 1.4 instead.
+
 ```json
 {
-	"id": "mygrid/misty-sun-87",
+	"id": "...",
+	"api_id": "mygrid/misty-sun-87",
```
SpComb: Perhaps some better name for this field?

SpComb:
> We should just fix the contents of id field and not introduce api_id. This might cause some unfortunate side-effects (mainly to plugins) but that's what happens when clients rely on buggy behaviour.

Done.

Some testing with a 1.3.2.rc2 CLI against this 1.4.0.dev server, most commands seem to work, apart from the two fixes in the CLI:

### `kontena node show`

```
core-01:
  id: development/core-01
  agent version: 1.4.0.dev
```

The `id` here used to be the Docker ID, and with a 1.4.0.dev CLI it's still the node ID, but with an older CLI against the 1.4.0.dev server, it's not.

### `kontena node label remove core-02 test=2`

```
 [error] 404 : Not found (/v1/nodes/development/development/core-02)
```

This is the only broken CLI command that I could find with a search for `node['id']`, fixed in the 1.4.0.dev CLI, but this will be a regression with an older CLI against the newer 1.4 server.

SpComb:
The e2e specs pass (apart from know issues with CLI `plugin` support, and #2537.

Looking through all the CLI `nodes/...` API calls, I don't expect to find any other CLI issues that need fixing for this.

Need to test some of the plugins next.

SpComb:
There's a slight problem for new nodes that connect, which are missing any `name` (e.g. they have the wrong version). The resulting `id` is wrong:

```
{"id":"development/","node_id":"4VU2:7NNE:2S67:L5IS:SOK7:SCHM:BGYS:7VFU:D66L:27R4:H6AA:3KBA","connected":false,"created_at":"2017-06-30T09:47:11.135Z","updated_at":"2017-06-30T09:47:11.135Z","connected_at":null,"last_seen_at":null,"name":null,...}
```

The `HostNode.to_path` should probably use `name || node_id`.

SpComb:
The plugins are very hard to test without a `kontena/agent:1.4.0.dev` release, because the plugin-provisioned agent will not be able to connect to the 1.4.0.dev server.

Hacking it manually by connecting an identically named vagrant dev node, and the `kontena digitalocean node create` plugin seems to survive this change:

```
$ kontena digitalocean node create core-03
> Choose a datacenter region:  Amsterdam 2
> Select CoreOS channel Stable
> Choose droplet size:  1GB/1CPU/30GB ($10/mo)
> How many droplets?:  1
 [done] Creating DigitalOcean droplet core-03      
 [done] Waiting for node core-03 join to grid development      
Warning: Server version is 1.4.0.dev. You are using CLI version 1.3.2.rc2.
```

```
I, [2017-06-30T09:57:04.678017 #12]  INFO -- WebsocketBackend: node UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC agent version 1.4.0.dev connected at 2017-06-30 09:57:04 UTC, 0.04s ago
I, [2017-06-30T09:57:04.696783 #12]  INFO -- Agent::NodePlugger: Connected node development/ at 2017-06-30 09:57:04 UTC

172.17.0.1 - - [30/Jun/2017:09:57:10 +0000] "GET /v1/grids/development/nodes HTTP/1.1" 200 4496 0.2793
172.17.0.1 - - [30/Jun/2017:09:57:11 +0000] "PUT /v1/nodes/development/core-03 HTTP/1.1" 200 1268 0.1291
```

The DO plugin labels were succesfully set on the new node:

```
NAME               VERSION     STATUS    INITIAL   LABELS
⊛ core-03          1.4.0.dev   online    -         region=ams2,az=ams2,provider=digitalocean
```

SpComb:
Fixed the `to_path` for initializing nodes: `{"id":"development/UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","node_id":"UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","connected":false,"created_at":"2017-06-30T09:57:26.657Z","updated_at":"2017-06-30T09:57:26.657Z","connected_at":null,"last_seen_at":null,"name":null,...}`

#### `DEBUG=api kontena node ls`
```
DEBUG [Request]: GET http://localhost:9292/v1/grids/development/nodes | Headers: {Accept: application/json, Authorization: Bearer} 
DEBUG [Response]: Headers: {Content-Type: application/json, X-Kontena-Version: 1.4.0.dev}  | Status: 200 | Body: 
{"nodes":[{"id":"development/UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","node_id":"UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","connected":false,"created_at":"2017-06-30T09:57:26.657Z","updated_at":"2017-06-30T09:57:26.657Z","connected_at":null,"last_seen_at":null,"name":null,...
NAME               VERSION     STATUS    INITIAL   LABELS
⊛ core-01          1.4.0.dev   online    1 / 1     test
⊝ core-02          1.4.0.dev   offline   -         test=label,test=2
⊛ core-03          1.4.0.dev   online    -         region=ams2,az=ams2,provider=digitalocean
⊝ (initializing)               offline   -
```

#### `DEBUG=api kontena node rm --force UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3`

```
DEBUG [Request]: GET http://localhost:9292/v1/nodes/development/UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3 | Headers: {Accept: application/json, Authorization: Bearer} 
DEBUG [Response]: Headers: {Content-Type: application/json, X-Kontena-Version: 1.4.0.dev}  | Status: 200 | Body: 
{"id":"development/UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","node_id":"UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3","connected":false,"created_at":"2017-06-30T09:57:26.657Z","updated_at":"2017-06-30T09:57:26.657Z","connected_at":null,"last_seen_at":null,"name":null,"os":null,"engine_root_dir":null,"driver":null,"network_drivers":[],"volume_drivers":[],"kernel_version":null,"labels":[],"mem_total":null,"mem_limit":null,"cpus":null,"public_ip":null,"private_ip":null,"overlay_ip":"10.81.0.4","agent_version":null,"docker_version":null,"peer_ips":["192.168.66.101","192.168.66.102","82.181.210.184"],"node_number":4,"initial_member":false,"grid":{"id":"development","name":"development","initial_size":1,"token":"tBsoco4sBClMtIOD/bHeIqS1KYM0kZMG7B960jNWWQXMOxqhOf8W1QgiNJlW9u6B/qS+rfS2ZLpC/ZJsyPNj0A==","stats":{"statsd":null},"trusted_subnets":[],"subnet":"10.81.0.0/16","supernet":"10.80.0.0/12"}}
    *   Removing UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3 node from development grid  .. \DEBUG [Request]: DELETE http://localhost:9292/v1/nodes/development/UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3 | Headers: {Accept: application/json, Authorization: Bearer} 
    *   Removing UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3 node from development grid  .. |DEBUG [Response]: Headers: {Content-Type: application/json, X-Kontena-Version: 1.4.0.dev}  | Status: 200 | Body: 
{}
 [done] Removing UXYC:NLTG:UBBL:WUO4:COY7:CLHS:XELH:5JGA:C5Q6:T6DV:OKLX:UFG3 node from development grid      
```

The fixed `id` works, even for non-initialized nodes.

cli/lib/kontena/cli/nodes/labels/remove_command.rb
```diff
@@ -12,7 +12,7 @@ class RemoveCommand < Kontena::Command
     def execute
       node = client.get("nodes/#{current_grid}/#{node_id}")
       data = { labels: Array(node['labels']).reject {|label| label_list.include?(label) } }
-      client.put("nodes/#{current_grid}/#{node['id']}", data)
+      client.put("nodes/#{current_grid}/#{node['name']}", data)
```
jnummelin: Shouldn't `node['id']` work here?

unrelated-side-note:
Why are we using `PUT` to remove a node? :)

server/docs/source/index.html.md
```diff
@@ -428,7 +429,7 @@ Update a node details.
 
 ### Endpoint
 
-`PUT /v1/nodes/{node_id}`
+`PUT /v1/nodes/{id}`
```
jnummelin: Inconsistent use of `:id` vs. `{id}`, although all the endpoints in the API docs seem to have these used inconsistently :(

cli/lib/kontena/cli/nodes/labels/remove_command.rb
SpComb: This is `kontena node label remove`, not `kontena node remove` :)

> Shouldn't `node['id']` work here?

Sure, without the `current_grid`. The `#{current_grid}/#{name}` just matches what all the other commands use. Could potentially change all of the commands to use `nodes/#{node['id']}`?

SpComb: There is actually one edge-case where `nodes/#{node['id']}` works but `nodes/#{current_grid}/#{node['name']}` won't: if it's a new node that just connected with a grid token, and it hasn't updated its node_info yet.

server/docs/source/index.html.md
SpComb: I think the "consistent" choice here would be `{node_id}`, but then that is super-confusing re `node['id']` vs `node['node_id']` - it needs to be the first one.

cli/lib/kontena/cli/nodes/labels/remove_command.rb
jnummelin: I'd change all the commands to use the "new" `nodes/#{node['id']}`

server/docs/source/index.html.md
jakolehm: @jnummelin pls open an issue about `id` inconsistency.

cli/lib/kontena/cli/nodes/labels/remove_command.rb
SpComb: Looks like all the commands that use `nodes/#{current_grid}/#{node['name']}` are missing specs for that part... need to write those up too.

cli/lib/kontena/cli/nodes/show_command.rb
```diff
@@ -4,16 +4,16 @@ class ShowCommand < Kontena::Command
...
     def execute
       require_api_url
       require_current_grid
       token = require_token
 
-      node = client(token).get("nodes/#{current_grid}/#{node_id}")
-      puts "#{node['name']}:"
-      puts "  id: #{node['id']}"
+      node = client(token).get("nodes/#{current_grid}/#{self.node}")
+      puts "#{node['id']}:"
```
SpComb: Also changed this, for consistency with other commands like `kontena service show`...

SpComb:
Have a bunch of new CLI specs: :cake:

server/docs/source/index.html.md
jnummelin: #2556 opened

cli/lib/kontena/cli/nodes/show_command.rb
jnummelin: 👍

cli/lib/kontena/cli/nodes/labels/remove_command.rb
```diff
@@ -2,17 +2,17 @@ module Kontena::Cli::Nodes::Labels
...
 
     requires_current_master
     requires_current_master_token
     requires_current_grid
 
     def execute
-      node = client.get("nodes/#{current_grid}/#{node_id}")
+      node = client.get("nodes/#{current_grid}/#{self.node}")
       data = { labels: Array(node['labels']).reject {|label| label_list.include?(label) } }
-      client.put("nodes/#{current_grid}/#{node['id']}", data)
```
SpComb: AFAIK this is the only API usage that got broken by the API `id` change.

cli/lib/kontena/cli/nodes/labels/add_command.rb
```diff
@@ -2,17 +2,17 @@ module Kontena::Cli::Nodes::Labels
   class AddCommand < Kontena::Command
     include Kontena::Cli::Common
 
-    parameter "NODE_ID", "Node id"
+    parameter "NODE", "Node"
```
jakolehm: Should description be `Node name` ?

SpComb: Not necessarily, because it can be either the node `name`, or the `node_id` - both of these will work.

Using the `node_id` is is only really relevant for odd cases like `kontena node rm 44C7:P5OM:NBJT:WXHV:6EDU:67T5:YDMX:4YPU:PF6D:VUH5:7LE7:5RC7`, if the agent has connected but never sent the node info... there is no name to remove it with. This reqiures picking out the node_id from `DEBUG=api kontena node list`, though :P

jakolehm: Should cli show node_id as a name if name is nil? Otherwise this might be a bit confusing to use...

SpComb: You mean in `kontena node list`? Perhaps, currently it will just show `(initializing)` in this case, which makes it very hard to get rid of that node if it never manages to update.

This PR or separate issue?

SpComb: The other, perhaps even better option, would be to send the `Kontena-Node-Name` has a websocket request header, so there would never be any need to create `HostNode`s without any name..

SpComb: Separate issue:  #2558

SpComb:
@jakolehm PTAL and either :white_check_mark: or :x:, AFAIK this is otherwise ready for merge?

cli/lib/kontena/cli/nodes/labels/add_command.rb
jakolehm: If listings shows "Name" and that is used here as primary parameter then parameter description should say "Node name".

SpComb: Yes, it's name 99.98% of the time. We can keep the `node_id` variant as an undocumented secret :)

Issue #2253, opened 2017-05-03
node cloud-config systemd-networkd units need updating for CoreOS changes


SpComb:
As a fix for #1264, the CoreOS cloud-config used by Kontena plugins ships the following `/etc/systemd/networkd` units to prevent systemd-networkd from managing the `weave` and `veth` interfaces, causing netlink races:

* `50-docker.network -> /dev/null`
* `50-docker-veth.network -> /dev/null`
* `zz-default.network` with `[Match] Name=... `

However, these network units have also been updated by CoreOS to fix the same issues in a different manner:

* https://github.com/coreos/systemd/pull/73 added support for `[Link] Unamanged=true`
* https://github.com/coreos/coreos-overlay/pull/2300 changed the `50-docker.network` and `50-docker-veth.network` units to use `[Link] Unamanged=true`, fixing the issue
* https://github.com/coreos/coreos-overlay/pull/2306 renamed the `50-docker-veth.network` to `90-docker-veth.network` meaning that our `/etc/systemd/network/50-docker-veth.network` no longer masks it out (?)

The current Kontena CoreOS cloud-configs are also broken on VMWare platforms, since CoreOS also ships a [`yy-vmware.network` unit](https://github.com/coreos/init/blob/master/systemd/network/yy-vmware.network) which does the same thing as `zz-default.network`, meaning that our network units overrides do not prevent the `yy-vmware.network` from matching the `docker0` bridge, since we mask out the `50-docker.network`.

Instead of masking out the `50-docker.network` units and overriding the `zz-default.network` to limit the `[Match]` criteria, we should instead utilize the new support for `[Link] Unamanged=true`, by unmasking the `50-docker.network` and `*-docker-veth.network`, instead adding in a new `50-weave.network` that also marks the `weave` and `datapath` bridges as unmanaged. Then we should be able to revert back to the upstream default `zz-default.network` unit.

SpComb:
Weave-related interfaces that need to matched and explicitly configured as unmanaged on CoreOS to prevent the `zz-default.network` from matching:

* `datapath` `Driver=openvswitch`
* `weave`  `Driver=bridge`
* `dummy0` `Driver=dummy` (but `udevadm info /sys/class/net/dummy0` is missing any `DEVTYPE`?)
* `vethwe-datapath`  `Driver=veth`
* `vethwe-bridge` `Driver=veth`
* `vxlan-6784` `Driver=vxlan`

The suggested network units for this:

#### `50-weave.network`
```
[...]

PR #2494, opened 2017-06-19
cli: update grid cloud-config to use [Link] Unmanaged=true

SpComb:
See #2253

Remove the masking of the `50-docker.network` and (non-existant) `50-docker-veth.network` which have been fixed in CoreOS, and add a similar `50-weave.network` + `90-dummy.network` to also leave the weave interfaces unmanaged. This configuration should also work better on VMware, where the `yy-vmware.network` would still have matched the Docker and Weave interfaces otherwise.

This keeps the `zz-default.network` override with `[Match] Name=...`, because IMO as long as we're overriding it to configure the DNS settings, it's also worth limiting the match scope. However, longterm I think it would be best to leave the CoreOS `zz-default.network` untouched, and override `/etc/systemd/resolved.conf` to configure the `kontena.local` DNS instead?

### DIFF
--- a/cli/lib/kontena/machine/cloud_config/cloudinit.yml
+++ b/cli/lib/kontena/machine/cloud_config/cloudinit.yml
@@ -19,10 +19,23 @@ write_files:
       fs.inotify.max_user_instances = 8192
 coreos:
   units:
-    - name: 50-docker.network
-      mask: true
-    - name: 50-docker-veth.network
-      mask: true
+    - name: 50-weave.network
+      content: |
+        [Match]
+        Name=weave datapath vxlan-6784
+
+        [Network]
+        Description=Network interfaces managed by weave
+
+        [Link]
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/machine/cloud_config/cloudinit.yml
```diff
@@ -19,10 +19,23 @@ write_files:
...
+
+        [Link]
+        Unmanaged=true
+    - name: 90-dummy.network
+      content: |
+        [Match]
+        Name=dummy0
+
+        [Link]
+        Unmanaged=true
```
jakolehm: Should this be `Unmanaged=yes` ?

SpComb: `yes`, `true`, `on` and `1` all work

PR #2496, opened 2017-06-20
Fix healtcheck port validation to check port in unix port range

### LINKED ISSUE #2491: health_check port can be set to 0
Should maybe validate and fail the stack..?

jnummelin:
fixes #2491

### DIFF
--- a/server/spec/mutations/grid_services/create_spec.rb
+++ b/server/spec/mutations/grid_services/create_spec.rb
@@ -352,6 +352,40 @@ describe GridServices::Create do
       expect(outcome.success?).to be(false)
     end
 
+    it 'fails to save health_check, port 0' do
+      outcome = described_class.new(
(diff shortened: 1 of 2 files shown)


server/spec/mutations/grid_services/create_spec.rb
```diff
@@ -352,6 +352,40 @@
...
+          stateful: false,
+          health_check: {
+            protocol: 'tcp',
+            interval: 10,
+            timeout: 5,
+            initial_delay: 10,
+            port: 0
+          }
+      ).run
+      expect(outcome.success?).to be(false)
```
SpComb: General nitpick: I would suggest also checking the `expect(outcome.errors.message).to eq 'health_check' => ...`, otherwise this spec might not end up testing the health check validation, if it turns out to fail on some other parameter.

PR #2478, opened 2017-06-15
cli node list: spec and fix initializing node with nil name

SpComb:
Fixes #2447

PR #2504, opened 2017-06-20
Create HostNode with per-node token for websocket auth

### LINKED ISSUE #2501: Per-node tokens for agent-server websocket authentication
The current agent-server websocket RPC mechanism uses a grid-level shared secret token for authentication, together with the agent-provided Docker ID for identifying nodes. This is useful for simplying autoscaling deployments, allowing any number of grid nodes to be provisioned using a common cloud-config. However, this is bad from a security standpoint, because the compromise of any node will leak the shared grid token, which is also very difficult to replace (#1973), requiring all nodes to be re-provisioned.

The agent provisioning and agent-server websocket connections should be changed to use predetermined node IDs and per-node tokens. Instead of the server dynamically creating new host nodes as agents connect, the host nodes should be created beforehand, using a user-supplied name and a server-generated token. 

The agents would then be provisioned with the following configuration:

* `KONTENA_URI=wss://...`
* `KONTENA_NODE_ID=grid/node-name`
* `KONTENA_NODE_TOKEN=<generated>`

Removing a node would then invalidate the node token, preventing the agent from reconnecting. Assuming further support for weave secret rotation, the grid token could then be replaced without affecting the remaining nodes.

With per-node authentication tokens, the agent-server RPC protocol should also be improved to remove the current use of explicit `node_id` RPC parameters, strictly limiting each agent to only have access to the `GridServiceInstance` and `Container` objects associated with the authenticated `HostNode`. This would mean that grid secrets would only be accessible by the specific nodes that the service is deployed to.

The server must still retain support for the legacy grid token based RPC websocket authentication, both for existing host nodes, as well as simplified autoscaling deployments using a fixed cloud-config. However, if all host nodes are deployed using node tokens, and given further weave secret related improvements, this would serve to avoid the potential #1973 security headaches.

SpComb:
See #2501

* New `kontena node create ` -> `POST /v1/grids/:grid/nodes` -> `HostNodes::Create` to create a new `HostNode` model with a generated `token` and given `name`

* New `kontena node env` -> `GET /v1/nodes/:id/token`, used to generate the agent config for manual provisioning:

    ```
    KONTENA_URI=ws://localhost:9292//
    KONTENA_NODE_TOKEN=EajcuCR5uiBnSqF+yIvvelnLfeEmHjuo/BmsbXNwBcQOeTT5jHebEZGoS/7t3bUDH6bAboK6onwktDNv0ZX7OA==
    ```

    Similar to `kontena grid env` when using grid tokens.

* New `kontena node reset-token --reset-connection` -> `PUT /v1/nodes/:id/token` to update node tokens, and reset the agent websocket connection

    This includes converting pre-existing nodes configured with grid tokens to use node tokens, and `kontena node reset-token --clear-token` to revert if unable to reconfigure the agent.

* `kontena node rm` will remove an online node if it has a node token, forcing the agent to disconnect, and preventing it from reconnecting (unless it also has the grid token)

The new `kontena node` commands require the `grid_admin` role, and normal grid users do not have access to the node tokens.

The agent connects with a `Kontena-Node-Id` header containing the `Docker.info['ID']`. Existing agents configured with just `KONTENA_TOKEN` will continue to connect using the `Kontena-Grid-Token` header, and the server will use the agent-provided node ID to dynamically create a new `HostNode`, if no node for that node ID exists yet.

New agents provisioned using `KONTENA_NODE_TOKEN`  will connect using the new `Kontena-Node-Token` header instead. The server maps this to an unique `HostNode.token`, and associates the node ID sent by the agent with this host node.

The `node_id`s must be unique, and cannot change later: the server will reject the connection if an agent later connects with the same node token but a different node ID, or some agent connects with the same node ID but a different node token. This protects against accidentally provisioning multiple nodes with the same `KONTENA_NODE_TOKEN`.

The server will also reject any connection using a `Kontena-Grid-Token` if there exists a host node with a matching node ID, and that host node has a node token - the agent is expected to be configured with the node token, not the grid token. This is intended to differentiate which nodes can be force-removed using `kontena node rm` when online.

## TODO

* If `kontena node create` requires `grid_admin`, then should `kontena node rm` also require `grid_admin`?

### Revokable node tokens

`kontena node rm` will now allow removing online nodes that have node tokens. This will cause the agent to get disconnected, and quit. Removing the node also removes the node token, preventing the agent from reconnecting.

#### CLI `kontena node rm core-03`
```
 [warn] Node core-03 is still connected using a node token, but will be force-disconnected
Destructive command. To proceed, type "core-03" or re-run this command with --force option.
> Enter 'core-03' to confirm:  core-03
 [done] Removing core-03 node from development grid   
```

#### Server
```
172.17.0.1 - - [29/Jun/2017:12:11:45 +0000] "DELETE /v1/nodes/development/core-03 HTTP/1.1" 200 2 0.2940
D, [2017-06-29T12:12:11.967379 #20] DEBUG -- WebsocketBackend: keepalive ping 0.00s of 5.00s timeout from client XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5
D, [2017-06-29T12:12:11.970415 #20] DEBUG -- WebsocketBackend: keepalive ping 0.00s of 5.00s timeout from client CAR3:26EI:CKHW:M3US:6L3D:W7CI:Y6KU:CJ5D:6ERB:DUHB:2ZAB:6Y3Q
D, [2017-06-29T12:12:11.971989 #24] DEBUG -- WebsocketBackend: keepalive ping 0.00s of 5.00s timeout from client UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC
W, [2017-06-29T12:12:11.973306 #24]  WARN -- WebsocketBackend: Close connection of removed node UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC
W, [2017-06-29T12:12:11.974365 #24]  WARN -- WebsocketBackend: skip unplug of missing node UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC
D, [2017-06-29T12:12:11.977278 #24] DEBUG -- WebsocketBackend: ignore close of unplugged client
```

#### Agent
```
W, [2017-06-29T12:12:03.911902 #1]  WARN -- Kontena::RpcClient: RPC request /nodes/update failed: RuntimeError: Missing HostNode: UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC
D, [2017-06-29T12:12:05.528670 #1] DEBUG -- Kontena::RpcClient: waited 0.0s of 30.0s until: request /node_service_pods/list has response wth id=772911908 yielded Array
W, [2017-06-29T12:12:05.530386 #1]  WARN -- Kontena::Workers::ServicePodManager: failed to get list of service pods from master: RuntimeError: Node not found
D, [2017-06-29T12:12:05.557995 #1] DEBUG -- Kontena::RpcClient: waited 0.0s of 30.0s until: request /node_volumes/list has response wth id=1836941490 yielded Array
W, [2017-06-29T12:12:05.559568 #1]  WARN -- Kontena::Workers::Volumes::VolumeManager: failed to get list of service pods from master: RuntimeError: Node not found
E, [2017-06-29T12:12:11.974091 #1] ERROR -- Kontena::WebsocketClient: master indicates that this agent should not reconnect: host node UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC has been removed
I, [2017-06-29T12:12:11.974818 #1]  INFO -- Kontena::Workers::LogWorker: stopped log streaming
Shutting down ...
I, [2017-06-29T12:12:12.104871 #1]  INFO -- Kontena::Workers::ContainerHealthCheckWorker: stopped to check status from /ingress-lb.lb-3
```

#### Agent restart
```
I, [2017-06-29T12:16:24.316962 #1]  INFO -- Kontena::WebsocketClient: initialized with node token zb6vz/qod..., node ID UXTT:TPWP:6OQM:3UKS:EPB4:MOQD:DYYT:N42X:VSR5:IIM3:VDCP:22IC
I, [2017-06-29T12:16:25.484987 #1]  INFO -- Kontena::WebsocketClient: connecting to master at ws://192.168.66.1:9292
I, [2017-06-29T12:16:25.637163 #1]  INFO -- Kontena::WebsocketClient: connection established
W, [2017-06-29T12:16:25.639085 #1]  WARN -- Kontena::WebsocketClient: connection closed with code 4002: Invalid node token
```

### Reset token with `kontena node reset`

```
W, [2017-07-14T06:03:30.598770 #1]  WARN -- Kontena::WebsocketClient: connection closed with code 4031: host node core-02 has been disconnected
I, [2017-07-14T06:03:30.833901 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://192.168.66.1:9293
I, [2017-07-14T06:03:30.885708 #1]  INFO -- Kontena::WebsocketClient: connection established
W, [2017-07-14T06:03:30.888061 #1]  WARN -- Kontena::WebsocketClient: connection closed with code 4002: Invalid node token
```


server/app/models/host_node.rb
```diff
@@ -68,7 +71,6 @@ def to_path
   # @param [Hash] attrs
   def attributes_from_docker(attrs)
     self.attributes = {
-      node_id: attrs['ID'],
```
SpComb: Considering re-adding this as some new `docker_id` field, but I'm not sure if it's needed.

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -1,13 +1,11 @@
 require_relative '../logging'
-require_relative '../helpers/node_helper'
 require_relative '../helpers/iface_helper'
 require_relative '../helpers/weave_helper'
 
 module Kontena::NetworkAdapters
   class Weave
     include Celluloid
     include Celluloid::Notifications
-    include Kontena::Helpers::NodeHelper
```
SpComb: AFAICT the current class didn't use the `NodeHelper` methods (`master_url` or `node_id`).

SpComb:
Needs some new `initializing` state for nodes, in between the `kontena node create` and the agent actually connecting.

The `node['name'].nil?` used in #2478  doesn't really work for that.

jakolehm:
Just wondering if we really need `Kontena-Node-Id` when `KONTENA_NODE_TOKEN` is used?

SpComb:
> Just wondering if we really need `Kontena-Node-Id` when `KONTENA_NODE_TOKEN` is used?

Do you mean the configurable `KONTENA_NODE_ID` for the agent, or the `Kontena-Node-Id` header?

Agreed that the `HostNode.token` must be unique across all grids, so having the agent connect with the pre-determined node ID isn't a requirement... I think it might be a good idea for other reasons, but we can implement the HostNode token auth without changing the node ID. I'll split those changes out of this PR... there may be some subtle side-effects of the node ID changes that are best dealt with separately.

server/app/models/host_node.rb
SpComb: I'm not entirely sure what this mechanism to update the `HostNode.node_id` in the `/nodes/update` RPC is meant for, but I feel like it does more harm than good. The node ID shouldn't ever change.

agent/lib/kontena/websocket_client.rb
```diff
@@ -29,16 +28,18 @@ class WebsocketClient
 
     # @param [String] api_uri
     # @param [String] api_token
```
jakolehm: yardoc is outdated

agent/lib/kontena/websocket_client.rb
```diff
@@ -29,16 +28,18 @@ class WebsocketClient
...
+      @node_token = node_token
+
       @rpc_server = Kontena::RpcServer.supervise(as: :rpc_server)
       @rpc_client = Kontena::RpcClient.supervise(as: :rpc_client, args: [self])
       @abort = false
       @connected = false
       @connecting = false
       @ping_timer = nil
-      info "initialized with token #{@api_token[0..10]}..."
+      info "initialized"
```
jakolehm: Probably would make sense to say something about token (grid vs node)?

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,69 @@ def call(env)
...
-
-    # check host node
+  # @raise [CloseError]
+  # @return [HostNode]
+  def find_node(req)
     node_id = req.env['HTTP_KONTENA_NODE_ID'].to_s
-    labels = req.env['HTTP_KONTENA_NODE_LABELS'].to_s.split(',')
+    node_labels = req.env['HTTP_KONTENA_NODE_LABELS'].to_s.split(',')
+
+    if grid_token = req.env['HTTP_KONTENA_GRID_TOKEN']
```
jakolehm: If/elsif internals should probably be own methods.. just for readability.

server/app/models/host_node.rb
jakolehm: Probably some legacy, good idea to cleanup this.

server/app/views/v1/host_nodes/_host_node.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 json.id node.node_id
+json.token node.token
```
jakolehm: Should we limit visibility to some roles?

SpComb: The `grid` API also exposes the grid token for everybody...  #2013 splits the grid token out into a separate API, which might also be an option here (`/v1/nodes/:grid/:node/token`)?

SpComb: Moved the token out to a separate `/v1/nodes/:grid/:node/token` API, and limited both `POST /v1/grids/:grid/nodes` and  `GET /v1/nodes/:grid/:name/token` to grid admins.

docs/using-kontena/nodes.md
```diff
@@ -74,3 +113,47 @@ Nodes labeled as `ephemeral` will automatically be removed by the Kontena Master
...
+
+#### List Nodes
+
+```
+$ kontena node list
+NAME          VERSION     STATUS    INITIAL   LABELS
+⊝ test-node               offline   -
+⊛ core-02     1.4.0.dev   online    -         test=label
+⊛ core-01     1.4.0.dev   online    1 / 1     test
+```
```
SpComb: This example is broken re #2512

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,91 @@ def call(env)
...
+    return node
+  end
+
+  # @param node_id [String] request header Kontena-Node-Id
+  # @param node_token [String] request header Kontena-Node-Token
+  # @param init_attrs [Hash] initialize attributes on new node
+  # @raise [CloseError]
+  # @return [HostNode] with node_id set
+  def find_node_by_node_token(node_id, node_token, init_attrs)
+    node = HostNode.find_by(token: node_token.to_s)
```
SpComb: Using the `token` for lookup opens up the possibility for having multiple nodes with the same `node_id`, if the same agent is re-configured with a new `KONTENA_NODE_TOKEN` generated from `kontena node create`...

How much of a problem is this? It would need some checks to reject the initial connection if a node already exists with the `node_id` + an unique index on the `node_id`?

server/app/views/v1/host_nodes/_host_node.json.jbuilder
SpComb: This does mean that `POST /v1/grids/:grid/nodes` no longer returns the `{"token": ...}`, so any provisioning tool will have to do an extra `GET /v1/nodes/:grid/:name/token` request for it.

server/app/views/v1/host_nodes/token.json.jbuilder
```diff
@@ -0,0 +1,2 @@
+json.id @node.node_id
```
SpComb: Using the `node_id` here feels particularly broken, because in the most common use-cases of this API, it's simply going to be `null`.... :face_with_head_bandage:

agent/bin/kontena-agent
```diff
@@ -9,19 +9,25 @@ Docker.validate_version!
...
   abort('Configuration error: The KONTENA_URI env is required')
 end
 if api_uri.match(/^http.*/)
   api_uri = api_uri.sub('http', 'ws')
 end
 
+if grid_token && !grid_token.empty?
+
+elsif node_token && !node_token.empty?
+
```
jnummelin: Looks a bit weird having empty if/elsif blocks. I get the point, but maybe some other styling would read better?

docs/references/environment-variables.md
```diff
@@ -27,7 +27,8 @@ toc_order: 2
 ## Kontena Agent
 
 - `KONTENA_URI`: Kontena Master websocket uri (required)
-- `KONTENA_TOKEN`: Kontena Grid token (required)
+- `KONTENA_TOKEN`: Kontena [Grid token](../using-kontena/grids.md#Grid Token) (required)
+- `KONTENA_NODE_TOKEN`: Kontena [Node token](../using-kontena/nodes.md#Node Token) token (required)
```
jnummelin: Both are not required? Only one of the tokens is required?

server/app/middlewares/websocket_backend.rb
jnummelin: IMO it's perfectly ok requirement to have unique `node_id`s. It sounds like it's gonna get quite tricky to manage all the cases properly so maybe not worth the effort?

SpComb: With the current grid tokens, the `node_id`s are defacto unique, although there is no unique index. There's a very short race condition window between the `grid.host_nodes.find_by(node_id: node_id)` and `grid.host_nodes.create!(node_id: node_id, labels: labels)` (if two agents connect with the same node ID at exactly the same time).

There is also no unique index on the `name` either, even though it's used for the API URLs.

SpComb:
TODO: allow `kontena node rm` for nodes using node tokens for websocket auth. That will disconnect the websocket and force the agent offline, not allowing it to return. Needs some changes to detect if a node is using grid or node tokens, because agents using grid tokens will just reconnect and get re-created.

server/app/views/v1/host_nodes/token.json.jbuilder
SpComb: I would suggest per #2483 that new node APIs use `id: @node.to_path` for consistency with the other APIs, not the deprecated `node_id` used in the existing node APIs.

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,95 @@ def call(env)
...
+
+    node = grid.host_nodes.find_by(node_id: node_id)
+
+    if !node
+      node = grid.host_nodes.create!(node_id: node_id, **init_attrs)
+
+      logger.info "new node #{node} connected using grid token"
+
+    elsif node.token
+      raise CloseError.new(4005), "Invalid grid token, node was created using a node token"
```
SpComb: This should only affect new nodes created via `HostNodes::Create`, and not any existing nodes.

SpComb:
TODO: use `kontena node update --generate-token` to replace node tokens

cli/lib/kontena/cli/nodes/update_command.rb
```diff
@@ -3,18 +3,42 @@ class UpdateCommand < Kontena::Command
...
+    end
 
     def execute
-      require_api_url
-      require_current_grid
-      token = require_token
+      self.update_token if self.token or self.generate_token?
 
       data = {}
-      data[:labels] = label_list if label_list
```
SpComb: The `if label_list` doesn't really work for multivalued options... an empty array is true.

This used to wipe out the node labels, now hurts because the `--label` is not the only option anymore.

cli/lib/kontena/cli/nodes/update_command.rb
```diff
@@ -3,18 +3,42 @@ class UpdateCommand < Kontena::Command
...
+    requires_current_master_token
+    requires_current_grid
+
+    parameter "NAME", "Node name"
+
+    option ["--token"], "TOKEN", "Node token"
+    option "--generate-token", :flag, "Generate new node token"
+
     option ["-l", "--label"], "LABEL", "Node label", multivalued: true
+    option "--clear-labels", :flag, "Clear node labels"
```
jakolehm: Are we using `--clear-xyz` somewhere else? Just check that this is consistent with other commands.

docs/using-kontena/nodes.md
```diff
@@ -74,3 +122,51 @@ Nodes labeled as `ephemeral=yes` will automatically be removed by the Kontena Ma
...
+
+The server will generate a new node token by default, if not using `--token` to supply a pre-generated node token.
+
+#### List Nodes
+
+```
+$ kontena node list
+NAME        VERSION     STATUS    INITIAL   LABELS
+⊛ core-01   1.4.0.dev   online    1 / 1     test
```
jakolehm: Probably docs should use something else than `dev` versions?

server/app/middlewares/websocket_backend.rb
jakolehm: IMHO we should keep node_id unique for now. That said it would make sense to add that unique index there to block possible race conditions. Name should be also unique within a grid.

server/app/views/v1/host_nodes/token.json.jbuilder
jakolehm: `@node.to_path` is better... but it's a bit confusing to return the same id as in node resource. Maybe it's fine here because token is "embedded" resource for a node.

server/app/middlewares/websocket_backend.rb
SpComb: Fixed to use a (sparse) unique index for the `node_id`, and validate that the node token and ID both matches the same node.

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,104 @@ def call(env)
...
+
+      if initializing_node
+        logger.info "new node #{node} connected using node token with node_id #{node_id}"
+
+        node.reload
+
+      else
+        logger.warn "new node #{node} connected using node token with node_id #{node_id}, but the node token was already used by #{node} with node ID #{node.node_id}"
+
+        raise CloseError.new(4003), "Invalid node token, already used by a different node"
```
SpComb: This happens if two different nodes (with different Docker IDs) are both provisioned with the same `kontena node create` -> node token: one of them connects first, and claims the node by updating the `node_id` from `nil` to its own node ID.

The second one connects with a different node ID (which doesn't exist on any other node), and tries to do the `find_one_and_update` on the same node identified by the token, which fails, because the `node_id` is no longer `nil`.

This doesn't help if two nodes are cloned with the same node ID and token: they will both be able to connect, and be detected by the keepalive ping `connected_at` conflict detection.

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,104 @@ def call(env)
...
+      end
+
+    elsif node.id == node_by_id.id
+      logger.debug "node #{node} connected using node token with node_id #{node_id}"
+
+    else
+      logger.warn "node #{node} connected using node token with node_id #{node_id}, but that node ID already exists for #{node_by_id}"
+
+      raise CloseError.new(4006), "Invalid node ID, already used by a different node"
```
SpComb: This happens if the same node, or two cloned nodes that have the same Docker ID, both connect with unique `kontena node create` -> node tokens: one of them connects first, and claims the node ID but setting it on the node identified by the token.

The second node to connect will find its own node by the token, and then find the other node by the node ID. The two will have different BSON ID:s, and this will fail.

There's a brief race condition window between two such nodes connecting at the same time, and both of them lookup their own node ID before doing the `find_one_and_update`. The second one go down in flames with a `Mongo::Error::OperationFailure`, like in the grid token case.

server/app/middlewares/websocket_backend.rb
```diff
@@ -67,29 +75,104 @@ def call(env)
...
-      ws.close(4001, "Invalid grid token")
-      return
+    grid = Grid.find_by(token: grid_token.to_s)
+
+    raise CloseError.new(4001), "Invalid grid token" unless grid
+
+    node = grid.host_nodes.find_by(node_id: node_id)
+
+    if !node
+      node = grid.host_nodes.create!(node_id: node_id, **init_attrs)
```
SpComb: There's a brief race condition window between the `grid.host_nodes.find_by(node_id: node_id)` and here, if two agents connect with the same node ID at the same time. With the unique index on the `node_id`, this will fail with a `Mongo::Error::OperationFailure`.

That error gets logged on the server, but it kills the `WebsocketBackend#on_open`, and the agent will presumably just be left hanging without seeing any error or response from the server :disappointed:

cli/lib/kontena/cli/nodes/update_command.rb
SpComb: The `kontena grid update` uses `--no-default-affinity` for something similar, should I change this to `kontena node update --no-labels`? The `--clear` just sounds more logical in this case.

server/app/middlewares/websocket_backend.rb
SpComb: Note that this doesn't prevent creating a node with `kontena node create`, and then configuring the agent with a grid token instead of the node token: the newly created node won't have any `node_id` yet, so the server will just create a new node for the agent. If the two nodes happen to have the same name, then the newly created node will get renamed by the server. The node created using `kontena node create` will just remain disconnected.

This only does anything if a node is provisioned using `kontena node create` + `KONTENA_NODE_TOKEN` and connects, and then later gets changed to use a grid token instead - that connection will get rejected.

server/app/views/v1/host_nodes/token.json.jbuilder
SpComb: Done. I kinda want to change the `.../health` JSON `id` too, but not here.

server/app/models/host_node.rb
```diff
@@ -45,9 +47,10 @@ class HostNode
   after_save :reserve_node_number, :ensure_unique_name
 
   index({ grid_id: 1 })
-  index({ node_id: 1 })
+  index({ node_id: 1 }, { unique: true, sparse: true })
```
SpComb: Do these kinds of changes to indexes actually get applied to the mongodb, or does this need a migration?

server/app/middlewares/websocket_backend.rb
jakolehm: Maybe dlock could help here?

server/app/models/host_node.rb
jakolehm: It needs a migration.

server/app/middlewares/websocket_backend.rb
SpComb: I doubt it's worth it, the race condition is very theoretical, and agents are supposed to use unique node IDs anyways. It's a bad race condition in the current code because it would result in invalid nodes with duplicate IDs being created in mongo, which would cause all kinds of weird behavior with the WebsocketBackend / RPC handler lookups.

Less of a problem with the unique index now, because it won't cause persistent effects. Missing error handling for a very unlikely race condition in a misconfiguration scenario = not worth fixing really.

SpComb: The fix would be for the websocket `open` handler to behave better in case of unexpected errors, not leave the connection hanging.

server/app/models/host_node.rb
SpComb: Added. Need to just drop the old index and re-create with the new options, or it fails:

```
I, [2017-06-30T07:38:48.892181 #243]  INFO -- Mongodb::Migrator: migrating HostNodeIndexes exception: Index with name: node_id_1 already exists with different options (85)
```

server/app/middlewares/websocket_backend.rb
SpComb: > Name should be also unique within a grid.

Agreed, the current `after_save :ensure_unique_name` is rather odd, but I don't expect that to cause persistent inconsistencies, just temporary ones. The `HostNodes::Create` validation on the name should help avoid any issues... worth fixing, but not here.

SpComb:
TODO: have the agent prefer `KONTENA_NODE_TOKEN` over `KONTENA_TOKEN` if both are provided. The `KONTENA_TOKEN` shouldn't be used if the node has a node token (results in duplicate `created` + `initializing` nodes, or websocket open errors if the agent has already connected earlier with a node ID + node token).

cli/lib/kontena/cli/nodes/update_command.rb
```diff
@@ -3,16 +3,35 @@ class UpdateCommand < Kontena::Command
...
+      spinner "Updating node #{self.node.colorize(:cyan)} token" do
+        client.put("nodes/#{current_grid}/#{self.node}/token", data)
+      end
+    end
 
     def execute
-      require_api_url
-      require_current_grid
-      token = require_token
+      self.update_token if self.token or self.generate_token?
```
jakolehm: Cli should prompt and warn user here.. once user has accepted side-effect and token is updated cli should say some guide how to proceed to fix connection to master.

SpComb: I'm also changing the `POST /v1/nodes/:id/token` to reset the websocket connection and kick the agent offline... all the side-effects makes me think that this should really be a separate command.

SpComb: Done... the `kontena node update` usage now looks like this...

```
Usage:
    kontena node update [OPTIONS] NODE

Parameters:
    NODE                          Node name

Options:
    --grid GRID                   Specify grid to use
    -l, --label LABEL             Node label
    --clear-labels                Clear node labels
    --token TOKEN                 Node token
    --generate-token              Generate new node token
    --[no-]reset-connection       Reset agent websocket connection (default: true)
    --force                       Force token update
    -D, --debug                   Enable debug (default: $DEBUG)
```

Time to split the token out as a separate command?

SpComb: Split out the `PUT /v1/nodes/:id/token` as `kontena node reset`, with appropriate confirm prompt:

```
> Resetting the node token will disconnect the agent (unless using --no-reset-connection), and require you to reconfigure the kontena-agent > Resetting the node token will disconnect the agent (unless using --no-reset-connection), and require you to reconfigure the kontena-agent > Resetting the node token will disconnect the agent (unless using --no-reset-connection), and require you to reconfigure the kontena-agent using the new `kontena node env` values before it will be able to reconnect. Are you sure? Yes
    *   Resetting node core-02 websocket connection token .. \DEBUG Excon opts: {:omit_default_port=>true, :connect_timeout=>5, :read_timeout=>30, :write_timeout=>5, :ssl_verify_peer=>true, :instrumentor=>Kontena::DebugInstrumentor}
DEBUG [Request]: PUT http://localhost:9292/v1/nodes/development/core-02/token | Headers: {Accept: application/json, Content-Type: application/json, Authorization: Bearer}  | Body: 
{"reset_connection":true}
    *   Resetting node core-02 websocket connection token .. |DEBUG [Response]: Headers: {Content-Type: application/json, X-Kontena-Version: 1.4.0.dev}  | Status: 200 | Body: 
{"id":"development/core-02","token":"lkQL...8g=="}
 [done] Resetting node core-02 websocket connection token     
```

cli/lib/kontena/cli/nodes/reset_command.rb
```diff
@@ -0,0 +1,31 @@
...
+    include Kontena::Cli::GridOptions
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    parameter "NODE", "Node name"
+
+    option ["--token"], "TOKEN", "Use given node token instead of generating a random token"
+    option ["--clear-token"], :flag, "Clear token"
```
SpComb: This is intended as a rollback when unable to reconfigure a pre-existing node to use the node token after using `kontena node reset` to generate a node token.

PR #2511, opened 2017-06-22
Enhanced HostNode status, health

### LINKED ISSUE #2510: CLI: node list sorts nodes incorrectly
The `kontena node list` command should sort nodes by `node_number`, such that initial nodes are first. That does not happen:

```
NAME        VERSION     STATUS   INITIAL   LABELS
⊛ core-02   1.4.0.dev   online   -         test=label
⊛ core-01   1.4.0.dev   online   1 / 1     test
```

This is because it uses [`grid_nodes.sort { |n| n['node_number'] }`](https://github.com/kontena/kontena/blob/master/cli/lib/kontena/cli/nodes/list_command.rb#L73), whereas `Array#sort` expects a `{ |a, b| a <=> b}` block. It should use `sort_by` instead.

The CLI specs `OutputHelpers#output_table` matcher is also broken, so the specs are still passing even though they should catch this.

### LINKED ISSUE #2558: Difficult to get rid of a node that is missing a name and stuck as (initializing)
When a new agent first connects, the server will create a new `HostNode` with the node ID sent by the agent, but without any `name`. These nodes without any name show up in `kontena node list` as `(initializing)`:

```
NAME               VERSION     STATUS   INITIAL   LABELS
⊛ core-01          1.4.0.dev   online   1 / 1     test
⊛ core-02          1.4.0.dev   online   -         test=label,test=2,ephemeral=true
⊛ (initializing)               online   -
```

If the agent never manages to send the initial node info update, then the node will be stuck in that state. While the server API allows deleting such nodes using `DELETE /v1/nodes/:grid/:node_id` (vs the typical `DELETE /v1/nodes/:grid/:name`), and the cli `kontena node remove` command also accepts such a node ID as an alternative to the name, figuring out the node ID is very difficult - the only way is probably using `DEBUG=api kontena node list`, and picking it out from the raw JSON.

* Fix `kontena node list` to show the node ID instead of the name, if missing?
* Change the agent to send the hostname as a `Kontena-Node-Name` header in the initial websocket connect, so that the server can immediately create the new `HostNode`s with a name already in place, without needing to wait for the node info update RPC?

SpComb:
## Changes

* Add a new `HostNode.updated` field, which is set to `false` when a node connects, and `true` when it sends the `/nodes/update` RPC

    This is used to determine if the `HostNode` attributes are out of date or invalid, particularly when a new node is connecting for the first time.

* Add a new `HostNode.disconnected_at` field, which is used together with `connected_at` to show node status uptime/downtime

* Add a new `HostNode.connection_error, connection_error_code` fields, which are used to store the last websocket close reason when disconnecting / rejecting a connection

* Rewrite `/v1/nodes/:id/health` to call `HostNodes::HealthCheck`, with better error handling for offline nodes / RPC errors

    Also includes the websocket connection error, if the node is offline.

* Change `kontena node list` to show the online/offline/connecting states with a since-duration

* Change `kontena node list` to show initializing/connecting nodes with the `node_id`, if they don't have any `name` yet (fixes #2558)

* Change `kontena node health` to query the `/v1/nodes/:id/health` API, showing both websocket connection and etcd health errors

### `HostNode.status`

The new `updated` field is used to determine a new nodes API `{"status": ...}` field, which is used for the CLI `kontena node list`:

* `created`: The `HostNode` has been created using `kontena node create`, but no agent has connected using the node token (he `node_id` is still `nil`)

    When the node connects, the node status will first briefly become `offline` as soon as the server `WebsocketBackend#find_node` sets the `node_id`. Once the `defer { NodePlugger#plugin! ... }` runs, this quickly becomes `connecting`.

* `offline`: The agent has connected but the `NodePlugger#plugin!` has not yet run, or the agent has disconnected, and the `NodeUnplugger#unplug!` has marked it as disconnected

    The node attributes may or not be be valid.

   When the `NodePlugger#plugin!` runs, the node status will become `connecting`.

* `connecting`: The agent has connected (`WebsocketBackend` -> `NodePlugger#plugin!`), but the node has not yet sent any `/nodes/update` RPC, so some fields may still be missing or out of date

    The node has not yet sent its node info, so the node attributes may not be valid or up to date.

    The node status will become `online` if it sends the node info, or `offline` if it disconnects without sending any node info.

* `online`: The agent has connected, and sent the node info, so the `HostNode` fields should be accurate

    The node has sent its node info, so the node attributes should be valid, and up to date.

    When the agent disconnects, the node status will become `offline`.

2017-07-12-154116_674x545_scrot

### `HostNode.websocket_error`

The `connected_at`/`disconnected_at` and `connection_error`/`connection_error_code` can be used to show why an agent was disconnected, or its connection was rejected on open:

* `Websocket disconnected at 2017-07-12 12:52:37 UTC with code 1006: `
* `Node core-02 connection error: Websocket disconnected at 2017-07-12 12:53:05 UTC with code 4030: ping timeout after 5.00s`
* `Node core-02 connection error: Websocket connection rejected at 2017-07-12 12:28:10 UTC with code 4010: agent version 1.4.0.dev.x is not compatible with server version 1.4.0.dev`

This does not include errors like `Invalid grid token, node was created using a node token`, because those are raised before the `node` is set...

### CLI `kontena node list`

Use the new node `status`, `connected_at`/`disconnected_at` fields:

2017-07-12-163719_851x480_scrot

### CLI `kontena node health`

Use the `/v1/nodes/.../health` API to show more health detains, including websocket errors and etcd health:

```
⊛ Node is online for 46m21s
⊛ Node core-01 etcd is healthy
```

```
⊝ Node development/core-02 connection error: Websocket disconnected at 2017-08-16 14:50:46 UTC with code 1006: 
```

```
⊛ Node is online for 5m45s
⊗ Node core-02 etcd is unhealthy: Failed to open TCP connection to 172.17.0.1:2379 (Connection refused - connect(2) for "172.17.0.1" port 2379)
```

```
⊛ Node is online for 4m24s
⊗ Node core-02 etcd is unhealthy: proxy: zero endpoints currently available
```


cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -70,7 +81,7 @@ def node_data
           end
         end
 
-        grid_nodes.sort { |n| n['node_number'] }
+        grid_nodes.sort_by { |n| n['node_number'] }
```
SpComb: Fixes  #2510

cli/spec/support/output_helpers.rb
```diff
@@ -52,23 +54,25 @@ module OutputHelpers
         line = 0
         @real.zip(@expected) do |real, expected|
           line += 1
-          error = values_match?(real.split(/\S+/), expected)
```
SpComb: @kke I don't think `values_match?` was actually returning an error String here... not 100% sure, but I think it returns a "good/bad" boolean...?

Also, the `split(/\S+/)` also resulted in an Array containing only whitespace strings:

```
       line 1: expected [":warning node-1", "1.1-dev", "online", "1 / 3", "-"], received ["", " ", "                      ", "         "]
       line 2: expected [":warning node-2", "1.1-dev", "online", "2 / 3", "-"], received ["", " ", "           ", "            ", " ", " ", "     "]
       line 3: expected [":offline (initializing)", "", "offline", "-", "-"], received ["", " ", "           ", "            ", " ", " ", "     "]
```

cli/spec/support/output_helpers.rb
```diff
@@ -52,23 +54,25 @@ module OutputHelpers
...
-          if error
-            @error ||= ""
-            @error << "\n#{error}"
-            @error.strip!
+          fields = real.split(/\s{2,}/)
```
SpComb: @kke is this a reasonable way to split the tty-table output fields?

cli/lib/kontena/cli/nodes/list_command.rb
SpComb: Split this change out to #2512

cli/spec/support/output_helpers.rb
SpComb: Split this out to #2512

cli/spec/support/output_helpers.rb
SpComb: Split this out to #2512

cli/spec/kontena/cli/nodes/list_command_spec.rb
```diff
@@ -26,14 +26,18 @@
...
-
+      context "with a single online node" do
         before do
           allow(client).to receive(:get).with('grids/test-grid/nodes').and_return(
             {
               'nodes' => [
                 {
+                  "id" => 'testAAA',
                   "connected" => true,
+                  'connected_at' => (Time.now - 50.0).to_s, # XXX: flaky spec if this rolls over
```
SpComb:
```
  1) Kontena::Cli::Nodes::ListCommand#show_grid_nodes For a initial_size=3 grid with two online nodes and one offline node outputs two nodes with warning and one offline

     Failure/Error:

       expect{subject.run([])}.to output_table [

         [':warning node-1', '1.1-dev', 'online 50s',  '1 / 3', '-'],

         [':warning node-2', '1.1-dev', 'online 10s',  '2 / 3', '-'],

         [':offline node-3', '1.1-dev', 'offline 10s', '3 / 3', '-'],

       ]

       on line 1:

        expected: [":warning node-1", "1.1-dev", "online 50s", "1 / 3", "-"]

        received: [":warning node-1", "1.1-dev", "online 51s", "1 / 3", "-"]

       on line 2:

        expected: [":warning node-2", "1.1-dev", "online 10s", "2 / 3", "-"]

        received: [":warning node-2", "1.1-dev", "online 11s", "2 / 3", "-"]

       on line 3:

        expected: [":offline node-3", "1.1-dev", "offline 10s", "3 / 3", "-"]

        received: [":offline node-3", "1.1-dev", "offline 11s", "3 / 3", "-"]

     # ./spec/kontena/cli/nodes/list_command_spec.rb:241:in `block (5 levels) in <top (required)>'
```

SpComb:
Simplified the `status` by forgetting about the `initializing` status, newly connecting nodes will just briefly be `offline` without having any fields set yet.

SpComb:
Rebased against git master after #2504

server/app/middlewares/websocket_backend.rb
```diff
@@ -213,17 +212,16 @@ def on_open(ws, req)
     EM.defer { Agent::NodePlugger.new(node).plugin! connected_at }
 
   rescue CloseError => exc
-    logger.warn "reject websocket connection: #{exc}"
+    logger.warn "reject websocket connection for node #{node || node_id || '<nil>'}: #{exc}"
     ws.close(exc.code, exc.message)
 
+    # XXX: this only applies to the version/clock skew errors, not the token -> node errors
+    Agent::NodePlugger.new(node).reject!(connected_at, exc.code, exc.message) if node
```
SpComb: This needs to use something like `min(connected_at, Time.now)` for the case of a agent sending a `connected_at` timestamp in the future, to avoid bumping the `HostNode.connected_at` to an incorrect timestamp. This would prevent the agent from re-connecting with the correct timestamp, until the actual wallclock time catches up with the invalid `connected_at` timestamp in the future...

cli/lib/kontena/cli/helpers/health_helper.rb
```diff
@@ -49,5 +49,49 @@ def node_health(node, grid_health)
         return node['connected'] ? :ok : :offline
       end
     end
+
+    # @return [Symbol]
+    # @return [String]
+    def node_etcd_health(node_health)
```
jakolehm: Probably not good idea to pollute helper with resource specific helpers? Although we have done this in the past...

docs/using-kontena/nodes.md
```diff
@@ -148,10 +148,10 @@ The server will generate a new node token by default, if not using `--token` to
...
 ```
 $ kontena node list
-NAME        VERSION     STATUS    INITIAL   LABELS
-⊛ core-01   1.4.0       online    1 / 1     provider=vagrant
-⊛ core-02   1.4.0       online    -         provider=vagrant
-⊝ core-03   1.4.0       offline   -         -
+NAME        VERSION     STATUS       INITIAL   LABELS
+⊛ core-01   1.4.0       online       1 / 1     provider=vagrant
+⊛ core-02   1.4.0       online       -         provider=vagrant
+⊝ core-03   1.4.0       connecting   -         -
```
jakolehm: Is this missing timing info?

server/app/models/host_node.rb
```diff
@@ -31,6 +31,10 @@ class HostNode
   field :agent_version, type: String
   field :docker_version, type: String
   field :connected_at, type: Time
+  field :disconnected_at, type: Time
+  field :updated, type: Boolean, default: false # true => node sent /nodes/update after connecting; false => node attributes may be out of date even if connected
+  field :connection_error_code, type: Integer
```
jakolehm: Should we combine connection error code & message to single field (hash or embedded object)?

cli/lib/kontena/cli/helpers/health_helper.rb
SpComb: It's used for both `kontena node health` and `kontena etcd health`, so it makes sense here... OTOH the `time_since` should really be elsewhere, it's just stuffed in here because it's used for both `kontena node list` and `kontena node health`.

cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -19,10 +19,19 @@ def node_name(node, grid)
...
-        node['availability'] == 'drain' ? pastel.yellow('drain') : pastel.green('online')
+      case node_status = node['status']
+      when 'created'
+        "#{pastel.dark('created')} #{time_since(node['created_at'], terse: true)}"
+      when 'connecting'
+        "#{pastel.cyan('connecting')} #{time_since(node['connected_at'], terse: true)}"
+      when 'online'
+        "#{pastel.green('online')} #{time_since(node['connected_at'], terse: true)}"
+      when 'drain'
+        "#{pastel.yellow('drain')} #{time_since(node['connected_at'], terse: true)}"
```
SpComb: The `time_since` here is maybe misleading, perhaps we should just remove it

```
NAME        VERSION     STATUS      INITIAL   LABELS
⊛ core-01   1.4.0.dev   online 4m   1 / 1     test
⊛ core-02   1.4.0.dev   drain 9s    -         test=label
```

jnummelin: yeah, maybe that can be shown in the node details (`kontena node show xyz`)

server/app/middlewares/websocket_backend.rb
```diff
@@ -172,30 +173,28 @@ def find_node(req)
   # @param [Faye::WebSocket] ws
   # @param [Rack::Request] req
   def on_open(ws, req)
+    node_id = req.env['HTTP_KONTENA_NODE_ID']
+    agent_version = req.env['HTTP_KONTENA_VERSION'].to_s
+    connected_at = req.env['HTTP_KONTENA_CONNECTED_AT']
+    connected_at = Time.parse(connected_at) if connected_at
```
jnummelin: Wouldn't this leave `connected_at` as `nil` if the `HTTP_KONTENA_CONNECTED_AT` is not set in request? And having it nil would probably cause some side effects. :) Is it possible that this is not set in the request?

server/app/models/host_node.rb
```diff
@@ -110,10 +113,58 @@ def attributes_from_docker(attrs)
...
+    end
+  end
+
+  # @return [String]
+  def websocket_error
+    if self.connected
+      return nil
+    elsif !self.websocket_connection
+      return "Websocket is not connected"
+    elsif !self.disconnected_at || self.connected_at > self.disconnected_at
```
jnummelin: Feels a bit dirty to determine state from timestamps.

server/app/views/v1/host_nodes/health.json.jbuilder
```diff
@@ -3,8 +3,14 @@ json.name @node.name
...
+json.status @node.status
+json.connected_at @node.connected_at
+json.disconnected_at @node.disconnected_at
 json.connected @node.connected
+
 json.etcd_health do
-  json.health @etcd_health[:health]
-  json.error @etcd_health[:error]
+  json.health @node_health.dig(:etcd_health, :health)
+  json.error @node_health.dig(:etcd_health, :error)
```
jnummelin: Isn't `error` always `nil` per https://github.com/kontena/kontena/pull/2511/files#diff-7f8c6dbe4b176e8af2a941184ced8569R26

jnummelin:
Is the state diagram still up-to-date? I think it would be nice addition to docs also.

server/app/views/v1/host_nodes/health.json.jbuilder
SpComb: The agent can return an `{error: ...}`: https://github.com/kontena/kontena/blob/master/agent/lib/kontena/rpc/etcd_api.rb#L62

This kind of "agent got an etcd API error" is distinct from "server got an agent RPC error".

jnummelin: true, misread a bit on the different level of error fields.

SpComb: It's not entirely not confusing.

server/app/mutations/host_nodes/health_check.rb
```diff
@@ -0,0 +1,36 @@
...
+      @rpc_client ||= self.host_node.rpc_client(10)
+    end
+
+    def query_etcd_health
+      begin
+        return rpc_client.request("/etcd/health")
+      rescue RpcClient::TimeoutError => error
+        add_error(:etcd_health, :timeout, error.message)
+
+        return {health: nil, error: nil}
```
SpComb: This ends up getting ignored after the `add_error`, and the `{error: ...}` here is misleading.

cli/lib/kontena/cli/nodes/list_command.rb
SpComb: Not really, there's no timestamp for how long the node has been draining?

Fixed to just show `drain` without any duration.

cli/lib/kontena/cli/helpers/health_helper.rb
SpComb: Moved the `time_since` to a separate `Kontena::Cli::Helpers::TimeHelper`.

The `node_etcd_health` is still there in the `HealthHelper`, because I think that's the right place for it.

server/app/middlewares/websocket_backend.rb
SpComb: Yes, because it's now used in the `rescue CloseError` that can be raised by the agent version check, and that can be triggered by pre-1.2 agents... changed this to just explicitly initialize it to nil.

server/app/models/host_node.rb
SpComb: Yes, particularly since the `connected_at` is from the agent clock, and the `disconnected_at` is from the server clock...

This used to differentiate between the close_code/reason set in `WebsocketBackend#on_open` -> `Agent::NodePlugger#reject!` vs `WebsocketBackend#on_close` -> `Agent::NodeUnplugger#unplug!`... I could add some new `self.websocket_connection.opened` flag instead, set to false if the open was rejected?

server/app/views/v1/host_nodes/health.json.jbuilder
SpComb: Changed this yet again to avoid the `@node_health` vs `@node_errors` duality - the `health.json` template is only used if the server is able to query the node health.

Offline nodes and RPC timeouts how result in HTTP 422 responses with a standard `{error: ...}` hash from the `GET /v1/nodes/:id/health` API.

SpComb:
> Is the state diagram still up-to-date? I think it would be nice addition to docs also.

Yes, it is. However, it's generated from a dot file that I had open in some online graphviz editor... closed that tab already. Not sure where in the docs it would make sense to have that kind of diagram... in the `using-kontena/nodes.md`, or some new `references/node-status.md`?

server/app/models/host_node.rb
SpComb: Added a `HostNode.websocket_connection.opened`... WDYT?

server/app/middlewares/websocket_backend.rb
```diff
@@ -183,30 +184,29 @@ def find_node(req)
   # @param [Faye::WebSocket] ws
   # @param [Rack::Request] req
   def on_open(ws, req)
+    node_id = req.env['HTTP_KONTENA_NODE_ID']
+    connected_at = nil
+
     node = find_node(req)
 
+    send_master_info(ws)
```
SpComb: Always sending the `/agent/master_info` this early has a slight effect on the agent behavior in terms of when the `websocket:connected` notification is emitted. Currently the server might close the websocket with a 4010 invalid version websocket close immediately after the master_info, but the agent will shut down in this case. Now the agent will also get the master_info immediately before the 4020 invalid clock websocket close, and the agent will not shutdown, but reconnect instead.

AFAIK this just means that the agent log worker would bounce a lot if the agent has an invalid clock.

jnummelin:
> Not sure where in the docs it would make sense to have that kind of diagram... in the using-kontena/nodes.md, or some new references/node-status.md?

I'd put it in `using-kontena/nodes.md`

jnummelin:
Tested the clock-skew case by making agent clock go ahead of time. :) This does cause quite a bit of CPU burning as there's lot of errors logged from RpcClient also:
```
kontena-agent | W, [2017-08-30T05:29:11.225343 #1]  WARN -- Kontena::WebsocketClient: not connected (RuntimeError)
kontena-agent | /app/lib/kontena/websocket_client.rb:211:in `ws'
kontena-agent | /app/lib/kontena/websocket_client.rb:248:in `send_request'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
kontena-agent | /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
```
Is this because agent still gets master info before the connection reject and something starts to send Rpc messages already?

cli/lib/kontena/cli/helpers/health_helper.rb
```diff
@@ -49,5 +49,17 @@ def node_health(node, grid_health)
         return node['connected'] ? :ok : :offline
       end
     end
+
+    # @param node_etcd_health [Hash{health: String, error: String}]
+    # @return [Symbol, String]
+    def node_etcd_health(node_etcd_health)
+      if node_etcd_health['health']
```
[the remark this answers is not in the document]
SpComb: Well, it's a boolean in the API and RPC protocol schema, but I see what you mean, because the actual etcd `/health` API seriously returns `{"health": "false"}` as a string...

Not sure what to do about it though... the agent `Etcd::Health#health` validates it.

server/app/middlewares/websocket_backend.rb
SpComb: > Is this because agent still gets master info before the connection reject and something starts to send Rpc messages already?

Yes, it will wake up the `LogWorker` and drain the queue, sending async RPC notifications, all of which now fail with a warning. In that specific case of invalid clocks, it used to get the close without the `master_info`.

I guess the best answer is to only explicit send the `master_info` where appropriate: the "invalid version" close (where it also breaks the logging, but it's necessary for the agent auto-upgrade to work) and then for the succesful `plugin!` case. I just wanted to get rid of the code duplication by moving it out of the `NodePlugger`, where it suffers from a race condition in the `handle_invalid_agent_version` case  :/

cli/lib/kontena/cli/helpers/health_helper.rb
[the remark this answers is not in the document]
jnummelin: As said: Not really an issue as there shouldn't be a case where its something else than a boolean, more like an observation of my own wft moments. :)

jnummelin:
Much less noisy now. 👍

jnummelin:
I'm not 100% sure about the status times, is it really something we want to see on the list? 
image


Or should it be only visible in node details? @jakolehm WDYT?

jakolehm:
I think we should _not_ show status time in that list.. it's a bit too noisy and not very clear what it means?

SpComb:
> I'm not 100% sure about the status times, is it really something we want to see on the list?

Intention is to be able to tell at a glance which nodes have recently reconnected. If something odd happened last night, you might notice that e.g. all of your nodes reconnected four hours ago.

jnummelin:
I see the point of having the status "timing" on the list. So let's leave it there, if it seems to cause too much confusion, we can get rid of it easily. :)

cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -70,7 +82,7 @@ def node_data
         unless quiet?
           grid_health = grid_health(grid, grid_nodes)
           grid_nodes.each do |node|
-            node['name'] = health_icon(node_health(node, grid_health)) + " " + (node['name'] || '(initializing)')
+            node['name'] = health_icon(node_health(node, grid_health)) + " " + (node['name'] || node['node_id'])
```
SpComb: This really shouldn't be necessary anymore after #2694 already fixing #2558, but it doesn't hurt either.

PR #2512, opened 2017-06-22
Fix cli node list sorting, specs

### LINKED ISSUE #2510: CLI: node list sorts nodes incorrectly

SpComb:
Fixes #2510
Fixes the `output_lines` matcher used by the `node list` specs - it was broken and passing even though the output lines were wrong.

### DIFF
--- a/cli/lib/kontena/cli/nodes/list_command.rb
+++ b/cli/lib/kontena/cli/nodes/list_command.rb
@@ -70,7 +70,7 @@ module Kontena::Cli::Nodes
           end
         end
 
-        grid_nodes.sort { |n| n['node_number'] }
+        grid_nodes.sort_by { |n| n['node_number'] }
       end
     end
 
--- a/cli/spec/support/output_helpers.rb
+++ b/cli/spec/support/output_helpers.rb
@@ -32,11 +32,13 @@ module OutputHelpers
     supports_block_expectations
 
     match do |block|
+      @errors = []
+
       @expected = lines
       begin
         @real = CaptureStdoutLines.capture(block)
       rescue Exception => error
-        @error = error
+        @errors = [error.to_s]
         return false
       end
 
@@ -52,23 +54,25 @@ module OutputHelpers
         line = 0
         @real.zip(@expected) do |real, expected|
           line += 1
-          error = values_match?(real.split(/\S+/), expected)
-          if error
-            @error ||= ""
-            @error << "\n#{error}"
-            @error.strip!
+          fields = real.split(/\s{2,}/)
+          unless values_match?(fields, expected)
+            @errors << [
+              "on line #{line}:",
+              " expected: #{expected}",
+              " received: #{fields}",
+            ].join("\n")
           end
         end
       else
-        @error = "expected #{@expected.size} lines but got #{@real.size} lines instead:\n"
-        @error += "Expected:\n#{@expected.map(&:inspect).join("\n")}"
-        @error += "Received:\n#{@real.map(&:inspect).join("\n")}"
+        @errors << "expected #{@expected.size} lines but got #{@real.size} lines instead:\n"
+          + " Expected:\n#{@expected.map(&:inspect).join("\n")}"
+          + " Received:\n#{@real.map(&:inspect).join("\n")}"
       end
-      @error.nil?
+      @errors.empty?
     end
 
     failure_message do |block|
-      @error
+      @errors.join "\n"
     end
 
     chain :with_header do |header|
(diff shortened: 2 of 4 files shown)


cli/spec/support/output_helpers.rb
```diff
@@ -52,23 +54,25 @@ module OutputHelpers
         line = 0
         @real.zip(@expected) do |real, expected|
           line += 1
-          error = values_match?(real.split(/\S+/), expected)
-          if error
-            @error ||= ""
-            @error << "\n#{error}"
-            @error.strip!
+          fields = real.split(/\s{2,}/)
```
SpComb: @kke is this a reasonable way to split the tty-table output fields? This just leaves out empty fields (`nil` or `''`), it would be nice to also have those included in `fields`.

kke: Maybe there's some way to trick tty-table to use a formatter that adds some recognizable separators, such as `\0` or `\t`. 

Perhaps that should be an option anyway, something like `-0|--print0` to output a null separated table that would be a lot easier to `cut` too.

cli/spec/support/output_helpers.rb
```diff
@@ -52,23 +54,25 @@ module OutputHelpers
...
+          unless values_match?(fields, expected)
```
kke: This works great 👍

cli/spec/support/output_helpers.rb
```diff
@@ -52,23 +54,25 @@ module OutputHelpers
...
+              " received: #{fields}",
+            ].join("\n")
           end
         end
       else
-        @error = "expected #{@expected.size} lines but got #{@real.size} lines instead:\n"
-        @error += "Expected:\n#{@expected.map(&:inspect).join("\n")}"
-        @error += "Received:\n#{@real.map(&:inspect).join("\n")}"
+        @errors << "expected #{@expected.size} lines but got #{@real.size} lines instead:\n"
+          + " Expected:\n#{@expected.map(&:inspect).join("\n")}"
```
kke: I don't think these `+` lines are actually appended to the message

cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -70,7 +70,7 @@ def node_data
           end
         end
 
-        grid_nodes.sort { |n| n['node_number'] }
+        grid_nodes.sort_by { |n| n['node_number'] }
```
kke: LGTM on the main ingredient here.

cli/spec/support/output_helpers.rb
kke: For now it seems to work (a way better than the previous one, I wonder how it worked at all) as long as no field has two spaces in it.

cli/spec/support/output_helpers.rb
SpComb: Right, didn't test that. Fixed now:

```
  1) Kontena::Cli::Nodes::ListCommand#show_grid_nodes For a initial_size=3 grid with a single node outputs node with error
     Failure/Error:
       expect{subject.run([])}.to output_table [
         [':error node-1', '1.1-dev', 'online', '1 / 3', '-'],
         [':error node-2', '1.1-dev', 'online', '2 / 3', '-'],
       ]

       expected 2 lines but got 1 lines instead:
        Expected:
       [":error node-1", "1.1-dev", "online", "1 / 3", "-"]
       [":error node-2", "1.1-dev", "online", "2 / 3", "-"]
        Received:
       [":error node-1", "1.1-dev", "online", "1 / 3", "-"]
     # ./spec/kontena/cli/nodes/list_command_spec.rb:48:in `block (5 levels) in <top (required)>'
     # ./spec/spec_helper.rb:45:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:43:in `catch'
     # ./spec/spec_helper.rb:43:in `block (2 levels) in <top (required)>'
```

PR #2518, opened 2017-06-26
Fix volume driver mismatches

### LINKED ISSUE #2261: Using Kontena volumes fails with Docker v2 volume plugins
When using Docker volume plugins, Docker uses version/tag in volume driver information
```
$ docker volume inspect s3fs-test
[
    {
        "Driver": "rexray/s3fs:latest",
        "Labels": {},
        "Mountpoint": "/var/lib/docker/plugins/165a4e198e4ff02d1a3f1167cdfe132e046efa45db6a2384e03fdfd4f6e9a381/rootfs",
        "Name": "s3fs-test",
        "Options": {},
        "Scope": "global",
        "Status": {
            "availabilityZone": "",
            "fields": null,
            "iops": 0,
            "name": "s3fs-test",
            "server": "s3fs",
            "service": "s3fs",
            "size": 0,
            "type": ""
        }
    }
]
```

When deploying a stack, `Kontena::Workers::Volumes::VolumeManager` checks exisiting volumes without the version information and the comparison will fail causing:

```
GridServiceInstanceDeployer::ServiceError: Kontena::Workers::Volumes::VolumeManager::DriverMismatchError: Volume driver not as expected. Expected rexray/s3fs, existing volume has rexray/s3fs:latest
```

If you try to create volume with version information `kontena volume create --driver rexray/s3fs:latest ...` Kontena scheduler cannot find a proper node because version/tag information is not used in the comparison.

jnummelin:
This PR makes node filtering to actually use volume plugin/driver version as part of the criteria. (commit by @jakolehm, cherry picked that from a WIP branch). 

This also changes the agents `VolumeManager` to omit the version/tag when checking volume existence as there might be different plugin versions used within the grid.

Fixes #2261

### DIFF
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -111,9 +111,10 @@ module Kontena::Workers::Volumes
       begin
         debug "volume #{volume_name} exists"
         volume = Docker::Volume.get(volume_name)
-        if volume && volume.info['Driver'] == driver
+        drivers_match = drivers_match?(driver, volume.info['Driver'])
+        if volume && drivers_match
           return true
-        elsif volume && volume.info['Driver'] != driver
+        elsif volume && !drivers_match
           raise DriverMismatchError.new("Volume driver not as expected. Expected #{driver}, existing volume has #{volume.info['Driver']}")
         end
       rescue Docker::Error::NotFoundError
@@ -122,6 +123,10 @@ module Kontena::Workers::Volumes
       rescue => error
         abort error
       end
+    end
+
+    def drivers_match?(expected, actual)
+      expected.split(':', 2)[0] == actual.split(':', 2)[0]
     end
 
     def terminate_volumes(current_ids)
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -129,7 +129,7 @@ module Kontena::Workers::Volumes
       if expected.include?(':')
         expected == actual
       else
-        expected.split(':', 2)[0] == actual.split(':', 2)[0]
+        expected == actual.split(':', 2)[0]
       end
     end
 
--- a/agent/lib/kontena/workers/volumes/volume_manager.rb
+++ b/agent/lib/kontena/workers/volumes/volume_manager.rb
@@ -126,7 +126,11 @@ module Kontena::Workers::Volumes
     end
 
     def drivers_match?(expected, actual)
-      expected.split(':', 2)[0] == actual.split(':', 2)[0]
+      if expected.include?(':')
+        expected == actual
+      else
+        expected.split(':', 2)[0] == actual.split(':', 2)[0]
+      end
     end
 
     def terminate_volumes(current_ids)
--- a/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
+++ b/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
@@ -149,6 +149,11 @@ describe Kontena::Workers::Volumes::VolumeManager, :celluloid => true do
       }.to raise_error(Kontena::Workers::Volumes::VolumeManager::DriverMismatchError)
     end
 
+    it 'return true if volume exists with different plugin version' do
+      expect(Docker::Volume).to receive(:get).with('foo').and_return(double(:volume, :info => {'Driver' => 'rexray/s3fs:foobar'}))
+      expect(subject.volume_exist?('foo', 'rexray/s3fs:latest')).to be_truthy
+    end
+
   end
 
   describe '#terminate_volumes' do
--- a/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
+++ b/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
@@ -154,6 +154,11 @@ describe Kontena::Workers::Volumes::VolumeManager, :celluloid => true do
       expect(subject.volume_exist?('foo', 'rexray/s3fs:latest')).to be_truthy
     end
 
+    it 'return true if volume exists with plugin version and requested volume does not specify version' do
+      expect(Docker::Volume).to receive(:get).with('foo').and_return(double(:volume, :info => {'Driver' => 'rexray/s3fs:latest'}))
+      expect(subject.volume_exist?('foo', 'rexray/s3fs')).to be_truthy
+    end
+
   end
 
   describe '#terminate_volumes' do
--- a/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
+++ b/agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
@@ -149,9 +149,9 @@ describe Kontena::Workers::Volumes::VolumeManager, :celluloid => true do
       }.to raise_error(Kontena::Workers::Volumes::VolumeManager::DriverMismatchError)
     end
 
-    it 'return true if volume exists with different plugin version' do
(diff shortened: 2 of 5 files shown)


SpComb:
LGTM on the agent `drivers_match?` part, but I don't think the `Scheduler::Filter::VolumeDriver` change is a good idea...

It will break deployments of any existing volumes created using `kontena volume create --driver foo/bar`, without a version?

It requires volumes to be created using specific `kontena volume create --driver foo/bar:x.y` versions? That seems like a broken feature without some kind of `kontena volume upgrade --driver-version x.z` to deal with driver version upgrades? Also missing validations for `Volumes::Create#driver`...

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -124,6 +125,10 @@ def volume_exist?(volume_name, driver)
       end
     end
 
+    def drivers_match?(expected, actual)
+      expected.split(':', 2)[0] == actual.split(':', 2)[0]
```
SpComb: Hmm.... if the `expected` from the Kontena volume driver actually contains a specific version, then perhaps this should reject volumes with an "incompatible" (?) version? Typically, I would expect the `expected` to just be a `plugin/name` without any version, and then the actual driver version can be ignored like this.

agent/spec/lib/kontena/workers/volumes/volume_manager_spec.rb
```diff
@@ -149,6 +149,11 @@
       }.to raise_error(Kontena::Workers::Volumes::VolumeManager::DriverMismatchError)
     end
 
+    it 'return true if volume exists with different plugin version' do
```
SpComb: Also needs a spec for the problematic `expected: 'rexray/s3fs', actual: 'rexray/s3fs:latest' case?

SpComb:
I think the right fix for both the agent `drivers_match?` and server `Scheduler::Filter::VolumeDriver` is to support an optional version suffix in the kontena volume `--driver`. If the version suffix is there, then match it against the actual driver version, otherwise, omit the actual driver version from both comparisons?

jnummelin:
> I think the right fix for both the agent drivers_match? and server Scheduler::Filter::VolumeDriver is to support an optional version suffix in the kontena volume --driver. If the version suffix is there, then match it against the actual driver version, otherwise, omit the actual driver version from both comparisons?

I agree. Already made couple of commits to reflect that thinking.

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -124,6 +125,14 @@ def volume_exist?(volume_name, driver)
       end
     end
 
+    def drivers_match?(expected, actual)
+      if expected.include?(':')
+        expected == actual
+      else
+        expected.split(':', 2)[0] == actual.split(':', 2)[0]
```
SpComb: Just `expected == actual.split(':', 2)[0]`?

server/app/services/scheduler/filter/volume_plugin.rb
```diff
@@ -30,6 +29,20 @@ def for_service(service, instance_number, nodes)
         nodes
       end
 
+      # Checks if all needed drivers are present on a node
+      # @param [Hash] needed_drivers
+      # @param [HostNode] node
+      def all_drivers_present?(needed_drivers, node)
+        needed_drivers.reject { |nd|
```
SpComb: `needed_drivers.all? { |nd| node.volume_drivers.any? { |vd| ... } }`

SpComb:
The `Volumes::Create` mutation should perhaps also validate the `driver` param with a regex.

SpComb:
:white_check_mark: Tested that all of `--driver local`, `--driver vieux/sshfs` and `--driver vieux/sshfs:latest` work.

server/app/services/scheduler/filter/volume_plugin.rb
jakolehm: I agree with @SpComb, current implementation is too complex.

cli/lib/kontena/cli/nodes/show_command.rb
```diff
@@ -26,7 +26,7 @@ def execute
       puts "  kernel: #{node['kernel_version']}"
       puts "  drivers:"
       puts "    storage: #{node['driver']}"
-      puts "    volume: #{node['volume_drivers'].to_a.map{|v| v['name']}.join(',')}"
+      puts "    volume: #{node['volume_drivers'].to_a.map{|v| [v['name'], v['version']].compact.join(':')}.join(',')}"
```
jakolehm: Imho we should not expose plugin version here to user. It's misleading in too many ways.

jnummelin:
Dropping the version/tag support would mean:
- Master needs to fill in the tag when sending volume info to agent
- volume create mutation needs to validate the volume so that it does not container the tag

Other than that, the implementation reverts pretty close to the original implementation. Even the agent check for driver match would revert back to the original as master would have filled in the correct version for that node.

jnummelin:
obsoleted by #2526

SpComb:
TBH I'm not really comfortable with the operational complexity of trying to understand how the additional logic in #2526 to select the driver version will behave. I'm concerned that it will break if the plugin versions installed on the host nodes change. Based on my own testing, I still think the approach of optional driver versions here is correct, and fits best with how the actual Docker plugins work?

For the typical usecase, where you use builtin drivers (`local`), or use the `latest` version of plugin drivers (`docker plugin install vieux/sshfs` -> `vieux/sshfs:latest`), then you can just use `kontena volume create --driver vieux/sshfs` without any specific version. You can still use `docker plugin install --alias vieux/sshfs:latest vieux/sshfs:1.0`, or `docker plugin upgrade vieux/sshfs vieux/sshfs:next` to manage the installed plugin versions, which upgraded across nodes without needing to re-create the Kontena volumes.

If you have more specific plugin version requirements like `rexray/s3fs:0.8.2`, then you can install those using either the upstream labels like `edge` or `latest` (see [Docker Hub](https://hub.docker.com/r/rexray/s3fs/tags/)), or some local `--alias rexray/s3fs:testing`, and then use `kontena volume create --driver rexray/s3fs:edge`, `kontena volume create --driver rexray/s3fs:testing` etc. You can even have multiple versions of the plugins installed, using e.g. `latest` for your production volumes, and have some separate `testing` label that you `docker plugin upgrade` to the latest release before upgrading the default `latest` plugin?

agent/lib/kontena/workers/volumes/volume_manager.rb
```diff
@@ -124,6 +125,14 @@ def volume_exist?(volume_name, driver)
       end
     end
 
+    def drivers_match?(expected, actual)
+      if expected.include?(':')
+        expected == actual
+      else
+        expected == actual.split(':', 2)[0]
```
SpComb: One optional for expected Kontena volume drivers without a version suffix would be to check that the actual volume driver is either a built-in like `local` without any version suffix, or a plugin with the `latest` version suffix.

SpComb:
> I think we need to remove volume driver version specify. It's too confusing to users and also freezes plugin version (hard to install newer versions to fresh nodes without removing old installation first).

My understanding of the issue with driver versions here is something like the following:

* node-1 `docker plugin install rexray/s3fs:0.8.2`
* node-1 `docker plugin upgrade rexray/s3fs:0.8.2 rexray/s3fs:0.9.1`
* node-2 `docker plugin install rexray/s3fs:0.9.1`

This results in the both nodes running the same 0.9.1 version, but the Docker volume driver on node-1 is still `rexray/s3fs:0.8.2`, whereas on node-2 it is `rexray/s3fs:0.9.1`. There would be no way to create/deploy the same Kontena volume to both nodes in this case.

PR #2526, opened 2017-06-28
Automatically determine volume driver version from installed plugins

jnummelin:
replaces #2518 

Docker plugin versioning causes lot of mismatches. E.g. if you install `rexray/s3fs:0.8.2` and the later on upgrade that to `0.9.2` the plugin is still reported to be on version `0.8.2`.

If Kontena volumes would be created with a "pinned"  plugin version it would be impossible to run multiple versions of a plugin within the grid. Then again, if Kontena would just ignore the version everywhere it would mean that all plugins would need to always be installed with `--alias` to be able to create the actual volumes in first place.

Hence this PR changes things so that:
- Volumes cannot be created with a tag (`:` is not allowed in the name)
- "Fully qualified" driver is sent to agent with both `ServicePod` and `VolumeInstance`. The driver version is looked up when serializing from `HostNode.volume_drivers`. This means that agent creates the volume with the nodes reported driver with version tag. `VolumeManager` driver match check checks fully qualified drivers.

### DIFF
--- a/docs/using-kontena/volumes.md
+++ b/docs/using-kontena/volumes.md
@@ -27,7 +27,7 @@ See volume [scopes](#volume-scopes) for details on the different scopes.
 The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
 
 **NOTE:**
-There should be always only one version of a v2 volume plugin installed on a node at any given time. Having multiple versions installed will make volumes seen multiple times on a node and thus makes the automatic selection of correct volume and driver impossible. It is also advisable to install the plugins with `--alias` as that allows smoother (and less confusing) upgrades of the plugins.
+There should be always only one version of a v2 volume plugin installed on a node at any given time. Having multiple versions installed will make volumes seen multiple times on a node and thus makes the automatic selection of correct volume and driver impossible. If not using the default latest version, then it is advisable to install the plugins with `--alias`, as that allows smoother (and less confusing) upgrades of the plugins.
 
 ### Listing volumes
 
--- a/docs/using-kontena/volumes.md
+++ b/docs/using-kontena/volumes.md
@@ -26,6 +26,9 @@ See volume [scopes](#volume-scopes) for details on the different scopes.
 
 The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
 
+**NOTE:**
+There should be always only one version of a v2 volume plugin installed on a node at any given time. Having multiple versions installed will make volumes seen multiple times on a node and thus makes the automatic selection of correct volume and driver impossible. It is also advisable to install the plugins with `--alias` as that allows smoother (and less confusing) upgrades of the plugins.
+
 ### Listing volumes
 
 ```
--- a/docs/using-kontena/volumes.md
+++ b/docs/using-kontena/volumes.md
@@ -20,16 +20,18 @@ Volumes can be managed from the Kontena CLI using the `kontena volume xyz` comma
 
 ### Creating volumes
 
-`kontena volume create --driver rexray --scope instance my-volume`
+`kontena volume create --driver rexray/s3fs --scope instance my-volume`
 
 See volume [scopes](#volume-scopes) for details on the different scopes.
+
+The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` for is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
 
 ### Listing volumes
 
 ```
 $ kontena volume ls
 
-NAME                      SCOPE                     DRIVER                    CREATED AT               
+NAME                      SCOPE                     DRIVER                    CREATED AT
 redis-data                grid                      local                     2017-04-06T06:57:34.374Z
 test-s3fs                 grid                      rexray/s3fs               2017-04-05T13:06:26.252Z
 ```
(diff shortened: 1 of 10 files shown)


server/app/models/volume.rb
```diff
@@ -39,4 +39,13 @@ def services
     self.grid.grid_services.where("service_volumes.volume_id" => self.id)
   end
 
+  def driver_for_node(host_node)
```
jakolehm: Missing specs (and docs) ... also wondering if we really need this middleman method here.

jnummelin: The middleman is there since `ServicePodSerializer` does not have access to the volume instance.

jnummelin: The middleman is there since the same logic is needed in two places, `ServicePodSerializer` and in `VolumeSerializer`

docs/using-kontena/volumes.md
```diff
@@ -20,16 +20,18 @@ Volumes can be managed from the Kontena CLI using the `kontena volume xyz` comma
 
 ### Creating volumes
 
-`kontena volume create --driver rexray --scope instance my-volume`
+`kontena volume create --driver rexray/s3fs --scope instance my-volume`
 
 See volume [scopes](#volume-scopes) for details on the different scopes.
 
+The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` for is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
```
jakolehm: `i.e. rexray/s3fs:latest for is **not** supported` -> `i.e. rexray/s3fs:latest is **not** supported` ?

SpComb:
> "Fully qualified" driver is sent to agent with both ServicePod and VolumeInstance. The driver version is looked up when serializing from HostNode.volume_drivers. This means that agent creates the volume with the nodes reported driver with version tag. VolumeManager driver match check checks fully qualified drivers.

Do we have to qualify the driver with the version tag when creating the volume on the agent? Can't we just omit the version tag from the volume driver when creating it, and ignore the volume driver version tag when checking the existing volume?

jnummelin:
> Can't we just omit the version tag from the volume driver when creating it
Not really:
```
$ docker plugin ls
ID                  NAME                DESCRIPTION               ENABLED
7fb5f9f88a95        rexray/s3fs:0.8.2   REX-Ray for Amazon S3FS   true

$ docker volume create --driver rexray/s3fs jussi-s3fs-test
Error response from daemon: create jussi-s3fs-test: create jussi-s3fs-test: Error looking up volume plugin rexray/s3fs: legacy plugin: plugin not found
```
The "unqualified" driver only works for either alias or `latest` version.

jakolehm:
@SpComb world would be much better place if one could just install everything with `latest` (or `edge`). In reality users have to install specific versions because they don't want to have surprises or newer version "just does not work". This brings us to a point where we cannot really have proper ux if we require user to specify those versions.

We can always add this "feature" later but if we now decide to add it, we don't have luxury to take it off later.

SpComb:
> @SpComb world would be much better place if one could just install everything with latest (or edge). In reality users have to install specific versions because they don't want to have surprises or newer version "just does not work". This brings us to a point where we cannot really have proper ux if we require user to specify those versions.

In terms of the actual Docker plugin install, yes, installing specific remote version tags per `docker plugin install rexray/s3fs:0.8.2` without any local `--alias` results in crazyness later when upgrading plugin versions, because the resulting local `0.8.2` version tag never changes, even though you can upgrade the plugin itself to a newer version. This isn't a problem if you install from upstream release tags like some hypothetical `rexray/s3fs:stable`, or supply your own using `--alias`, and those can be the same across all nodes in a grid.

If we really want to let users provision nodes using `docker plugin install` for specific non-`latest` version tags without using `--alias`, such that different nodes will have different local plugin version labels that don't match the actual upgraded plugin versions, then this PR makes sense... strictly assuming that nodes will never have multiple plugin versions installed, and that different versions of the same plugin are always compatible with eachother.

> We can always add this "feature" later but if we now decide to add it, we don't have luxury to take it off later.

Right, volumes using `--driver rexray/s3fs:0.8.2` are broken enough ATM per #2261 that we can deprecate those without breaking existing installs. Just need to document that this assumes only a single plugin version, and undefined behavior if you have multiple plugin versions installed - but that already seems to apply to Docker itself (https://github.com/kontena/kontena/issues/2261#issuecomment-311698390).

We can loosen the `driver` validation restrictions later and add support for explicit version tags per  #2518, and fix this to prefer the `latest` version if there are multiple plugin versions installed.

I would personally just have done these fixes in the other order: the simple #2518 fix for explicit version tags first, and then this more complex #2526 automatic driver version selection later.

docs/using-kontena/volumes.md
```diff
@@ -20,16 +20,18 @@ Volumes can be managed from the Kontena CLI using the `kontena volume xyz` comma
 
 ### Creating volumes
 
-`kontena volume create --driver rexray --scope instance my-volume`
+`kontena volume create --driver rexray/s3fs --scope instance my-volume`
 
 See volume [scopes](#volume-scopes) for details on the different scopes.
 
+The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
```
SpComb: Document the plugin version selection, and specifically, the case of having multiple plugin versions installed. Specifically, that you must only ever install one plugin version.

docs/using-kontena/volumes.md
```diff
@@ -26,6 +26,9 @@ See volume [scopes](#volume-scopes) for details on the different scopes.
 
 The driver has to be specified **without** a version tag, i.e. `rexray/s3fs:latest` is **not** supported. This is because using a hard version info on volume specification will make running multiple volume plugin versions within the grid impossible and also makes upgrading volume plugins really hard. Kontena will match the reported plugin version automatically and thus is able to create and manage the volumes.
 
+**NOTE:**
+There should be always only one version of a v2 volume plugin installed on a node at any given time. Having multiple versions installed will make volumes seen multiple times on a node and thus makes the automatic selection of correct volume and driver impossible. It is also advisable to install the plugins with `--alias` as that allows smoother (and less confusing) upgrades of the plugins.
```
SpComb: If not using the default `latest` version, then it is advisable to install the plugins with `--alias`, as that allows smoother (and less confusing) upgrades of the plugins.

PR #2527, opened 2017-06-28
Fix server .ruby-version

jakolehm:


SpComb:
What is this used for? Heroku?

jakolehm:
Rvm, chruby etc can switch to correct ruby version automatically... so basically this helps with dev environments.

PR #2528, opened 2017-06-28
Add rbtrace support to server & agent

jakolehm:
Includes [rbtrace](https://github.com/tmm1/rbtrace) gem to server & agent that can be enabled with `RBTRACE` environment variable (disabled by default).

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -7,7 +7,7 @@ RUN apk update && apk --update add tzdata ruby ruby-irb ruby-bigdecimal \
 ADD Gemfile /app/
 ADD Gemfile.lock /app/
 
-RUN apk --update add --virtual build-dependencies ruby-dev build-base openssl-dev git && \
+RUN apk --update add --virtual build-dependencies ruby-dev build-base openssl-dev libffi-dev git && \
     gem install bundler --no-ri --no-rdoc && \
     cd /app ; bundle install --without development test && \
     apk del build-dependencies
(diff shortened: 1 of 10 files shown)


jnummelin:
Is there much overhead in practice for enabling this always? The docs are bit vague:
> rbtrace is designed to have minimal overhead, and should be safe to run in production.

I'm just thinking if the actual overhead is minimal why not run it always?

SpComb:
I wonder what the security implications of loading `rbttrace` are... what controls access to the traced process? I'm seeing some `msgget(rbtracer.mqi_key, 0666 | IPC_CREAT)` and a `/tmp/rbtrace-%d.sock`.

They'll be running with a Docker container though, so that will probably be enough to hide them away from any threat model.

jakolehm:
Enabling this by default sounds a bit risky, we can change it later if we can prove that this does not have any side-effects.

jnummelin:
I'm not sure if the GC tracing really works, tracing servers puma process with `rbtrace --pid 16 --gc` gives nothing out. :/ `16` process is both the leader and the only worker that has an agent connected at. The slow method tracing does work though, so `rbtrace` itself seems to be working.

jnummelin:
Has anyone else tested the GC tracing, do I have PEBKAC issue?

jnummelin:
@jakolehm @SpComb Have you tried the GC tracing with this, does it work?

jakolehm:
Rebased.

@jnummelin haven't tried GC tracing.

PR #2257, opened 2017-05-03
Upgrade to Mongoid 5.2

jakolehm:
This upgrade will bring Mongo >= 3.2 compatibility (because moped will change to official mongodb driver).

https://docs.mongodb.com/ruby-driver/master/reference/driver-compatibility/

PR #2534, opened 2017-06-30
Fix mongoid last sort

jakolehm:
Quarantees order when last item is fetched from a collection.

### DIFF
--- a/server/spec/models/host_node_stat_spec.rb
+++ b/server/spec/models/host_node_stat_spec.rb
@@ -11,6 +11,21 @@ describe HostNodeStat do
   it { should have_index_for(host_node_id: 1).with_options(background: true) }
   it { should have_index_for(host_node_id:1, created_at: 1).with_options(background: true) }
 
+  describe '.latest' do
+    let(:grid) { Grid.create!(name: 'test-grid') }
+    let(:node) { HostNode.create!(grid: grid, name: 'test-node') }
+
+    it 'returns latest stat item' do
+      node.host_node_stats.create
+      last = node.host_node_stats.create
+      expect(described_class.latest).to eq(last)
+    end
+
+    it 'returns nil if no stats' do
+      expect(described_class.latest).to be_nil
+    end
+  end
+
   describe 'aggregations' do
     let(:grid) { Grid.create!(name: 'test-grid') }
     let(:other_grid) { Grid.create!(name: 'other-grid') }
--- a/server/app/models/concerns/sortable_stat.rb
+++ b/server/app/models/concerns/sortable_stat.rb
@@ -1,0 +1,9 @@
+module SortableStat
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def latest
+      asc(:id).last
+    end
+  end
+end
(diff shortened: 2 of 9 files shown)


SpComb:
Is it possible to have a rspec for these, or it is hard to repro?

jakolehm:
I'm surprised that our tests did not catch this. It might be a bit hard to reproduce this in tests but I can try.

jakolehm:
Refactored & did some cleanup because linter was showing quite many red lines. @jnummelin @SpComb PTAL.

server/spec/models/host_node_stat_spec.rb
```diff
@@ -11,6 +11,21 @@
...
   it { should have_index_for(host_node_id:1, created_at: 1).with_options(background: true) }
 
+  describe '.latest' do
+    let(:grid) { Grid.create!(name: 'test-grid') }
+    let(:node) { HostNode.create!(grid: grid, name: 'test-node') }
+
+    it 'returns latest stat item' do
+      node.host_node_stats.create
+      last = node.host_node_stats.create
+      expect(described_class.latest).to eq(last)
```
SpComb: Confirm that the spec works, aka even with just the two items, removing the `asc(:id)` allows this to fail:

```
  1) HostNodeStat.latest returns latest stat item
     Failure/Error: expect(described_class.latest).to eq(last)
       
       expected: #<HostNodeStat _id: 5955f959de3578646bd0b488, created_at: 2017-06-30 07:10:17 UTC, load: nil, memory: nil, filesystem: nil, usage: nil, cpu: nil, network: nil, grid_id: nil, host_node_id: BSON::ObjectId('5955f959de3578646bd0b484')>
            got: #<HostNodeStat _id: 5955f959de3578646bd0b487, created_at: 2017-06-30 07:10:17 UTC, load: nil, memory: nil, filesystem: nil, usage: nil, cpu: nil, network: nil, grid_id: nil, host_node_id: BSON::ObjectId('5955f959de3578646bd0b484')>
       
       (compared using ==)
       
       Diff:
       @@ -1,2 +1,2 @@
       -#<HostNodeStat _id: 5955f959de3578646bd0b488, created_at: 2017-06-30 07:10:17 UTC, load: nil, memory: nil, filesystem: nil, usage: nil, cpu: nil, network: nil, grid_id: nil, host_node_id: BSON::ObjectId('5955f959de3578646bd0b484')>
       +#<HostNodeStat _id: 5955f959de3578646bd0b487, created_at: 2017-06-30 07:10:17 UTC, load: nil, memory: nil, filesystem: nil, usage: nil, cpu: nil, network: nil, grid_id: nil, host_node_id: BSON::ObjectId('5955f959de3578646bd0b484')>
       
     # ./spec/models/host_node_stat_spec.rb:21:in `block (3 levels) in <top (required)>'
```

SpComb:
This is probably something that broke with the  #2257 mongoid/mongo-driver upgrade in 1.3.0?

> first and last no longer add an _id sort when no sorting options have been provided. In order to guarantee that a document is the first or last, it needs to now contain an explicit sort.

https://docs.mongodb.com/mongoid/5.2/tutorials/mongoid-upgrade/

Issue #2450, opened 2017-06-08
Support for container limiting options in newer docker


matti:
Bunch of new options in https://docs.docker.com/engine/admin/resource_constraints/

Currently Kontena only supports `cpu_shares`

jakolehm:
Probably `cpus` makes most sense (in addition to some new memory options).

jnummelin:
We just need to be bit careful adding these, CoreOS for example is still on 1.12.6 docker in stable channel :/

PR #2541, opened 2017-07-01
Service cpus limit

jakolehm:
> Specify how much of the available CPU resources a service instance can use. For example, if the host machine has two CPUs and you set `cpus: 1.5`, the service instance will be guaranteed to be able to access at most one and a half of the CPUs. Kontena scheduler will automatically target deployment to nodes that have at least number of specified CPUs available.

Related #2450

### DIFF
--- a/cli/lib/kontena/cli/services/create_command.rb
+++ b/cli/lib/kontena/cli/services/create_command.rb
@@ -16,7 +16,9 @@ module Kontena::Cli::Services
     option ["-v", "--volume"], "VOLUME", "Add a volume or bind mount it from the host", multivalued: true
     option "--volumes-from", "VOLUMES_FROM", "Mount volumes from another container", multivalued: true
     option ["-a", "--affinity"], "AFFINITY", "Set service affinity", multivalued: true
-    option "--cpus", "CPUS", "Number of CPUs"
+    option "--cpus", "CPUS", "Number of CPUs" do |cpus|
+      Float(cpus)
+    end
     option ["-c", "--cpu-shares"], "CPU_SHARES", "CPU shares (relative weight)"
     option ["-m", "--memory"], "MEMORY", "Memory limit (format: <number><optional unit>, where unit = b, k, m or g)"
     option ["--memory-swap"], "MEMORY_SWAP", "Total memory usage (memory + swap), set \'-1\' to disable swap (format: <number><optional unit>, where unit = b, k, m or g)"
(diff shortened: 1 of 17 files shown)


jakolehm:
@SpComb PTAL.

server/app/services/scheduler/filter/cpu.rb
```diff
@@ -0,0 +1,25 @@
...
+      # @return [Array<HostNode>]
+      def for_service(service, instance_number, nodes)
+        return nodes if service.cpus.nil?
+
+        candidates = nodes.dup.select { |n|
+          n.cpus.to_f >= service.cpus.to_f
+        }
+
+        if candidates.empty?
+          raise Scheduler::Error, "Did not find any nodes with #{service.cpus} CPUs available"
```
SpComb: The filtering logic is also a little dubious... `--cpus 4` doesn't imply that the service needs at least four CPUs to run... there's no reservation of CPU resources that would imply that the CPUs are somehow "available" for the service.

If the host node has other CPU-hungry services scheduled, then it won't get it's four CPUs anyways.

SpComb:
@jakolehm waiting for `status/merge`, unless you still want to address the minor comments.

cli/lib/kontena/cli/services/create_command.rb
```diff
@@ -69,6 +70,7 @@ def parse_service_data_from_options
       data[:volumes_from] = volumes_from_list unless volumes_from_list.empty?
       data[:memory] = parse_memory(memory) if memory
       data[:memory_swap] = parse_memory(memory_swap) if memory_swap
+      data[:cpus] = cpus.to_f if cpus
```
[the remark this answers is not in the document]
jakolehm: Good point, fixed.

server/app/services/scheduler/filter/cpu.rb
jakolehm: >  guaranteed to be able to access

SpComb: What do you mean? If you're talking about the [`docker run`](https://docs.docker.com/engine/admin/resource_constraints/#configure-the-default-cfs-scheduler) docs, then your quote is very much out of context:

> the container will be guaranteed to be able to access ***at most*** one and a half of the CPUs

cli/lib/kontena/cli/services/create_command.rb
```diff
@@ -16,6 +16,9 @@ class CreateCommand < Kontena::Command
     option ["-v", "--volume"], "VOLUME", "Add a volume or bind mount it from the host", multivalued: true
     option "--volumes-from", "VOLUMES_FROM", "Mount volumes from another container", multivalued: true
     option ["-a", "--affinity"], "AFFINITY", "Set service affinity", multivalued: true
+    option "--cpus", "CPUS", "Number of CPUs" do |cpus|
+      Float(cpus)
```
SpComb: :+1 `ERROR: option '--cpus': invalid value for Float(): "five"`

Issue #2071, opened 2017-04-05
Race condition on reserve_node_number with concurrent node creates


jakolehm:
From the logs:

```
E, [2017-04-05T10:55:04.134330 #98] ERROR -- WebsocketBackend: Moped::Errors::OperationFailure: The operation: #<Moped::Protocol::Command
api_1      |   @length=83
api_1      |   @request_id=847
api_1      |   @response_to=0
api_1      |   @op_code=2004
api_1      |   @flags=[]
api_1      |   @full_collection_name="kontena_development.$cmd"
api_1      |   @skip=0
api_1      |   @limit=-1
api_1      |   @selector={:getlasterror=>1, "w"=>1}
api_1      |   @fields=nil>
api_1      | failed with error 11000: "E11000 duplicate key error index: kontena_development.host_nodes.$grid_id_1_node_number_1 dup key: { : ObjectId('58e4cb98ea110f005400001a'), : null }"
api_1      | 
api_1      | See https://github.com/mongodb/mongo/blob/master/docs/errors.md
api_1      | for details about this error.
```

SpComb:
This is now very easy to trigger with the new node create API:

```
82.181.164.219 - - [28/Jul/2017:13:05:10 +0000] "POST /v1/grids/test/nodes HTTP/1.1" 500 35 0.1186 
82.181.164.219 - - [28/Jul/2017:13:05:10 +0000] "POST /v1/grids/test/nodes HTTP/1.1" 201 882 0.2571 
```

```
Mongo::Error::OperationFailure (E11000 duplicate key error index: kontena_server.host_nodes.$grid_id_1_node_number_1 dup key: { : ObjectId('597b1f61afa8810007f99320'), : null } (11000)):
```

PR #2542, opened 2017-07-02
sort_by_nil_fix - fix sort_by nil

7sedam7:


SpComb:
Hmm, what situation does this happen in? It is actually possible for the `node_number` to be `nil`, but that should only happen for a very short period of time while a node is connecting.

Are you perhaps running into something like the race condition in #2071?

Just a little surprised that this popped up now, right after restoring the original `sort_by` in 1.3.2.

7sedam7:
Signed CLA.

@SpComb  I think it's not #2071, I do not know how I got to it, but I was not able to get rid of it, so I just changed it for my self and added PR in case it's ok for you as well.

SpComb:
Your git commit author doesn't match up with your GitHub account's email addresses, so the cla-assistant.io check didn't pass yet... you need to either add that email address to your github profile and re-check, or rewrite the commit author and force-push?

cli/lib/kontena/cli/nodes/list_command.rb
```diff
@@ -70,7 +70,7 @@ def node_data
           end
         end
 
-        grid_nodes.sort_by { |n| n['node_number'] }
+        grid_nodes.sort_by { |n| n['node_number'].to_i }
```
SpComb: Could also have a spec case with `{ 'node_number' => nil }` (see `cli/spec/kontena/cli/nodes/list_command_spec.rb`)

SpComb:
@7sedam7 this is still pending on the CLA check passing, because your git commit author does not match up with your github email address.

SpComb:
The server should now be fixed in #2694 to always assign a `node_number` to nodes, so this shouldn't be a problem on the CLI anymore... I hope.

PR #2543, opened 2017-07-03
Run post_start hooks before wait for port

jakolehm:
Regression in 1.3.

### DIFF
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -65,6 +65,8 @@ module Kontena
         Celluloid::Notifications.publish('service_pod:start', service_pod.name)
         Celluloid::Notifications.publish('container:publish_info', service_container)
 
+        self.run_hooks(service_container, 'post_start')
+
         if service_pod.wait_for_port
           info "waiting for port #{service_pod.name_for_humans}:#{service_pod.wait_for_port} to respond"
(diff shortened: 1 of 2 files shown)


SpComb:
Looks like a regression in 1.2 from #1607... the server used to wait for the agent `ServicePods::Creator` to run before doing the wait_for_port, and that included running the `post_start` hooks. 

The new code in 1.2 blocks the agent `ServicePods::Creator` on the wait_for_port before running the `post_start` hooks.

jakolehm:
Ok, yes it seems that this buggy behaviour has been introduced in 1.2. I think we should fix this because it kinda harms some of the stacks that have post_start hooks (eg they initialize some files/settings that enable executable to open port).

SpComb:
Just makes me wonder if there are also post-1.2 stacks where the `post_start` hooks rely on the service responding on the port... like `post_start: - cmd: curl http://localhost:8080/whatever`? :confused:

SpComb:
The https://github.com/kontena/kontena/blob/master/docs/references/kontena-yml.md#hooks could also mention something about the order that `post_start` and `wait_for_port` run in.

jakolehm:
@jnummelin improved docs. Should we include this into 1.3.3?

jnummelin:
Docs look good to me 👍 

> Should we include this into 1.3.3?

Sure, why not :)

Issue #2277, opened 2017-05-08
kontena service deploy should give a better error with stateful instances on offline nodes


SpComb:
If a stateful service has instances deployed to host nodes that are offline, then the resulting `kontena service deploy` does not actually say that:

```
$ kontena service deploy redis-ha-stateful
 [fail] Deploying service redis-ha-stateful      
 [error] Kontena::Errors::StandardErrorArray : Cannot find applicable node for service instance development/null/redis-ha-stateful-1: Strategy Scheduler::Strategy::HighAvailability did not find any node:
```

This error happens because the `GridServiceSchedulerWorker#deployer` only considers connected host nodes as elegible for scheduling, so the strategy is unable to find the existing service instance host node.

The `GridServiceInstanceDeployer` would give a much better error if trying to deploy an instance to the offline host node, but that only happens if the host node goes offline in the very short time window between scheduling and updating the `GridServiceInstance` for deployment.

SpComb:
The way the user is expected to proceed in this situation is to either use `kontena service rm --instance` to force the offline stateful instance to get re-scheduled to a different node, or use `kontena node rm` to force all service instances on that node to get re-scheduled.

This does not affect stateless services, as the scheduler will just chose a new node instead, and not fail on the missing offline node.

PR #2549, opened 2017-07-03
Ephemeral nodes should not get any stateful services

### LINKED ISSUE #2487: stateful: true service should not get deployed on instances with ephemeral[=yes] label

jnummelin:
fixes #2487

### DIFF
--- a/docs/using-kontena/nodes.md
+++ b/docs/using-kontena/nodes.md
@@ -74,3 +74,5 @@ Nodes labeled as `ephemeral=yes` will automatically be removed by the Kontena Ma
 The nodes should not be [initial nodes](grids.md#initial-nodes), and they should not have any stateful services deployed on them.
 
 Ephemeral nodes are intended to be used for autoscaled nodes, which may be provisioned automatically, and then cleaned up once terminated.
+
+Ephemeral nodes will not be available for sheduling of stateful services as the service state is only persisted on local disk.
--- a/docs/using-kontena/nodes.md
+++ b/docs/using-kontena/nodes.md
@@ -75,4 +75,4 @@ The nodes should not be [initial nodes](grids.md#initial-nodes), and they should
 
 Ephemeral nodes are intended to be used for autoscaled nodes, which may be provisioned automatically, and then cleaned up once terminated.
 
-Ephemeral nodes will not be available for sheduling of stateful services as the service state is only persisted on local disk.
+Ephemeral nodes will not be available for scheduling of stateful services as the service state is only persisted on local disk.
--- a/server/app/services/grid_service_scheduler.rb
+++ b/server/app/services/grid_service_scheduler.rb
@@ -16,6 +16,7 @@ class GridServiceScheduler
   def initialize(strategy)
     @strategy = strategy
     @filters = [
+      Scheduler::Filter::Ephemeral.new,
       Scheduler::Filter::VolumePlugin.new,
       Scheduler::Filter::VolumeInstance.new,
       Scheduler::Filter::Affinity.new,
--- a/server/app/services/scheduler/filter/ephemeral.rb
+++ b/server/app/services/scheduler/filter/ephemeral.rb
@@ -1,0 +1,26 @@
+module Scheduler
+  module Filter
+    class Ephemeral
+      include Logging
+
+      ##
+      # Filters nodes based on ephemeral status
+      # @param [GridService] service
+      # @param [Integer] instance_number
+      # @param [Array<HostNode>] nodes
+      # @raise [Scheduler::Error]
+      def for_service(service, instance_number, nodes)
+        return nodes unless service.stateful?
+
+        nodes = nodes.reject{|n| n.ephemeral?}
+
+        if nodes.empty?
+          raise Scheduler::Error, "Did not find any non-ephemeral nodes for stateful service"
+        end
+
+        nodes
+      end
+
+    end
+  end
+end
--- a/server/app/services/scheduler/filter/ephemeral.rb
+++ b/server/app/services/scheduler/filter/ephemeral.rb
@@ -12,7 +12,7 @@ module Scheduler
       def for_service(service, instance_number, nodes)
         return nodes unless service.stateful?
 
-        nodes = nodes.reject{|n| n.ephemeral?}
+        nodes = nodes.reject{|n| n.ephemeral? && !n.grid_service_instances.find_by(grid_service: service, instance_number: instance_number)}
 
         if nodes.empty?
           raise Scheduler::Error, "Did not find any non-ephemeral nodes for stateful service"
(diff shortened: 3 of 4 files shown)


jnummelin:
eh, there's _slight_  problem if there's a service which already has instances scheduled for `ephemeral` nodes:
```
$ k service deploy redis
⊛ Deployed instance e2e/null/redis-1 to node moby

$ k node label add moby ephemeral=true

$ k service deploy redis
 [fail] Deploying service redis      
 [error] 500 : Cannot find applicable node for service instance e2e/null/redis-1: Did not find any non-ephemeral nodes for stateful service:
```

SpComb:
Hmm right, I was only testing with `strategy: daemon` and then making sure that new services do not get deployed to ephemeral nodes. 

Hmm... odd error though, how many nodes do you have in your grid? If it's just one node, then that error looks sensible? Otherwise for the case of a stateful service instance deployed to a node that gets labled as ephemeral, I would expect it to behave like #2277 for an offline node: `Strategy Scheduler::Strategy::HighAvailability did not find any node: `

Labeling a node as ephemeral after the fact is a bit of an odd use-case, that label should come from the Docker labels when the agent first connects.

SpComb:
> Otherwise for the case of a stateful service instance deployed to a node that gets labled as ephemeral, I would expect it to behave like #2277 for an offline node: Strategy Scheduler::Strategy::HighAvailability did not find any node:

And the workaround should be the same: `kontena service rm --instance` the affected stateful instance on the ephemeral node.

jnummelin:
> Hmm... odd error though, how many nodes do you have in your grid? If it's just one node, then that error looks sensible?

Kinda yes, but still some part of my brain would expect that an existing instance could be re-deployed to the same node where the data already is.

> Labeling a node as ephemeral after the fact is a bit of an odd use-case, that label should come from the Docker labels when the agent first connects.

I agree that this is bit of odd case, but I think there's a real possibility of this happening since < 1.4 releases have not had any deployment controls over ephemeral nodes. The worst part is that the only workaround involves manual removal of instance(s) which could very well lead to data loss.

SpComb:
Indeed:

#### `kontena service show stateful-1/foo`
Stateful service is deployed to `core-02`:

```
development/stateful-1/foo:
  created: 2017-07-03T14:09:12.247Z
  updated: 2017-07-03T14:09:12.247Z
  stack: development/stateful-1
  desired_state: running
  image: redis:latest
  revision: 1
  stack_revision: 1
  stateful: yes
  scaling: 1
  strategy: ha
  stop_grace_period: 10s
  deploy_opts:
  dns: foo.stateful-1.development.kontena.local
  instances:
    stateful-1/foo/1:
      scheduled_to: core-02
      deploy_rev: 2017-07-03 14:09:12 UTC
      rev: 2017-07-03 14:09:12 UTC
      state: running
      containers:
        stateful-1.foo-1 (on core-02):
          dns: foo-1.stateful-1.development.kontena.local
          ip: 10.81.128.108
          public ip: 91.150.10.190
          status: running
```

#### `kontena node label add core-02 ephemeral=true`

Label core-02 as ephemeral

#### `kontena stack deploy stateful-1`

Scheduling fails per #2277 
```
 [done] Triggering deployment of stack stateful-1     
 [fail] Waiting for deployment to start     
        Cannot find applicable node for service instance development/stateful-1/foo-1: Strategy Scheduler::Strategy::HighAvailability did not find any node
 [error] RuntimeError : deploy failed
```

#### `kontena service rm --instance=1 stateful-1/foo`
Force stateful service instance to get re-scheduled:

```
Destructive command. To proceed, type "stateful-1/foo/1" or re-run this command with --force option.
> Enter 'stateful-1/foo/1' to confirm:  stateful-1/foo/1
 [done] Removing service instance stateful-1/foo/1      
```

#### `kontena stack deploy stateful-1`

Able to schedule instance to new node:

```
 [done] Triggering deployment of stack stateful-1     
 [done] Deploying service foo     
```

#### `kontena service show stateful-1/foo`
```
development/stateful-1/foo:
  created: 2017-07-03T14:09:12.247Z
  updated: 2017-07-03T14:09:12.247Z
  stack: development/stateful-1
  desired_state: running
  image: redis:latest
  revision: 1
  stack_revision: 1
  stateful: yes
  scaling: 1
  strategy: ha
  stop_grace_period: 10s
  deploy_opts:
  dns: foo.stateful-1.development.kontena.local
  instances:
    stateful-1/foo/1:
      scheduled_to: core-01
      deploy_rev: 2017-07-03 14:09:53 UTC
      rev: 2017-07-03 14:09:53 UTC
      state: running
      containers:
        stateful-1.foo-1 (on core-02):
          dns: foo-1.stateful-1.development.kontena.local
          ip: 10.81.128.108
          public ip: 91.150.10.190
          status: running
        stateful-1.foo-1 (on core-01):
          dns: foo-1.stateful-1.development.kontena.local
          ip: 10.81.128.30
          public ip: 91.150.10.190
          status: running
```

The dup container disappers once the agent notices that it should be terminated.

jnummelin:
I _think_  the `Ephemeral` filter should consider an `ephemeral` node being valid deployment target for stateful service IF there's already existing instance on that node.

SpComb:
> I think the Ephemeral filter should consider an ephemeral node being valid deployment target for stateful service IF there's already existing instance on that node.

Then that sounds vaugely like the offline node grace period stickyness, and should be implemented at the strategy level.... although that offline grace period thing is implemented in the `GridScheduler` and does not apply to the actual deploy, which is the exactly the missing thing that makes the resulting error in #2277 so nasty.

jnummelin:
> Then that sounds vaugely like the offline node grace period stickyness, and should be implemented at the strategy level....

The problem is that filters are run first and thus ephemeral nodes will not get to strategy level at all for stateless services. I don't see a problem if the filter would return ephemeral node if it already has an stateful instance, the strategy should then pick that up.

jnummelin:
With the latest commit that allows ephemeral node for scheduling only if it already has an instance:
```
k service create --stateful redis redis
 [done] Creating redis service      

$ k service deploy redis
⊛ Deployed instance e2e/null/redis-1 to node moby

$ k node label add moby ephemeral=true

$ k service deploy redis
⊛ Deployed instance e2e/null/redis-1 to node moby

$ k service create --stateful another redis
 [done] Creating another service      

$ k service deploy another
 [fail] Deploying service another      
 [error] 500 : Cannot find applicable node for service instance e2e/null/another-1: Did not find any non-ephemeral nodes for stateful service:
```

SpComb:
> With the latest commit that allows ephemeral node for scheduling only if it already has an instance:

BTW I think offline nodes should also behave similarly re #2277: existing stateful instances should just keep being scheduled to offline nodes, then the deploy will fail on them being offline with an informative error message.

docs/using-kontena/nodes.md
```diff
@@ -74,3 +74,5 @@ Nodes labeled as `ephemeral=yes` will automatically be removed by the Kontena Ma
 The nodes should not be [initial nodes](grids.md#initial-nodes), and they should not have any stateful services deployed on them.
 
 Ephemeral nodes are intended to be used for autoscaled nodes, which may be provisioned automatically, and then cleaned up once terminated.
+
+Ephemeral nodes will not be available for sheduling of stateful services as the service state is only persisted on local disk.
```
jakolehm: `sheduling` -> `scheduling`

PR #2550, opened 2017-07-03
Read-only container instances

jnummelin:
`read-only` containers are a great tool setting up the first line of defence:
https://diogomonica.com/2016/11/19/increasing-attacker-cost-using-immutable-infrastructure/

### DIFF
--- a/cli/lib/kontena/cli/stacks/service_generator.rb
+++ b/cli/lib/kontena/cli/stacks/service_generator.rb
@@ -64,7 +64,7 @@ module Kontena::Cli::Stacks
       data['build'] = parse_build_options(options) if options['build']
       data['health_check'] = parse_health_check(options)
       data['stop_grace_period'] = options['stop_grace_period'] if options['stop_grace_period']
-      data['read_only'] = options['read_only'] if options['read_only']
+      data['read_only'] = options['read_only'] || false
       data
     end
 
--- a/cli/lib/kontena/cli/stacks/service_generator.rb
+++ b/cli/lib/kontena/cli/stacks/service_generator.rb
@@ -64,6 +64,7 @@ module Kontena::Cli::Stacks
       data['build'] = parse_build_options(options) if options['build']
       data['health_check'] = parse_health_check(options)
       data['stop_grace_period'] = options['stop_grace_period'] if options['stop_grace_period']
+      data['read_only'] = options['read_only'] if options['read_only']
       data
     end
 
--- a/test/spec/features/stack/install_spec.rb
+++ b/test/spec/features/stack/install_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe 'stack install' do
-  
+
   after(:each) do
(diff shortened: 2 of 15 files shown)


cli/lib/kontena/cli/stacks/service_generator.rb
```diff
@@ -64,6 +64,7 @@ def parse_data(options)
       data['build'] = parse_build_options(options) if options['build']
       data['health_check'] = parse_health_check(options)
       data['stop_grace_period'] = options['stop_grace_period'] if options['stop_grace_period']
+      data['read_only'] = options['read_only'] if options['read_only']
```
SpComb: Hmm... this only sends `{"read_only": true}` if it's set to true, but it never sends `{"read_only": false}`? How does that work if you `kontena stack upgrade` to change a service from `read_only: true` -> `read_only: false`?

Confirmed:
```
DEBUG [Request]: PUT http://localhost:9292/v1/stacks/development/readonly | Headers: {Accept: application/json, Content-Type: application/json, Authorization: Bearer}  | Body: 
{"name":"readonly","stack":"test/readonly","expose":null,"version":"0.0.1","source":"stack: test/readonly\nservices:\n  foo:\n    image: redis\n    read_only: false\n","registry":"file://","services":[{"instances":null,"image":"redis:latest","links":[],"external_links":[],"ports":[],"volumes":[],"volumes_from":[],"affinity":[],"stateful":false,"deploy_opts":{"wait_for_port":null,"min_health":null,"interval":null},"hooks":{},"health_check":{"port":null,"protocol":null,"uri":null,"timeout":null,"interval":null,"initial_delay":null},"log_driver":null,"log_opts":null,"name":"foo"}],"volumes":[],"variables":{}}
```

After changing stack file `read_only: false`, and `kontena stack upgrade`, the container is still readonly:
```
$ kontena service exec --shell readonly/foo 'echo foo > /foo'
/bin/sh: 1: cannot create /foo: Read-only file system
```

SpComb:
Also missing from `kontena service show`.

cli/lib/kontena/cli/stacks/service_generator.rb
jnummelin: yep, brainfart :) fixed that in latest commits

test/spec/features/stack/install_spec.rb
```diff
@@ -64,4 +64,21 @@
...
+      with_fixture_dir("stack/read_only") do
+        run 'kontena stack install redis.yml'
+      end
+      k = run 'kontena service show simple/redis'
+      expect(k.code).to eq(0)
+      expect(k.out.match(/read_only: yes/)).to be_truthy
+      with_fixture_dir("stack/read_only") do
+        run 'kontena stack upgrade simple redis_read_only_false.yml'
+      end
+      k = run 'kontena service show simple/redis'
```
SpComb: A `kontena container inspect` to look for `HostConfig.ReadonlyRootfs: true` would be even better (more end-to-end)

PR #2565, opened 2017-07-07
[WIP] Support tls-sni-01 authorization for LE

jnummelin:
This PR brings in support for `tls-sni-01` authorization for LE certs. `tls-sni-01` makes possible to build automation on top of.

## TODO

- [ ] API docs update
- [ ] Test renewal


cli/lib/kontena/cli/certificate/authorize_command.rb
```diff
@@ -7,16 +7,26 @@ class AuthorizeCommand < Kontena::Command
 
     parameter "DOMAIN", "Domain to authorize"
 
+    option "--auth-type", "AUTH_TYPE", "Authorization type, either dns-01 or tls-sni-01", default: 'dns-01'
+    option "--lb-link", "LB_LINK", "Link to loadbalancer where the certificate will be used on"
```
jakolehm: Maybe just `--lb` or `--loadbalancer` ?

docs/using-kontena/certificates.md
```diff
@@ -27,20 +27,35 @@ The email is needed for Let's Encrypt to notify when certificates are about to e
...
+$ kontena certificate authorize --auth-type dns-01 api.example.com
+Authorization successfully created. Use the following details to create necessary validations:
+Record name: _acme-challenge.api.example.com
+Record type: TXT
+Record content: jEeBHU0WtHhnf0ZRXBbN0nYnjcBWlSS7TFiXvjFs62k
 ```
 
 To verify that you really control the requested domain, create a DNS TXT record for the domain `_acme-challenge.api.example.com` with content specified in the response.
 
+### TLS-SNI based authorization
```
jakolehm: I would swap `TLS-SNI` to first because it's probably easier to use?

server/app/models/certificate.rb
```diff
@@ -0,0 +1,23 @@
...
+  include Mongoid::Document
+
+  field :domain, type: String
+  field :valid_until, type: DateTime
+  field :alt_names, type: Array
+
+  field :cert_type, type: String
+
+  # Need to have references to the secrets to know which ones to automatically update
+  has_one :private_key, class_name: 'GridSecret'
```
jakolehm: I think we should use "soft" references here.

server/app/models/grid_secret.rb
```diff
@@ -7,6 +7,8 @@ class GridSecret
   field :name, type: String
   field :encrypted_value, type: String, encrypted: true
 
+  belongs_to :certificate
```
jakolehm: Not needed if reference is "soft".

server/app/views/v1/certificates/_certificate.json.jbuilder
```diff
@@ -0,0 +1,8 @@
+json.id certificate.id
```
jakolehm: Maybe add also timestamps?

jnummelin: there is `valid_until`

server/app/models/certificate.rb
jnummelin: by "soft" I assume you mean just the names of the secrets?

server/app/views/v1/certificates/_certificate.json.jbuilder
jnummelin: Timestamps added now.

server/app/models/certificate.rb
jakolehm: Yes.

server/app/mutations/grid_certificates/authorize_domain.rb
```diff
@@ -9,32 +9,89 @@ class AuthorizeDomain < Mutations::Command
...
+
+    # Adds the tls-sni secret to the service
+    def add_secret_to_service(secret, grid_service)
+      unless grid_service.secrets.index {|s| s['secret'] == secret.name}
+        info "adding tls-sni secret to service: #{grid_service.qualified_name}"
+        secrets = grid_service.secrets.map{ |s| {'secret' => s['secret'], 'name' => s['name'], 'type' => s['type']}}
+        secrets << {secret: secret.name, name: 'SSL_CERTS', type: 'env'}
+        outcome = GridServices::Update.run(grid_service: grid_service, secrets: secrets)
+        unless outcome.success?
+          add_error(:tls_sni_secret, :error, outcome.errors.message)
```
SpComb: With our [`mutations/command_errors.rb`](https://github.com/kontena/kontena/blob/master/server/lib/mutations/command_errors.rb#L13) patch, these should just be:

```ruby
add_error(:tls_sni_secret, outcome.errors)
```

server/app/mutations/grid_certificates/common.rb
```diff
@@ -46,5 +46,28 @@ def get_authz_for_domain(grid, domain)
...
+    def upsert_secret(name, value)
+      cert_secret = self.grid.grid_secrets.find_by(name: name)
+      if cert_secret
+        outcome = GridSecrets::Update.run(grid_secret: cert_secret, value: value)
+      else
+        outcome = GridSecrets::Create.run(grid: self.grid, name: name, value: value)
+      end
+
+      unless outcome.success?
+        add_error(:cert_store, :failure, "Certificate storing to vault failed: #{outcome.errors.message}")
```
SpComb: `add_error(:cert_store, outcome.errors)`

server/app/models/grid_domain_authorization.rb
```diff
@@ -9,6 +9,9 @@ class GridDomainAuthorization
   field :domain, type: String
   field :challenge, type: Hash
   field :challenge_opts, type: Hash # TODO encrypt?
+  field :authorization_type, type: String
+
+  has_one :grid_service
```
SpComb: This field is never set in the `GridCertificates::AuthorizeDomain` mutation, so it's not used, and always nil?

server/app/models/grid_service.rb
```diff
@@ -43,6 +43,8 @@ class GridService
   belongs_to :grid
   belongs_to :image
   belongs_to :stack
+  belongs_to :grid_domain_authorization
```
SpComb: This only allows one `GridDomainAuthorization` per (linked LB) service? If you want to have this field, then it needs to be the other way around: `GridDomainAuthorization belongs_to :grid_service`, and `GridService has_many :grid_domain_authorizations`. It should probably also be called something like `:linked_service` to make it clearer.

It doesn't look like this relation is actually used for anything, though.

server/app/mutations/grid_certificates/authorize_domain.rb
```diff
@@ -9,32 +9,89 @@ class AuthorizeDomain < Mutations::Command
...
+      challenge
+    end
+
+    # Adds the tls-sni secret to the service
+    def add_secret_to_service(secret, grid_service)
+      unless grid_service.secrets.index {|s| s['secret'] == secret.name}
+        info "adding tls-sni secret to service: #{grid_service.qualified_name}"
+        secrets = grid_service.secrets.map{ |s| {'secret' => s['secret'], 'name' => s['name'], 'type' => s['type']}}
+        secrets << {secret: secret.name, name: 'SSL_CERTS', type: 'env'}
+        outcome = GridServices::Update.run(grid_service: grid_service, secrets: secrets)
```
SpComb: Might be nicer if this was a new `GridServices::AddSecret` mutation? I might even go so far as to consider having a `GridServices::AddCertificate` with a new `GridService.certificates` relation to the `Certificate` model, and have the `Rpc::ServicePodSerializer` include those as secrets.

server/app/mutations/grid_certificates/get_certificate.rb
```diff
@@ -54,10 +57,8 @@ def execute
           end
         end
 
-        Timeout::timeout(30) {
-          info 'waiting for DNS validation...'
-          sleep 1 until challenge.verify_status == 'valid'
-          info 'DNS validation complete'
+        wait_until!("domain verification for #{domain} is valid", interval: 1, timeout: 30, threshold: 10) {
+          challenge.verify_status == 'valid'
```
SpComb: I see that this is existing code, but what happens if the verify fails?

server/app/routes/v1/grids/grid_certificates.rb
```diff
@@ -0,0 +1,20 @@
+V1::GridsApi.route('grid_certificates') do |r|
+
+  # POST /v1/grids/:name/certificates
+  r.post do
+    data = parse_json_body
+    data[:current_user] = current_user
+    data[:grid] = @grid
+    data[:grid] = @grid
+    outcome = GridCertificates::GetCertificate.run(data)
```
SpComb: I wonder if the `GridCertificates::GetCertificate` should be called something other than `GET`... maybe `GridCertificates::CreateCertificate`, because it seems to do an entirely new CSR each time.

Is there a `GridCertificates::RenewCertificate`?

server/app/mutations/grid_certificates/authorize_domain.rb
SpComb: This doesn't do a deploy, so the amount of time between the domain authorization challenge being created and the verification passing can be unknown... something with the 20s grid scheduler job, plus the actual LB service deploy queuing and running times?

The `GetCertificate` doesn't validate that the SNI challenge is in place, and it only has a 30s timeout for the the LE verification to pass?

server/app/routes/v1/grids/grid_certificates.rb
```diff
@@ -0,0 +1,20 @@
+V1::GridsApi.route('grid_certificates') do |r|
+
+  # POST /v1/grids/:name/certificates
+  r.post do
+    data = parse_json_body
+    data[:current_user] = current_user
```
SpComb: Not used

server/app/routes/v1/grids/grid_certificates.rb
```diff
@@ -0,0 +1,20 @@
...
+    data[:grid] = @grid
+    data[:grid] = @grid
```
SpComb: Twice

server/app/routes/v1/grids/grid_certificates.rb
```diff
@@ -0,0 +1,20 @@
+V1::GridsApi.route('grid_certificates') do |r|
+
+  # POST /v1/grids/:name/certificates
```
SpComb: Should this also have a `GET` method, for `kontena certificate list`, which would show what certs are expiring?

server/app/views/v1/certificates/_certificate.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id certificate.to_path
```
SpComb: This `id` is misleading... there is (unfortunately) no `.../certificates/:grid/:domain` API that this ID could refer to :(

BTW: The `GridDomainAuthorization#to_path` is even more broken, it uses a non-existent `self.name` field.

server/app/models/grid_domain_authorization.rb
jnummelin: It is set: https://github.com/kontena/kontena/pull/2565/files#diff-075d4bb6ae3e06ed6cca0d0d823de622R54 or in:
https://github.com/kontena/kontena/pull/2565/files#diff-075d4bb6ae3e06ed6cca0d0d823de622R56

server/app/models/grid_service.rb
jnummelin: Yes, something else is needed to support many authz per LB.

It's not currently used for anything. The idea was to use this in near future in the automated cert update, to update the SNI secret into correct LB service.

server/app/models/grid_domain_authorization.rb
SpComb: I mean the `grid_service` relation, not the `authorization_type`

SpComb:
Also consider updating the `acme-client` dependency from 0.3.4 to something more recent (0.6.1)?

server/app/models/grid_domain_authorization.rb
jnummelin: Oh, obviously. :) yeah, not used yet for anything. Idea was to use that for the LB link so it's possible to link the SNI secret to LB when renewing.

server/app/mutations/grid_certificates/get_certificate.rb
SpComb: From the [README](https://github.com/unixcharles/acme-client):
```
# If authorization.verify_status is 'invalid', you can get at the error
# message only through the failed challenge.
authorization.verify_status # => 'invalid'
authorization.http01.error # => {"type" => "...", "detail" => "..."}
```

It would be nice to see those errors? It doesn't seem like they would get raised as `Acme::Client:Error`s?

SpComb: Separate issue with the existing code: #2688

It's just way more likely to fail here compared to the old `dns-01` validation, because the server would itself also validate the DNS challenge record, which means that the domain authorization challenge is unlikely to fail.

However, particularly with the delayed `LE_TLS_SNI_*` challenge cert deployment from the `GridCertificates::AuthorizeDomain`, and the lack of our own client-side validation for the TLS-SNI challenge, this is now far more likely to fail here.

server/app/mutations/grid_certificates/authorize_domain.rb
SpComb: My suggestion: don't modify the lb service secrets like this, instead, just `GridServiceAuthorization.create(linked_service: @lb_service)` and bump the `@lb_service` `revision`, `updated_at` fields. The `Rpc::ServicePodSerializer` can then pick up any `grid_service_authorizations` and serialize them as `SSL_CERT` secrets.

For bonus points, create a new `GridServiceDeploy` and return the deploy ID, so that the CLI can spin waiting for the challenge cert to get deployed.

SpComb: The `GridServiceAuthorization`s should probably also be deleted once their verification has passed in `GridCertificates::GetCertificate`?

SpComb:
What about automating the TLS-SNI challenge by linking the `GridDomainAuthorization` to the LB `GridService`, and deploying the challenge cert as a `SSL_CERTS` secret in the `Rpc::ServicePodSerializer`?

jnummelin:
> What about automating the TLS-SNI challenge by linking the GridDomainAuthorization to the LB GridService, and deploying the challenge cert as a SSL_CERTS secret in the Rpc::ServicePodSerializer?

That would be one option. As per the internal slack discussion, I wanted to keep the model and API changes at minimum as we were thinking of complete cert API "re-write".

What I was thinking was to update the secrets mutation in a way that they would accept `update_services: false` (or something alike) as option. Cert renewal could use that to defer automatic deployment from secret update and trigger actual deploy for the LB and track it's progress. Now it needs to "guess" deployment readiness based on `deploy_rev` fields.

jnummelin:
#2732 replaces this

PR #2590, opened 2017-07-17
Configurable agent node ID, labels

### LINKED ISSUE #1771: Set node labels from kontena-agent container labels

SpComb:
Fixes #??? by providing the `KONTENA_NODE_ID=` env for hosts with volatile Docker IDs (DfM)
Fixes #1771 by providing the `KONTENA_NODE_LABELS=` env

* Configure `Kontena::Agent` `@node_id`, `@node_labels` from `ENV['KONTENA_NODE_*'] || `Docker.info[...]`

* Change agent `WebsocketClient` to send configured node ID/labels

* Fix agent worker RPC calls to either use the configured node ID, or the `Node.id` (from the serialized `/agent/node_info` sent by the server)

    Several workers were using `Docker.info['ID']` directly.

* Change the `/nodes/update` and `/nodes/stats` RPCs to take an explicit `node_id` parameter


agent/lib/kontena/agent.rb
```diff
@@ -4,14 +4,31 @@ module Kontena
   class Agent
     include Logging
 
+    NODE_ID_REGEXP = /\A[A-Z0-9]{4}(:[A-Z0-9]{4}){11}\z/
```
SpComb: This has to still match with what the server expects the node_id to look like:

```ruby
        if node_id.include?(':')
          node = grid.host_nodes.find_by(node_id: node_id)
        else
          node = grid.host_nodes.find_by(name: node_id)
        end
```

agent/spec/lib/kontena/workers/node_stats_worker_spec.rb
```diff
@@ -138,12 +129,13 @@
   describe '#publish_node_stats' do
 
     before(:each) do
+      subject.instance_variable_set('@node', node)
```
jakolehm: Must be a better way?

SpComb: It's set in the `start` -> `observe` block from the node info sent by the server... could refactor that out and call that from `before`.

#2591 also removes the need for the `@node` here entirely.

SpComb: There's actually a bug here... the worker could call `start_publish_loop` -> `publish_node_stats` -> `@node.id` before the `start` -> `observe` -> `@node = ...` happens.

SpComb: The specs were already doing `subject.instance_variable_set("@statsd", statsd)` for the `send_statsd_metrics`... refactored the worker to have a `configure` method, and call that in the rspec context.

agent/lib/kontena/workers/node_stats_worker.rb
```diff
@@ -116,13 +125,15 @@ def publish_node_stats
...
+
+    # @param [Hash] data
+    def send_node_stats(data)
+      rpc_client.async.notification('/nodes/stats', [@node.id, data])
     end
 
     # @param [Hash] event
     def send_statsd_metrics(event)
-      return unless statsd
       key_base = "#{docker_info['Name']}"
```
SpComb: I would very much also like to remove the `Docker.info` in here..

* The statsd `docker_info['Name']` could be changed to use the `@node.name` instead? It looks like ATM nodes with duplicate hostnames would result in overlapping statsd keys, whereas the `@node.name` is unique?

* The `docker_info['DockerRootDir']` is also used for the `filesystem.name` path... but that's inaccurate, because the disk stats are actually from `Vmstat.disk('/')`, which might be a different filesystem?

jakolehm: Docker root dir is best guess here.. you can open an issue about this if you think it's needed.

PR #2591, opened 2017-07-17
[WiP] Refactor agent RPCs to use implicit node IDs

SpComb:
Scope all RPC requests/responses/notifications by the authenticated node ID from the websocket connection. This is a security feature, preventing malicious nodes from being able to leak/manipulate the state of other nodes by their node ID. This also simplifies the agent/server RPC callers/handlers.

* Change the server `RpcServer` handlers to be per-node, based on the authenticated node from the websocket connection

* Removes any explicit node ID parameters from agent -> server RPCs, using the rpc handler `@node` instead

* Server `RpcClient`-> agent RPC responses are now also routed using a per-node `rpc_client/:node_id/:rpc_id` pubsub name

# TODO

* Fix specs
* Consider moving the server `Rpc::ContainerHandler` buffering to a separate actor, with periodic flushing... having it be per-node means the stats/logs buffers are even longer now?

### DIFF
--- a/agent/lib/kontena/actors/container_coroner.rb
+++ b/agent/lib/kontena/actors/container_coroner.rb
@@ -6,10 +6,8 @@ module Kontena::Actors
 
     INVESTIGATION_PERIOD = 20
 
-    # @param [String] node_id
     # @param [Boolean] autostart
-    def initialize(node_id, autostart = true)
-      @node_id = node_id
+    def initialize(autostart = true)
       async.process if autostart
     end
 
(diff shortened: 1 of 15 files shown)


SpComb:
@jakolehm WDYT, does this make sense, and should I continue with fixing up the specs?

SpComb:
The `ServicePodManager` doesn't use the `@node` anymore for anything, but I don't want to remove it either yet... it serves to wait until the agent websocket connection is open and the host node is connected.... if anything, the initial pod sync should happen even later, to allow the server to reschedule containers running on the disconnected node back to the same node when it reconnects.

PR #2614, opened 2017-07-26
Build deb package with omnibus

jakolehm:
Binary deb package for Debian/Ubuntu.

Fixes #1954

PR #2617, opened 2017-07-26
Release 1.4.0.pre1

SpComb:
Initial pre-release for testing the new node provisioning workflow.


SpComb:
Updated the `v1.4` branch to also include #2614

PR #2621, opened 2017-07-27
Release 1.4.0.pre2

SpComb:


SpComb:
Added #2619

SpComb:
Updated to include #2624

PR #2624, opened 2017-07-27
Remove development deps from Omnibus Gemfile

jakolehm:
Faster `bundle install`


cli/omnibus/Gemfile
```diff
@@ -1,22 +1,9 @@
...
 
 # Use Chef's software definitions. It is recommended that you write your own
 # software definitions, but you can clone/fork Chef's to get you started.
-# gem 'omnibus-software', github: 'opscode/omnibus-software'
+gem 'omnibus-software', github: 'opscode/omnibus-software'
 
-# This development group is installed by default when you run `bundle install`,
-# but if you are using Omnibus in a CI-based infrastructure, you do not need
-# the Test Kitchen-based build lab. You can skip these unnecessary dependencies
-# by running `bundle install --without development` to speed up build times.
```
SpComb: Should the `build/travis/omnibus_deb.sh` be changed to use `bundle install --without development` instead of removing these?

jakolehm: Imho better to remove this, we are not using kitchen.

PR #2632, opened 2017-07-27
Update server/docs dependencies for ruby 2.4

SpComb:
Old json-1.8.3 is not compatible with ruby 2.4: https://travis-ci.org/kontena/kontena/jobs/258121470#L1944


server/docs/Dockerfile.build
```diff
@@ -0,0 +1,18 @@
...
+RUN gem install bundler --no-ri --no-rdoc
+
+ADD Gemfile Gemfile.lock /home/build/
+WORKDIR /home/build
+
+RUN bundle config build.nokogiri --use-system-libraries
+RUN bundle install
+
+VOLUME /home/build # expect -v $PWD:/home/build
+CMD middleman build
```
SpComb: Doesn't work for whatever reason, fails with

`$ docker build -t kontena/server-docs.build -f Dockerfile.build . && docker run --rm -it -v $PWD:/home/build kontena/server-docs.build`

```
/home/build not found
```

PR #2636, opened 2017-07-28
Fix node token clear to use DELETE /v1/nodes/:id/token

SpComb:
Change `kontena node reset-token --clear-token` to use `DELETE /v1/nodes/:id/token` which calls `HostNodes::UpdateToken.run :clear_token => true`. Also accepts the optional `reset_connection: true` parameter.

* Validate node tokens to be 16..64 chars in length, rejecting empty tokens
* Add `DELETE /v1/nodes/:id/token` to replace `PUT /v1/nodes/:id/token` `{ "token": "" }`
* Change `GET /v1/nodes/:id/token` to return HTTP 404 instead of `{ "token": null }` if no node token

### DIFF
--- a/cli/lib/kontena/cli/nodes/env_command.rb
+++ b/cli/lib/kontena/cli/nodes/env_command.rb
@@ -14,10 +14,15 @@ module Kontena::Cli::Nodes
       grid_uri = self.current_master['url'].sub('http', 'ws')
     end
 
-    def execute
-      token_node = client.get("nodes/#{current_grid}/#{name}/token")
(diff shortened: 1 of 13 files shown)


SpComb:
Main design question is: should `GET /v1/nodes/:id/token` return HTTP 404 after a DELETE, or for a legacy node that wasn't created with a node token?

jakolehm:
`404` sounds about right if node does not have a token.

SpComb:
> 404 sounds about right if node does not have a token.

Done.

Issue #2608, opened 2017-07-24
stack rm fails to remove containers that are in restart loop


matti:
stack with two services, the other service fails to start --> kontena stack rm --> docker keeps restarting the service for ~ 2 minutes until agent (?) removes the container

jakolehm:
Delay is most likely caused by `wait_for_port` option. Agent waits max 5 mins port to open.. after that it will read desired state again and remove service instance (container).

matti:
delay is veried to be `wait_for_port`

SpComb:
Fixed in 1.4 by #2776, the `kontena stack rm` now has ~immediate effect even if the initial deploy is in the `wait_for_port`.

```yaml
stack: test/waitforport
services:
  redis:
    image: redis
    deploy:
      wait_for_port: 6378
```

```
$ bundle exec bin/kontena stack install test/waitforport.yaml 
 [done] Creating stack waitforport      
 [done] Triggering deployment of stack waitforport     
 [done] Waiting for deployment to start     
 [fail] Deploying service redis     ^C
$ bundle exec bin/kontena stack rm --force waitforport
 [done] Removing stack waitforport      
```

```
D, [2018-01-09T10:04:59.011381 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of waitforport.redis-1: running
D, [2018-01-09T10:04:59.087148 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state is in-sync: running
I, [2018-01-09T10:04:59.104334 #1]  INFO -- Kontena::Workers::ServicePodWorker: waiting until container /waitforport.redis-1 port 10.81.128.33:6378 is responding
D, [2018-01-09T10:05:02.121406 #1] DEBUG -- Kontena::Workers::ServicePodWorker: waiting 4.0s of 300.0s until: container /waitforport.redis-1 port 10.81.128.33:6378 is responding
D, [2018-01-09T10:05:03.124945 #1] DEBUG -- Kontena::Workers::ServicePodWorker: waiting 5.0s of 300.0s until: container /waitforport.redis-1 port 10.81.128.33:6378 is responding
D, [2018-01-09T10:05:03.749324 #1] DEBUG -- Kontena::Workers::ServicePodWorker: waiting 9.8s of 300.0s until: container /waitforport.redis-1 port 10.81.128.33:6378 is responding
[...]

PR #2648, opened 2017-07-30
Recreate service pod worker on unexpected dead actor error

### LINKED ISSUE #2625: stack rm and install of a container that is in restart loop does  not create a new container
1. vault write invalid SSL cert
2. stack install kontena/lb-ingress and select the invalid cert
3. restart loop starts
4. vault upsert SSL cert to be valid
5. kontena stack rm --force lb-ingress; kontena stack install --values-from values.yml kontena/lb-ingress
6. stack is successfully installed, container is not removed (created timestamp is older than the stack rm && install and thus has the old invalid cert
7. repeat rm && install (scripted so that it's fast), happens every time

fix: wait "5 seconds" before installing again.

jakolehm:
Cannot reproduce this situation but in theory there is a condition where `ServicePodManager` calls dead `ServicePodWorker`. Previously this just cleared worker from `workers` pool, which is fine if service still exists in master. If it does not exist in master this can lead into situation where worker is gone but actual container still exists -> container can block new deployments because no worker is doing container remove/cleanup.

Probably fixes #2625, #2608

### DIFF
--- a/agent/lib/kontena/workers/service_pod_manager.rb
+++ b/agent/lib/kontena/workers/service_pod_manager.rb
@@ -122,7 +122,10 @@ module Kontena::Workers
           workers[service_pod.id].async.update(service_pod)
         end
       rescue Celluloid::DeadActorError
+        error "tried to call dead worker: #{service_pod.name}"
         workers.delete(service_pod.id)
+        info "recreating worker for: #{service_pod.name}"
+        ensure_service_worker(service_pod)
       end
     end
 


agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -122,7 +122,10 @@ def ensure_service_worker(service_pod)
           workers[service_pod.id].async.update(service_pod)
         end
       rescue Celluloid::DeadActorError
+        error "tried to call dead worker: #{service_pod.name}"
         workers.delete(service_pod.id)
+        info "recreating worker for: #{service_pod.name}"
+        ensure_service_worker(service_pod)
```
SpComb: I suspect this isn't the right place to have this logic... it only specifically applies to `Celluloid::DeadActorError`s raised from `async.update` calls .... with the actor linking, a crash in the async `update` call would first show up as a `on_worker_exit`, which removes it from the workers?

* the `ServicePodWorker` crashes in the `on_container_event` immediately after the `initialize` but before the manager calls `async.update`
* the `ServicePodWorker` crashes in the `async.update`, but before celluloid is able to propagate the linked actor crash to the manager `trap_exit :on_worker_exit`, the manager `populate_workers_from_master` loop calls `ensure_service_worker`

I don't exactly see when this specific rescue would ever happen, apart from some very rare race condition scenarios?

jakolehm: Probably true, I did not manage to get this rescue to happen in my local tests.

PR #2666, opened 2017-08-07
CLI: Use the "press any key" dialog from tty-prompt library

### LINKED ISSUE #2651: Use keypress from tty-prompt
It was reverted back to own old implementation because there was a problem. See if https://github.com/piotrmurach/tty-prompt/issues/56 fixes it now.

kke:
Fixes #2651

Replaced the own custom "press any key to continue" dialog with the implementation from tty-prompt.

Upgraded to TTY-Prompt 0.13.1 which [solves the earlier problem](https://github.com/piotrmurach/tty-prompt/issues/56) in `#keypress` that left the countdown running in another thread after pressing a key and finally caused a timeout exception to halt the operation going on in main thread.


piotrmurach:
@kke I've released `v0.13.1` with a patch that fixes issues on reading input on Windows connected with the `#keypress` prompt.

PR #2667, opened 2017-08-07
Make stack yml interpolation default ENVs extendable for plugin devs

### LINKED ISSUE #2644: Allow selected ENVs to be passed through to stack
make this extendable: https://github.com/kontena/kontena/blob/master/cli/lib/kontena/cli/stacks/yaml/reader.rb#L62-L78

kke:
Fixes #2644

I think this is what you were after? You should now be able to overwrite the default_envs method and get more "factory envs" into the interpolation even during variable parsing.

ping @matti


matti:
This makes it a bit easier, but I think I was looking for a way to not overwrite anything - not sure what that would be.

Anyway this is a good change just for the code quality reasons.

SpComb:
You can already use `from: env: ASDF` to access arbitrary envs via opto, and declaring those as opto variables also means that they're persisted on the server for use in stack upgrades.

matti:
When you have 20 stacks and every single one has the 5 envs in them, you end up with massive repetition in stack ymls.

kke:
You may want to take a peek at https://github.com/jannishuebl/orchparty and especially https://github.com/jannishuebl/orchparty/issues/10 

Looks quite promising and a lot like the stack DSL I have been prototyping a bit.

PR #2669, opened 2017-08-07
CLI: Load environment variables from .kontena.env in current or parent directories

### LINKED ISSUE #2482: CLI: Implement .kontena-env support
It would be extra cool to have something like `.kontena-env` that could be used to switch master, grid, debug mode and perhaps even the cli version when changing to another directory.

Implementation should be as simple as `bin/kontena` trying to read the file in the current directory and sending the stuff to env.

Some universal utility for something like this is probably already available.

kke:
Fixes #2482

This PR adds dependency to `dotenv` (which handles loading of env-files) for loading any env variables from a `.kontena.env` file in the current working directory or any of its' parents.


kke:
Dotenv also supports some level of substitutions, such as:

`DATABASE_URL="postgres://$(whoami)@localhost/my_database"`

More at https://github.com/bkeepers/dotenv

jakolehm:
How much overhead this adds?

kke:
Overhead? For spin up time or..?

kke:
Dotenv has no dependencies

jakolehm:
Yes, with overhead I meant spinup time.

kke:
Of course depends on what you have in the env-file, but for a simple `KONTENA_MASTER=master-name` looks like ~2ms.

kke:
Dotenv does not look for the file in any parent directories by default.

That could be done by using `Dotenv.load(*Pathname.new('.').realpath.ascend.map { |p| p.join('.kontena-env') }.map(&:to_s))`

kke:
Updated with latest dotenv and loading from parent directories, also changed the `.kontena-env` to `.kontena.env` to enable highlighting in editors.

image

kke:
Just use autoenv  or such.

PR #2676, opened 2017-08-08
Fix agent Docker.info caching

### LINKED ISSUE #2237: kontena-agent needs a restart to notice new v1 volume plugins
I'm using the volume driver https://portworx.com/ for volume persistence.
There are two problems in using this custom driver with kontena:

1 - The driver itself is registered by a portworx container. When that container is not started, kontena-agent do not recognize pxd driver. All containers that declare volumes with this driver are not deployed, and if for some reason the driver is unregistred even for a moment, kontena stop the containers (I'm not sure if that is the case but I'm noticing instability and services are dropping, when that happen the command `kontena node show` does _not_ show the pxd volume driver). The driver container need to start **before** kontena-agent, but here comes the second problem
2 - portworx use etcd or consul to orquestrate a cluster of persistent storage and replication volumes. I don't want to maintain two etcd installs, and use kontena etcd, which works great, but portworx driver needs to start **after** kontena-agent which is Chicken and egg problem :-)

What I do now:

1 - Start the node
2 - Wait for pxd service
3 - `kontena node show` will not recognize the driver
4 - Restart kontena-agent. Wait a few moments
5 - `kontena node show` will recognize the driver
6 -  Manual redeploy stack with `kontena stack deploy`

What I whould like to do:

Use a volume with **grid** or **stack** scope and disable the check for the volume driver in order to force the deployment (not so good solution, but I think is necessary to control this behavior in problems like this). I'm using a affinity filter, it should not have problems.

Or, I would like to let the check for the driver on, but kontena needs to be way more tolerant in order to work with not static drivers like this. (maybe checking more often on volume driver information, and having a delay before removing/stopping/rescheduling container service) 

Whould be great to run portworx as a stack with kontena, it was my first try, but I'm running as a systemd service to see if improves stability

SpComb:
Fixes #2237

The agent `NodeInfoWorker` would cache the `Docker.info` at startup, and use the same values for each `publish_node_info`. This would fail to notice any new volume drivers that appear after agent startup.

### DIFF
--- a/agent/lib/kontena/workers/node_info_worker.rb
+++ b/agent/lib/kontena/workers/node_info_worker.rb
@@ -49,21 +49,22 @@ module Kontena::Workers
 
     def publish_node_info
       debug 'publishing node information'
-      node_info = docker_info.dup
+      node_info = Docker.info()
       node_info['PublicIp'] = self.public_ip
       node_info['PrivateIp'] = self.private_ip
       node_info['AgentVersion'] = Kontena::Agent::VERSION
       node_info['Drivers'] = {
-        'Volume' => volume_drivers,
-        'Network' => network_drivers
+        'Volume' => volume_drivers(node_info),
+        'Network' => network_drivers(node_info),
       }
       rpc_client.request('/nodes/update', [@node_id, node_info])
     rescue => exc
       error "publish_node_info: #{exc.message}"
     end
 
+    # @param docker_info [Hash]
     # @return [Array<Hash>]
-    def volume_drivers
+    def volume_drivers(docker_info)
       drivers = []
       plugins.each do |plugin|
         config = plugin['Config']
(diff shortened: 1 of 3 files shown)


agent/lib/kontena/workers/node_info_worker.rb
```diff
@@ -49,21 +49,22 @@ def on_node_info(topic, node)
 
     def publish_node_info
       debug 'publishing node information'
-      node_info = docker_info.dup
+      node_info = Docker.info()
```
jakolehm: No need for `()`.

SpComb: I vaugely like the parens for things that are actual method calls with side-effects, not just cheap attribute accesses... TBH I also think those should be verbs, like `Docker.get_info()`

PR #2678, opened 2017-08-09
Fix nil resource usage error on node show cmd

jnummelin:
When node just joins grid and has not yet reported any resource usages CLI fails `node show` command with nasty `[error] NoMethodError : undefined method '[]' for nil:NilClass`
Backtrace:
```
/Users/jussi/code/kontena/cli/lib/kontena/cli/nodes/show_command.rb:47:in `execute': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/jussi/code/kontena/cli/lib/kontena/command.rb:202:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
	from /Users/jussi/code/kontena/cli/lib/kontena/command.rb:202:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
	from /Users/jussi/code/kontena/cli/lib/kontena/command.rb:202:in `run'
	from /Users/jussi/.rvm/gems/ruby-2.3.3/gems/clamp-1.1.2/lib/clamp/command.rb:132:in `run'
	from ./bin/kontena:18:in `<main>'
```

Fixed by using `dig` rather than chained `[]` calls

### DIFF
--- a/cli/lib/kontena/cli/nodes/show_command.rb
+++ b/cli/lib/kontena/cli/nodes/show_command.rb
@@ -43,9 +43,10 @@ module Kontena::Cli::Nodes
         mem_used = mem['used'] - (mem['cached'] + mem['buffers'])
         puts "    memory: #{to_gigabytes(mem_used, 2)} of #{to_gigabytes(mem['total'], 2)} GB"
       end
-      if node['resource_usage']['filesystem']
+      fs = node.dig('resource_usage','filesystem')
+      if fs
         puts "    filesystem:"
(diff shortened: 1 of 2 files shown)


kke:
(That whole puts mess could have been avoided by building a hash and `YAML.dump`ing it since the output is practically valid YAML as is anyway)

PR #2681, opened 2017-08-10
[WIP] Docker plugin installation through CLI

### LINKED ISSUE #2320: Volume driver management
```
$ kontena volume driver add rexray/ebs:0.8.2 [--label region=eu-west-1] --kind block [name]
```

When user adds a driver to Kontena Master it is automatically distributed as a volume plugin to matching nodes (based on label filter).

Volume driver subcommand should have basic list/show/rm features.

jnummelin:
This PR makes installing Docker plugins on nodes possible through Kontena CLI.

## Questions

As each plugin (regardless of it's type) is installed exactly the same way, should the command be moved to be e.g. `kontena node plugin install ...`?

## TODO

- [ ] Collect each nodes result
- [ ] Make each node run installation in parallel
- [ ] Docs

fixes #2320


server/app/mutations/volumes/plugin_install.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+      string :label
+    end
+
+    def validate
+
+    end
+
+    def execute
+      # TODO
```
jakolehm: Why are we sending rpc calls here? Imho more robust way is to pass plugin information to agent via agent pull (node info maybe?)

jakolehm:
> As each plugin (regardless of it's type) is installed exactly the same way, should the command be moved to be e.g. kontena node plugin install ...?

No. It's not very convenient to run install command to every node... whole point of this install command is to automate things as much as possible. Meaning: you can run plugin install without any nodes, nodes will get matching plugins installed automatically when they join.

jnummelin:
> No. It's not very convenient to run install command to every node
Regardless of the installation way (pull/push mode), I think it makes sense to have the command somewhere else than under `volumes`. Yes, volumes are the main use case for this but any other type of plugin is installed on the same way. So I thought the command could be under `kontena node plugins ...`. It's true that most of the sub-commands under `kontena node` are for single node, but would it be too bad to have even the pull mode installation as `kontena node plugin install ...`?

server/app/mutations/volumes/plugin_install.rb
jnummelin: Rcp push was more simple and straightforward. Yes, it has the drawback of installing plugins only on request and not automatically after a new node joins the grid.

jakolehm:
Other `kontena node` commands are for single node, it's confusing to mix there something else. Maybe  `kontena grid plugin` would make more sense?

SpComb:
Bumping WIP PR to 1.5, needs some major design refactoring (declarative pull vs RPC push).

jnummelin:
rebased

PR #2684, opened 2017-08-11
Remove unnecessary request parsing in auth middleware

kke:
The token authentication middleware in server API had some leftover call to `Rack::Request.new(env)` and the result was being assigned to an instance variable that was never used.


SpComb:
What makes it thread unsafe?

kke:
Rack Middleware should not mutate state.

In theory that could have leaked requests from one thread to another and in case it crashed, they might have shown up in some error output. But in practice it was probably harmless, except for the extra work it had to do for every request.

PR #2686, opened 2017-08-11
CLI: Exit with error from node ssh --any if no nodes are online

kke:
This PR makes `kontena node ssh --any` exit with proper error message when there are no connected nodes. 

Before:

```console
$ kontena node ssh --any
 [error] NoMethodError : undefined method `[]' for nil:NilClass
         See /Users/kimmo/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception
```

After:

```console
$ kontena node ssh --any
 [error] There are no online nodes
```

### DIFF
--- a/cli/spec/kontena/cli/nodes/ssh_command_spec.rb
+++ b/cli/spec/kontena/cli/nodes/ssh_command_spec.rb
@@ -16,17 +16,23 @@ describe Kontena::Cli::Nodes::SshCommand do
     allow(client).to receive(:get).with('nodes/test-grid/test-node').and_return(node)
   end
 
-  context '--any' do
-    it "fails if using both --any and a node name as a command" do
-      expect(subject).to_not receive(:exec)
-      expect{subject.run(['--any', 'ls', '-l'])}.to exit_with_error.and output(/Cannot combine --any with a node name/).to_stderr
+  describe '--any flag' do
+    context 'used together with a node name' do
+      it "fails and outputs an error message" do
+        expect(subject).to_not receive(:exec)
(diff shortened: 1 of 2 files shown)


cli/spec/kontena/cli/nodes/ssh_command_spec.rb
```diff
@@ -16,9 +16,18 @@
     allow(client).to receive(:get).with('nodes/test-grid/test-node').and_return(node)
   end
 
-  it "fails if using both --any and a node name as a command" do
-    expect(subject).to_not receive(:exec)
-    expect{subject.run(['--any', 'ls', '-l'])}.to exit_with_error.and output(/Cannot combine --any with a node name/).to_stderr
+  context '--any' do
```
SpComb: If you want to group up the examples that are using `--any`, then this should probably be `describe`, not `context`. They're technically the same thing, but `context` should be for what the state the server nodes are in, and `describe` for what functionality the specs test: http://lmws.net/describe-vs-context-in-rspec

```ruby
describe '--any' do
  it 'fails if given a node name' do
    ...
  end
  context 'with all nodes offline' do
    before do
      allow(client).to receive(:get).with(...).and_return(...)
    end

    it 'fails' do
      ...
    end
  end
end
```

The resulting rspec descriptions should read nicely:

* `Kontena::Cli::Nodes::SshCommand --any fails if given a node name`
* `Kontena::Cli::Nodes::SshCommand --any with all nodes offline fails`

kke:
Failing specs are in master

PR #2687, opened 2017-08-11
Fix node show specs to include availability attribute

jnummelin:
Fixes broken CLI specs in master branch.


SpComb:
Broken in #2306 merge after #2678 added a second spec case with separate expected output

PR #2689, opened 2017-08-13
Implement container restart policies in the agent ServicePodWorker

jakolehm:
Less mayhem when Docker engine (or host) is restarted. Agent is already capable of handling restarts because it tries to ensure desired state.


jnummelin:
When service container exits, how long does it take to restart it?

jakolehm:
0-30s (which is `ServicePodManager` loop delay). One option is to trigger `ServicePodWorker#apply` after it notices that container has died and desired state is still "running".

jnummelin:
I'd prefer "immediate" actions to restart the container. The worst case 30s is IMO bit long to wait for service to restart.

jakolehm:
Yes, immediate first action is better, we just need to implement backoff logic.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -16,12 +16,17 @@ class ServicePodWorker
     attr_reader :node, :prev_state, :service_pod
     attr_accessor :service_pod, :container_state_changed
 
+    DIE_EVENTS = ['die', 'kill'].freeze
```
jnummelin: I think only `die` should be handled. `docker kill/stop xyz` emits also `die` event so AFAIK it should be safe to only handle `die` event. Handling both makes the restart logic to be executed twice in case of `docker kill/stop xyz` as that emits both `kill`and `die`.

`docker kill nginx-1` results in:
```
kontena-agent | I, [2017-08-15T10:15:51.164164 #1]  INFO -- Kontena::Workers::ServicePodWorker: restarting nginx-1 because it has stopped
kontena-agent | D, [2017-08-15T10:15:51.164448 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of nginx-1: running
kontena-agent | D, [2017-08-15T10:15:51.177731 #1] DEBUG -- Kontena::RpcClient: waited 0.0s of 30.0s until: request /containers/event has response wth id=2044946063 yielded Array
kontena-agent | D, [2017-08-15T10:15:51.187778 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state is in-sync: running
kontena-agent | D, [2017-08-15T10:15:51.247931 #1] DEBUG -- Kontena::RpcClient: waited 0.1s of 30.0s until: request /containers/save has response wth id=63854677 yielded Array
kontena-agent | I, [2017-08-15T10:15:51.381047 #1]  INFO -- Kontena::Workers::ServicePodWorker: restarting nginx-1 because it has stopped (delay: 1s)
kontena-agent | D, [2017-08-15T10:15:51.393973 #1] DEBUG -- Kontena::RpcClient: waited 0.0s of 30.0s until: request /containers/event has response wth id=382750331 yielded Array
kontena-agent | D, [2017-08-15T10:15:52.382805 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of nginx-1: running
kontena-agent | D, [2017-08-15T10:15:53.866767 #1] DEBUG -- Kontena::RpcClient: waited 0.0s of 30.0s until: request /containers/event has response wth id=1151646354 yielded Array
```

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -62,17 +67,63 @@ def on_container_event(topic, e)
...
+      end
+
+      return if apply_in_progress?
+
+      # cancel counters because we are called again
+      cancel_restart_timers
+
+      # backoff restarts
+      backoff = @restarts ** 2
+      backoff = MAX_RESTART_BACKOFF if backoff > MAX_RESTART_BACKOFF
```
jnummelin: The max time is actually limited by the server sync-loop interval. And as the server sync-loop essentially also restarts died container, it makes the whole backoff period quite un-predictable. Would it be possible to make the sync-loop to respect the backoff policy? I.e. in the sync-loop "skip" starting the container if there's existing restart loop ongoing.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -62,17 +67,63 @@ def on_container_event(topic, e)
...
+
+    def cancel_restart_timers
+      @restart_counter_reset_timer.cancel if @restart_counter_reset_timer
+      @restart_backoff_timer.cancel if @restart_backoff_timer
+    end
+
+    def apply_restart
+      apply
+      # reset restart counter if instance stays up 10s
+      @restart_counter_reset_timer = after(10) {
```
jnummelin: I assume that the timers will be properly terminated when the pod is removed and the actor terminated? Or would some cleanup be required on actor finalizer?

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Good point, changed to match manager loop interval.

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Now it only acts on `die` event.

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Celluloid should cleanup actor timers, but I need to confirm that it actually works as expected.

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Oh, misread your proposal. It might be quite hard sync to skip apply if there is restart still going on (because backoff)?

jnummelin: yes, and it should be able to differentiate sync-loop vs. actual triggered deploy also. sounds like too many flags for little benefits. :)

jnummelin:
Wonder how this works with existing `restart=always` containers. Need to test a bit...

jakolehm:
> Wonder how this works with existing restart=always containers. Need to test a bit...

I think we can do a migration that removes policy?

jnummelin:
Seems to work pretty ok, tested with existing `unless-stopped` policy container. There seems to be some duplication in state logging but I don't see any duplicated start/restart events happening.

jnummelin:
> I think we can do a migration that removes policy?

We could, but as it seems to work ok even without one I don't see it as mandatory.

jakolehm:
@jnummelin PTAL

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -62,17 +67,71 @@ def on_container_event(topic, e)
...
+    def restarting?
+      @restarts > 0
     end
 
     def destroy
       @service_pod.mark_as_terminated
       apply
     end
 
+    def apply_in_progress?
```
SpComb: The `@apply_in_progress = true/false` is set from an `exclusive { ... }` block.... it should not be possible for the `on_container_event` -> `handle_restart_on_die` to ever be `apply_in_progress?`...? AFAIK celluloid will queue up any container event notifications while the exclusive apply is running, and then process all of those once the exclusive apply block returns, and at that point, they will no longer be `@apply_in_progress`?

SpComb: Yes... a `kontena service deploy --force` will trigger a total of three `apply`s now, because all of the container events from the initial apply queue up and get processed:

```
D, [2017-08-21T11:31:45.207297 #1] DEBUG -- Kontena::RpcServer: rpc request: Kontena::Rpc::ServicePodsApi#notify_update [[]]
D, [2017-08-21T11:31:45.557896 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-test-1: running
I, [2017-08-21T11:31:45.596364 #1]  INFO -- Kontena::Workers::ServicePodWorker: re-creating redis-test-1
I, [2017-08-21T11:31:45.597022 #1]  INFO -- Kontena::ServicePods::Creator: creating service: redis-test-1
I, [2017-08-21T11:31:47.963434 #1]  INFO -- Kontena::ServicePods::Creator: removing previous version of service: redis-test-1
D, [2017-08-21T11:31:48.488478 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: redis-test-1
D, [2017-08-21T11:31:48.630464 #1] DEBUG -- Kontena::ServicePods::Creator: container created: redis-test-1
I, [2017-08-21T11:31:49.001441 #1]  INFO -- Kontena::ServicePods::Creator: service started: redis-test-1
D, [2017-08-21T11:31:49.115770 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=kill => container_state_changed!
D, [2017-08-21T11:31:49.116685 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=die => container_state_changed!
D, [2017-08-21T11:31:49.118750 #1] DEBUG -- Kontena::Workers::ServicePodWorker: handle_restart_on_die: pod running?=true apply_in_progress?=false restarts=0
I, [2017-08-21T11:31:49.118931 #1]  INFO -- Kontena::Workers::ServicePodWorker: restarting redis-test-1 because it has stopped
D, [2017-08-21T11:31:49.119201 #1] DEBUG -- Kontena::Workers::ServicePodWorker: handle_restart_on_die => apply_restart if true
D, [2017-08-21T11:31:49.119442 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-test-1: running
D, [2017-08-21T11:31:49.264550 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state is in-sync: running
D, [2017-08-21T11:31:49.304540 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=stop => container_state_changed!
D, [2017-08-21T11:31:49.304910 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=destroy => container_state_changed!
D, [2017-08-21T11:31:49.305751 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=create => container_state_changed!
D, [2017-08-21T11:31:49.305928 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=start => container_state_changed!
D, [2017-08-21T11:31:59.304620 #1] DEBUG -- Kontena::Workers::ServicePodWorker: redis-test-1 stayed up 10s, resetting restart backoff counter
D, [2017-08-21T11:32:01.854993 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-test-1: running
D, [2017-08-21T11:32:01.899455 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state is in-sync: running
```

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -273,5 +334,12 @@ def sync_state_to_master(current_state, error = nil)
     def log_service_pod_event(type, data, severity = Logger::INFO)
       super(service_pod.service_id, service_pod.instance_number, type, data, severity)
     end
+
+    private
+
+    def shutdown
+      @restart_backoff_timer.cancel if @restart_backoff_timer
```
SpComb: I doubt cleaning up the timers in the actor shutdown is necessary... looking at the code, the `@timers` are ignored once the `Celluloid::Actor#run` loop calls `shutdown`?

```
Celluloid::Cell#shutdown Kontena::Workers::ServicePodWorker @ 
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:306:in `shutdown'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:169:in `run'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:131:in `block in start'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/thread_handle.rb:14:in `block in initialize'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:78:in `block in get_thread'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/group/spawner.rb:50:in `block in instantiate'
```

SpComb:
This also triggers on `kontena service restart`... it doesn't cause duplicate restarts, but the logging is misleading... perhaps those RPC restart requests should eventually also be routed to the `ServicePodManager` so that it knows what's going on, like the health check restarts?

```
D, [2017-08-21T11:50:51.052170 #1] DEBUG -- Kontena::RpcServer: rpc request: Kontena::Rpc::ServicePodsApi#restart ["598c54866435f900112357ce", 3]
I, [2017-08-21T11:50:51.160676 #1]  INFO -- Kontena::ServicePods::Restarter: restarting service: /redis-test-3
D, [2017-08-21T11:50:51.180151 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=kill => container_state_changed!
D, [2017-08-21T11:50:51.344473 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event status=die => container_state_changed!
D, [2017-08-21T11:50:51.355736 #1] DEBUG -- Kontena::Workers::ServicePodWorker: handle_restart_on_die: pod running?=true apply_in_progress?=false restarts=0
I, [2017-08-21T11:50:51.356536 #1]  INFO -- Kontena::Workers::ServicePodWorker: restarting redis-test-3 because it has stopped
D, [2017-08-21T11:50:51.346579 #1] DEBUG -- Kontena::RpcClient: waited 0.1s of 30.0s until: request /containers/event has response wth id=1254242128 yielded Array
D, [2017-08-21T11:50:51.356769 #1] DEBUG -- Kontena::Workers::ServicePodWorker: handle_restart_on_die => apply_restart if true
D, [2017-08-21T11:50:51.357205 #1] DEBUG -- Kontena::Workers::ServicePodWorker: state of redis-test-3: running
```

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -243,7 +304,7 @@ def current_state
 
       if service_container.running?
         'running'
-      elsif service_container.restarting?
+      elsif restarting?
         'restarting'
```
SpComb: Unable to get this to report `state: 'restarting'` in my testing... this `current_state` call happens immediately after the `ensure_started`, and at that point the container is still likely to be `running`. 

For this state to be useful and show up in `kontena service ls`, perhaps the worker should also `sync_state_to_master` at the point when it schedules the restart backoff. Also, consider keeping the `restarting` state for the `@restart_counter_reset_timer` duration, and only have it return to `running` once that expires?

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Remove the `@apply_in_progress` if the apply is exclusive, it doesn't work and it's misleading.

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Remove the unnecessarily timer-cancel/finalizer unless there's some actual reason for it, AFAIK it's not needed.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -62,17 +67,71 @@ def on_container_event(topic, e)
...
+    def max_restart_backoff
+      Kontena::Workers::ServicePodManager::LOOP_INTERVAL
+    end
+
+    def cancel_restart_timers
+      @restart_counter_reset_timer.cancel if @restart_counter_reset_timer
+      @restart_backoff_timer.cancel if @restart_backoff_timer
+    end
+
+    def apply_restart
```
SpComb: I don't see why there needs to be a separate `apply_restart`, why not include the `@restart_counter_reset_timer` -> `@restarts = 0` logic into the normal `apply`?

ATM if the `on_container_event` -> `handle_restart_on_die` schedules an `apply_restart`, but the `ServicePodManager` loop kicks in before the `@restart_backoff_timer` fires, then the `update` -> `apply` will actually restart the container, *without* resetting the `@restarts` counter. I'm not sure if the `@restart_backoff_timer` will actually end up resetting the `@restarts` counter in that case; it depends on what celluloid will run first after the exclusive apply block exits: does it run the `@restart_backoff_timer` with `@container_state_changed` still set to false, or does it first process the `on_container_event` with the `start` event that sets `@container_state_changed` back to true, allowing the `@restart_backoff_timer` to call `apply_restart` and reset the `@restarts` counter? I don't want to have to test it out...

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: If it is needed, then do we need to add something similar to the other actors that use timers as well?

jakolehm:
@SpComb logic is now simplified, PTAL.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -16,12 +17,16 @@ class ServicePodWorker
...
 
+    # @param [Node] node
+    # @param [ServicePod] service_pod
     def initialize(node, service_pod)
       @node = node
       @service_pod = service_pod
       @prev_state = nil # sync'd to master
       @container_state_changed = true
       @deploy_rev_changed = false
+      @apply_in_progress = false
```
SpComb: Dead code

jakolehm: Removed.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -62,7 +68,42 @@ def on_container_event(topic, e)
       if attrs['io.kontena.service.id'] == @service_pod.service_id &&
           attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number
```
SpComb: The event also includes the `attrs['io.kontena.container.deploy_rev']`. Ignore the event if that doesn't match `@service_pod.deploy_rev`?

jakolehm: That makes sense.

jakolehm: Implemented better event check.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -60,9 +66,46 @@ def deploy_rev_changed?
     def on_container_event(topic, e)
       attrs = e.actor.attributes
       if attrs['io.kontena.service.id'] == @service_pod.service_id &&
-          attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number
+          attrs['io.kontena.container.type'] == 'service' &&
```
SpComb: Isn't it `io.kontena.container.type=container`?

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: Hmm, no, the `io.kontena.container.deploy_rev` doesn't work. The `@service_pod.deploy_rev` will change on each deploy without the container getting recreated, so the event `deploy_rev` label can be something much older. It's not useful for detecting old events from before the apply:

```
I, [2017-09-04T13:21:23.760931 #1]  INFO -- Kontena::Workers::ServicePodWorker: starting restart-test-1
D, [2017-09-04T13:21:24.148467 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event start: container.deploy_rev=2017-09-04 12:56:59 UTC service_pod.deploy_rev=2017-09-04 13:21:21 UTC)
D, [2017-09-04T13:21:24.149617 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event restart: container.deploy_rev=2017-09-04 12:56:59 UTC service_pod.deploy_rev=2017-09-04 13:21:21 UTC)
D, [2017-09-04T13:21:29.989317 #1] DEBUG -- Kontena::Workers::ServicePodWorker: on_container_event die: container.deploy_rev=2017-09-04 12:56:59 UTC service_pod.deploy_rev=2017-09-04 13:21:21 UTC)
```

I think it needs the `@container_id` approach, set from `apply`. They are not the same thing.

SpComb: Testing with something like this seems to work?

```patch
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -65,14 +65,10 @@ module Kontena::Workers
 
     # @param [String] topic
     # @param [Docker::Event] e
-    def on_container_event(topic, e)
-      attrs = e.actor.attributes
-      if attrs['io.kontena.service.id'] == @service_pod.service_id &&
-          attrs['io.kontena.container.type'] == 'service' &&
-          attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number &&
-          attrs['io.kontena.container.deploy_rev'].to_s == @service_pod.deploy_rev
+    def on_container_event(topic, event)
+      if @container && event.id == @container.id
         @container_state_changed = true
-        handle_restart_on_die if e.status == 'die'.freeze
+        handle_restart_on_die if event.status == 'die'.freeze
       end
     end
 
@@ -119,7 +115,7 @@ module Kontena::Workers
       cancel_restart_timers
       exclusive {
         begin
-          ensure_desired_state
+          @container = ensure_desired_state
           # reset restart counter if instance stays up 10s
           @restart_counter_reset_timer = after(10) {
             info "#{@service_pod.name_for_humans} stayed up 10s, resetting restart backoff counter" if restarting?
@@ -140,15 +136,16 @@ module Kontena::Workers
       }
     end
 
+    # @return [Docker::Container, nil]
     def ensure_desired_state
       if service_pod.running? && service_container.nil?
         info "creating #{service_pod.name}"
-        ensure_running
+        service_container = ensure_running
       elsif service_pod.running? && (service_container && service_container_outdated?(service_container))
         info "re-creating #{service_pod.name}"
-        ensure_running
+        service_container = ensure_running
       elsif service_container && service_pod.running? && !service_container.running?
         info "starting #{service_pod.name}"
         ensure_started
@@ -158,6 +155,7 @@ module Kontena::Workers
       elsif service_pod.terminated?
         info "terminating #{service_pod.name}"
         ensure_terminated if service_container
+        service_container = nil
       elsif service_pod.desired_state_unknown?
         info "desired state is unknown for #{service_pod.name}, not doing anything"
       elsif state_in_sync?(service_pod, service_container)
@@ -165,8 +163,11 @@ module Kontena::Workers
       else
         warn "unknown state #{service_pod.desired_state} for #{service_pod.name}"
       end
+
+      service_container
     end
 
+    # @return [Docker::Container]
     def ensure_running
       Kontena::ServicePods::Creator.new(service_pod).perform
     rescue => exc
```

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -57,14 +63,52 @@ def deploy_rev_changed?
...
-      attrs = e.actor.attributes
+    # @param topic [String]
+    # @param event [Docker::Event]
+    def on_container_event(topic, event)
+      attrs = event.actor.attributes
       if attrs['io.kontena.service.id'] == @service_pod.service_id &&
+          attrs['io.kontena.container.type'] == 'container'.freeze &&
           attrs['io.kontena.service.instance_number'].to_i == @service_pod.instance_number
         @container_state_changed = true
+        handle_restart_on_die if event.status == 'die'.freeze
```
SpComb: Separate issue for fixing the `on_container_event` to ignore old events, and not think that the container died and needs to be restarted immediately after each redeploy: #2766

PR #2560, opened 2017-07-05
Rewrite agent websocket client

SpComb:
Fixes #2500 by replacing the EventMachine-based `faye-websocket` client with the new `kontena-websocket-client` running as a Celluloid actor, which has full support for SSL certificate verification.

* Drop `eventmachine` and `faye-websocket` dependencies

    Now with 100% less EventMachine.

* Add `kontena-websocket-client` dependency

    Supports SSL certification verification.

* Agent `WebsocketClient` is now a celluloid actor

    The new `Kontena::Websocket::Client` should implement open/ping/close timeouts with keepalive pings at least as robustly as the old EM-based client, probably even more robustly.

* Fix agent RPC server/client websocket message sends to log a warning and abort the celluloid call on websocket send errors

### Default behavior with `KONTENA_URI=wss://` + invalid cert

Intended to be backwards-compatible, but show a warning with the validation error + cert details.

```
I, [2017-07-13T12:31:42.360771 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://kontena.test:9293
W, [2017-07-13T13:48:19.964864 #1]  WARN -- Kontena::WebsocketClient: insecure connection established with SSL errors: certificate verify failed: self signed certificate: /CN=kontena.test (issuer /CN=kontena.test)
```

### Default behavior with `KONTENA_URI=wss://` + valid cert

Backwards compatible, but suggesting to configure `KONTENA_SSL_VERIFY=true` to protect from MITM attacks:

```
I, [2017-07-14T05:31:49.094554 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://aws-master-1.kontena.io/
W, [2017-07-14T05:31:54.470200 #1]  WARN -- Kontena::WebsocketClient: secure connection established without KONTENA_SSL_VERIFY=true: /CN=aws-master-1.kontena.io (issuer /C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3)
```

### Error case with `KONTENA_SSL_VERIFY=true` (not signed by default CAs)

```
I, [2017-07-13T12:30:35.566300 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://kontena.test:9293
E, [2017-07-13T12:30:35.619952 #1] ERROR -- Kontena::WebsocketClient: unable to connect to SSL server with KONTENA_SSL_VERIFY=true: certificate verify failed: self signed certificate
```

### Error case with verify + CA, but wrong host
```
[...]

PR #2690, opened 2017-08-14
Agent: replace wait_until with condition

jakolehm:


agent/lib/kontena/rpc_client.rb
```diff
@@ -52,21 +53,28 @@ def notification(method, params)
...
+      @requests[id] = cond
+
+      start = Time.now
       websocket_client.send_request(id, method, params)
 
-      if !wait_until("request #{method} has response wth id=#{id}", timeout: timeout, interval: 0.01) { @requests[id] }
+      begin
+        cond.wait(timeout)
+      rescue Celluloid::ConditionError
+        warn "request timeout after waiting #{timeout} seconds"
```
jakolehm: @SpComb added this because specs checked warn, is it really necessary to log because this also raises `TimeoutError`?

SpComb: Several places call `rpc_client.async.request(...)`, so those RPC failures will be silent unless the `RpcClient` logs them.

OTOH it looks like there are now two warns here, both the `wait_until` warning and the `rescue` -> `warn + abort`.

agent/lib/kontena/rpc_client.rb
```diff
@@ -52,21 +53,28 @@ def notification(method, params)
...
+      id = request_id
+      cond = Celluloid::Condition.new
+      @requests[id] = cond
+
+      start = Time.now
       websocket_client.send_request(id, method, params)
 
-      if !wait_until("request #{method} has response wth id=#{id}", timeout: timeout, interval: 0.01) { @requests[id] }
+      begin
+        cond.wait(timeout)
```
SpComb: There's theoretically a race condition here if the `websocket_client.send_request` -> websocket write -> server rpc handler -> websocket read -> handle_response -> condition.signal` happens before the `websocket_client.send_request` returns and this `request` task calls `cond.wait`.

I don't know if that's really possible in practice, though.

SpComb: Background thought here: implementing some kind of `Observable#wait` API to replace the `Celluloid::Condition` would be a much better solution here :)

* Avoid the dual use of the `@requests[id]` for storing both a `Celluloid::Condition` and `[result, error]` - it would just be an `Observable` with an `observable_value`
* Avoid the theoretical race condition by having the `wait` return immediately once the observable value is updated

agent/lib/kontena/rpc_client.rb
SpComb: IOW that would be a bug from #2560, but I'm not sure exactly how the warning from the rescue shows up? If that looks reasonable, then I don't see why we need this warning... maybe improve the `raise TimeoutError` to include the timeout seconds instead?

BTW: the actual wait time can be longer than the timeout, it would be preferrable to log/raise with the `Time.now - start` instead.

agent/lib/kontena/rpc_client.rb
SpComb: #2704 implements the necessary Observable/Observer changes to allow using them as a replacement for `Celluloid::Condition`.

jakolehm:
Let's proceed with #2704 .

SpComb:
Note that in this case the `Observable` behaves exactly like a Future. Indeed, it would make more sense to use `Celluloid::Future` here, rather than the `Celluloid::Condition`.

I don't think it's documented, but it looks like you can use `@future = Celluloid::Future.new; @future.value` and then `@future << response` from a different task.

SpComb:
Added a benchmark over in #2704 - the observer and future seem virtually identical - which isn't surprising, they're both implemented using `Celluloid.receive`. The `Celluloid::Condition` is implemented using the lower-level actor interfaces, so it makes more sense that it performs better than the `Celluloid.receive` interface:

```
wait        1.740000   0.150000   1.890000 (  2.666116)
condition   1.160000   0.270000   1.430000 (  2.455603)
observer    1.240000   0.310000   1.550000 (  2.442276)
future      1.230000   0.240000   1.470000 (  2.470067)
                    delay
wait             0.007141
condition        0.000882
observer         0.001777
future           0.001125
```

PR #2691, opened 2017-08-14
[WIP] Improve container health check logic

jakolehm:
Let `ServicePodWorker` to handle & report container health. `ContainerHealthCheckWorker` only responsibility is to check & report status.

### DIFF
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -13,6 +13,8 @@ module Kontena::Workers
 
     finalizer :log_exit
 
+    Event = Struct.new(:service_id, :instance_number, :health)
+
     # @param [Docker::Container] container
     def initialize(container)
       @container = container
@@ -53,10 +55,16 @@ module Kontena::Workers
       if msg['status'] == 'unhealthy'
         name = @container.labels['io.kontena.container.name']
         # Restart the container, master will handle re-scheduling logic
-        info "About to restart container #{name} as it's reported to be unhealthy"
-        emit_service_pod_event("service:health_check", "restarting #{name} because it's reported as unhealthy", Logger::WARN)
-        
-        restart_container
+        info "#{name} is unhealthy"
+        emit_service_pod_event("service:health_check", "#{name} is unhealthy", Logger::WARN)
+
+        publish('container:health_check', Event.new(
+          @container.service_id, @container.instance_number, 'unhealthy'.freeze
+        ))
+      elsif msg['status'] == 'healthy'
+        publish('container:health_check', Event.new(
+          @container.service_id, @container.instance_number, 'healthy'.freeze
+        ))
       end
     end
 
@@ -104,10 +112,6 @@ module Kontena::Workers
         data['status'] = 'unhealthy'
       end
       data
-    end
-
-    def restart_container
-      Kontena::ServicePods::Restarter.new(@container.service_id, @container.instance_number).perform
     end
 
     # @param [String] type
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -13,8 +13,33 @@ module Kontena::Workers
     include Kontena::Helpers::RpcHelper
     include Kontena::Helpers::EventLogHelper
 
-    attr_reader :node, :prev_state, :service_pod
+    attr_reader :node, :prev_state, :service_pod, :health
     attr_accessor :service_pod, :container_state_changed
+
+    class Health
+      HEALTHY = 'healthy'.freeze
+      UNHEALTHY = 'unhealthy'.freeze
+
+      attr_reader :status, :count
+
+      def initialize
+        @status = 'unknown'.freeze
+      end
+
+      def status=(status)
+        if @status != status
+          @count = 1
+        else
+          @count += 1
+        end
+
+        @status = status
+      end
+
+      def unhealthy?
+        @status == UNHEALTHY
+      end
+    end
 
     def initialize(node, service_pod)
       @node = node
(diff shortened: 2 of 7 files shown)


agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -13,16 +13,43 @@ class ServicePodWorker
...
+
+      attr_reader :status, :count
+
+      def initialize
+        @status = 'unknown'.freeze
+      end
+
+      def status=(status)
+        if @status != status
+          @count = 1
```
SpComb: Initialize `@count = 0` just for sanity

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -65,6 +92,16 @@ def on_container_event(topic, e)
       end
     end
 
+    # @param [String] topic
+    # @param [ContainerHealthCheckWorker::Event] event
+    def on_container_health_check(topic, event)
+      return if event.service_id != @service_pod.service_id
```
SpComb: Also `instance_number`

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -13,16 +13,43 @@ class ServicePodWorker
...
+      HEALTHY = 'healthy'.freeze
+      UNHEALTHY = 'unhealthy'.freeze
+
+      attr_reader :status, :count
+
+      def initialize
+        @status = 'unknown'.freeze
+      end
+
+      def status=(status)
```
SpComb: Some `update` verb instead of `status=` with side effects?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -95,6 +132,8 @@ def ensure_desired_state
       if service_pod.running? && service_container.nil?
         info "creating #{service_pod.name}"
         ensure_running
+      elsif service_pod.running? && (service_container && service_container.running? && health.unhealthy?)
+        ensure_healing
```
SpComb: Shouldn't the `service_container_outdated?` check come before the `unhealthy?` check? This would block deployments for an unhealthy service?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -140,6 +179,32 @@ def ensure_started
...
     end
 
+    def ensure_healing
+      if health.count < 5
+        info "restarting #{service_pod.name} because it's unhealthy"
+        restart_instance
+      elsif health.count == 5
+        info "re-creating #{service_pod.name} because it's unhealthy"
+        ensure_running(pull: false)
+      elsif health.count % 5 == 0
```
SpComb: Wut, this `health.count % 5` condition seems very odd... how this will behave depends on what the relative phase of the `ServicePodManager` loop and the `ContainerHealthCheckWorker` interval loop? i.e. I have no idea how this will behave? It might not actually do anything, because the `else` condition is a no-op?

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -13,16 +13,43 @@ class ServicePodWorker
...
+
+      def initialize
+        @status = 'unknown'.freeze
+      end
+
+      def status=(status)
+        if @status != status
+          @count = 1
+        else
+          @count += 1
```
SpComb: Initialize the instance with `@count = 0` just for sanity?

agent/lib/kontena/workers/service_pod_worker.rb
jakolehm: Yeah, I left this braindead logic here on purpose. Should agent backoff like in restarts or something else?

SpComb: Probably needs to be merged with #2689, because the restart/backoff mechanism is probably going to be the same for both container die events, and healthcheck failures.

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -53,10 +55,16 @@ def handle_action(msg)
...
         name = @container.labels['io.kontena.container.name']
         # Restart the container, master will handle re-scheduling logic
-        info "About to restart container #{name} as it's reported to be unhealthy"
-        emit_service_pod_event("service:health_check", "restarting #{name} because it's reported as unhealthy", Logger::WARN)
-        
-        restart_container
+        info "#{name} is unhealthy"
+        emit_service_pod_event("service:health_check", "#{name} is unhealthy", Logger::WARN)
+
+        publish('container:health_check', Event.new(
```
jnummelin: "Broadcasting" per-container health statuses feels bit of a wasted effort. These events should always be handled by a single pod worker. I could think of two possible alternatives:
- use something like `container:health_check:<service_id>:<instance_number>` as the channel key
- "route" the events through pod manager, it knows which pod worker should handle each event

SpComb: Or have the `ServicePodWorker` supervise the `ContainerHealthCheckWorkers` instead of spawning them from Docker events.

Then it can pass in `Celluloid.current_actor`, and the health check worker can send these as direct calls to the pod worker.

jnummelin: yep, that's even better I think. that way these are also tied to each others lifecycle nicely.

SpComb:
Needs rewrite (?) to match up with #2689

agent/lib/kontena/workers/service_pod_worker.rb
SpComb: This will also suffer from #2766, I imagine it could see "unhealthy" events for the old container after creating a new one.

SpComb:
Keeping this on the 1.4.0 milestone, because the way the `ContainerHealthCheckWorker` restarts interact with the `ServicePodWorker` has changed, so it should be fixed.

SpComb:
This is a change that needs to happen in some form, but not for 1.4.x anymore, I think.

I also have an WIP branch that does something similar to integrate the readyness/health-checks into the `ServicePodWorker` with state=running -> status=starting -> status=running steps, but it needs some more effort.

PR #2693, opened 2017-08-15
Agent: Allow overriding KONTENA_NODE_NAME

### LINKED ISSUE #2465: Allow to set node name via environment variable
Now it comes always from hostname, might be long and cryptic (and not so easy to change).

SpComb:
Fixes #2465

### DIFF
--- a/agent/README.md
+++ b/agent/README.md
@@ -21,7 +21,6 @@ Set development env variables to .env file:
 ```
 KONTENA_URI=ws://kontena.local:4040
 KONTENA_TOKEN=secret_grid_token
-AGENT_NAME=vagrant_agent_1
 ```
 
 Start agent using Docker Compose:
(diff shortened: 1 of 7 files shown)


agent/README.md
```diff
@@ -21,7 +21,6 @@ Set development env variables to .env file:
 ```
 KONTENA_URI=ws://kontena.local:4040
 KONTENA_TOKEN=secret_grid_token
-AGENT_NAME=vagrant_agent_1
```
SpComb: Not to be confused with the `KONTENA_NODE_NAME`... looking through the git history, the `AGENT_NAME` (or `KONTENA_AGENT_NAME` refers to the name of the `kontena-agent` container, but these are not used anymore.

SpComb:
Also worth noting that the `KONTENA_NODE_NAME` is only used for new nodes connecting with a grid token: it is not used for nodes connecting with a node token (the name comes from the `kontena node create`), and changing the `KONTENA_NODE_NAME` later will not rename the node.

PR #2694, opened 2017-08-15
Always create HostNode with unique name and node_number

### LINKED ISSUE #2071: Race condition on reserve_node_number with concurrent node creates

### LINKED ISSUE #2558: Difficult to get rid of a node that is missing a name and stuck as (initializing)

SpComb:
Fixes #2071, fixes #2542 by creating the nodes with a `node_number` set
Fixes #2558 by always having a node name available for `kontena node list`

* Send and validate both `Kontena-Node-ID` and `Kontena-Node-Name` headers in agent websocket connections
* Use `Grid.create_node!` to allocate the `node_number` before saving, retrying on `node_number` conflicts
* Create new `HostNode`s with the node name set for grid token connections, adding unique suffix if required
* Add validation and unique index on the `HostNode.name` (namespaced per grid)

* Validate the `HostNode.node_number` and change the unique index to be non-sparse

    It was already broken, seemingly failing with duplicate key errors on `node_number: null` with the `grid_id` set?

* Fix specs to use `grid.create_node!`

### DIFF
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -149,11 +149,15 @@ class WebsocketBackend
   # @raise [CloseError]
   # @return [HostNode]
   def find_node(req)
-    node_id = req.env['HTTP_KONTENA_NODE_ID'].to_s
+    node_id = req.env['HTTP_KONTENA_NODE_ID']
     node_labels = req.env['HTTP_KONTENA_NODE_LABELS'].to_s.split(',')
     init_attrs = {
       labels: node_labels,
     }
+
+    if node_id.nil? || node_id.empty?
+      raise CloseError.new(4000), "Missing Kontena-Node-ID"
+    end
 
     if grid_token = req.env['HTTP_KONTENA_GRID_TOKEN']
       return find_node_by_grid_token(node_id, grid_token, init_attrs)
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -75,6 +75,25 @@ class WebsocketBackend
     end
   end
 
+  # @param grid [Grid]
+  # @param node_id [String]
+  # @param name [String]
+  # @param attrs [Hash]
+  # @raise [Mongo::Error::OperationFailure] concurrent race
+  # @return [HostNode]
+  def create_node(grid, node_id, name, attrs)
+    node = grid.host_nodes.new(node_id: node_id, name: name, **attrs)
+    node.ensure_unique_name
+
+    if node.name != name
+      logger.warn "rename node #{node_id} on name collision: #{node.name}"
+    end
+
+    # XXX: this can still race with concurrent creates, and result in unique name index conflicts
+    node.save!
+    node
+  end
+
   # @param node_id [String] request header Kontena-Node-Id
   # @param grid_token [String] request header Kontena-Grid-Token
   # @param node_name [String] initialize name for new node
@@ -90,7 +109,7 @@ class WebsocketBackend
     node = grid.host_nodes.find_by(node_id: node_id)
 
     if !node
-      node = grid.host_nodes.create!(node_id: node_id, name: node_name, **init_attrs)
+      node = create_node(grid, node_id, node_name, init_attrs)
 
       logger.info "new node #{node} connected using grid token"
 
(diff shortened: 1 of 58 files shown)


server/app/middlewares/websocket_backend.rb
```diff
@@ -75,12 +75,32 @@ def call(env)
...
+  def create_node(grid, node_id, name, attrs)
+    node = grid.host_nodes.new(node_id: node_id, name: name, **attrs)
+    node.ensure_unique_name
+
+    if node.name != name
+      logger.warn "rename node #{node_id} on name collision: #{node.name}"
+    end
+
+    # XXX: this can still race with concurrent creates, and result in unique name index conflicts
+    node.save!
```
SpComb: This can still fail with `Mongo::Error::OperationFailure`, but so can the `reserve_node_number`: #2071

The server `WebsocketBackend` does not close the agent websocket connection on unknown errors, so the agent will hang.

SpComb:
WiP attempting to also fix #2071, and use the `node_number` as the name suffix at create-time.

server/app/models/grid.rb
```diff
@@ -54,8 +56,43 @@ def to_json(args = {})
...
+
+      return node
+
+    rescue Mongo::Error::OperationFailure => exc
+      raise unless exc.message =~ /^E11000 duplicate key error/
+
+      if self.host_nodes.where(node_number: node.node_number).exists?
+        warn "retry node #{name} node_number allocation on error: #{exc}"
+        node.name = name
+        retry
```
SpComb: This retry is still missing specs.

server/app/models/grid.rb
```diff
@@ -83,11 +92,10 @@ def create_node!(name, ensure_unique_name: false, **attrs)
 
       if self.host_nodes.where(node_number: node.node_number).exists?
         warn "retry node #{name} node_number allocation on error: #{exc}"
-        node.name = name
         retry
       elsif ensure_unique_name && self.host_nodes.where(name: node.name).exists?
         warn "rename node #{name} on name conflict: #{exc}"
-        node.name = "#{name}-#{node.node_number}"
```
SpComb: This logic was broken, it would re-use the wrong `node_number` for the suffix if that node_number happened to be allocated for a different node while this was retrying.

server/app/models/grid.rb
```diff
@@ -83,11 +92,10 @@ def create_node!(name, ensure_unique_name: false, **attrs)
...
+        duplicate_name = true
```
SpComb: This logic was broken, it would re-use the wrong `node_number` for the suffix if that node_number happened to be allocated for a different node while this was retrying.

SpComb:
@jakolehm PT(another)L

PR #2589, opened 2017-07-17
Ubuntu kontena-agent: Use /usr/bin/dockerd for compatibility with Docker 1.12 - 17.06

SpComb:
Fixes #2588 by replacing `ExecStart=/usr/bin/docker daemon` with `ExecStart=/usr/bin/dockerd`

* :x:  `docker.io` `1.10.3-0ubuntu6` does NOT support `/usr/bin/dockerd`
* :white_check_mark:  `docker.io` `1.12.6-0ubuntu1~16.04.1` supports `/usr/bin/dockerd`
* :grey_exclamation:  `docker-engine` `1.10.*` is no longer available
* :x: `docker-engine` `1.11.2-0~xenial` does NOT support `/usr/bin/dockerd`
* :white_check_mark: `docker-engine` `1.12.6-0~ubuntu-xenial` supports `/usr/bin/dockerd`
* :white_check_mark:  `docker-engine` `1.13.1-0~ubuntu-xenial` supports `/usr/bin/dockerd`
* :white_check_mark: `docker-engine` `17.05.0~ce-0~ubuntu-xenial` supports `/usr/bin/dockerd`

* :white_check_mark: `docker-ce` `17.03.2~ce-0~ubuntu-xenial` supports `/usr/bin/dockerd`
* :exclamation: `docker-ce` `17.06.0~ce-0~ubuntu` fails with `/usr/bin/docker daemon`

Breaks compatibility with Docker < 1.12, bumping the minimum version from 1.10 to 1.12.

See https://github.com/kontena/kontena/issues/2588#issuecomment-315717468:

PR #2695, opened 2017-08-16
Upgrade docker-api gem to 1.33.6

jakolehm:
Fixes https://github.com/moby/moby/issues/33434


SpComb:
Given that we're bumping the minimum supported Docker version to 1.12 (#2589), we could also bump the `Docker::API_VERSION` to 1.24... although maybe that doesn't need to be in this PR.

SpComb:
Upgrade for the https://github.com/swipely/docker-api/pull/485 fix?

Issue #2584, opened 2017-07-14
[1.3] kontena grid update does not notify nodes


SpComb:
The server [`Grids::Update#notify_nodes`](https://github.com/kontena/kontena/blob/v1.3.3/server/app/mutations/grids/update.rb#L84) call to [`Agent::NodePlugger.new`](https://github.com/kontena/kontena/blob/v1.3.3/server/app/services/agent/node_plugger.rb#L10) was broken by the refactoring in #2144:

```
ArgumentError (wrong number of arguments (given 2, expected 1)):
  /app/app/services/agent/node_plugger.rb:10:in `initialize'
  /app/app/mutations/grids/update.rb:38:in `new'
  /app/app/mutations/grids/update.rb:38:in `block (2 levels) in notify_nodes'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/contextual/mongo.rb:678:in `yield_document'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/contextual/mongo.rb:130:in `block in each'
  /usr/lib/ruby/gems/2.4.0/gems/mongo-2.4.1/lib/mongo/collection/view/iterable.rb:45:in `block in each'
  /usr/lib/ruby/gems/2.4.0/gems/mongo-2.4.1/lib/mongo/cursor.rb:110:in `block in each'
  /usr/lib/ruby/gems/2.4.0/gems/mongo-2.4.1/lib/mongo/cursor.rb:110:in `each'
  /usr/lib/ruby/gems/2.4.0/gems/mongo-2.4.1/lib/mongo/cursor.rb:110:in `each'
  /usr/lib/ruby/gems/2.4.0/gems/mongo-2.4.1/lib/mongo/collection/view/iterable.rb:44:in `each'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/query_cache.rb:221:in `each'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/contextual/mongo.rb:129:in `each'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/contextual.rb:20:in `each'
  /app/app/mutations/grids/update.rb:37:in `block in notify_nodes'
[...]

PR #2699, opened 2017-08-16
Server: Replace use of Celluloid::Future for async threads 

### LINKED ISSUE #2126: Server uses Celluloid::Future without logging of errors
Several server mutations and the `MongoPubsub::Subscription` use a `Celluloid::Future.new { ... }` block to run the code in the background, without any top-level error handling, or any call to `value` to re-raise errors.

For example, the error handling in `GridServices::Stop` means that it does revert the `GridService.state` on errors, but the `raise exc` just gets swallowed silently without any logging:

```ruby
    def execute
      ...
      Celluloid::Future.new{
        begin
          ...
        rescue => exc
          self.grid_service.set_state(prev_state)
          raise exc
        end
      }
    end
```

SpComb:
Fixes #2126 to log any errors raised by the async thread blocks

Using `Celluloid::Future` to execute async blocks in threads without actually checking the future `value` is broken, in that it silently ignores any resulting errors. This results in bugs like #2584 which are very hard to notice.

Fix the server to just use a simple `AsyncHelper#async_thread do ... end` which at least logs the errors, making them slightly more likely to be noticed during development. It does not actually handle the errors, though.

### DIFF
--- a/server/app/services/mongo_pubsub.rb
+++ b/server/app/services/mongo_pubsub.rb
@@ -1,10 +1,13 @@
 require_relative 'logging'
+require_relative '../helpers/async_helper.rb'
 
 class MongoPubsub
   include Celluloid
   include Logging
 
   class Subscription
+    include AsyncHelper
+
     attr_reader :channel
 
     # @param [String] channel
@@ -45,7 +48,7 @@ class MongoPubsub
 
     def process
       @process = true
-      Celluloid::Future.new {
+      async_thread do
         while @process == true && @stopped == false
           data = @queue.shift
           if data
@@ -54,7 +57,7 @@ class MongoPubsub
             @process = false
           end
         end
-      }
+      end
     end
 
     # @param [Hash] data
--- a/server/spec/mutations/grid_services/start_spec.rb
+++ b/server/spec/mutations/grid_services/start_spec.rb
@@ -1,5 +1,6 @@
+describe GridServices::Start do
+  include AsyncMock
 
-describe GridServices::Start, celluloid: true do
   let(:grid) {
     grid = Grid.create!(name: 'test-grid')
     grid
@@ -13,14 +14,13 @@ describe GridServices::Start, celluloid: true do
       container = redis_service.containers.create!(name: 'redis-1', container_id: '34')
       expect(subject).to receive(:start_service_instances).and_return(true)
 
-      subject.run.result.value
+      outcome = subject.run
     end
 
     it 'sets service state to running' do
       redis_service.containers.create!(name: 'redis-1', container_id: '34')
       allow(subject).to receive(:start_service_instances).and_return(true)
       outcome = subject.run
-      outcome.result.value
       expect(redis_service.reload.state).to eq('running')
     end
 
@@ -29,10 +29,8 @@ describe GridServices::Start, celluloid: true do
       redis_service.containers.create!(name: 'redis-1', container_id: '34')
       subject = described_class.new(grid_service: redis_service)
       expect(subject).to receive(:start_service_instances).and_raise(StandardError.new('error'))
-      outcome = subject.run
-      expect(outcome.success?).to be_truthy
       expect {
-        outcome.result.value
+        outcome = subject.run
       }.to raise_exception(StandardError)
       expect(redis_service.state).to eq(prev_state)
     end
--- a/server/app/helpers/async_helper.rb
+++ b/server/app/helpers/async_helper.rb
@@ -1,0 +1,17 @@
+module AsyncHelper
+  include Logging
+
+  # Run block in a new thread. Log exceptions.
+  #
+  # @yield
+  # @return [Thread] don't care about it
+  def async_thread(&block)
+    Thread.new {
+      begin
+        yield
+      rescue Exception => exc
+        error exc
+      end
+    }
+  end
+end
--- a/server/app/mutations/grid_services/restart.rb
+++ b/server/app/mutations/grid_services/restart.rb
@@ -1,13 +1,15 @@
 module GridServices
   class Restart < Mutations::Command
+    include AsyncHelper
+
     required do
       model :grid_service
     end
 
     def execute
-      Celluloid::Future.new{
+      async_thread do
         self.restart_service_instances
-      }
+      end
     end
 
     def restart_service_instances
--- a/server/app/mutations/grid_services/start.rb
+++ b/server/app/mutations/grid_services/start.rb
@@ -1,12 +1,14 @@
 module GridServices
   class Start < Mutations::Command
+    include AsyncHelper
+
     required do
       model :grid_service
     end
 
     def execute
       prev_state = self.grid_service.state
-      Celluloid::Future.new {
+      async_thread do
         begin
           self.grid_service.set_state('running')
           self.start_service_instances
@@ -14,7 +16,7 @@ module GridServices
           self.grid_service.set_state(prev_state)
           raise exc
         end
-      }
+      end
     end
 
     def start_service_instances
--- a/server/app/mutations/grid_services/stop.rb
+++ b/server/app/mutations/grid_services/stop.rb
@@ -1,12 +1,14 @@
 module GridServices
   class Stop < Mutations::Command
+    include AsyncHelper
+
     required do
       model :grid_service
     end
 
     def execute
       prev_state = self.grid_service.state
-      Celluloid::Future.new{
+      async_thread do
         begin
           self.grid_service.set_state('stopped')
           self.stop_service_instances
@@ -14,7 +16,7 @@ module GridServices
           self.grid_service.set_state(prev_state)
           raise exc
         end
-      }
+      end
     end
 
     def stop_service_instances
--- a/server/app/mutations/grids/update.rb
+++ b/server/app/mutations/grids/update.rb
@@ -2,6 +2,7 @@ require_relative 'common'
 
 module Grids
   class Update < Mutations::Command
+    include AsyncHelper
     include Common
 
     required do
@@ -27,19 +28,19 @@ module Grids
         return
       end
 
-      self.notify_nodes
+      async_thread do
+        self.notify_nodes
+      end
 
       self.grid
     end
 
     def notify_nodes
-      Celluloid::Future.new {
-        grid.host_nodes.connected.each do |node|
-          plugger = Agent::NodePlugger.new(node)
-          plugger.send_node_info
-        end
-        GridScheduler.new(grid).reschedule
-      }
+      grid.host_nodes.connected.each do |node|
+        plugger = Agent::NodePlugger.new(node)
+        plugger.send_node_info
+      end
+      GridScheduler.new(grid).reschedule
     end
   end
 end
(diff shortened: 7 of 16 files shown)


server/app/services/mongo_pubsub.rb
```diff
@@ -45,7 +48,7 @@ def queue_message(data)
 
     def process
       @process = true
-      Celluloid::Future.new {
+      async_thread do
```
SpComb: This `MongoPubsub::Subscriber` needs some actual error handling, because if the `MongoPubsub#subscribe do ... end` block raises an error, then the process thread will die, leaving the `Subscription` in the `@process = true` state. The `MongoPubsub#tail!` actor thread will continue calling `subscription.queue_message` to push messages to the `@queue`, but there won't be anyone there to read them.

server/spec/mutations/grid_services/start_spec.rb
```diff
@@ -29,10 +29,8 @@
       redis_service.containers.create!(name: 'redis-1', container_id: '34')
       subject = described_class.new(grid_service: redis_service)
       expect(subject).to receive(:start_service_instances).and_raise(StandardError.new('error'))
-      outcome = subject.run
-      expect(outcome.success?).to be_truthy
       expect {
-        outcome.result.value
+        outcome = subject.run
       }.to raise_exception(StandardError)
```
SpComb: These specs are not that great with the `AsyncMock`, because the `Mutation::Command#run` does not actually ever raise this `StandardError`.

jakolehm:
`Celluloid::Future` uses a thread pool, how much this change will impact resource usage?

SpComb:
> `Celluloid::Future` uses a thread pool, how much this change will impact resource usage?

From reading the celluloid 0.17 code, I suspect that `Celluloid::Future.new { ... }` doesn't actually use a thread pool, it looks like each call spawns a new thread without any limit on the concurrency. Have not confirmed that, though, and the documentation / discussion on issues is unclear on the subject.

jakolehm:
https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/future.rb#L38

SpComb:
The `Celluloid::Future.new` implementation is very tricky to read, but I think the `return super unless block` means that the block never actually gets passed to `#initialize`, so that code using `Celluloid.internal_pool.get` is not actually used? Instead, it uses the [`Internals::ThreadHandle.new`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/future.rb#L13) path, which is a new thread, not a pool.

As I said, would need to test to actually figure it out... :confused:

SpComb:
Does not look like a threadpool to me, every `Celluloid::Future.new` gets a new thread:

```
irb(main):028:0> require 'pp'; threads = {}; (1..100).map { |i| Celluloid::Future.new { [ i, Thread.current.to_s] } }.map{|f| f.value }.each { |i, thread| (threads[thread] ||= []) << i }; pp threads
{"#<Celluloid::Thread:0x00562cea91ff60>"=>[1],
 "#<Celluloid::Thread:0x00562cea91f880>"=>[2],
 "#<Celluloid::Thread:0x00562cea91e9a8>"=>[3],
 "#<Celluloid::Thread:0x00562cea91e390>"=>[4],
 "#<Celluloid::Thread:0x00562cea91de18>"=>[5],
 "#<Celluloid::Thread:0x00562cea91d878>"=>[6],
 "#<Celluloid::Thread:0x00562cea91d2d8>"=>[7],
 "#<Celluloid::Thread:0x00562cea91ccc0>"=>[8],
 "#<Celluloid::Thread:0x00562cea91c770>"=>[9],
 "#<Celluloid::Thread:0x00562cea91c1a8>"=>[10],
 "#<Celluloid::Thread:0x00562cea91e930>"=>[11],
 "#<Celluloid::Thread:0x00562cea91cd38>"=>[12],
 "#<Celluloid::Thread:0x00562cea917d38>"=>[13],
 "#<Celluloid::Thread:0x00562cea9178b0>"=>[14],
 "#<Celluloid::Thread:0x00562cea917400>"=>[15],
 "#<Celluloid::Thread:0x00562cea916f78>"=>[16],
 "#<Celluloid::Thread:0x00562cea916ac8>"=>[17],
 "#<Celluloid::Thread:0x00562cea916640>"=>[18],
 "#<Celluloid::Thread:0x00562cea9161b8>"=>[19],
 "#<Celluloid::Thread:0x00562cea915d30>"=>[20],
 "#<Celluloid::Thread:0x00562cea915880>"=>[21],
 "#<Celluloid::Thread:0x00562cea9153f8>"=>[22],
 "#<Celluloid::Thread:0x00562cea914f48>"=>[23],
 "#<Celluloid::Thread:0x00562cea914a70>"=>[24],
 "#<Celluloid::Thread:0x00562cea914570>"=>[25],
 "#<Celluloid::Thread:0x00562cea9140c0>"=>[26],
 "#<Celluloid::Thread:0x00562cea90fac0>"=>[27],
 "#<Celluloid::Thread:0x00562cea90f4d0>"=>[28],
 "#<Celluloid::Thread:0x00562cea90ee40>"=>[29],
 "#<Celluloid::Thread:0x00562cea90e8a0>"=>[30],
 "#<Celluloid::Thread:0x00562cea90dfb8>"=>[31],
 "#<Celluloid::Thread:0x00562cea90c960>"=>[32],
 "#<Celluloid::Thread:0x00562cea90f930>"=>[33],
 "#<Celluloid::Thread:0x00562cea90ddb0>"=>[34],
 "#<Celluloid::Thread:0x00562cea90d810>"=>[35],
 "#<Celluloid::Thread:0x00562cea90d2c0>"=>[36],
 "#<Celluloid::Thread:0x00562cea90cc30>"=>[37],
 "#<Celluloid::Thread:0x00562cea90c280>"=>[38],
 "#<Celluloid::Thread:0x00562cea907e10>"=>[39],
 "#<Celluloid::Thread:0x00562cea907960>"=>[40],
 "#<Celluloid::Thread:0x00562cea9074d8>"=>[41],
 "#<Celluloid::Thread:0x00562cea907050>"=>[42],
 "#<Celluloid::Thread:0x00562cea906bc8>"=>[43],
 "#<Celluloid::Thread:0x00562cea906740>"=>[44],
 "#<Celluloid::Thread:0x00562cea9062b8>"=>[45],
 "#<Celluloid::Thread:0x00562cea905e30>"=>[46],
 "#<Celluloid::Thread:0x00562cea9059a8>"=>[47],
 "#<Celluloid::Thread:0x00562cea905520>"=>[48],
 "#<Celluloid::Thread:0x00562cea905098>"=>[49],
 "#<Celluloid::Thread:0x00562cea904be8>"=>[50],
 "#<Celluloid::Thread:0x00562cea904738>"=>[51],
 "#<Celluloid::Thread:0x00562cea9042b0>"=>[52],
 "#<Celluloid::Thread:0x00562cea8fbd40>"=>[53],
 "#<Celluloid::Thread:0x00562cea8fb818>"=>[54],
 "#<Celluloid::Thread:0x00562cea8fb340>"=>[55],
 "#<Celluloid::Thread:0x00562cea8fae90>"=>[56],
 "#<Celluloid::Thread:0x00562cea8fa9e0>"=>[57],
 "#<Celluloid::Thread:0x00562cea8fa530>"=>[58],
 "#<Celluloid::Thread:0x00562cea8f9c48>"=>[59],
 "#<Celluloid::Thread:0x00562cea8f9540>"=>[60],
 "#<Celluloid::Thread:0x00562cea8f9090>"=>[61],
 "#<Celluloid::Thread:0x00562cea8f8c08>"=>[62],
 "#<Celluloid::Thread:0x00562cea8f8780>"=>[63],
 "#<Celluloid::Thread:0x00562cea8f82f8>"=>[64],
 "#<Celluloid::Thread:0x00562cea8dfc58>"=>[65],
 "#<Celluloid::Thread:0x00562cea8df640>"=>[66],
 "#<Celluloid::Thread:0x00562cea8df1b8>"=>[67],
 "#<Celluloid::Thread:0x00562cea8dea60>"=>[68],
 "#<Celluloid::Thread:0x00562cea8ddb88>"=>[69],
 "#<Celluloid::Thread:0x00562cea8dd700>"=>[70],
 "#<Celluloid::Thread:0x00562cea8dd278>"=>[71],
 "#<Celluloid::Thread:0x00562cea8dcdf0>"=>[72],
 "#<Celluloid::Thread:0x00562cea8dc968>"=>[73],
 "#<Celluloid::Thread:0x00562cea8dc4e0>"=>[74],
 "#<Celluloid::Thread:0x00562cea8dc058>"=>[75],
 "#<Celluloid::Thread:0x00562cea8d3b60>"=>[76],
 "#<Celluloid::Thread:0x00562cea8d36b0>"=>[77],
 "#<Celluloid::Thread:0x00562cea8d3228>"=>[78],
 "#<Celluloid::Thread:0x00562cea8d2d28>"=>[79],
 "#<Celluloid::Thread:0x00562cea8d28a0>"=>[80],
 "#<Celluloid::Thread:0x00562cea8d2418>"=>[81],
 "#<Celluloid::Thread:0x00562cea8d1f40>"=>[82],
 "#<Celluloid::Thread:0x00562cea8d1a90>"=>[83],
 "#<Celluloid::Thread:0x00562cea8d15b8>"=>[84],
 "#<Celluloid::Thread:0x00562cea8d1130>"=>[85],
 "#<Celluloid::Thread:0x00562cea8d07d0>"=>[86],
 "#<Celluloid::Thread:0x00562cea8cbe10>"=>[87],
 "#<Celluloid::Thread:0x00562cea8cb988>"=>[88],
 "#<Celluloid::Thread:0x00562cea8cb500>"=>[89],
 "#<Celluloid::Thread:0x00562cea8cb050>"=>[90],
 "#<Celluloid::Thread:0x00562cea8cab28>"=>[91],
 "#<Celluloid::Thread:0x00562cea8ca600>"=>[92],
 "#<Celluloid::Thread:0x00562cea8ca128>"=>[93],
 "#<Celluloid::Thread:0x00562cea8c9ca0>"=>[94],
 "#<Celluloid::Thread:0x00562cea8c97f0>"=>[95],
 "#<Celluloid::Thread:0x00562cea8c9368>"=>[96],
 "#<Celluloid::Thread:0x00562cea8c8eb8>"=>[97],
 "#<Celluloid::Thread:0x00562cea8c8a30>"=>[98],
 "#<Celluloid::Thread:0x00562ceabe7338>"=>[99],
 "#<Celluloid::Thread:0x00562ceabe6b18>"=>[100]}
```

SpComb:
PTAL @jakolehm? This is ugly, but AFAIK it's equivalent and better than the blind `Celluloid::Future.new`, and I can't think of a better approach without reworking how async mutations run entirely.

Fixing the `MongoPubsub` error handling is a separate issue, this shouldn't make it any worse...

jakolehm:
Yeah, it seems that celluloid 0.17 behaves a bit differently with plain futures.

SpComb:
Just to be clear I don't think this is the ideal solution, but at least now we might notice these errors. Once we start seeing those errors, we might have a better idea of what we should be doing to handle them.

PR #2701, opened 2017-08-17
Fix oneshot hooks; Change post_start hooks to run on start; Add pre_start, pre_stop hooks

jakolehm:
This PR implements service hooks that are executed before actual service container is started/stopped. Great for db migrations etc.

Other changes/improvements:
- improved `oneshot` hook tracking
- `post_start` is executed always when service instance is started (but not with restart)

### DIFF
--- a/agent/lib/kontena/rpc/service_pods_api.rb
+++ b/agent/lib/kontena/rpc/service_pods_api.rb
@@ -8,31 +8,13 @@ module Kontena
       # @param instance_number [Integer]
       # @return [Hash]
       def restart(service_id, instance_number)
-        service_pod = find_service_pod(service_id, instance_number)
-        raise Kontena::RpcServer::Error.new(404, "Service instance not found") unless service_pod
-
-        Kontena::ServicePods::Restarter.new(service_pod).perform
+        Kontena::ServicePods::Restarter.new(service_id, instance_number).perform
         {}
       end
 
       def notify_update(reason)
         Celluloid::Notifications.publish('service_pod:update', reason)
         {}
-      end
-
-      private
-
-      # @param service_id [String]
-      # @param instance_number [Integer]
-      def find_service_pod(service_id, instance_number)
-        worker = Celluloid::Actor[:service_pod_manager].workers.find { |w|
-          w.service_pod.service_id == service_id && w.service_pod.instance_number == instance_number
-        }
-        if worker
-          worker.service_pod
-        else
-          nil
-        end
       end
     end
   end
--- a/agent/lib/kontena/rpc/service_pods_api.rb
+++ b/agent/lib/kontena/rpc/service_pods_api.rb
@@ -4,17 +4,35 @@ module Kontena
   module Rpc
     class ServicePodsApi
 
-      # @param [String] service_id
-      # @param [Integer] instance_number
+      # @param service_id [String]
+      # @param instance_number [Integer]
       # @return [Hash]
       def restart(service_id, instance_number)
-        Kontena::ServicePods::Restarter.new(service_id, instance_number).perform
+        service_pod = find_service_pod(service_id, instance_number)
+        raise Kontena::RpcServer::Error.new(404, "Service instance not found") unless service_pod
+
+        Kontena::ServicePods::Restarter.new(service_pod).perform
         {}
       end
 
       def notify_update(reason)
         Celluloid::Notifications.publish('service_pod:update', reason)
         {}
+      end
+
+      private
+
+      # @param service_id [String]
+      # @param instance_number [Integer]
+      def find_service_pod(service_id, instance_number)
+        worker = Celluloid::Actor[:service_pod_manager].workers.find { |w|
+          w.service_pod.service_id == service_id && w.service_pod.instance_number == instance_number
+        }
+        if worker
+          worker.service_pod
+        else
+          nil
+        end
       end
     end
   end
--- a/agent/lib/kontena/service_pods/common.rb
+++ b/agent/lib/kontena/service_pods/common.rb
@@ -1,9 +1,36 @@
 require_relative '../helpers/event_log_helper'
+require_relative '../helpers/weave_helper'
 
 module Kontena
   module ServicePods
     module Common
       include Kontena::Helpers::EventLogHelper
+      include Kontena::Helpers::WeaveHelper
+
+      # Docker create configuration for ServicePod
+      # @param [ServicePod] service_pod
+      # @return [Hash] Docker create API JSON object
+      def config_container(service_pod)
+        service_config = service_pod.service_config
+
+        unless service_pod.net == 'host'
+          network_adapter.modify_create_opts(service_config)
+        end
+
+        service_config
+      end
+
+      # @param [Hash] opts
+      def create_container(opts)
+        Docker::Container.create(opts)
+      end
+
+      # @param [Docker::Container] container
+      def cleanup_container(container)
+        container.stop('timeout' => container.stop_grace_period)
+        container.wait
+        container.delete(v: true)
+      end
 
       # @param [String] service_id
       # @param [Integer] instance_number
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -16,11 +16,13 @@ module Kontena
 
       attr_reader :service_pod, :image_credentials, :hook_manager
 
-      # @param [ServicePod] service_pod
-      def initialize(service_pod)
+      # @param service_pod [ServicePod]
+      # @param hook_manager [LifecycleHookManager]
+      def initialize(service_pod, hook_manager)
         @service_pod = service_pod
         @image_credentials = service_pod.image_credentials
-        @hook_manager = Kontena::ServicePods::LifecycleHookManager.new(service_pod)
+        @hook_manager = hook_manager
+        @hook_manager.track(service_pod)
       end
 
       # @return [Docker::Container]
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -46,6 +46,8 @@ module Kontena
           log_service_pod_event("service:create_instance", "removed previous version of service #{service_pod.name_for_humans} instance")
         end
 
+        self.run_pre_start_hooks
+
         service_config = config_container(service_pod)
 
         debug "creating container: #{service_pod.name}"
@@ -65,7 +67,7 @@ module Kontena
         Celluloid::Notifications.publish('service_pod:start', service_pod.name)
         Celluloid::Notifications.publish('container:publish_info', service_container)
 
-        self.run_hooks(service_container, 'post_start')
+        self.run_post_start_hooks(service_container)
 
         if service_pod.wait_for_port
           info "waiting for port #{service_pod.name_for_humans}:#{service_pod.wait_for_port} to respond"
@@ -78,21 +80,23 @@ module Kontena
         service_container
       end
 
-      # @param [Docker::Container] service_container
-      # @param [String] type
       # @return [Boolean]
-      def run_hooks(service_container, type)
-        service_pod.hooks.each do |hook|
-          if hook['type'] == type
-            info "running #{type} hook: #{hook['cmd']}"
-            command = ['/bin/sh', '-c', hook['cmd']]
-            log_hook_output(service_container.id, ["running #{type} hook: #{hook['cmd']}"], 'stdout')
-            stdout, stderr, exit_code = service_container.exec(command)
-            log_hook_output(service_container.id, stdout, 'stdout')
-            log_hook_output(service_container.id, stderr, 'stderr')
-            if exit_code != 0
-              raise "Failed to execute hook: #{hook['cmd']}"
+      def run_pre_start_hooks
+        hooks_for('pre_start').each do |hook|
+          service_config = config_container(service_pod.dup)
+          service_config['Cmd'] = ['/bin/sh', '-c', hook['cmd']]
+          service_container = nil
+          begin
+            service_container = create_container(service_config)
+            info "running pre_start hook: #{hook['cmd']}"
+            service_container.tap(&:start).attach { |stream, chunk|
+              log_hook_output(service_container.id, [chunk], stream)
+            }
+            if service_container.state['ExitCode'] != 0
+              raise "Failed to execute pre_start hook: #{hook['cmd']}, exit code: #{service_container.state['ExitCode']}"
             end
+          ensure
+            cleanup_container(service_container) if service_container
           end
         end
         true
@@ -100,6 +104,33 @@ module Kontena
         log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
         error exc.message
         false
+      end
+
+      # @param [Docker::Container] service_container
+      # @return [Boolean]
+      def run_post_start_hooks(service_container)
+        hooks_for('post_start').each do |hook|
+          info "running post_start hook: #{hook['cmd']}"
+          command = ['/bin/sh', '-c', hook['cmd']]
+          log_hook_output(service_container.id, ["running #{type} hook: #{hook['cmd']}"], 'stdout')
+          _, _, exit_code = service_container.exec(command) { |stream, chunk|
+            log_hook_output(service_container.id, [chunk], stream)
+          }
+          if exit_code != 0
+            raise "Failed to execute post_start hook: #{hook['cmd']}"
+          end
+        end
+        true
+      rescue => exc
+        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
+        error exc.message
+        false
+      end
+
+      # @param [String] type
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        service_pod.hooks.select{ |h| h['type'] == type }
       end
 
       # @param [String] id
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -42,6 +42,7 @@ module Kontena
         wait_network_ready?
 
         if service_container
+          hook_manager.on_pre_stop(service_container)
           info "removing previous version of service: #{service_pod.name_for_humans}"
           cleanup_container(service_container)
           log_service_pod_event("service:create_instance", "removed previous version of service #{service_pod.name_for_humans} instance")
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -89,9 +89,7 @@ module Kontena
           begin
             service_container = create_container(service_config)
             info "running pre_start hook: #{hook['cmd']}"
-            service_container.tap(&:start).attach { |stream, chunk|
-              log_hook_output(service_container.id, [chunk], stream)
-            }
+            service_container.tap(&:start).attach
             if service_container.state['ExitCode'] != 0
               raise "Failed to execute pre_start hook: #{hook['cmd']}, exit code: #{service_container.state['ExitCode']}"
             end
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -102,8 +102,7 @@ module Kontena
         true
       rescue => exc
         log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
-        error exc.message
-        false
+        raise exc
       end
 
       # @param [Docker::Container] service_container
--- a/agent/lib/kontena/service_pods/creator.rb
+++ b/agent/lib/kontena/service_pods/creator.rb
@@ -1,8 +1,8 @@
 require 'docker'
 require 'celluloid'
 require_relative 'common'
+require_relative 'lifecycle_hook_manager'
 require_relative '../logging'
-require_relative '../helpers/weave_helper'
 require_relative '../helpers/port_helper'
 require_relative '../helpers/rpc_helper'
 
@@ -15,12 +15,13 @@ module Kontena
       include Kontena::Helpers::PortHelper
       include Kontena::Helpers::RpcHelper
 
-      attr_reader :service_pod, :image_credentials
+      attr_reader :service_pod, :image_credentials, :hook_manager
 
       # @param [ServicePod] service_pod
       def initialize(service_pod)
         @service_pod = service_pod
         @image_credentials = service_pod.image_credentials
+        @hook_manager = Kontena::ServicePods::LifecycleHookManager.new(service_pod)
       end
 
       # @return [Docker::Container]
@@ -42,11 +43,11 @@ module Kontena
 
         if service_container
           info "removing previous version of service: #{service_pod.name_for_humans}"
-          self.cleanup_container(service_container)
+          cleanup_container(service_container)
           log_service_pod_event("service:create_instance", "removed previous version of service #{service_pod.name_for_humans} instance")
         end
 
-        self.run_pre_start_hooks
+        hook_manager.on_pre_start
 
         service_config = config_container(service_pod)
 
@@ -67,7 +68,7 @@ module Kontena
         Celluloid::Notifications.publish('service_pod:start', service_pod.name)
         Celluloid::Notifications.publish('container:publish_info', service_container)
 
-        self.run_post_start_hooks(service_container)
+        hook_manager.on_post_start(service_container)
 
         if service_pod.wait_for_port
           info "waiting for port #{service_pod.name_for_humans}:#{service_pod.wait_for_port} to respond"
@@ -78,71 +79,6 @@ module Kontena
         end
 
         service_container
-      end
-
-      # @return [Boolean]
-      def run_pre_start_hooks
-        hooks_for('pre_start').each do |hook|
-          service_config = config_container(service_pod.dup)
-          service_config['Cmd'] = ['/bin/sh', '-c', hook['cmd']]
-          service_container = nil
-          begin
-            service_container = create_container(service_config)
-            info "running pre_start hook: #{hook['cmd']}"
-            service_container.tap(&:start).attach
-            if service_container.state['ExitCode'] != 0
-              raise "Failed to execute pre_start hook: #{hook['cmd']}, exit code: #{service_container.state['ExitCode']}"
-            end
-          ensure
-            cleanup_container(service_container) if service_container
-          end
-        end
-        true
-      rescue => exc
-        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
-        raise exc
-      end
-
-      # @param [Docker::Container] service_container
-      # @return [Boolean]
-      def run_post_start_hooks(service_container)
-        hooks_for('post_start').each do |hook|
-          info "running post_start hook: #{hook['cmd']}"
-          command = ['/bin/sh', '-c', hook['cmd']]
-          log_hook_output(service_container.id, ["running #{type} hook: #{hook['cmd']}"], 'stdout')
-          _, _, exit_code = service_container.exec(command) { |stream, chunk|
-            log_hook_output(service_container.id, [chunk], stream)
-          }
-          if exit_code != 0
-            raise "Failed to execute post_start hook: #{hook['cmd']}"
-          end
-        end
-        true
-      rescue => exc
-        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
-        error exc.message
-        false
-      end
-
-      # @param [String] type
-      # @return [Array<Hash>]
-      def hooks_for(type)
-        service_pod.hooks.select{ |h| h['type'] == type }
-      end
-
-      # @param [String] id
-      # @param [Array<String>] lines
-      # @param [String] type
-      def log_hook_output(id, lines, type)
-        lines.each do |chunk|
-          data = {
-              id: id,
-              time: Time.now.utc.xmlschema,
-              type: type,
-              data: chunk
-          }
-          rpc_client.async.notification('/containers/log', [data])
-        end
       end
 
       # @param [String] type
@@ -178,31 +114,6 @@ module Kontena
           end
         end
         true
-      end
-
-      # @param [Docker::Container] container
-      def cleanup_container(container)
-        container.stop('timeout' => container.stop_grace_period)
-        container.wait
-        container.delete(v: true)
-      end
-
-      # Docker create configuration for ServicePod
-      # @param [ServicePod] service_pod
-      # @return [Hash] Docker create API JSON object
-      def config_container(service_pod)
-        service_config = service_pod.service_config
-
-        unless service_pod.net == 'host'
-          network_adapter.modify_create_opts(service_config)
-        end
-
-        service_config
-      end
-
-      # @param [Hash] opts
-      def create_container(opts)
-        Docker::Container.create(opts)
       end
 
       # Make sure that image exists
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -24,6 +24,7 @@ module Kontena
             service_config = config_container(service_pod.dup, hook['cmd'])
             service_config['name'] = "#{service_config['name']}-#{SecureRandom.urlsafe_base64(5)}"
             service_container = create_container(service_config)
+            log_hook_output(service_container.id, "running pre_start hook: #{hook['cmd']}", 'stdout')
             service_container.tap(&:start).attach
             if service_container.state['ExitCode'] != 0
               raise "Failed to execute pre_start hook: #{hook['cmd']}, exit code: #{service_container.state['ExitCode']}"
@@ -44,9 +45,9 @@ module Kontena
         hooks_for('post_start').each do |hook|
           info "running post_start hook: #{hook['cmd']}"
           command = build_cmd(hook['cmd'])
-          log_hook_output(service_container.id, ["running post_start hook: #{hook['cmd']}"], 'stdout')
+          log_hook_output(service_container.id, "running post_start hook: #{hook['cmd']}", 'stdout')
           _, _, exit_code = service_container.exec(command) { |stream, chunk|
-            log_hook_output(service_container.id, [chunk], stream)
+            log_hook_output(service_container.id, chunk, stream)
           }
           if exit_code != 0
             raise "Failed to execute post_start hook: #{hook['cmd']} (exit code: #{exit_code}"
@@ -65,9 +66,9 @@ module Kontena
         hooks_for('pre_stop').each do |hook|
           info "running pre_stop hook: #{hook['cmd']}"
           command = build_cmd(hook['cmd'])
-          log_hook_output(service_container.id, ["running pre_stop hook: #{hook['cmd']}"], 'stdout')
+          log_hook_output(service_container.id, "running pre_stop hook: #{hook['cmd']}", 'stdout')
           _, _, exit_code = service_container.exec(command) { |stream, chunk|
-            log_hook_output(service_container.id, [chunk], stream)
+            log_hook_output(service_container.id, chunk, stream)
           }
           if exit_code != 0
             raise "Failed to execute pre_stop hook: #{hook['cmd']} (exit code: #{exit_code}"
@@ -107,18 +108,16 @@ module Kontena
       end
 
       # @param [String] id
-      # @param [Array<String>] lines
+      # @param [String] line
       # @param [String] type
-      def log_hook_output(id, lines, type)
-        lines.each do |chunk|
-          data = {
-              id: id,
-              time: Time.now.utc.xmlschema,
-              type: type,
-              data: chunk
-          }
-          rpc_client.async.notification('/containers/log', [data])
-        end
+      def log_hook_output(id, line, type)
+        data = [{
+          id: id,
+          time: Time.now.utc.xmlschema,
+          type: type,
+          data: line
+        }]
+        rpc_client.async.notification('/containers/log_batch', [data])
       end
 
       # @param [String] type
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -66,7 +66,6 @@ module Kontena
       # @return [Array<String>]
       def build_cmd(cmd)
         command = ['/bin/sh', '-c', cmd]
-        command.unshift('/w/w') if service_pod.can_expose_ports?
 
         command
       end
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -6,10 +6,13 @@ module Kontena
       include Kontena::Helpers::RpcHelper
       include Common
 
-      attr_reader :service_pod
+      attr_reader :node, :service_pod
 
-      # @param [ServicePod] service_pod
-      def initialize(service_pod)
+      def initialize(node)
+        @node = node
+      end
+
+      def track(service_pod)
         @service_pod = service_pod
       end
 
@@ -84,7 +87,16 @@ module Kontena
       # @param [String] type
       # @return [Array<Hash>]
       def hooks_for(type)
-        service_pod.hooks.select{ |h| h['type'] == type }
+        hooks = service_pod.hooks.select { |h|
+          h['type'] == type && !h['oneshot'] || (h['oneshot'] && !oneshot_cache.include?(h))
+        }
+        hooks.each do |h|
+          if h['oneshot'] && !oneshot_cache.include?(h)
+            mark_oneshot_hook(h)
+          end
+        end
+
+        hooks
       end
 
       # @param [String] cmd
@@ -126,6 +138,21 @@ module Kontena
       # @param [Integer] severity
       def log_service_pod_event(type, data, severity = Logger::INFO)
         super(service_pod.service_id, service_pod.instance_number, type, data, severity)
+      end
+
+      # @param hook [Hash]
+      def mark_oneshot_hook(hook)
+        pod = {
+          service_id: service_pod.service_id,
+          instance_number: service_pod.instance_number
+        }
+        rpc_client.async.request('/node_service_pods/mark_oneshot_hook', [node.id, pod, hook])
+        oneshot_cache << hook
+      end
+
+      # @return [Array<Hash>]
+      def oneshot_cache
+        @oneshot_cache ||= []
       end
     end
   end
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -1,3 +1,5 @@
+require 'securerandom'
+
 module Kontena
   module ServicePods
     class LifecycleHookManager
@@ -20,6 +22,7 @@ module Kontena
           begin
             info "running pre_start hook: #{hook['cmd']}"
             service_config = config_container(service_pod.dup, hook['cmd'])
+            service_config['name'] = "#{service_config['name']}-#{SecureRandom.urlsafe_base64(5)}"
             service_container = create_container(service_config)
             service_container.tap(&:start).attach
             if service_container.state['ExitCode'] != 0
@@ -47,6 +50,27 @@ module Kontena
           }
           if exit_code != 0
             raise "Failed to execute post_start hook: #{hook['cmd']} (exit code: #{exit_code}"
+          end
+        end
+        true
+      rescue => exc
+        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
+        error exc.message
+        false
+      end
+
+      # @param [Docker::Container] service_container
+      # @return [Boolean]
+      def on_pre_stop(service_container)
+        hooks_for('pre_stop').each do |hook|
+          info "running pre_stop hook: #{hook['cmd']}"
+          command = build_cmd(hook['cmd'])
+          log_hook_output(service_container.id, ["running pre_stop hook: #{hook['cmd']}"], 'stdout')
+          _, _, exit_code = service_container.exec(command) { |stream, chunk|
+            log_hook_output(service_container.id, [chunk], stream)
+          }
+          if exit_code != 0
+            raise "Failed to execute pre_stop hook: #{hook['cmd']} (exit code: #{exit_code}"
           end
         end
         true
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -1,0 +1,93 @@
+module Kontena
+  module ServicePods
+    class LifecycleHookManager
+      include Kontena::Helpers::RpcHelper
+      include Common
+
+      attr_reader :service_pod
+
+      # @param [ServicePod] service_pod
+      def initialize(service_pod)
+        @service_pod = service_pod
+      end
+
+      # @return [Boolean]
+      def on_pre_start
+        pre_start_hooks = hooks_for('pre_start')
+        return if pre_start_hooks.size == 0
+
+        service_config = config_container(service_pod.dup)
+        service_config['Entrypoint'] = '/bin/sh'
+        service_config['Cmd'] = ['-c', 'sleep 600']
+        service_container = create_container(service_config)
+        service_container.start!
+        pre_start_hooks.each do |hook|
+          info "running pre_start hook: #{hook['cmd']}"
+          command = ['/bin/sh', '-c', hook['cmd']]
+          log_hook_output(service_container.id, ["running pre_start hook: #{hook['cmd']}"], 'stdout')
+          _, _, exit_code = service_container.exec(command) { |stream, chunk|
+            log_hook_output(service_container.id, [chunk], stream)
+          }
+          if exit_code != 0
+            raise "Failed to execute pre_start hook: #{hook['cmd']} (exit code: #{exit_code}"
+          end
+        end
+        true
+      rescue => exc
+        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
+        raise exc
+      ensure
+        cleanup_container(service_container) if service_container
+      end
+
+      # @param [Docker::Container] service_container
+      # @return [Boolean]
+      def on_post_start(service_container)
+        hooks_for('post_start').each do |hook|
+          info "running post_start hook: #{hook['cmd']}"
+          command = ['/bin/sh', '-c', hook['cmd']]
+          log_hook_output(service_container.id, ["running post_start hook: #{hook['cmd']}"], 'stdout')
+          _, _, exit_code = service_container.exec(command) { |stream, chunk|
+            log_hook_output(service_container.id, [chunk], stream)
+          }
+          if exit_code != 0
+            raise "Failed to execute post_start hook: #{hook['cmd']} (exit code: #{exit_code}"
+          end
+        end
+        true
+      rescue => exc
+        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
+        error exc.message
+        false
+      end
+
+      # @param [String] type
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        service_pod.hooks.select{ |h| h['type'] == type }
+      end
+
+      # @param [String] id
+      # @param [Array<String>] lines
+      # @param [String] type
+      def log_hook_output(id, lines, type)
+        lines.each do |chunk|
+          data = {
+              id: id,
+              time: Time.now.utc.xmlschema,
+              type: type,
+              data: chunk
+          }
+          rpc_client.async.notification('/containers/log', [data])
+        end
+      end
+
+      # @param [String] type
+      # @param [String] data
+      # @param [Integer] severity
+      def log_service_pod_event(type, data, severity = Logger::INFO)
+        super(service_pod.service_id, service_pod.instance_number, type, data, severity)
+      end
+    end
+  end
+end
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -24,7 +24,7 @@ module Kontena
         service_container.start!
         pre_start_hooks.each do |hook|
           info "running pre_start hook: #{hook['cmd']}"
-          command = ['/bin/sh', '-c', hook['cmd']]
+          command = ['/w/w', '/bin/sh', '-c', hook['cmd']]
           log_hook_output(service_container.id, ["running pre_start hook: #{hook['cmd']}"], 'stdout')
           _, _, exit_code = service_container.exec(command) { |stream, chunk|
             log_hook_output(service_container.id, [chunk], stream)
@@ -46,7 +46,7 @@ module Kontena
       def on_post_start(service_container)
         hooks_for('post_start').each do |hook|
           info "running post_start hook: #{hook['cmd']}"
-          command = ['/bin/sh', '-c', hook['cmd']]
+          command = ['/w/w', '/bin/sh', '-c', hook['cmd']]
           log_hook_output(service_container.id, ["running post_start hook: #{hook['cmd']}"], 'stdout')
           _, _, exit_code = service_container.exec(command) { |stream, chunk|
             log_hook_output(service_container.id, [chunk], stream)
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -16,9 +16,10 @@ module Kontena
         pre_start_hooks = hooks_for('pre_start')
         return if pre_start_hooks.size == 0
 
-        service_config = config_container(service_pod.dup)
-        service_config['Entrypoint'] = '/bin/sh'
-        service_config['Cmd'] = ['-c', 'sleep 600']
+        pre_pod = service_pod.dup
+        pre_pod.entrypoint = '/bin/sh'
+        pre_pod.cmd = ['-c', 'sleep 600']
+        service_config = config_container(pre_pod)
         service_container = create_container(service_config)
         service_container.start!
         pre_start_hooks.each do |hook|
--- a/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
+++ b/agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
@@ -146,7 +146,7 @@ module Kontena
           service_id: service_pod.service_id,
           instance_number: service_pod.instance_number
         }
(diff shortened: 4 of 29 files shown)


agent/lib/kontena/service_pods/creator.rb
```diff
@@ -46,6 +46,8 @@ def perform
           log_service_pod_event("service:create_instance", "removed previous version of service #{service_pod.name_for_humans} instance")
         end
 
+        self.run_pre_start_hooks
```
SpComb: Abort the service pod create if the pre start hooks fail?

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -78,21 +80,44 @@ def perform
...
+      def run_pre_start_hooks
+        hooks_for('pre_start').each do |hook|
+          service_config = config_container(service_pod.dup)
+          service_config['Cmd'] = ['/bin/sh', '-c', hook['cmd']]
+          service_container = nil
+          begin
+            service_container = create_container(service_config)
+            info "running pre_start hook: #{hook['cmd']}"
+            service_container.tap(&:start).attach { |stream, chunk|
+              log_hook_output(service_container.id, [chunk], stream)
```
SpComb: Do these get logged twice, because the `LogWorker` may also notice these exec containers?

agent/lib/kontena/service_pods/creator.rb
jakolehm: Yes.

agent/lib/kontena/service_pods/creator.rb
jnummelin: AFAIK `exec` streams will not be picked by Docker for the container log streams

jnummelin: uh, scratch that. these are not execs... So yes, these will be picked up twice I think

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -78,21 +80,43 @@ def perform
...
-            log_hook_output(service_container.id, ["running #{type} hook: #{hook['cmd']}"], 'stdout')
-            stdout, stderr, exit_code = service_container.exec(command)
-            log_hook_output(service_container.id, stdout, 'stdout')
-            log_hook_output(service_container.id, stderr, 'stderr')
-            if exit_code != 0
-              raise "Failed to execute hook: #{hook['cmd']}"
+      def run_pre_start_hooks
+        hooks_for('pre_start').each do |hook|
+          service_config = config_container(service_pod.dup)
+          service_config['Cmd'] = ['/bin/sh', '-c', hook['cmd']]
```
jnummelin: Should this override the entrypoint? or perhaps override that entrypoint is `['/bin/sh', '-c']` and then put the cmd in place. I think users will expect to execute script like things with hooks and now if my entrypoint is some binary then these will most likely fail.

Also could there be only one temp container for all the pre-hooks? That would mean to modify the entrypoint to be some "pause" thing and then all the hooks would be `exec`'s.

SpComb: It needs the `/w/w` entrypoint if the pre-start hooks require weave overlay network access.

agent/lib/kontena/service_pods/creator.rb
jakolehm: Fixed.

agent/lib/kontena/service_pods/creator.rb
jakolehm: Fixed.

agent/lib/kontena/service_pods/creator.rb
jakolehm: @jnummelin good (entry)point. Maybe it's doable to set `/bin/sh` to entrypoint before network manager modifies it.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -78,21 +80,41 @@ def perform
...
+          service_container = nil
+          begin
+            service_container = create_container(service_config)
```
SpComb: Using the complete `service_pod.service_config` including all labels etc. means that running the pre-start hooks will also have some side-effects, like `kontena.local` DNS and LB registrations? That means that client traffic might start getting routed to the pre-start container, which will likely fail?

agent/lib/kontena/service_pods/creator.rb
jnummelin: > It needs the `/w/w` entrypoint if the pre-start hooks require weave overlay network access.

Yes, definitely. But it still needs to "overwrite" the images entrypoint in order to be able to execute anything.

What about having a single temp container to execute all pre-start hooks? Now if I have 3 hooks to execute, there would be 9 (3 pre-start containers, 3 weave exec, 3 (or even 6?) weave attach) temp containers during the startup of a service. Feels a bit too much IMHO. :/

agent/lib/kontena/service_pods/creator.rb
jakolehm: Does it really matter, routing might fail anyway? If we later move this pre_start logic to exec inside actual service container situation stays the same => network is up but nothing is listening while pre_start hooks are executed.

agent/lib/kontena/service_pods/creator.rb
SpComb: In order to ensure that the pre-start hook logs actually get picked up, it might be better to fix this the other way: create the hook containers with `io.kontena.container.skip_logs=1` and use the `attach do log_hook_output` instead?

agent/lib/kontena/service_pods/creator.rb
SpComb: It will cause some churn that could be avoided, but I suppose it's still acceptable.

jakolehm: @SpComb do you suggest that we remove dns related labels?

agent/lib/kontena/service_pods/common.rb
```diff
@@ -1,9 +1,36 @@
...
 module Kontena
   module ServicePods
     module Common
       include Kontena::Helpers::EventLogHelper
+      include Kontena::Helpers::WeaveHelper
+
+      # Docker create configuration for ServicePod
+      # @param [ServicePod] service_pod
+      # @return [Hash] Docker create API JSON object
+      def config_container(service_pod)
```
kke: I think this Common could be converted to a class such as `Kontena::ServicePods::DockerContainer`

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+
+      # @return [Boolean]
+      def on_pre_start
+        pre_start_hooks = hooks_for('pre_start')
+        return if pre_start_hooks.size == 0
+
+        service_config = config_container(service_pod.dup)
+        service_config['Entrypoint'] = '/bin/sh'
+        service_config['Cmd'] = ['-c', 'sleep 600']
```
SpComb: Compared to the previous version which left the `/w/w` entrypoint in place, this means that the `pre_start` hooks may or may not run with weave attached, depending on the timing?

Although I suppose the same also applies to the `post_start` hooks?

jakolehm: Right, actually this is a bug. Entrypoint should be set before `config_container` call.

jakolehm: Fixed:

```
2017-08-22T08:08:34.000Z [nginx.nginx-1]: running pre_start hook: ifconfig
2017-08-22T08:08:34.000Z [nginx.nginx-1]: eth0      Link encap:Ethernet  HWaddr 02:42:AC:11:00:02  
          inet addr:172.17.0.2  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::42:acff:fe11:2/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:5 errors:0 dropped:0 overruns:0 frame:0
          TX packets:3 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:446 (446.0 B)  TX bytes:266 (266.0 B)

ethwe     Link encap:Ethernet  HWaddr 9E:BC:F2:95:09:5C  
          inet addr:10.81.128.32  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::9cbc:f2ff:fe95:95c/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1376  Metric:1
          RX packets:3 errors:0 dropped:0 overruns:0 frame:0
          TX packets:4 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:266 (266.0 B)  TX bytes:308 (308.0 B)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

2017-08-22T08:08:34.000Z [nginx.nginx-1]: running pre_start hook: echo "hello world 2"
2017-08-22T08:08:34.000Z [nginx.nginx-1]: hello world 2
2017-08-22T08:08:45.000Z [nginx.nginx-1]: running post_start hook: echo "hello world 3"
2017-08-22T08:08:45.000Z [nginx.nginx-1]: hello world 3
```

jakolehm: But yes, there is a race cond between exec and weave attach.. not sure how we can fix that?

jnummelin: "Prefix" each pre_start command with `/w/w`?

jakolehm: hacks

but yeah, probably works.

jnummelin: yep, not beautiful at all :/

jakolehm: fixed

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -16,14 +16,15 @@ def on_pre_start
...
+        pre_pod = service_pod.dup
+        pre_pod.entrypoint = '/bin/sh'
+        pre_pod.cmd = ['-c', 'sleep 600']
+        service_config = config_container(pre_pod)
         service_container = create_container(service_config)
         service_container.start!
         pre_start_hooks.each do |hook|
           info "running pre_start hook: #{hook['cmd']}"
-          command = ['/bin/sh', '-c', hook['cmd']]
+          command = ['/w/w', '/bin/sh', '-c', hook['cmd']]
```
SpComb: The `/w/w` is only for the default `net: bridge` containers, but not `net: host`

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      # @return [Boolean]
+      def on_pre_start
+        pre_start_hooks = hooks_for('pre_start')
+        return if pre_start_hooks.size == 0
+
+        pre_pod = service_pod.dup
+        pre_pod.entrypoint = '/bin/sh'
+        pre_pod.cmd = ['-c', 'sleep 600']
```
jnummelin: So this makes the effective entrypoint to be `/w/w` and cmd will be `['/bin/sh', '-c', 'sleep 600']`. `/bin/sh -c` itself as entrypoint will NOT handle signals properly, not 100% sure if it behaves any better when invoked through `/w/w`. If the sleep does not properly get the signals it will make the containers stop to wait unnecessarily for the `timeout`

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
SpComb: :+1: for also fixing the `post_start` hooks to be run with weave attached, though.

Worstcase is if the weave attach tends to win the race and users write hooks that assume that weave is attached... and they work until the timing subtly changes and they start breaking spontaneously.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,94 @@
...
+        service_config = config_container(pre_pod)
+        service_container = create_container(service_config)
+        service_container.start!
+        pre_start_hooks.each do |hook|
+          info "running pre_start hook: #{hook['cmd']}"
+          command = ['/w/w', '/bin/sh', '-c', hook['cmd']]
```
jnummelin: Now if we prefix each command with `/w/w` we don't need the entrypoint "hacking" for this tmp container?

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jnummelin: With quick testing it seems that it will NOT handle signals any better through `/w/w` :/

SpComb: What about `pre_pod.entrypoint = nil`, `pre_pod.cmd = ['sh', '-c', 'exec sleep 600']`?

BTW the `sleep 600` imposes an implicit 10-minute timeout on the `pre_start` hooks like this, I assume the execs will fail if the main container process exits?

SpComb: Nope, coreutils `sleep` will [restart `nanosleep` on `EINTR`](https://fossies.org/linux/coreutils/lib/xnanosleep.c#l_53), so you can't kill it when running as pid1, since it's missing the default term action for SIGTERM.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        service_pod.hooks.select{ |h| h['type'] == type }
+      end
+
+      # @param [String] cmd
+      # @return [Array<String>]
+      def build_cmd(cmd)
+        command = ['/bin/sh', '-c', cmd]
+        command.unshift('/w/w') if service_pod.can_expose_ports?
```
SpComb: Doesn't this result in double-`/w/w` if called like `on_pre_start`-> `config_container` -> `build_cmd`? The `Common#config_contianer` -> `Weave#modify_create_opts` will set `entrypoint: '/w/w'` and then this will be `cmd: ['/w/w', ...]`?

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+
+        command
+      end
+
+      # @param [Hash] service_config
+      # @param [String] cmd
+      # @return [Hash]
+      def config_container(service_config, cmd)
+        service_config = super(service_config)
```
SpComb: Do we still want to override the `entrypoint`, in case of images/service configs that set a non-shell entrypoint? I don't think this is doing that anymore?

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+        service_config['HostConfig'].delete('RestartPolicy')
+        service_config['Labels']['io.kontena.container.type'] = 'service_hook'
```
SpComb: Interesting... should the `WeaveWorker#register_container_dns` and `Kontena::LoadBalancers::Registrator` be changed to skip unless `service_container?`

agent/lib/kontena/service_pods/starter.rb
```diff
@@ -8,28 +9,28 @@ class Starter
...
+        service_container = get_container(service_pod.service_id, service_pod.instance_number)
         unless service_container.running?
           info "starting service: #{service_container.name_for_humans}"
           log_service_pod_event(
-            self.service_id, self.instance_number,
+            service_pod.service_id, service_pod.instance_number,
             "service:start_instance", "starting service instance #{service_container.name_for_humans}"
           )
-          service_container.restart!('timeout' => service_container.stop_grace_period)
+          hook_manager.on_pre_start(service_container)
```
SpComb: I personally disagree about running these hooks also during container restarts, but I can accept this as a design choice. 

BTW should this also then run the `post_start` hooks? It hasn't run those before. It will become more significant once the agent container restart policies are implemented, and all container restarts will start using `Kontena::ServicePods::Starter`. ATM docker restart policies will not run the start hooks.

docs/references/kontena-yml.md
```diff
@@ -159,6 +159,24 @@ affinity:
 
 #### hooks
 
+**pre_start**
+
+`pre_start` hooks define commands that are executed before a container is started on a node. If multiple hooks are provided they are executed in the order defined. Note that these commands are executed *before* `wait_for_port` check.
```
SpComb: * If any pre_start hook fails, then the remaining hooks will be skipped, and the container will not be created
* Any changes to the container filesystem from pre-start hooks will be reset when starting the actual container
* The pre-start hooks are run with a different overlay network IP address, and cannot be used to e.g. register the container IP somewhere (use post-start hooks)

agent/lib/kontena/service_pods/restarter.rb
```diff
@@ -8,29 +8,27 @@ class Restarter
       include Kontena::Logging
       include Common
 
-      attr_reader :service_id, :instance_number
+      attr_reader :service_pod
 
-      # @param [String] service_id
-      # @param [Integer] instance_number
-      def initialize(service_id, instance_number)
```
SpComb: The RPC handler also needs updating:

```
D, [2017-08-24T11:19:12.785001 #1] DEBUG -- Kontena::RpcServer: rpc request: Kontena::Rpc::ServicePodsApi#restart ["599eb591b2c7c30010a2a4fb", 1]
E, [2017-08-24T11:19:12.785469 #1] ERROR -- Kontena::RpcServer: ArgumentError: wrong number of arguments (given 2, expected 1)
E, [2017-08-24T11:19:12.787071 #1] ERROR -- Kontena::RpcServer: /app/lib/kontena/service_pods/restarter.rb:14:in `initialize'
/app/lib/kontena/rpc/service_pods_api.rb:11:in `new'
/app/lib/kontena/rpc/service_pods_api.rb:11:in `restart'
/app/lib/kontena/rpc_server.rb:45:in `handle_request'
```

agent/lib/kontena/service_pods/starter.rb
SpComb: Does not expect a container argument:

```
I, [2017-08-24T11:20:26.422253 #1]  INFO -- Kontena::ServicePods::Starter: starting service: /hooks-prestarts/redis-1
W, [2017-08-24T11:20:26.428535 #1]  WARN -- Kontena::Workers::ServicePodWorker: failed to sync hooks-prestarts.redis-1 at 2017-08-24 11:16:40 UTC: wrong number of arguments (given 1, expected 0)
W, [2017-08-24T11:20:26.429172 #1]  WARN -- Kontena::Workers::ServicePodWorker: wrong number of arguments (given 1, expected 0) (ArgumentError)
/app/lib/kontena/service_pods/lifecycle_hook_manager.rb:15:in `on_pre_start'
/app/lib/kontena/service_pods/starter.rb:29:in `perform'
/app/lib/kontena/workers/service_pod_worker.rb:131:in `ensure_started'
/app/lib/kontena/workers/service_pod_worker.rb:103:in `ensure_desired_state'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
/app/lib/kontena/workers/service_pod_worker.rb:33:in `update'
```

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+      def on_pre_start
+        pre_start_hooks = hooks_for('pre_start')
+        return if pre_start_hooks.size == 0
+
+        pre_start_hooks.each do |hook|
+          begin
+            info "running pre_start hook: #{hook['cmd']}"
+            service_config = config_container(service_pod.dup, hook['cmd'])
+            service_container = create_container(service_config)
+            service_container.tap(&:start).attach
```
SpComb: Testing with hooks that execute very quickly (just `echo ...`), I somehow managed to get this stuck here:

```
W, [2017-08-24T11:32:27.447220 #1]  WARN -- Kontena::Agent: Thread 1wvzk9pw 
W, [2017-08-24T11:32:27.447410 #1]  WARN -- Kontena::Agent: /usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/socket.rb:263:in `select'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/socket.rb:263:in `select_with_timeout'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/socket.rb:175:in `rescue in read_nonblock'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/socket.rb:156:in `read_nonblock'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/socket.rb:36:in `read'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/response.rb:163:in `parse'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/middlewares/response_parser.rb:7:in `response_call'
/usr/lib/ruby/gems/2.4.0/gems/docker-api-1.33.6/lib/excon/middlewares/hijack.rb:45:in `response_call'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/connection.rb:389:in `response'
/usr/lib/ruby/gems/2.4.0/gems/excon-0.58.0/lib/excon/connection.rb:253:in `request'
/usr/lib/ruby/gems/2.4.0/gems/docker-api-1.33.6/lib/docker/connection.rb:40:in `request'
/usr/lib/ruby/gems/2.4.0/gems/docker-api-1.33.6/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
/usr/lib/ruby/gems/2.4.0/gems/docker-api-1.33.6/lib/docker/container.rb:123:in `attach'
/app/lib/kontena/service_pods/lifecycle_hook_manager.rb:24:in `block in on_pre_start'
/app/lib/kontena/service_pods/lifecycle_hook_manager.rb:19:in `each'
/app/lib/kontena/service_pods/lifecycle_hook_manager.rb:19:in `on_pre_start'
/app/lib/kontena/service_pods/creator.rb:50:in `perform'
/app/lib/kontena/workers/service_pod_worker.rb:120:in `ensure_running'
/app/lib/kontena/workers/service_pod_worker.rb:100:in `ensure_desired_state'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
/app/lib/kontena/workers/service_pod_worker.rb:33:in `update'
```

`docker inspect` shows that the hook container srated and finished in ~60ms with an exit code of 0. Is there some kind of race between the `start` and `attach`?

agent/lib/kontena/service_pods/starter.rb
SpComb: Should these hooks also run in the `kontena service restart` -> `Kontena::ServicePods::Restarter`?

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jnummelin: IMO yes. That would make DNS and LB config less "bouncy" with these hooks containers.

agent/lib/kontena/service_pods/starter.rb
jnummelin: If we make `pre_start` hooks to be run on every restart, `post_start` should also be run similarly I think

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jakolehm: Yeah, fixed.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jakolehm: Done.

jakolehm:
@SpComb @jnummelin PTAL

SpComb:
Needs (trivial) merge with #2776

agent/lib/kontena/rpc/service_pods_api.rb
```diff
@@ -4,18 +4,36 @@ module Kontena
...
         {}
       end
+
+      private
+
+      # @param service_id [String]
+      # @param instance_number [Integer]
+      def find_service_pod(service_id, instance_number)
+        worker = Celluloid::Actor[:service_pod_manager].workers.find { |w|
+          w.service_pod.service_id == service_id && w.service_pod.instance_number == instance_number
```
SpComb: Hmm, feels like a bad idea to share the `Hash` across actor threads... this should really be something like `Celluloid::Actor[:service_pod_manager].lookup_service_pod(service_id, instance_number)`

SpComb: The two calls to `w.service_pod` are both also sync actor calls to each `ServicePodWorker`. If those actors are blocked in the exclusive `apply`, then this can also end up blocking the RPC server for a long time.

SpComb:
I think my suggestion for 1.4.0 would be to leave out the `restart` and `start` changes. Keep the existing `post_start` behavior as-is (only run it after container create, not after each container restart), and add the `pre_start` hook to the `ServicePods::Creator`. The `pre_stop` hook should run for both `ServicePods::Stopper`, and also `ServicePods::Terminator` if the container is indeed still running.

agent/lib/kontena/rpc/service_pods_api.rb
jakolehm: Yes, this was a bad idea. Reverted these changes.

jakolehm:
`pre_stop` is now executed also when instance is terminated.

@SpComb what do you mean by leaving out `start` changes?

SpComb:
Running the `start` hooks for container die/start events (per `ServicePodWorker` restart policy -> `ServicePods::Starter`) breaks `oneshot` hooks badly:

```
$ bin/kontena stack install test/hooks-oneshot.yaml 
 [done] Creating stack hooks-oneshot      
 [done] Triggering deployment of stack hooks-oneshot     
$ bin/kontena service logs hooks-oneshot/redis
2017-09-19T12:02:06.000Z [1]: running post_start hook: echo "post-start hook 2 (oneshot)"
2017-09-19T12:02:06.000Z [1]: post-start hook 2 (oneshot)
2017-09-19T12:02:08.000Z [1]: 
2017-09-19T12:02:08.000Z [1]: *** FATAL CONFIG FILE ERROR ***
2017-09-19T12:02:08.000Z [1]: Reading the configuration file, at line 2
2017-09-19T12:02:08.000Z [1]: >>> 'fail'
2017-09-19T12:02:08.000Z [1]: Bad directive or wrong number of arguments
2017-09-19T12:02:09.000Z [1]: running post_start hook: echo "post-start hook 2 (oneshot)"
2017-09-19T12:02:09.000Z [1]: post-start hook 2 (oneshot)
2017-09-19T12:02:12.000Z [1]: running post_start hook: echo "post-start hook 2 (oneshot)"
2017-09-19T12:02:12.000Z [1]: post-start hook 2 (oneshot)
2017-09-19T12:02:13.000Z [1]: 
2017-09-19T12:02:13.000Z [1]: *** FATAL CONFIG FILE ERROR ***
2017-09-19T12:02:13.000Z [1]: Reading the configuration file, at line 2
2017-09-19T12:02:13.000Z [1]: >>> 'fail'
2017-09-19T12:02:13.000Z [1]: Bad directive or wrong number of arguments
2017-09-19T12:02:17.000Z [1]: running post_start hook: echo "post-start hook 2 (oneshot)"
2017-09-19T12:02:18.000Z [1]: post-start hook 2 (oneshot)
2017-09-19T12:02:19.000Z [1]: 
```

```yaml
stack: test/hooks-oneshot
services:
  redis:
    image: redis
    command: redis-server --fail
    hooks:
      post_start:
        - name: test-oneshot
          cmd: echo "post-start hook 2 (oneshot)"
          instances: '*'
          oneshot: true
```

jakolehm:
Let's put this on hold until we figure out how to not break `oneshot`.

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -14,10 +14,21 @@ def cached_pod(service_instance)
...
-      @lru_cache.getset(cache_key) {
+      hooks = service_instance.grid_service.hooks
+      pod_hash = @lru_cache.getset(cache_key) {
         debug "pod_cache miss"
         ServicePodSerializer.new(service_instance).to_hash if service_instance.grid_service
       }
+      # remove cached oneshot hooks
+      hooks.dup.each do |hook|
+        if hook.done_for?(service_instance.instance_number)
+          pod_hash[:hooks].delete_if { |h|
```
SpComb: Mutating a cached/serialized blob is a bit of a hack... this is a cache, so IMO the sane way to implement this is as "only cache the serialized service pod if it does not contain oneshot hooks":

```ruby
      return nil unless service_instance.grid_service

      hooks = service_instance.grid_service.hooks

      unless pod_hash = @lru_cache[cache_key]
        debug "pod_cache miss"
        pod_hash = ServicePodSerializer.new(service_instance).to_hash
      end

      has_oneshot_hooks = hooks.any? { |hook|
        hook.oneshot && pod_hash[:hooks].any? { |h|
           h[:type] == hook.type && h[:cmd] == hook.cmd
        }
      }

      if has_oneshot_hooks
        debug "do not cache service #{pod_hash[:service_name]} pod with oneshot hooks"
      else
        @lru_cache[cache_key] = pod_hash
      end

      pod_hash
```

```
172.17.0.1 - - [19/Sep/2017:14:13:43 +0000] "POST /v1/grids/development/stacks HTTP/1.1" 201 1623 0.1662
D, [2017-09-19T14:13:43.531175 #20] DEBUG -- GridServices::Update: not updating service development/hooks-oneshot/redis without changes
172.17.0.1 - - [19/Sep/2017:14:13:43 +0000] "POST /v1/stacks/development/hooks-oneshot/deploy HTTP/1.1" 200 151 0.1393
I, [2017-09-19T14:13:43.582886 #20]  INFO -- StackDeployWorker: removing following services: 
I, [2017-09-19T14:13:43.593367 #20]  INFO -- StackDeployWorker: deploying service development/hooks-oneshot/redis...
I, [2017-09-19T14:13:43.737389 #8]  INFO -- GridServiceSchedulerWorker: starting development/hooks-oneshot/redis deploy
I, [2017-09-19T14:13:43.743659 #8]  INFO -- GridServiceDeployer: starting to deploy development/hooks-oneshot/redis
I, [2017-09-19T14:13:43.826322 #8]  INFO -- GridServiceDeployer: throttling service instance development/hooks-oneshot/redis deploy because of min_health limit (1 instances in-progress)
I, [2017-09-19T14:13:43.831448 #8]  INFO -- GridServiceInstanceDeployer: Deploying service instance development/hooks-oneshot/redis-1 to node core-01 at 2017-09-19 14:13:43 UTC...
D, [2017-09-19T14:13:43.856740 #20] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2017-09-19T14:13:43.869170 #20] DEBUG -- Rpc::ServicePodSerializer: fire development/hooks-oneshot/redis-1 oneshot hook post_start/test-oneshot
D, [2017-09-19T14:13:43.872349 #20] DEBUG -- Rpc::NodeServicePodHandler: do not cache service hooks-oneshot.redis pod with oneshot hooks
```

SpComb: BTW the `h[:type] == hook.type && h[:cmd] == hook.cmd` comparison isn't exactly unique... the hooks should really be serialized to include the `oneshot: true/false` flag.

jakolehm: The reason why I chose the mutate cached object path is perf.. without it we don't get any benefits from caching if there are oneshot hooks.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -84,7 +87,16 @@ def on_pre_stop(service_container)
       # @param [String] type
       # @return [Array<Hash>]
       def hooks_for(type)
-        service_pod.hooks.select{ |h| h['type'] == type }
+        hooks = service_pod.hooks.select { |h|
+          h['type'] == type && !h['oneshot'] || (h['oneshot'] && !oneshot_cache.include?(h))
+        }
+        hooks.each do |h|
+          if h['oneshot'] && !oneshot_cache.include?(h)
+            mark_oneshot_hook(h)
```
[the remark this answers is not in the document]
jakolehm: Just thinking should we remove `async` from there.. then if RPC fails it would not mark hook as done in cache.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,159 @@
...
+        super(service_pod.service_id, service_pod.instance_number, type, data, severity)
+      end
+
+      # @param hook [Hash]
+      def mark_oneshot_hook(hook)
+        pod = {
+          service_id: service_pod.service_id,
+          instance_number: service_pod.instance_number
+        }
+        rpc_client.async.request('/node_service_pods/mark_oneshot_hook', [node.id, pod, hook])
```
SpComb: > Just thinking should we remove async from there.. then if RPC fails it would not mark hook as done in cache.

I didn't notice that this was `async`. IMO yes, it's safer to make this sync, because that means that errors cause oneshot hooks to never get executed, rather than having them get executed multiple times.

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -66,6 +63,27 @@ def set_state(id, pod)
...
+      node = @grid.host_nodes.find_by(node_id: id)
+      raise 'Node not found' unless node
+
+      service_instance = node.grid_service_instances.find_by(
+        grid_service_id: pod['service_id'], instance_number: pod['instance_number']
+      )
+      raise 'Instance not found' unless service_instance
+
+      oneshot_hook = service_instance.grid_service.hooks.to_a.find{ |h|
+        h.oneshot && h.type == hook['type'] && h.cmd == hook['cmd']
```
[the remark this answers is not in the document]
jakolehm: Not sure if we can trust that embedded `hooks` collection does not get wiped and replaced with new id?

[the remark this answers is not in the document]
SpComb: This should have been fixed in  #2097, and there's a [spec](https://github.com/kontena/kontena/blob/6b134c6f96d44ff05fe0056f188d8979f20a62f6/server/spec/mutations/grid_services/update_spec.rb#L368-L386) for it.

Besides, the `hook['id']` would only be used briefly between the `list` -> `mark_oneshot_hook` RPCs.

And the same `GridServiceHook` object also contains the `done` array, if the embedded object gets wiped, then we'll have bigger problems with the `oneshot` hooks anyways.

agent/lib/kontena/service_pods/creator.rb
```diff
@@ -63,50 +70,11 @@ def perform
...
-      # @param [String] type
-      def log_hook_output(id, lines, type)
-        lines.each do |chunk|
-          data = {
-              id: id,
-              time: Time.now.utc.xmlschema,
-              type: type,
-              data: chunk
-          }
-          rpc_client.async.notification('/containers/log', [data])
```
SpComb: Fixes #2842

server/app/services/rpc/node_service_pod_handler.rb
```diff
@@ -7,17 +7,25 @@ def initialize(grid)
...
       }
-      @lru_cache.getset(cache_key) {
+      serializer = ServicePodSerializer.new(service_instance)
+      pod_hash = @lru_cache.getset(cache_key) {
         debug "pod_cache miss"
-        ServicePodSerializer.new(service_instance).to_hash if service_instance.grid_service
+        serializer.to_hash
       }
+      # hooks cannot be cached
+      pod_hash[:hooks] = serializer.build_hooks
```
SpComb: re-serializing the hooks for each service lookup is simpler, but doesn't it also end up being more expensive than the alternative policy of not caching serialized pods that contain oneshot hooks?

Like in https://github.com/kontena/kontena/pull/2701#discussion_r139705110 (although that writes to the LRU cache every time)

jakolehm: Generating hooks should be lightweight because it does not fetch anything from the database (all data is already loaded).

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,159 @@
...
+        log_service_pod_event("service:create_instance", exc.message, Logger::ERROR)
+        error exc.message
+        false
+      end
+
+      # @param [String] type
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        hooks = service_pod.hooks.select { |h|
+          h['type'] == type && !h['oneshot'] || (h['oneshot'] && !oneshot_cache.include?(h))
```
SpComb: This is also returning `oneshot` `post_start` hooks for `type = 'pre_start'`, so they get run as `pre_start` hooks:

```yaml
stack: test/hooks-oneshot
services:
  redis:
    image: redis
    hooks:
      post_start:
        - name: test-first
          cmd: date && echo "first"
          instances: '*'
          oneshot: false
        - name: test-echo1
          cmd: date && echo "pre-sleep"
          instances: '*'
          oneshot: true
        - name: test-sleep1
          cmd: sleep 5
          instances: '*'
          oneshot: true
        - name: test-echo2
          cmd: date && echo "mid-sleep"
          instances: '*'
          oneshot: true
        - name: test-sleep2
          cmd: sleep 5
          instances: '*'
          oneshot: true
        - name: test-echo3
          cmd: date && echo "post-sleep"
          instances: '*'
          oneshot: true
        - name: test-last
          cmd: date && echo "last"
          instances: '*'
          oneshot: false
```

```
I, [2017-09-20T11:44:36.983241 #1]  INFO -- Kontena::ServicePods::Creator: creating service: hooks-oneshot.redis-1
I, [2017-09-20T11:44:36.983601 #1]  INFO -- Kontena::Workers::ImagePullWorker: pulling image: redis:latest
I, [2017-09-20T11:44:39.707346 #1]  INFO -- Kontena::Workers::ImagePullWorker: pulled image: redis:latest
I, [2017-09-20T11:44:39.883092 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: date && echo "pre-sleep"
I, [2017-09-20T11:44:41.903369 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: sleep 5
I, [2017-09-20T11:44:49.354710 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: date && echo "mid-sleep"
I, [2017-09-20T11:44:51.535713 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: sleep 5
I, [2017-09-20T11:44:58.802024 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: date && echo "post-sleep"
D, [2017-09-20T11:45:01.448684 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: hooks-oneshot.redis-1
D, [2017-09-20T11:45:01.543426 #1] DEBUG -- Kontena::ServicePods::Creator: container created: hooks-oneshot.redis-1
I, [2017-09-20T11:45:01.962542 #1]  INFO -- Kontena::ServicePods::Creator: service started: hooks-oneshot/redis-1
I, [2017-09-20T11:45:01.967796 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "first"
I, [2017-09-20T11:45:02.175671 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "last"
```

server/app/services/rpc/node_service_pod_handler.rb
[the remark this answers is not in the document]
SpComb: This really needs the `hook['id']`, there can be multiple identical commands in the service hooks, and this kind of `hooks.find{...}` will only mark the first such command:

```yaml
stack: test/hooks-oneshot
services:
  redis:
    image: redis
    hooks:
      post_start:
        - name: test-first
          cmd: date && echo "first"
          instances: '*'
          oneshot: false
        - name: test-oneshot1
          cmd: date && echo "pre-sleep"
          instances: '*'
          oneshot: true
        - name: test-sleep1
          cmd: sleep 5
          instances: '*'
          oneshot: true
        - name: test-echo2
          cmd: date && echo "mid-sleep"
          instances: '*'
          oneshot: true
        - name: test-sleep2
          cmd: sleep 5
          instances: '*'
          oneshot: true
        - name: test-echo3
          cmd: date && echo "post-sleep"
          instances: '*'
          oneshot: true
        - name: test-last
          cmd: date && echo "last"
          instances: '*'
          oneshot: false
```

Initial deploy:

```
D, [2017-09-20T12:02:16.747939 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: hooks-oneshot.redis-1
D, [2017-09-20T12:02:17.024074 #1] DEBUG -- Kontena::ServicePods::Creator: container created: hooks-oneshot.redis-1
I, [2017-09-20T12:02:18.504467 #1]  INFO -- Kontena::ServicePods::Creator: service started: hooks-oneshot/redis-1
I, [2017-09-20T12:02:18.661459 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "first"
I, [2017-09-20T12:02:19.038625 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "pre-sleep"
I, [2017-09-20T12:02:19.298197 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: sleep 5
I, [2017-09-20T12:02:24.727551 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "mid-sleep"
I, [2017-09-20T12:02:24.842957 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: sleep 5
I, [2017-09-20T12:02:29.964046 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "post-sleep"
I, [2017-09-20T12:02:30.086937 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "last"
```

Restart the agent to clear out the `LifecycleHookManager@oneshot_cache`, and re-deploy the service:

```
I, [2017-09-20T12:03:02.891027 #1]  INFO -- Kontena::ServicePods::Creator: removing previous version of service: hooks-oneshot/redis-1
D, [2017-09-20T12:03:03.406122 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: hooks-oneshot.redis-1
D, [2017-09-20T12:03:03.496747 #1] DEBUG -- Kontena::ServicePods::Creator: container created: hooks-oneshot.redis-1
I, [2017-09-20T12:03:03.962722 #1]  INFO -- Kontena::ServicePods::Creator: service started: hooks-oneshot/redis-1
I, [2017-09-20T12:03:04.051490 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "first"
I, [2017-09-20T12:03:04.276366 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: sleep 5
I, [2017-09-20T12:03:09.404685 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "last"
```

The second `sleep 5` hook gets executed again, even though it's `oneshot`

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,159 @@
...
+          service_id: service_pod.service_id,
+          instance_number: service_pod.instance_number
+        }
+        rpc_client.request('/node_service_pods/mark_oneshot_hook', [node.id, pod, hook])
+        oneshot_cache << hook
+      end
+
+      # @return [Array<Hash>]
+      def oneshot_cache
+        @oneshot_cache ||= []
```
kke: Could be `@oneshot_cache ||= Set.new`. It will still take `<<` and `.include?` but is faster and don't have to iterate.

jakolehm: Good point.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jakolehm: Should be fixed now.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+
+      # @param [String] type
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        hooks = service_pod.hooks.select { |h|
+          h['type'] == type && !h['oneshot'] || (h['oneshot'] && !oneshot_cache.include?(h))
+        }
+        hooks.each do |h|
+          if h['oneshot'] && !oneshot_cache.include?(h)
```
SpComb: This `oneshot_cache.include?` is subtly broken, because it's using the `{ 'type' => ..., 'cmd' => ..., 'oneshot' => true/false}` as a "unique" key for the `oneshot_cache`, but it isn't: the sequence of hooks for a type can contain the same `cmd` multiple times. With some `DEBUG` logging inside this `if` it's apparent that this is only marking the first occurence of the `sleep 5` command, and it's not marking the second occurence:

```
D, [2017-09-20T12:27:22.021211 #1] DEBUG -- Kontena::ServicePods::Creator: creating container: hooks-oneshot.redis-1
D, [2017-09-20T12:27:22.311482 #1] DEBUG -- Kontena::ServicePods::Creator: container created: hooks-oneshot.redis-1
I, [2017-09-20T12:27:23.338078 #1]  INFO -- Kontena::ServicePods::Creator: service started: hooks-oneshot/redis-1
D, [2017-09-20T12:27:23.367677 #1] DEBUG -- Kontena::ServicePods::LifecycleHookManager: marking oneshot hook: {"type"=>"post_start", "cmd"=>"date && echo \"pre-sleep\"", "oneshot"=>true}
D, [2017-09-20T12:27:23.477041 #1] DEBUG -- Kontena::ServicePods::LifecycleHookManager: marking oneshot hook: {"type"=>"post_start", "cmd"=>"sleep 5", "oneshot"=>true}
D, [2017-09-20T12:27:23.487460 #1] DEBUG -- Kontena::ServicePods::LifecycleHookManager: marking oneshot hook: {"type"=>"post_start", "cmd"=>"date && echo \"mid-sleep\"", "oneshot"=>true}
D, [2017-09-20T12:27:23.496961 #1] DEBUG -- Kontena::ServicePods::LifecycleHookManager: marking oneshot hook: {"type"=>"post_start", "cmd"=>"date && echo \"post-sleep\"", "oneshot"=>true}
I, [2017-09-20T12:27:23.509313 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "first hook (non-oneshot)"
I, [2017-09-20T12:27:23.854263 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "pre-sleep"
I, [2017-09-20T12:27:24.102533 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: sleep 5
I, [2017-09-20T12:27:29.284095 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "mid-sleep"
I, [2017-09-20T12:27:29.448416 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: sleep 5
I, [2017-09-20T12:27:34.595113 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "post-sleep"
I, [2017-09-20T12:27:34.769682 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: date && echo "last hook (non-oneshot)"
```

I suggest that the agent `oneshot_cache` would also use `h['id']` as the key. It's feels very wrong to have overlapping keys for distinct entries in the `oneshot_cache`.

It's also subtly non-broken for the `service_pod.hooks.select { |h| ... oneshot_cache.include?(h)  }` case, because this does all of the per-type `oneshot_cache.include?(h)` lookups before marking any of them.  There's a risk of it skipping the duplicate hooks if that were written any differently.

server/app/services/rpc/node_service_pod_handler.rb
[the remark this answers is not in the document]
SpComb: Turns out this wasn't actually because of the way that the `find` works here, it was because the agent has a similar problem, and doesn't even mark the second `sleep 5` hook: https://github.com/kontena/kontena/pull/2701/files#r139954994

agent/spec/lib/kontena/service_pods/lifecycle_hook_manager_spec.rb
```diff
@@ -1,17 +1,22 @@
...
-      { 'type' => 'post_start', 'cmd' => 'sleep 2'},
-      { 'type' => 'pre_stop', 'cmd' => 'sleep 3'}
-    ])
+    double(:service_pod,
+      service_id: 'a',
+      instance_number: 2,
+      hooks: [
+        { 'type' => 'pre_start', 'cmd' => 'sleep 1'},
+        { 'type' => 'post_start', 'cmd' => 'sleep 2'},
+        { 'type' => 'pre_stop', 'cmd' => 'sleep 3'}
```
SpComb: I would suggest also adding some `'oneshot' => true` case to the specs.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      # @param [String] type
+      # @return [Array<Hash>]
+      def hooks_for(type)
+        hooks = service_pod.hooks.select { |h|
+          h['type'] == type && !cached_oneshot_hook?(h)
+        }
+        hooks.each { |h|
+          mark_oneshot_hook(h) if !cached_oneshot_hook?(h)
```
SpComb: This is also reporting non-oneshot hooks now:

```
D, [2017-09-20T13:29:26.844311 #1] DEBUG -- Kontena::ServicePods::LifecycleHookManager: mark oneshot hook: {"id"=>"59c26d311c7aaf000fe67683", "type"=>"post_start", "cmd"=>"date && echo \"first hook (non-oneshot)\"", "oneshot"=>false}
E, [2017-09-20T13:29:26.922872 #1] ERROR -- Kontena::ServicePods::LifecycleHookManager: RuntimeError: Hook not found: {"id"=>"59c26d311c7aaf000fe67683", "type"=>"post_start", "cmd"=>"date && echo \"first hook (non-oneshot)\"", "oneshot"=>false}
```

```
E, [2017-09-20T13:29:26.902442 #12] ERROR -- RpcServer: RuntimeError: Hook not found: {"id"=>"59c26d311c7aaf000fe67683", "type"=>"post_start", "cmd"=>"date && echo \"first hook (non-oneshot)\"", "oneshot"=>false}
D, [2017-09-20T13:29:26.902574 #12] DEBUG -- RpcServer: /app/app/services/rpc/node_service_pod_handler.rb:81:in `mark_oneshot_hook'
/app/app/services/rpc_server.rb:63:in `handle_request'
/app/app/services/rpc_server.rb:47:in `process!'
```

server/spec/serializers/rpc/service_pod_serializer_spec.rb
```diff
@@ -267,10 +267,33 @@
...
 
+  describe '#build_hooks' do
+    it 'returns not-executed oneshot hook' do
+      service.hooks.create(
+        type: 'post_start',
+        cmd: 'sleep 1',
+        oneshot: true
+      )
+      hooks = subject.build_hooks
+      expect(hooks[0]).to eq({ type: 'post_start', cmd: 'sleep 1', oneshot: true})
```
SpComb:
```
  1) Rpc::ServicePodSerializer#build_hooks returns not-executed oneshot hook

     Failure/Error: expect(hooks[0]).to eq({ type: 'post_start', cmd: 'sleep 1', oneshot: true})

       

       expected: {:type=>"post_start", :cmd=>"sleep 1", :oneshot=>true}

            got: {:id=>"59c2670e0087301ce4dc16ef", :type=>"post_start", :cmd=>"sleep 1", :oneshot=>true}

       

       (compared using ==)

       

       Diff:

       @@ -1,4 +1,5 @@

        :cmd => "sleep 1",

       +:id => "59c2670e0087301ce4dc16ef",

        :oneshot => true,

        :type => "post_start",

     # ./spec/serializers/rpc/service_pod_serializer_spec.rb:285:in `block (3 levels) in <top (required)>'
```

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
jakolehm: Fixed.

jakolehm:
Flakey observer spec, restarted travis job.. let's see if we can get green.

agent/lib/kontena/service_pods/lifecycle_hook_manager.rb
```diff
@@ -146,6 +146,8 @@ def log_service_pod_event(type, data, severity = Logger::INFO)
 
       # @param hook [Hash]
       def mark_oneshot_hook(hook)
+        return unless hook['oneshot']
```
SpComb: What about `mark_oneshot_hook(h) if h['oneshot']`?

server/spec/serializers/rpc/service_pod_serializer_spec.rb
```diff
@@ -276,13 +276,13 @@
...
     it 'returns not-executed oneshot hook' do
-      service.hooks.create(
+      hook = service.hooks.create!(
         type: 'post_start',
         cmd: 'sleep 1',
         oneshot: true
       )
       hooks = subject.build_hooks
-      expect(hooks[0]).to eq({ type: 'post_start', cmd: 'sleep 1', oneshot: true})
+      expect(hooks[0]).to eq({ id: hook.id.to_s, type: hook.type, cmd: hook.cmd, oneshot: hook.oneshot})
```
SpComb: Nothing wrong here specifically, but maybe explicit `expect(...).to match { id: String, type: 'post_start', cmd: 'sleep 1', oneshot: true})` assertions are better for specs in general.

SpComb:
The `io.kontena.container.type=service_hook` seems to work. Weave itself will still add a DNS entry for the `pre_start` hook's container hostname itself, but the agent does not add the service name to DNS until the real container starts:

```
INFO: 2017/09/21 10:23:15.514846 [nameserver ee:d3:06:82:31:7f] adding entry for 1e6368b4cc27ae6d5a09d0cd2f05d73acf87ad5bac9cd523ddbabbae49875427: redis-1.hooks-prestart.development.kontena.local. -> 10.81.128.19
INFO: 2017/09/21 10:23:15.520655 [nameserver ee:d3:06:82:31:7f] container '1e6368b4cc27ae6d5a09d0cd2f05d73acf87ad5bac9cd523ddbabbae49875427' is not running: removing
INFO: 2017/09/21 10:23:15.520757 [nameserver ee:d3:06:82:31:7f] tombstoning hostname=redis-1.hooks-prestart.development.kontena.local., container=1e6368b4cc27ae6d5a09d0cd2f05d73acf87ad5bac9cd523ddbabbae49875427, ip=10.81.128.19
INFO: 2017/09/21 10:23:15.520815 [nameserver ee:d3:06:82:31:7f] tombstoning entry redis-1.hooks-prestart.development.kontena.local. -> 10.81.128.19
ERRO: 2017/09/21 10:23:15.857384 Vetoed installation of hairpin flow FlowSpec{keys: [EthernetFlowKey{src: ee:d3:06:82:31:7f, dst: ae:2d:8c:e1:48:40} InPortFlowKey{vport: 1}], actions: [OutputAction{vport: 1}]}
INFO: 2017/09/21 10:23:16.901911 [nameserver ee:d3:06:82:31:7f] adding entry for c3b28efccfec8a6bd32233f805d6ff5a492db6a7abd0ac5ae58bdc4ab1fba250: redis.hooks-prestart.development.kontena.local. -> 10.81.128.24
INFO: 2017/09/21 10:23:16.903026 [nameserver ee:d3:06:82:31:7f] adding entry for c3b28efccfec8a6bd32233f805d6ff5a492db6a7abd0ac5ae58bdc4ab1fba250: redis-1.hooks-prestart.development.kontena.local. -> 10.81.128.24
ERRO: 2017/09/21 10:23:17.888375 Vetoed installation of hairpin flow FlowSpec{keys: [EthernetFlowKey{src: ee:d3:06:82:31:7f, dst: ae:2d:8c:e1:48:40} InPortFlowKey{vport: 1}], actions: [OutputAction{vport: 1}]}
INFO: 2017/09/21 10:23:18.022298 Discovered local MAC 6e:02:5e:e5:9d:0d
INFO: 2017/09/21 10:23:18.276247 [nameserver ee:d3:06:82:31:7f] adding entry for c3b28efccfec8a6bd32233f805d6ff5a492db6a7abd0ac5ae58bdc4ab1fba250: redis-1.hooks-prestart.development.kontena.local. -> 10.81.128.24
```

Same for the LB registration:

```
I, [2017-09-21T10:30:47.352320 #1]  INFO -- Kontena::Workers::ServicePodWorker: starting hooks-prestart.redis-1
I, [2017-09-21T10:30:47.439858 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running pre_start hook: echo pre-start hook output
I, [2017-09-21T10:30:51.127663 #1]  INFO -- Kontena::ServicePods::LifecycleHookManager: running post_start hook: echo post-start hook output
I, [2017-09-21T10:30:51.147191 #1]  INFO -- Kontena::LoadBalancers::Registrator: adding container hooks-prestart.redis-1 to load balancer ingress-lb/lb
D, [2017-09-21T10:30:51.515441 #1] DEBUG -- Kontena::Workers::ServicePodWorker: sync hooks-prestart/redis-1 state at 2017-09-21 10:30:44 UTC: {:service_id=>"59c3930dc953bf000df4c1e3", :instance_number=>1, :rev=>"2017-09-21 10:30:44 UTC", :state=>"running", :error=>nil}
```

PR #2703, opened 2017-08-17
CLI: Only include PLATFORM in stack default envs when env[PLATFORM] present

kke:
Fixes a bug introduced by #2677 and #2667 that made stacks require a PLATFORM in env.

Before:

```
$ bin/kontena stack validate redis.yml
 [error] RuntimeError : Variable validation failed: {"PLATFORM"=>{:presence=>"Required value missing"}}
```

After:

👍

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -40,7 +40,7 @@ module Kontena::Cli::Stacks
       end
 
       def dump_variables(reader)
-        vals = reader.variables.to_h(values_only: true).reject {|k,_| k == 'STACK' || k == 'GRID' }
+        vals = reader.variables.to_h(values_only: true).reject {|k,_| k == 'STACK' || k == 'GRID' || k == 'PLATFORM' }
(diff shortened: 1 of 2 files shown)


kke:
Hmm.

```
    def set_env_variables(stack, grid)
      ENV['STACK'] = stack
      ENV['GRID'] = grid
      ENV['PLATFORM'] = grid
    end
```

It should always be set.

jnummelin:
Yes, the idea was that `PLATFORM` is always set, just like `GRID` and `STACK`

jnummelin:
Hmm, testing with `master` branch: 
```
$ k stack install ~/code/stacks/redis-simple.yml 
 [done] Creating stack simple-redis      
 [done] Triggering deployment of stack simple-redis     

$ k stack validate ~/code/stacks/redis-simple.yml 
 [error] RuntimeError : Variable validation failed: {"PLATFORM"=>{:presence=>"Required value missing"}}
         See /Users/jussi/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception

```
How come install can succeed if validation fails?

kke:
Fixed in #2707

PR #2582, opened 2017-07-14
[WiP] Refactor agent launchers, weave to use Observable

SpComb:
Fixes #2134 launchers call `Docker::Container#start!` and raise errors
Fixes #1639 weave expose migration with `WEAVE_DEBUG=1`
Fixes #1397 weave launcher calls `launch-router` / `attach-router`
Hopefully fixes #1395 allow weave actors to recover from crashes

## TODO

* Rewrite specs
* Testing
* Refactor the `ObservableObserver` `update` pattern with smart error handling
* Optimize launcher/weave observers to avoid performance regressions

PR #2011, opened 2017-03-22
Refactor agent to use Observable node info

SpComb:
Replace the event-based `Celluloid::Notifications` `subscribe('agent:node_info')` with a new reactive `Observable` -> `Observer` interface that allows actors to restart after a crash and immediately pick up the existing state.

### `StatsWorker`

```
kontena-agent | I, [2017-03-22T09:18:14.924600 #1]  INFO -- Kontena::Workers::NodeInfoWorker: initialized
kontena-agent | I, [2017-03-22T09:18:14.948380 #1]  INFO -- Kontena::Workers::StatsWorker: initialized
kontena-agent | I, [2017-03-22T09:18:14.949128 #1]  INFO -- Kontena::Workers::StatsWorker: waiting for cadvisor
kontena-agent | I, [2017-03-22T09:18:14.971066 #1]  INFO -- Kontena::Workers::StatsWorker: cadvisor is running, starting stats loop
kontena-agent | D, [2017-03-22T09:18:14.972339 #1] DEBUG -- Workers::NodeInfoWorker: observe: #<Kontena::Workers::StatsWorker:0x005639615085e8>.node=
kontena-agent | D, [2017-03-22T09:18:16.865729 #1] DEBUG -- RpcServer: rpc notification: Kontena::Rpc::AgentApi#master_info [{"version"=>"1.2.0.dev"}]
kontena-agent | D, [2017-03-22T09:18:16.865977 #1] DEBUG -- RpcServer: rpc notification: Kontena::Rpc::AgentApi#node_info [{"id"=>"XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5", "created_at"=>"2016-10-27 08:15:31 UTC", "updated_at"=>"2017-03-20 14:40:38 UTC", "name"=>"core-01", "labels"=>["test"], "overlay_ip"=>"10.81.0.1", "peer_ips"=>["192.168.66.102"], "node_number"=>1, "initial_member"=>true, "grid"=>{"id"=>"development", "name"=>"development", "initial_size"=>1, "trusted_subnets"=>[], "subnet"=>"10.81.0.0/16", "supernet"=>"10.80.0.0/12", "stats"=>{"statsd"=>{"server"=>"127.0.0.2", "port"=>8125}}}}]
kontena-agent | D, [2017-03-22T09:18:16.868890 #1] DEBUG -- Workers::NodeInfoWorker: publishing node information
[...]

PR #2704, opened 2017-08-17
Implement synchronous Observers and lightweight Observables

SpComb:
Closes #2690 replaces `Celluloid::Condition` with `Kontena::Observable` for use by `RpcClient` 
Prerequisite for #2582 agent refactoring to replace `WaitHelper#wait_until!` with `observe`

Implements new `Kontena::Observer` primitives/behaviors required for the #2582 refactoring:

* Change the `Kontena::Observable` to be a standalone class, using a mutex to synchronize observers and observable updates 
* Change the `Kontena::Observer` to be a standalone class, using the actor mailbox to wait for observable updates
* Replace the `observer.async.update_observe` update calls with custom mailbox messages, handled by the observing task using [`Celluloid.receive`](https://github.com/celluloid/celluloid/wiki/Protocol-Interaction)
* Changed async `observe(observable) do |value| ... end` to yield from the same task
* Implement sync `value = observe(observable)` that just returns instead of yielding
* The sync API also supports a timeout
* Use sync `@node = observe(Actor[:node_info_worker])` for service pod and volume managers
* Use lightweight `Observable` instances for the RPC client requests

With the mailbox-based observer protocol, the `Kontena::Observer` is now also usable outside of the top-level actor classes. This means that the `Kontena::Observer.observe` method can be used in normal object instances like the `Kontena::ServicePods::Creator`. The observer uses `Celluloid.current_actor.mailbox` to receive updates, and `Celluloid.receive` to suspend the observing task.

With the mutex-based observable synchronization, the `Kontena::Observable` is now also usable outside of the top-level actor class. It is used by the `RpcClient` actor as a kind of hybrid between a Future and a Condition.

Compared to the existing code, where the behavior of the `wait_until!` `sleep` would depend on what kind of class the `WaitHelper` was included into, and the `observer.async.update_observe` implementation only allowed the top-level actor object to observe anything.

The new observe interface requires the use of [`Celluloid.receive`](https://github.com/celluloid/celluloid/wiki/Protocol-Interaction) API to suspend the observing task while waiting for observable updates. This API isn't ideal, though, because apart from the special case of a synchronous observe of a single observable, the observer must be prepared to receive multiple observable update messages at any time. However, if the observing task ever suspends outside of the call to `Celluloid.receive`, then any observable messages received by the actor will just get discarded. Hence, the new `observe` implementation must make extensive use of `Celluloid.exclusive { ... }` - this allows any concurrent observable update messages to remain queued up in the actor mailbox until the task returns to the `Celluloid.receive` call.

- [x] Fix async observe race condition on `Kontena::Observable handles concurrent observers`
- [x] Fix sync observe wait for non-task callers (exclusive actor mode, or non-actor callers)
- [x] Figure out how to abort sync observe if the observable crashes
- [x] Implement exclusive async observe
- [x] The debug logging might have some overhead, and should probably use `debug { ... }` blocks
- [ ] The debug logging is too verbose for the `RpcClient` usecase (#2793)

#### `benchmark/observable-singleactor.rb`

Simulating the `Kontena::RpcClient` with 1000 requests at response delays between 0..1s:

```
wait        4.550000   0.860000   5.410000 (  5.910000)
condition   1.590000   0.660000   2.250000 (  2.853941)
observer    1.530000   0.450000   1.980000 (  2.706446)
future      1.480000   0.420000   1.900000 (  2.833874)
                    count          min          avg          max
wait                 1000     0.000682     0.011950     0.082162
condition            1000     0.000229     0.002769     0.085503
observer             1000     0.000262     0.003154     0.026616
future               1000     0.000195     0.002913     0.068220

wait        4.520000   0.780000   5.300000 (  6.023467)
                    count          min          avg          max
wait                 1000     0.000748     0.078273     0.650481

condition   1.420000   0.510000   1.930000 (  2.685398)
                    count          min          avg          max
condition            1000     0.000220     0.001424     0.042306

observer    1.910000   0.460000   2.370000 (  3.101071)
                    count          min          avg          max
observer             1000     0.000507     0.003363     0.064034

future      1.550000   0.430000   1.980000 (  2.776762)
                    count          min          avg          max
future               1000     0.000197     0.003098     0.058607
```

There is plenty of variance in the results, but it looks like the `Kontena::Observable` is comparable to the `Celluloid::Condition` and `Celluloid::Future` implementations, and each of those is significantly better than the existing `WaitHelper` based method. The `WaitHelper` implementation consumes significantly more CPU and has significantly higher latency.

Note that the condition implementation fails with timeouts when run with a nonzero `DELAY_MIN`... this tickles the race condition in https://github.com/kontena/kontena/pull/2690/files#r132900537

Note that the `WaitHelper` livelocks the actor with too many concurrent requests: https://github.com/kontena/kontena/pull/2704#discussion_r137786535

#### `benchmark/observable-actors.rb`

Simulating 1000 actors waiting/observing for some other actor to become ready:

```
wait       1.100000   0.140000   1.240000 (  1.197224)
observer   0.390000   0.040000   0.430000 (  1.101720)
                    count          min          avg          max
wait                 1000     0.095555     0.123384     0.151735
observer             1000     0.000818     0.072345     0.101241

wait       1.130000   0.110000   1.240000 (  1.189544)
                    count          min          avg          max
wait                 1000     0.096675     0.124369     0.151298

wait       1.140000   0.130000   1.270000 (  1.220306)
                    count          min          avg          max
wait                 1000     0.060735     0.120847     0.165023

observer   0.400000   0.050000   0.450000 (  1.187906)
                    count          min          avg          max
observer             1000     0.000534     0.105021     0.187681

observer   0.380000   0.060000   0.440000 (  1.190294)
                    count          min          avg          max
observer             1000     0.000507     0.111576     0.189190

wait       1.090000   0.140000   1.230000 (  1.178699)
observer   0.410000   0.060000   0.470000 (  1.257101)
                    count          min          avg          max
wait                 1000     0.068831     0.104866     0.153697
observer             1000     0.000603     0.133861     0.255509
```

The observer implementation has significantly less CPU usage and lower latency than the wait implementation.


SpComb:
Unfortunately there's still some race condition in the implementation, but it's caught by the specs:

```
  1) Kontena::Observable handles concurrent observers
     Failure/Error: expect(observers.map{|obs| obs.value}).to eq [update_count] * observer_count
     
       expected: [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
            got: [10, 10, 10, 10, 2, 2, 2, 3, 3, 4, 5, 6, 6, 7, 8, 8, 8, 8, 8, 9]
     
       (compared using ==)
     # ./spec/lib/kontena/observable_spec.rb:178:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:80:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:79:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:63:in `block (2 levels) in <top (required)>'
```

Based on the debug logs, it is related to the fake `Kontena::Observable::Message` sent at the end of `observe_async` if the observe is already ready... it somehow causes the values to reset back to the initial values returned by `add_observer`.

agent/lib/kontena/observer.rb
```diff
@@ -43,33 +68,92 @@ def values
...
-    # Observe has been updated, call if ready.
-    # Called from observe as an async task.
+    # Register celluloid actor handler for Kontena::Observable::Message.
     #
-    # @param observe [Observer::Observe]
-    def observed(observe)
-      observe.call if observe.ready?
+    # Updates the Observe, and calls if ready.
+    def register_observer_handler
+      @observer_handler ||= Thread.current[:celluloid_actor].handle(Kontena::Observable::Message) do |message|
```
SpComb: This needs to get memoized in some per-actor storage, not a per-instance variable. If multiple classes called by the same actor use `Kontena::Observer`, they would register multiple handlers... alternatively, the `handle` pattern would need to be something thats per-observer?

agent/lib/kontena/observer.rb
```diff
@@ -43,33 +68,92 @@ def values
...
+          !@done
+        end
+        def done!
+          @done = true
+          @task = nil
+        end
+
+        # Run block or resume task
+        def call
+          @task.resume(self)
```
SpComb: Need to deal with `@task` being nil if the `observe` timed out with a `Kontena::Observable::Message` queued... shouldn't kill the entire actor if the mailbox handler raises?

agent/lib/kontena/workers/service_pod_manager.rb
```diff
@@ -22,11 +21,8 @@ def initialize(autostart = true)
     end
 
     def start
-      observe(Actor[:node_info_worker]) do |node|
-        @node = node
-      end
+      @node = observe(Actor[:node_info_worker], timeout: 300.0)
```
SpComb: This is assuming that `Manager` actor is not interested in later `@node` updates...

agent/spec/lib/kontena/observer_spec.rb
```diff
@@ -35,45 +55,92 @@ def crash
...
+        observable.delay_update(object, delay: 0.5)
 
-      expect{subject.ready?}.to raise_error(Celluloid::DeadActorError)
+        # NOTE: the class must include the WaitHelper, so that it uses Celluloid#sleep
+        #       if the wait_until! uses Kernel#sleep and blocks the actor thread,
+        #       then this spec will fail, because the delayed update doesn't have a chance to run
+        expect(subject.observe(observable, timeout: 1.0)).to eq object
+      end
+
+      it 'does not lose wait messages' do
```
SpComb: I'm pretty sure it would be impossible to make this spec pass using the `Celluloid#receive` API (at 29c4db32160b5ad53c2bd89343eec94275b64c51):

```
D, [2017-08-17T20:20:39.172119 #28489] DEBUG -- TestObservable: waiter: Observer<TestObserver>...
DEBUG: wait Observable<TestObservable>... (timeout=1.0)
D, [2017-08-17T20:20:39.272178 #28489] DEBUG -- TestObservable: update: 1
D, [2017-08-17T20:20:39.272511 #28489] DEBUG -- TestObservable: notify: Observer<TestObserver> <- 1...
D, [2017-08-17T20:20:39.282949 #28489] DEBUG -- TestObservable: update: 2
D, [2017-08-17T20:20:39.293702 #28489] DEBUG -- TestObservable: update: 3
...
D, [2017-08-17T20:20:39.662842 #28489] DEBUG -- TestObservable: update: 38
```

```
Failures:

  1) Kontena::Observer For a single observable #wait_observable does not lose wait messages
     Failure/Error: expect(subject.wait_observable!(observable, timeout: 1.0)).to be_a Integer
     
     Timeout::Error:
       timeout waiting 1.00s until: Observable<TestObservable> is ready
     # ./vendor/bundle/ruby/2.3.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/responses.rb:29:in `value'
     # ./vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:45:in `value'
     # ./vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/proxy/sync.rb:22:in `method_missing'
     # ./spec/lib/kontena/observer_spec.rb:136:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:80:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:79:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:63:in `block (2 levels) in <top (required)>'
```

agent/lib/kontena/observer.rb
SpComb:
```
E, [2017-08-17T20:22:42.823957 #28622] ERROR -- : Actor crashed!
NoMethodError: undefined method `resume' for nil:NilClass
	/home/kontena/kontena/kontena/agent/lib/kontena/observer.rb:123:in `call'
	/home/kontena/kontena/kontena/agent/lib/kontena/observer.rb:143:in `block in register_observer_handler'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/handlers.rb:38:in `call'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/handlers.rb:20:in `handle_message'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:283:in `handle_message'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:156:in `block in run'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/timers-4.1.1/lib/timers/group.rb:66:in `wait'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:152:in `run'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:131:in `block in start'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-essentials-0.20.5/lib/celluloid/internals/thread_handle.rb:14:in `block in initialize'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:78:in `block in get_thread'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/group/spawner.rb:50:in `block in instantiate'
```

SpComb:
The race condition detected by `Kontena::Observable handles concurrent observers` should now be fixed, it was caused by proxied calls to `wrapped_object` in between the `value = observable.add_observer` and `observe.set(unwrap_observable(observable), value)`, which raced with observable update notifications, and would reset the observed value back to the initially observed value.

agent/lib/kontena/observer.rb
```diff
@@ -89,29 +268,103 @@ def update_observe(observe, observable, value)
...
+          observe.add(message.observable, message.value)
+        else
+          debug "observe async #{message.describe_observable}..."
+
+          observe.add(message.observable)
+        end
 
         # crash if observed Actor crashes, otherwise we get stuck without updates
         # this is not a bidrectional link: our crashes do not propagate to the observable
         self.monitor observable
```
SpComb: This won't work for standalone Observables... the sync observe is also missing linking to notice crashing observables, although it does have the timeout support.

SpComb:
Looks like we need to use `Celluloid.suspend` with a blocking wait method if we want to support being called from exclusive blocks...

```
W, [2017-08-18T14:11:23.966955 #1]  WARN -- Kontena::Workers::ServicePodWorker: Cannot suspend while in exclusive mode (RuntimeError)
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:61:in `suspend'
/app/lib/kontena/observer.rb:174:in `wait'
/app/lib/kontena/observer.rb:354:in `observe_sync'
/app/lib/kontena/observer.rb:247:in `observe'
/app/lib/kontena/service_pods/creator.rb:30:in `network_adapter'
/app/lib/kontena/service_pods/creator.rb:175:in `config_container'
/app/lib/kontena/service_pods/creator.rb:55:in `perform'
/app/lib/kontena/workers/service_pod_worker.rb:120:in `ensure_running'
/app/lib/kontena/workers/service_pod_worker.rb:97:in `ensure_desired_state'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
```

agent/spec/lib/kontena/observer_spec.rb
```diff
@@ -1,79 +1,198 @@
...
+        expect(subject).to be_observed
+        expect(subject.observed_values).to eq [object]
+      end
+
+      it "crashes if the observable does", :log_celluloid_actor_crashes => true do
+        subject.test_observe_async(observable)
+
+        expect{observable.crash}.to raise_error(RuntimeError)
+
+        expect{subject.ping; subject.observed?}.to raise_error(Celluloid::DeadActorError)
```
SpComb: Failing randomly

```
  1) Kontena::Observer For a single observable #observe crashes if the observable does

     Failure/Error: expect{subject.ping; subject.observed?}.to raise_error(Celluloid::DeadActorError)

       expected Celluloid::DeadActorError but nothing was raised

     # ./spec/lib/kontena/observer_spec.rb:119:in `block (4 levels) in <top (required)>'
```

agent/lib/kontena/observer.rb
```diff
@@ -43,75 +93,382 @@ def values
...
+
+      while !observe.ready?
+        receive_timeout = deadline ? deadline - Time.now : nil
+
+        if receive_timeout && receive_timeout < 0
+          raise Celluloid::TaskTimeout.new("wait deadline exceeded")
+        end
+
+        message = mailbox.receive(receive_timeout) { |msg|
+          debug "observe receive #{msg.class.name}"
```
SpComb: There seems to be some really nasty `O(N^2)` behavior here if the `wait_observe` blocks for a long time (`timeout: 300.0`), and lots of other messages (`Celluloid::Call::Async`) queue up - the agent starts chewing CPU.

agent/lib/kontena/observer.rb
```diff
@@ -1,37 +1,87 @@
...
+        @pending
+      end
+
+      # Have pending changes
+      def pending!
+        @pending = true
+      end
+
+      # Rsovle pending changes
+      def resolve!
```
SpComb: Considering `clean!` and `dirty!` instead of `resolve!` and `pending!`

agent/lib/kontena/observer.rb
SpComb: OTOH this blocking mailbox receive should match how the existing `wait_until! { Celluloid::Proxy::Cell.ready? }` calls are implemented in exclusive mode:

* [`Celluloid::Proxy::Sync#method_missing`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/proxy/sync.rb#L20-L22)
* [`Celluloid::Call::Sync#value`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/call/sync.rb#L44)
* [`Celluloid::Call::Sync#response`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/call/sync.rb#L41)
* [`Celluloid.suspend`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid.rb#L141)
* [`Celluloid::Call::Sync#wait`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/call/sync.rb#L51)

So I'd be surprised if this somehow performed any worse... I guess it's just the debug logging that made it chew up so much CPU?

agent/lib/kontena/observable.rb
```diff
@@ -43,33 +68,48 @@ def reset_observable
...
 
-          # XXX: is the Observable's Celluloid.current_actor guranteed to match the Actor[:node_info_worker] Celluloid::Proxy::Cell by identity?
-          observer.async.update_observe(observe, Celluloid.current_actor, @observable_value)
+          actor.mailbox << Message.new(observe, self, @observable_value)
         else
-          debug "observer died: #{observe}"
+          debug "dead: #{observe.describe_observer}"
 
           observers.delete(observe)
+          Celluloid.links.delete(actor)
```
SpComb: This might happen a bit too late for sync observers... if the `Observable` crashes after the sync Observe returns, but before updating again (and noticing that the `Observe` is dead), then the link will still be there, and crash the observing actor after the observe call has returned...

SpComb: Might need something to actually check the `observe.alive?` when the Observable actor shuts down, but extending the actor shutdown/cleanup logic from a mixin like `Kontena::Observable` is questionable... maybe with some hack to store the `Observe` instead of the calling actor in the links.

SpComb: Fixed by making sync observes non-persistent (again), so that they do not get linked if the observable has an immediate value, and they get unlinked immediately after the first update.

This breaks the `updates first value while waiting for second value`, but that's less important than the latent observable crashes propagating to observers that have already stopped being interested.

agent/spec/lib/kontena/observer_spec.rb
SpComb: Fixed, the `oservable.crash` [`Celluloid::Call::Sync#dispatch`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/call/sync.rb#L21) was sending the error response back to the calling rspec thread before raising within the observable actor thread, and having that send out the `ExitEvent` messages to the linked actors.

It was thus possible for the rspec thread to see the crash call response, and send the ping calls to the observer actor before the observable actor propagated the crash.

```
D, [2017-08-22T17:18:20.769611 #32054] DEBUG -- TestObservable: observer: Observer<TestObserver>...
D, [2017-08-22T17:18:20.769870 #32054] DEBUG -- TestObserver: observe async Observable<TestObservable>...
D, [2017-08-22T17:18:20.769926 #32054] DEBUG -- TestObserver: observe async Observable<!TestObservable>...
D, [2017-08-22T17:18:20.772056 #32054] DEBUG -- TestObserver: ping
D, [2017-08-22T17:18:20.772259 #32054] DEBUG -- TestObserver: ping
D, [2017-08-22T17:18:20.772418 #32054] DEBUG -- TestObserver: ping
E, [2017-08-22T17:18:20.771622 #32054] ERROR -- : Actor crashed!
RuntimeError: 
	/home/kontena/kontena/kontena/agent/spec/lib/kontena/observable_spec.rb:11:in `crash'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `public_send'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:28:in `dispatch'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:76:in `block in task'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
	/home/kontena/kontena/kontena/agent/vendor/bundle/ruby/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
E, [2017-08-22T17:18:20.772906 #32054] ERROR -- : Actor crashed!
```

SpComb:
Experimenting with a new `Kontena::Observable` API with better `observe` performance (no `Celluloid::Mailbox` overhead), less `Celluloid.current_actor` magic, a new `crash` state, and the `Kontena::Observable::Registry`.

agent/lib/kontena/logging.rb
```diff
@@ -23,47 +22,31 @@ def logger
...
-      else
-        logger.info(self.class.name) { string }
-      end
+    def info(message = nil, &block)
+      logger.add(Logger::INFO, message, self.class.name, &block)
     end
 
     # Send a warning message
-    # @param [String] string
+    # @param message [String]
```
jakolehm: Why order is swapped?

SpComb: You mean the YARD `@param` tag? That's how it's documented now: http://www.rubydoc.info/gems/yard/file/docs/Tags.md#param

>  @param name [Types] description 

Although I'm pretty sure that hasn't always been the case, the examples used to be like `@param [String] message string to print`, and I think both orders work... IMO this new (?) ordering is much better when you actually have a param description.

SpComb: I'd still keep the old ordering when changing existing code, except in this case, I'm changing all the `@param`s for everything in this class/module/file.

agent/spec/lib/kontena/observer_spec.rb
SpComb: This may not be the case anymore, now that the `observable.observe` is an atomic call protected by a mutex; once that returns, the observe task should be able to register the `Celluloid.receive` handler before any observable update messages arriving in the observing actor mailbox are processed.

Evidenced by the [`Celluloid::Future`](https://github.com/celluloid/celluloid/blob/v0.17.3/lib/celluloid/future.rb#L90) also using `Celluloid.receive` in a similar fashion.

SpComb: This is what a race here looks like with celluloid logging, although it's more apparent in the mocked out `Kontena::Observer For two observables #observe_sync accepts update for first value while requesting second value` spec:

```
D, [2017-08-29T13:35:52.708039 #16208] DEBUG -- TestObserverActor: observe sync Kontena::Observable<TestObservableActor>...
D, [2017-08-29T13:35:52.708334 #16208] DEBUG -- Kontena::Observable<TestObservableActor>: update: #[Double :test1]
D, [2017-08-29T13:35:52.708381 #16208] DEBUG -- Kontena::Observable<TestObservableActor>: notify and drop: Kontena::Observer::Observe::Sync<TestObserverActor> <- #[Double :test1]
D, [2017-08-29T13:35:52.708454 #16208] DEBUG -- : Discarded message (unhandled): #<Kontena::Observable::Message:0x000000011c9ae8>
D, [2017-08-29T13:35:52.708555 #16208] DEBUG -- TestObserverActor: observe sync Kontena::Observable<TestObservableActor> => #[Double :test2]
D, [2017-08-29T13:35:52.708598 #16208] DEBUG -- TestObserverActor: observe wait Kontena::Observable<TestObservableActor>?, Kontena::Observable<TestObservableActor>... (wait timeout=)
```

Note that somehow celluloid processes the observable message before the call to receive, so it gets discarded... that leaves the observe stuck in the timeout.

SpComb:
Massive refactoring done:

* `Kontena::Observer` class replaces the `Kontena::Observer::Observe`
* `Kontena::Observer::Helper` mixin provides a stub for `Kontena::Observer.observe` that just passes through `subject: self.class.name` for debug logging
* Unified `observe` implementation that only differs in use of `yield` vs `return`
* Async `observe` now yields exclusively from within the same task context, without spawning new tasks. Does not return anymore, but raises on Observable errors instead of crashing the actor
* No more `Thread.current[:celluloid_*]` internals (apart from the `Celluloid::Actor::System::ROOT_SERVICES` thing for `Kontena::Observable::Registry`, but just supervising it from anywhere before registering Observables)

agent/lib/kontena/workers/node_stats_worker.rb
```diff
@@ -32,13 +32,13 @@ def initialize(autostart = true)
...
     def start
-      observe(Actor[:node_info_worker]) do |node|
-        configure(node)
-      end
-
       every(PUBLISH_INTERVAL) do
         self.publish_node_stats
       end
+
+      observe(Actor[:node_info_worker].observable) do |node|
```
SpComb: Need to swap the order here, because `observe` does not return anymore.

agent/lib/kontena/network_adapters/weave.rb
```diff
@@ -33,7 +33,7 @@ def initialize(autostart = true)
     end
 
     def start
-      observe(Actor[:node_info_worker]) do |node|
+      observe(Actor[:node_info_worker].observable) do |node|
         launch(node)
```
SpComb: Breaking change with the `exclusive` yield:

```
E, [2017-08-29T15:45:59.682615 #1] ERROR -- Kontena::NetworkAdapters::Weave: RuntimeError: Cannot execute blocks on sender in exclusive mode
E, [2017-08-29T15:45:59.682726 #1] ERROR -- Kontena::NetworkAdapters::Weave: /usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:13:in `initialize'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:8:in `initialize'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/proxy/sync.rb:20:in `new'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/proxy/sync.rb:20:in `method_missing'
/app/lib/kontena/network_adapters/weave.rb:172:in `ensure_exposed'
/app/lib/kontena/network_adapters/weave.rb:268:in `post_start'
/app/lib/kontena/network_adapters/weave.rb:227:in `launch'
/app/lib/kontena/network_adapters/weave.rb:37:in `block in start'
/app/lib/kontena/observer.rb:290:in `block in observe'
/app/lib/kontena/observer.rb:225:in `block in receive'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:97:in `exclusive'
/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:421:in `exclusive'
/app/lib/kontena/observer.rb:224:in `receive'
/app/lib/kontena/observer.rb:280:in `observe'
/app/lib/kontena/observer.rb:92:in `observe'
/app/lib/kontena/observer.rb:29:in `observe'
/app/lib/kontena/network_adapters/weave.rb:36:in `start'
```

OTOH I think the answer is to fix the weave actor `launch` method to work in exclusive mode by working around the celluloid limitation on call blocks in exclusive mode.

agent/lib/kontena/workers/fluentd_worker.rb
```diff
@@ -20,7 +20,7 @@ def initialize(autostart = true)
     end
 
     def start
-      observe(Actor[:node_info_worker]) do |node|
+      observe(Actor[:node_info_worker].observable) do |node|
```
SpComb: I can fix this back to `observe(Actor[:node_info_worker])` by adding a `Kontena::Observable::Helper` mixin module that provides an `add_observer` method on the `NodeInfoWorker` actor... however, then that needs some "magic" to be reintroduced into the `add_observer` method, such that it returns an `Kontena::Observable::Message` with  real `observable: self` value used to identify future observable updates, because the observable that the `Kontena::Observer.observe` gets is not the actual observable.

Design choice: Yes / No?

SpComb:
Surprisingly for the "1000 observing actors" benchmark, the `Kontena::Observe` based method is now slower than the `wait_until!` polling:

```
wait       2.340000   0.370000   2.710000 (  3.229085)
observer   2.670000   0.130000   2.800000 (  3.723787)
                    delay
wait             0.888103
observer         0.932582
```

No idea why, it's very difficult to profile across 1k threads using `ruby-prof`... the `Celluloid.receive` API is bad, particularly for persistent async observes, but I wouldn't expect it to show up that badly in that benchmark, with a sync observe of a single observable?

agent/lib/kontena/workers/fluentd_worker.rb
SpComb: Side-effect of that `Kontena::Observable::Helper#add_observer` proxying: the `observable.add_observer` loop would no longer actually be atomic, but the use of `Celluloid.exclusive { ... }` around it should still make it safe against lost messages for `Celluloid.receive`.

Alternative: add a `Kontena::Observable#observable` method that `return self`, and then `observables = observables.map{|o| o.observable }` to do the sync calls before the `add_observer` loop... that's just even more magic?

agent/lib/kontena/observable.rb
```diff
@@ -1,75 +1,201 @@
...
+
+    # mixin for Celluloid actor classes
+    module Helper
+      # Create a new Observable using the including class name as the subject.
+      # Register the Observable with the Kontena::Observable::Registry.
+      # Links to the registry to crash the Observable if the owning actor crashes.
+      #
+      # @return [Kontena::Observable]
+      def observable
+        @observable ||= Kontena::Observable.new(self.class.name).tap do |observable|
```
SpComb: Turns out that this `@observable ||= ... tap` is racy even within a single actor, probably because of the sync `observable_registry.register` call inside the `tap` block:

```
I, [2017-08-30T11:50:54.140808 #10290]  INFO -- Kontena::Observable::Registry: initialized
D, [2017-08-30T11:50:54.243572 #10290] DEBUG -- Kontena::Observable::Registry: register Kontena::Observable<TestObservableActor>: TestObservableActor
D, [2017-08-30T11:50:54.244124 #10290] DEBUG -- Kontena::Observable::Registry: register Kontena::Observable<TestObservableActor>: TestObservableActor
D, [2017-08-30T11:50:54.244522 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: observe Kontena::Observable<TestObservableActor>...
D, [2017-08-30T11:50:54.244692 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: wait: Kontena::Observable<TestObservableActor>?
D, [2017-08-30T11:50:54.244744 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: receive timeout=...
D, [2017-08-30T11:50:54.244864 #10290] DEBUG -- Kontena::Observable<TestObservableActor>: update: 1
D, [2017-08-30T11:50:54.245628 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: observe Kontena::Observable<TestObservableActor>...
D, [2017-08-30T11:50:54.245697 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: wait: Kontena::Observable<TestObservableActor>?
D, [2017-08-30T11:50:54.245738 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: receive timeout=...
D, [2017-08-30T11:50:54.246087 #10290] DEBUG -- Kontena::Observable<TestObservableActor>: update: 2
D, [2017-08-30T11:50:54.246764 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: observe Kontena::Observable<TestObservableActor>...
D, [2017-08-30T11:50:54.246873 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: wait: Kontena::Observable<TestObservableActor>?
D, [2017-08-30T11:50:54.246918 #10290] DEBUG -- Kontena::Observer<TestObserverActor>: receive timeout=...
```

The actor ends up with two different `@observable`s, and updates one while the observers try and observe the other one:

```
Failures:

  1) Kontena::Observable registered by an Actor handles concurrent observers
     Failure/Error: expect(observer_actors.map{|actor| actor.observed_values.last}).to eq [update_count] * observer_count
     
       expected: [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
            got: [nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil]
```

agent/lib/kontena/observable.rb
```diff
@@ -1,75 +1,201 @@
...
-  # The value does not yet exist when initialized, it is nil
-  # Once the value is first updated, then other Actors will be able to observe it
-  # When the value later updated, other Actors will also observe those changes
-  module Observable
+  # The observable value is nil when initialized, and Observers will block waiting for it to become ready.
+  # Once the observable is first updated, then Observers will unblock and return/yield the updated value.
+  # When the observable is later updated, then Observers will return/yield the updated value.
+  # If the observable is crashed, then any Observers will also raise.
+  #
+  # TODO: are you allowed to reset an observable after crashing it, allowing observers to restart and re-observe?
```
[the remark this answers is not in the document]
SpComb: I think it should be possible for an observable actor to do something along the lines of "lightweight crashes":

```
observable.crash FooError.new(...)
observable.reset
# some time later...
observable.update foo
```

And have all observers raise, but allow them to re-observe the same observable again without needing to crash all the associated actors, spawn a new actor with a new observable, and have the observing actors lookup the new observable.

Not sure if that's a good idea, though.

It should work though, because the `Observer` implementation always checks for crashes immediately after each update, so it should be guaranteed to raise. Risk is somehow having the observer only check for `error?` after applying both the `crash` and `reset` updates, which would miss the rror.

agent/lib/kontena/observable.rb
```diff
@@ -1,75 +1,201 @@
...
-      notify_observers
+      set_and_notify(value)
     end
 
-    # The Observable no longer has a value
-    # This will notify any Observers, causing them to block yields until we update again
-    def reset_observable
-      @observable_value = nil
+    # Reset the observable value back into the initialized state.
+    # This will notify any Observers, causing them to block until we update again.
```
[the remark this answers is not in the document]
SpComb: Should probably avoid the use of the term `block` for this, and use something like `wait` instead. The observer debug logging used to say `observe blocked: Kontena::Observable<...>?` if it was waiting for something to get a value, but I already changed that to log `wait: ...`

If you do a sync `observe` after resetting the observable, then it will wait for the observable to get updated again.

agent/lib/kontena/workers/fluentd_worker.rb
SpComb: > ... might be a bit confusing as one cannot observe any random actor that's not an `Observable`, right?

Pedantic: no actor can be an `Observable` anymore, but the `Kontena::Observable::Helper' could act as some kind of "Observable proxy".

agent/lib/kontena/observable.rb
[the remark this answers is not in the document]
SpComb: The main alternative that I have in mind for this is some kind of higher-level layer where the observable actors to update with some kind of "unhealthy" error state, without actually crashing the observable. The observing actors would then enter some kind of error state of their own.

With that in mind, it might be a good idea to just remove the `reset` entirely. I'm not sure if it's actually useful at all.

agent/lib/kontena/observable.rb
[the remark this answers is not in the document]
jnummelin: Yeah, the word `block` made this really hard to understand :D So probably good idea to switch `block` to something else.

agent/lib/kontena/observable.rb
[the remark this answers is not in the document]
jnummelin: > The observing actors would then enter some kind of unhealthy error state of their own

Isn't that kinda opposite to one of the main actor principles of "let it crash" :) 

I think that kind of state dependencies will get one into trouble really fast. Better to keep things simple and just "let it  🔥 "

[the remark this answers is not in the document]
SpComb: Yes, actors should be able to crash, and the system should be able to recover itself into a safe state from actor crashes.

But there's also a difference between known and unknown errors. Should the `LogWorker` crash if the websocket disconnects? Can it crash without losing the contents of the `@queue`?

I'm mostly talking about the launcher actors here, that are responsible for the state of some external system. The kind of unhealthy states that I'm talking about here are along the lines of "etcd isn't working because weave was unable to connect to any of its peers", and I don't think it's appropriate for actors to start crashing in that case.

Should the observables start crashing in that case? Probably not, those kind of known errors should be handled as observable `{status: healthy/unhealthy}` states.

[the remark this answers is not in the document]
jnummelin: > But there's also a difference between known and unknown errors

Yes, absolutely. I was also more referring to the un-expected cases. 

> ... known errors should be handled as observable `{status: healthy/unhealthy}` states.

I don't think there's anything in observables that doesn't allow this already? OR are you saying that this kind of "error" update should be done through separate "interface" and not as normal observable update?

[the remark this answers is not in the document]
SpComb: > I don't think there's anything in observables that doesn't allow this already? OR are you saying that this kind of "error" update should be done through separate "interface" and not as normal observable update?

I think what I'm saying is that we should just remove the `reset` interface entirely... maybe. I can't say for certain before actually validating the design over in#2582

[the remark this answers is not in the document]
SpComb: Note that there isn't actually any code in the `Observer` to explicitly handle the reset case. The behavior of the `Observable#reset` method is just a side effect of how the `Observer#eval` `ready?` check is implemented.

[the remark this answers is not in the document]
SpComb: > I think what I'm saying is that we should just remove the reset interface entirely... maybe. I can't say for certain before actually validating the design over in #2582

It was added as an experiment #2011, nothing uses it yet, and based on how it has behaved so far in #2582, I don't expect it to be used.

agent/lib/kontena/observable.rb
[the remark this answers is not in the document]
SpComb: No more "block" terms now.

agent/lib/kontena/observer.rb
```diff
@@ -1,117 +1,306 @@
...
+
+        # must be atomic!
+        update(receive())
+      end
+    end
+
+    # No longer expecting updates.
+    #
+    def kill
+      @alive = false
```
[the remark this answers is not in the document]
SpComb: It's not the intention to be able to "kill" on observer that's receiving. This is only intended to be called from the observing task itself, and that only happens when the observe call returns (per `ensure`). The observing task can't be receiving then. Crashing an observable will wake up and kill the observers.

Maybe this is badly named. The `Kontena::Observe.observe` interface doesn't expose it, but there might be some usecase for using `Kontena::Observer.new `/ `observer.observe` / `observe.each` separately, and in that case, the caller is responsible for calling `observer.kill` to not leak the observer. If the observer stops receiving, then the observed observables will still have that observer in their observers hash, and will continue sending messages to the observing actor's mailbox that just get discarded.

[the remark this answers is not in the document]
SpComb: The `while alive?` loop in the `Observer#each` method is kinda cosmetic... there is no known case where that would ever be false. It could be changed to `while true`.

agent/lib/kontena/observer.rb
```diff
@@ -1,117 +1,335 @@
...
+    # Yield each set of ready? observed values while alive, or raise on error?
+    #
+    # The yield is exclusive, because suspending the observing task would mean that
+    # any observable messages would get discarded.
+    #
+    # @param timeout [Float] timeout between each yield
+    def each(timeout: nil)
+      @deadline = Time.now + timeout if timeout
+
+      while alive?
```
SpComb: > Meaning that if I `kill` my observer, it might still get an update as it could have been receiving?

This `while alive?` loop is kinda cosmetic... there is no known case where that would ever be false, because the observer gets killed by the `ensure` block in the `Kontena::Observer.observe` method calling this function.

This should be changed to `while true`?

agent/spec/lib/kontena/observable_spec.rb
```diff
@@ -1,166 +1,275 @@
...
+      # make sure they get the values in order
+      expect(observer_actors.map{|actor| actor.ordered?}).to eq [true] * observer_count
+
+      # some observers observed before the first update, some after
+      # this is potentially racy, but it's important, or this spec doesn't test what it should!
+      min_first = observer_actors.map{|actor| actor.observed_values.first}.min
+      max_first = observer_actors.map{|actor| actor.observed_values.first}.max
+
+      expect(min_first).to be < max_first
+      expect(max_first).to be > 1
```
SpComb: These spec failures are not actually observable bugs: they just mean that the spec conditions did not trigger the expected race conditions. It should probably retry instead of failing the spec: if there's a 1% chance of missing the race, then retrying three times should be a 0.001% chance.

```
  1) Kontena::Observable registered by an Actor handles concurrent observers

     Failure/Error: expect(min_first).to be < max_first

       expected: < 1

            got:   1

     # ./spec/lib/kontena/observable_spec.rb:226:in `block (3 levels) in <top (required)>'
```

agent/lib/kontena/observable.rb
```diff
@@ -1,75 +1,209 @@
...
+
+    # @return [Object, nil] last updated value, or nil if not observable?
+    def get
+      @value
+    end
+
+    # Observable has updated, and has not reset. It might be crashed?
+    #
+    # @return [Boolean]
+    def observable?
```
SpComb: Rename this to `ready?` - that's the term used elsewhere, and `observable?` doesn't really make any sense here anymore.

agent/lib/kontena/observer.rb
```diff
@@ -1,117 +1,346 @@
+module Celluloid
+  # XXX: allow Celluloid.exclusive { ... } to be used outside of actor context: it's a no-op
+  def exclusive(&block)
```
SpComb: Shouldn't be monkey-patching `Celluloid` here in this file, either:

* Move this to `lib/celluloid/exclusive_patch.rb`
* Wrap this in `Kontena::Observer.exclusive`

agent/lib/kontena/observer.rb
```diff
@@ -1,117 +1,346 @@
...
+    #   Yields again whenever any Observable updates.
+    #   Does not yield if any Observable resets, until ready again.
+    #   Raises if any of the observed Actors crashes.
+    #   Does not return.
+    #
+    #   Suspends the task in between yields.
+    #   Yields in exclusive mode.
+    #   Preserves Observable update ordering: each Observable update will be seen in order.
+    #
+    #   TODO: spec async yield timeout
```
SpComb: Let me spec and document this now, before I forget.

agent/spec/lib/kontena/observable_spec.rb
SpComb: This seems to occasionally fail in travis, so I think it needs the retry loop to not become a flaky spec.

agent/benchmark/observable-actors.rb
```diff
@@ -0,0 +1,71 @@
...
+end
+
+supervise_registry = Kontena::Observable::Registry.supervise :as => :observable_registry
+test_actor = nil
+test_observable = nil
+
+benchmark(
+  {
+    'wait'      => ->(id) { TestWaitActor.new(test_actor).future.watch },
+    'observer'  => ->(id) { TestObserverActor.new(test_observable).future.watch },
```
SpComb: These results are invalid... the benchmark time includes the actor setup, and the results depend heavily on the ordering:

```
wait       2.120000   0.290000   2.410000 (  3.031371)
observer   2.660000   0.150000   2.810000 (  3.697485)
                    delay
wait             0.814029
observer         0.900439
```

```
observer   1.520000   0.150000   1.670000 (  2.545030)
wait       3.310000   0.280000   3.590000 (  4.184796)
                    delay
observer         0.431958
wait             1.279973
```

```

```               user     system      total        real
wait       2.160000   0.290000   2.450000 (  3.057220)
                    delay
wait             0.821810
```
```
observer   1.510000   0.100000   1.610000 (  2.505361)
                    delay
observer         0.406092
```
```

File: agent/benchmark/observable-singleactor.rb [hunk 24]
```diff
@@ -0,0 +1,176 @@
...
+test_wait = TestWaiterActor.new(test_client)
+test_condition = TestConditionActor.new(test_client)
+test_observer = TestObserverActor.new(test_client)
+test_future = TestFutureActor.new(test_client)
+
+COUNT = getenv('COUNT', 1000) { |v| Integer(v) }
+
+benchmark(
+  'wait'      => -> {
+    map_futures(1..COUNT) {|id| sleep 0.001; test_wait.future.request(id) }
```
SpComb: Without the `sleep`, this benchmark deadlocks... I think the actor thread prioritizes timers over mailbox messages, so with a thousand tasks sleeping, the actor spends all its time running the timers, and never processes the mailbox messages..

agent/lib/kontena/observer.rb
```diff
@@ -1,117 +1,335 @@
...
+      observer = self.new(subject, Celluloid.mailbox)
+
+      persistent = true
+      persistent = false if !block_given? && observables.length == 1 # special case: sync observe of a single observable does not need updates
+
+      # must not suspend and risk discarding any messages in between observing and receiving from each()!
+      Celluloid.exclusive {
+        # this block should not make any suspending calls, but use exclusive mode to guarantee that regardless
+        observables.each do |observable|
+          observer.observe(observable)
```
SpComb: Oops... this is missing the `persistent` flag.

Issue #825, opened 2016-06-30
Service instance cannot start with too long instance name


janivihervas:
Noticed this with a long project name running `kontena app deploy`: service with longer than 48 character name fails to start, so the instance name must be 50 characters or less. Creation and deployment runs fine, but the instance cannot start. I tested this also locally with Docker for Mac (client and server versions `1.12.0-rc2` and api versions `1.24`) and was able to run a container with arbitrary long container name. 

```
$ kontena version
cli: 0.14.3
master: 0.14.3
```

This fails:

```
$ kontena service create --grid dev 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49 kontena/lb:latest
$ kontena service deploy 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49
$ kontena service show 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49
dev/0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49:
  status: running
  image: kontena/lb:latest
  stateful: no
  scaling: 1
  strategy: ha
  deploy_opts:
    min_health: 0.8
  dns: 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49.dev.kontena.local
  net: bridge
  instances:
    0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49-1:
...
      dns: 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxxx49-1.dev.kontena.local
...
      status: stopped
      reason: [8] System error: invalid argument
```

This works:

```
$ kontena service create --grid dev 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxx48 kontena/lb:latest
$ kontena service deploy 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxx48
$ kontena service show 0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxx48
dev/0xxxxxxxxx10xxxxxxxx20xxxxxxxx30xxxxxxxx40xxxx48:
  status: running
  image: kontena/lb:latest
  stateful: no
  scaling: 1
  strategy: ha
  deploy_opts:
[...]

PR #2707, opened 2017-08-18
Allow stacks to depend on other stacks

### LINKED ISSUE #2702: Stack dependencies
Stack should have top-level keyword `depends` that describes stack dependencies.

```yaml
stack: user/myapp
depends:
  redis:
    stack: kontena/redis:latest
  postgres:
    stack: kontena/stolon:0.1.0
services:
  api:
    image: user/myapp:1.0
    environment:
      - REDIS_HOST={{ redis }}
      - POSTGRES_HOST={{ postgres }}
```

kke:
Fixes #2702

This PR adds `depends:` to stack top level keywords. You can now create stacks that depend on other stacks.

The target stack can be anything that stack install takes (a local filename, an url or a registry reference).

Dependencies can be nested.

Variables can be passed from parent stacks to child stacks.

Example:

```yaml
stack: user/stackname
depends:
  redis:
    stack: kontena/redis
    variables: 
       version: 3.2-alpine
services:
  web_app:
    image: your-web-app
    environment:
      - "REDIS_HOST=${redis}"
```

When installing a stack such as above, two stacks will be installed, one called `stackname-redis` and `stackname`. 

When you `kontena stack remove` or `upgrade` a stack that has dependencies, all of them will be removed or upgraded also.

This PR also adds the option to pass variable values using command line parameters when installing or upgrading stacks, for example:

```console
$ kontena stack install -v redis.version=3.2-alpine kontena/redis
```

More details in https://github.com/kontena/kontena/pull/2707#issuecomment-326252779

### Missing:
* API doc changes
* Usage doc changes


cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -23,6 +23,13 @@ def execute
       stack = stack_read_and_dump(filename, name: name, values: values)
 
       stack['name'] = name if name
+
+      stack['dependencies'].each do |dependency|
```
kke: Some of the main ingredient is here

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -153,10 +154,20 @@ def execute(service_name = nil)
             end
           end
           result[:volumes]       = errors.count.zero? ? parse_volumes : {}
+          result[:dependencies]  = dependencies
         end
         result
       end
 
+      def dependencies
```
kke: And the rest of it here

matti:
one trillion beer/vissys deserved

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -62,9 +62,8 @@ def load_from_url(url)
       def default_envs
         @default_envs ||= {
           'GRID' => env['GRID'],
-          'STACK' => env['STACK'],
-          'PLATFORM' => env['PLATFORM']
-        }
+          'STACK' => env['STACK']
+        }.merge(env['PLATFORM'].to_s.empty? ? {} : { 'PLATFORM' => env['PLATFORM'] })
```
jakolehm: Not very happy with these oneliners.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -109,7 +108,9 @@ def raw_yaml
       def variables
         return @variables if @variables
         @variables = ::Opto::Group.new(
-          (internals_interpolated_yaml['variables'] || {}).merge(default_envs.each_with_object({}) { |env, obj| obj[env[0]] = { type: :string, value: env[1] } }),
+          (internals_interpolated_yaml['variables'] || {}).merge(
+            default_envs.each_with_object({}) { |env, obj| obj[env[0]] = { type: :string, value: env[1] } }
```
jakolehm: This is also too complex.

jnummelin:
@kke Short description how the deps work and how to express them in stack yaml would help to see "the big picture"

kke:
@jnummelin: 

### Big picture

With a stack such as : https://github.com/kontena/kontena/blob/df1fd8c2e72a73e9cfe85291293f7941fea92711/cli/spec/fixtures/stack-with-dependencies.yml

The structure is:
```
deptest1
     |
     +- deptest1child1 (as dep_1)
     |             |
     |             + deptest1child1child1 (as dep_1)
     |
     +- deptest1child2
 ```

Stack install will install stacks `deptest1`, `deptest1-dep_1`, `deptest1-dep_1-dep_1` and `deptest1-dep_2`. In `stack ls` or `show` they look like any regular stacks.

### Overview

* Variables can be passed to dependencies from parent.
* `stack rm` will warn before removing a stack that has a parent
* `stack rm` will remove all dependency stacks when removing a stack with children
* `stack upgrade` will warn before removing a stack because it was removed from the dependency chain (for example if dep_1 switches dependency) 
* `stack upgrade` will install (instead of upgrade) any new stacks that appear in the dependency chain (it compares master dependency tree with the dependency tree generated from the yaml)

### New features:

* `stack install`, upgrade, etc now accept a multivalued `-v var_name=var_value` which can be used to set variable values from the command line. Also possible to set dependency variables by `-v dependency_name.var_name=var_value`. If you use it with `--values-from`, the `-v` ones will overwrite anything in the `--values-from` file.
* Install/remove et al have options such as `--keep-dependent` and `--ignore-dependencies``
* Validate can validate the whole tree, if you don't want to, you use `--no-dependencies`.
* Validate can render just the dependency tree with `--dependency-tree`

### Notable changes:

* Master `_stack.json.jbuilder` now returns parent name and children names
* Master stack mutations now accept the parent_name as a parameter and store it with the stack if present
* There's a middleman called `StackFileLoader` which manages loading of yamls from files, urls and registry
* There's a library called `StackName` which parses something such as `user/stackname:version` and returns an object that responds to `.user`, `.stack`, `.version` 
* `YAML::Reader` now only takes one argument, the filename/uri/registry-string. Any options such as `skip_validation:` have been moved to `#execute`
* The strange voodoo in `Cli::Stacks::Common` with `reader_from_stack`, `stack_read_and_dump` et al have been removed
* The odd "second manipulation" of `reader.execute` outcome that was in the `Cli::Stacks::Common` has been removed and `reader.execute` now returns the result directly in the same format as the master accepts it

**Comment by jnummelin** [comment]
> Stack install will install stacks deptest1, deptest1-dep_1, deptest1-dep_1-dep_1 and deptest1-dep_2. In stack ls or show they look like any regular stacks.

And also internally behave like any other stack? I'm thinking mostly the `expose` and DNS'es that get generated, they are as expected?

In the fixture stack, there's:
```
depends:
  dep_1:
    stack: stack-with-dependencies-dep-1.yml
```

`dep_1` is used as variable, what's the value of that?

**Comment by kke** [comment]
> And also internally behave like any other stack? I'm thinking mostly the expose and DNS'es that get generated, they are as expected?

Yes

> dep_1 is used as variable, what's the value of that?

It has the generated name, in this case `deptest1-dep_1`. For use with something like `REDIS_HOST=redis.${redis}.${GRID}.kontena.local`

**Comment by kke** [comment]
I can't understand why specs don't run on Ruby 2.1 and 2.2.

**Comment by kke** [comment]
Looks like the culprit is Hashie, which was added as the easy way out of having `stack[:foo]['name']` kind of keying irregularities mess up everything.

I guess it needs to be removed.

Kind of interesting how it makes `bundle exec rspec` die, it shouldn't even be loaded at that point and in any case, it doesn't monkey-patch Hash. And why it only dies in 2.1 and 2.2.. Weird.

TODO: change all the symbol-usages to strings (i tried the other way around already, but had to give up, should've gone with strings.)

File: server/app/models/stack.rb [hunk 5]
```diff
@@ -14,10 +15,41 @@ class Stack
 
   index({ grid_id: 1 })
   index({ name: 1 })
+  index({ parent_name: 1 })
 
   validates_presence_of :name
   validates_uniqueness_of :name, scope: [:grid_id]
 
+  # this can't be here because child stacks are created before parents exist
```
jakolehm: Should we remove this?

server/app/views/v1/stacks/_stack.json.jbuilder
```diff
@@ -15,3 +15,5 @@ json.variables latest_rev.variables
 json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
+json.parent stack.initial? ? nil : { name: stack.parent_name }
+json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
```
jakolehm: Needs update to rest api docs.

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -2,28 +2,46 @@
 require_relative '../services/services_helper'
 require_relative 'service_generator_v2'
 require_relative '../../stacks_client'
+require 'kontena/cli/stacks/yaml/stack_file_loader'
```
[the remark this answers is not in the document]
kke: `require_relative` can cause files to be loaded multiple times if you are in a symlinked directory, especially if you mix both variations. Plain `require` is usually the best choice and should work as expected when the LOAD_PATH is set up correctly.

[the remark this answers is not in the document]
jakolehm: Only problem I see here is that we cannot "know" which version is required (it's coming from LOAD_PATH).. and it's a bit slower.

Anyway we should pick one route and use that.. it's not very good practice to mix `require_relative` and `require` ?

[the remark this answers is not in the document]
kke: (Actually `require` is faster than `require_relative` when `$LOAD_PATH` has only a few items and only marginally slower when `$LOAD_PATH` has 50+ items.)

cli/lib/kontena/cli/stacks/remove_command.rb
```diff
@@ -10,12 +10,46 @@ class RemoveCommand < Kontena::Command
...
+      confirm_command(name)
+    end
+
     def execute
-      confirm_command(name) unless forced?
+      stack = fetch_stack
+      confirm_remove(stack) unless forced?
+      stack.fetch('children', Hash.new).each do |child_stack|
+        caret"Removing dependency #{pastel.cyan(child_stack['name'])}"
+        cmd = ['stack', 'remove', '--ignore-not-found']
```
SpComb: The `{"children": [ ... ]}` array returned by the server API will only include stacks that actually exist?

Is this `--ignore-not-found` really even needed? That would imply some bug where the same stack shows up twice in the dependency tree?

cli/lib/kontena/cli/stacks/stack_name.rb
```diff
@@ -0,0 +1,69 @@
...
+    def to_s
+      version ? "#{stack_name}:#{version}" : stack_name
+    end
+    alias to_str to_s
+
+    # True when version is a prerelease
+    # @return [NilClass,TrueClass,FalseClass] nil when no version, true when prerelease, false when not.
+    def pre?
+      return nil if version.nil?
+      Gem::Version.new(version).prerelease?
```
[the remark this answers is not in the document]
kke: Hmm yes, correct, forgot we have the "semver" gem, will switch to that.

server/app/mutations/stacks/create.rb
```diff
@@ -44,7 +48,8 @@ def validate_services
     def execute
       attributes = self.inputs.clone
       grid = attributes.delete(:grid)
-      stack = Stack.create(name: self.name, grid: grid)
+      parent = attributes.delete(:parent_name)
+      stack = Stack.create(name: self.name, grid: grid, parent_name: parent)
```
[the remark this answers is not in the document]
kke: Childs are created before parents, that's why.

cli/lib/kontena/cli/stacks/remove_command.rb
kke: In case you have force removed something from between or your stack-tree failed to install completely. The `--ignore-not-found` will remove everything and not die even if a stack from the dependency tree is missing for some reason.

kke: But yes, the children array is built from stacks that actually exist so the flag might actually be of no use.

kke:
> The nested service-X.stack-depstack-subsubstack.grid.kontena.local naming will start to run into the 64-char FQDN limit (#825) even sooner

Also this makes it possible to create invalid hostnames by having illegal characters in the child name.

> Ordering of and "linking" between dependent stacks, if a service in one dependent stack needs a service in another dependent stack - can you pass the name of a one dependent stack to a different dependent stack? Or even the name of a sub-dependency?

At this point you can only pass anything from parent to children, linking to children or grand children should work ok since the parent is always installed last.

> Consider the case of a meta-stack that does not include any services, just some specific collection of some DB + app stacks?

Yes, maybe validator should allow stacks that don't have any services or volumes.

> depends re embeds/includes vs requires/depends - do we still want to consider shared stacks that are depended on by multiple stacks, or is it better to just stick to this simpler embedding model?

I think this can be extended in the future, for example something like:
```yaml
depends: 
  redis:
    provides: redis
  db:
    stack: kontena/postgres
```

> Would there be any advantage to implementing the recursive stack create/update/delete operations on the server side, instead of the CLI?

Yes, but the implementation can be changed in the future.

cli/lib/kontena/cli/stacks/list_command.rb
```diff
@@ -35,9 +48,13 @@ def fields
     end
 
     def execute
-      print_table(stacks) do |row|
+    stacks = get_stacks
+    build_depths(stacks)
+    stacks = stacks.sort_by { |s| s['name'] }
```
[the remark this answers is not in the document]
kke: Does it make any difference?

The tree drawing thing confuses myself too and it works by strange magic or pure luck.

server/app/models/stack.rb
```diff
@@ -5,6 +5,7 @@ class Stack
   NULL_STACK = 'null'.freeze
 
   field :name, type: String
+  field :parent_name, type: String
```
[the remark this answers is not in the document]
kke: Yes that would be nice but the children exist before parents so it's a bit difficult. (well creating the parent could update the children's parent_id)

cli/lib/kontena/cli/stacks/list_command.rb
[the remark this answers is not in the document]
SpComb: I reckon it works, assuming that the parent stack always comes before any child stacks.

Hence the `sort`, because the child stack names always have the parent stack name as their prefix, so they will sort lexically after the parent stack.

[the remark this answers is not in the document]
kke: It won't matter if the depths are built before or after sorting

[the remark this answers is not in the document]
kke: At least I don't see how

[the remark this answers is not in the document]
SpComb: It's the same for single-level dependencies, but broken for higher-level dependencies unless sorted? 

With sorted `top1`, `child2`, `subchild3` => depths 1, 2, 3
With reversed `subchild3`, `child2`, `top1` => depths 2, 2, 1?

SpComb:
>> The nested service-X.stack-depstack-subsubstack.grid.kontena.local naming will start to run into the 64-char FQDN limit (#825) even sooner
> Also this makes it possible to create invalid hostnames by having illegal characters in the child name.

What do you mean? The stack names are still validated by the server.

>> Ordering of and "linking" between dependent stacks, if a service in one dependent stack needs a service in another dependent stack - can you pass the name of a one dependent stack to a different dependent stack? Or even the name of a sub-dependency?
> At this point you can only pass anything from parent to children, linking to children or grand children should work ok since the parent is always installed last.

I'm not sure about linking yet, but what I mean is this:

```yaml
depends:
  foo:
    stack: test/db
  bar:
    stack: test/app
    variables:
      db_address: "db://$foo/"
```

Are stacks installed in the right order and with the right kinds of variable interpolations to allow that to work? Or do you need to assume the naming schema and write it as `${STACK}-foo`?

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -2,136 +2,135 @@
...
+    def loader_class
+      ::Kontena::Cli::Stacks::YAML::StackFileLoader
+    end
+
+    module StackNameParam
+      # Include to add a STACK_NAME parameter
       def self.included(where)
-        where.parameter "STACK_NAME", "Stack name, for example user/stackname or user/stackname:version" do |name|
-          if name.include?(':')
-            name, @stack_version = name.split(':',2 )
```
SpComb: This is still used in the `kontena stack registry pull` command, at least:

```
NameError: undefined local variable or method `stack_version' for #<Kontena::Cli::Stacks::Registry::PullCommand:0x0000000295d3d0>
Did you mean?  stacks_client
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/registry/pull_command.rb:17:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:213:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:213:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:213:in `run'
  /home/kontena/kontena/kontena/cli/vendor/bundle/ruby/2.3.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:213:in `run'
  bin/kontena:18:in `<top (required)>'
```

kke:
> What do you mean? The stack names are still validated by the server.

At least underscores work.

>      db_address: "db://$foo/"
>Are stacks installed in the right order and with the right kinds of variable interpolations to allow that to work? Or do you need to assume the naming schema and write it as ${STACK}-foo?

I don't think any variables work there. Not even `${STACK}`.

server/app/mutations/stacks/create.rb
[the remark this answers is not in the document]
SpComb: Then it should validate that the `parent` stack does not exist.. :wink: 

```
$ kontena stack install test/deptest-nodeps.yaml
 [done] Creating stack deptest      
$ kontena stack install test/deptest.yaml 
> Installing dependency  as deptest-foo ..
> Redis version? : 3.2-alpine
 [done] Creating stack deptest-foo      
 [done] Deploying service db     
 [fail] Creating stack deptest      
	name: deptest already exists
```

Although with the way that this loose stack parent-child linking works, that partially-installed `deptest-foo` stack actually ends up as a child of the originally installed `deptest` stack - the end result is very close to just `kontena stack upgrade deptest test/deptest.yaml`, just without touching the deptest stack and giving an error.

#### `deptest-nodeps.yaml`
```yaml
stack: test/deptest
services:
  redis:
    image: redis
```
#### `deptest.yaml
```yaml
stack: test/deptest
depends:
  foo:
    stack: kontena/redis
services:
  redis:
    image: redis
```
```

File: cli/lib/kontena/cli/stacks/yaml/stack_file_loader.rb [hunk 14]
```diff
@@ -0,0 +1,96 @@
...
+
+      attr_reader :source, :parent
+
+      # @param source [String] stack file source string (filename, url, ..)
+      # @param parent [StackFileLoader] define a parent for recursion
+      # @return [StackFileLoader]
+      def initialize(source, parent = nil)
+        @source = source
+        @parent = parent
+        set_context if respond_to?(:set_context)
```
SpComb: This took a little bit of hunting to figure out what's going on here... have the `FileLoader` override `initialize instead`?

```ruby
def initialize(source, parent = nil)
  super(self.class.with_context(source, parent), parent)
end
```

cli/lib/kontena/cli/stacks/yaml/stack_file_loader/file_loader.rb
```diff
@@ -0,0 +1,40 @@
...
+      def self.match?(source, parent = nil)
+        ::File.exist?(with_context(source, parent))
+      end
+
+      def self.is_file?(parent)
+        parent.is_a?(FileLoader)
+      end
+
+      def self.with_context(source, parent = nil)
+        if is_file?(parent)
```
SpComb: Wut, is this like pulling in a stack from the registry, and then that `depends: stack: ~/.ssh/id_rsa`? Should this perhaps skip the match / fail with an exception if trying to depend on a local file stack from a remote stack?

kke: This trickery is to figure out parent directory so that `foo/xyz.yml` is searched from `parent_dir/foo/xyz.yml`.

The registry commands refuse to push with local dependencies or extends.  Of course you can push stacks like that using some other client.

(Trying to parse `id_rsa` as YAML returns a string and the parser will crash because it was expecting a hash.)

cli/lib/kontena/cli/stacks/list_command.rb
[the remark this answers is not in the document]
kke: Seems so. Fixed.

SpComb:
Now with more YAML files, it would be nice if the syntax errors also included the name of the YAML file.

```
 [error] Psych::SyntaxError : (<unknown>): could not find expected ':' while scanning a simple key at line 8 column 5
```

Had to look at the traceback to figure out that it was actually line 8 in the dependent stack YAML file, not the one given to `kontena stack install`.

kke:
> Now with more YAML files, it would be nice if the syntax errors also included the name of the YAML file.

#2771 should help in that, but this PR needs one of those for itself.

cli/lib/kontena/cli/stacks/service_generator.rb
```diff
@@ -27,7 +27,7 @@ def parse_data(options)
       data = {}
       data['instances'] = options['instances']
       data['image'] = parse_image(options['image'])
-      data['env'] = options['environment'] if options['environment']
+      data['environment'] = options['environment']
```
SpComb: Stack services are missing all of their envs.... the server param is still `data['env']`

No tests for this?

kke: There is now

SpComb:
>> Are stacks installed in the right order and with the right kinds of variable interpolations to allow that to work? Or do you need to assume the naming schema and write it as ${STACK}-foo?
> 
> I don't think any variables work there. Not even ${STACK}.

Right... something like

```yaml
depends:
  foo:
    stack: deptest-child.yaml
  bar:
    stack: deptest-child.yaml
    variables:
      peers: "foo=redis://$foo:6379/"
```

results in `{"Env": ["PEERS=foo=redis://$foo:6379/"]}` with a literal `$foo`, those dependent stack name variables only work for services in the parent stack. Same thing for `$STACK`, it's just literally `foo=redis://$STACK-foo:6379/`.

server/app/mutations/stacks/create.rb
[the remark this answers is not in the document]
kke: And you can install any stack with `--parent-name foostack` to suddenly make it someones dependency. Actually I think you can do `kontena stack install --parent-name foostack --name foostack stack.yml` and crash everything.

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -16,21 +17,59 @@ class InstallCommand < Kontena::Command
...
+    end
 
-      stack['name'] = name if name
+    def create_stack
       spinner "Creating stack #{pastel.cyan(stack['name'])} " do
-        create_stack(stack)
+        stack.delete('errors')
+        stack.delete('notifications')
+        stack['services'].each do |svc|
+          svc['env'] = svc.delete('environment')
```
SpComb: Wut no, just revert the `env` name in the `ServiceGenerator`?

Also applies to the stack upgrade `PUT` etc.

kke:
The $STACK / $GRID etc should now work in depends section.

SpComb:
There's some crazy bug with `depends` `variables` leaking across sibling stacks on `kontena stack upgrade` that I'm having a hard time figuring out:

```
$ kontena stack install test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=redis://foo:6379
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```
```
$ kontena stack upgrade deptest test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```

No change to the `yaml` in between the install and upgrade. Why did the `deptest-foo/redis` `PEERS` change?

#### DEBUG `kontena stack upgrade` `GET`
```json
{"id":"development/deptest-foo","name":"deptest-foo",...,"variables":{"peers":"redis://foo:6379","PARENT_STACK":"deptest"},"services":[{"id":"development/deptest-foo/redis",...,"env":["PEERS=redis://foo:6379"],...}]...}

```

#### DEBUG `kontena stack upgrade` `PUT`
```json
{"stack":"test/deptest-child","name":"deptest-foo",...,"services":[{"instances":null,"image":"redis:latest","env":["PEERS=foo=redis://$STACK-foo:6379/"],...,"name":"redis"}],...,"variables":{"peers":"foo=redis://$STACK-foo:6379/","PARENT_STACK":"deptest"}}
``` 

#### `deptest.yaml`
```yaml
stack: test/deptest
depends:
  foo:
    stack: deptest-child.yaml
  bar:
    stack: deptest-child.yaml
    variables:
      peers: "foo=redis://$STACK-foo:6379/"
services:
  redis:
    image: redis
    environment:
      - PEERS=foo=redis://$foo:6379/
```

#### `deptest-child.yml`
```yaml
stack: test/deptest-child
variables:
  peers:
    type: string
    required: false
    default: redis://foo:6379
services:
  redis:
    image: redis
    environment:
      - PEERS=${peers}
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -148,6 +148,11 @@ def execute
 
     def update_stack(name, data)
       return true if dry_run?
+      data.delete('errors')
+      data.delete('notifications')
+      data['services'].each do |svc|
+        svc['env'] = svc.delete('environment')
```
SpComb: I don't get this D:

Why use the wrong name in the `ServiceGenerator#parse_data`, and then mangle it in each command?

There's plenty of other things that are named subtly differently between the server API `data` and the stack YAML `options`.

kke: Because 'environment' is everywhere.

cli/lib/kontena/cli/stacks/yaml/stack_file_loader.rb
```diff
@@ -0,0 +1,95 @@
...
+      # @param parent [StackFileLoader] define a parent for recursion
+      # @return [StackFileLoader]
+      def initialize(source, parent = nil)
+        @source = source
+        @parent = parent
+      end
+
+      # @return [Hash] a hash parsed from the YAML content
+      def yaml
+        @yaml ||= reader.internals_interpolated_yaml
```
SpComb: Doesn't seem to work, I suppose the `ENV['STACK']` isnt' set at this point, so `$STACK` just interpolates to an empty string?

```
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://-foo:6379/
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: I don't see anything in the actual CLI code that touches the generated service `'env'`? The `Kontena::Cli::Stacks::YAML::Reader` specs are easy to fix.

I think this should be reverted back to `env` in the `ServiceGenerator`, unless there's some other reason to change it...

SpComb:
> There's some crazy bug with depends variables leaking across sibling stacks on kontena stack upgrade that I'm having a hard time figuring out:

Right, it's because of the default opto `to`/`from` resolvers being `env`, and those env names leak across stacks. It looks like the second child stack then picks up that the previous stack's identically named variable from `ENV`, because the default value is `nil`:

```
DEBUG ENV['peers']: 
DEBUG execute deptest-bar with parent_name=deptest values={"peers"=>"foo=redis://-foo:6379/"} defaults={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}
DEBUG upgrade deptest-bar with variables={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}


DEBUG ENV['peers']: foo=redis://-foo:6379/
DEBUG execute deptest-foo with parent_name=deptest values={} defaults={"peers"=>nil, "PARENT_STACK"=>"deptest"}
DEBUG upgrade deptest-foo with variables={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}

DEBUG ENV['peers']: foo=redis://-foo:6379/
DEBUG execute deptest with parent_name= values={} defaults={"foo"=>"deptest-foo", "bar"=>"deptest-bar"}
DEBUG upgrade deptest with variables={"foo"=>"deptest-foo", "bar"=>"deptest-bar"}
```

cli/lib/kontena/cli/stacks/remove_command.rb
```diff
@@ -10,12 +10,46 @@ class RemoveCommand < Kontena::Command
 
     parameter "NAME", "Stack name"
     option "--force", :flag, "Force remove", default: false, attribute_name: :forced
+    option "--keep-dependent", :flag, "Do not remove dependencies"
```
SpComb: This `--keep-dependent` flag doesn't seem to be implemented?

server/app/mutations/stacks/create.rb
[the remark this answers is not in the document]
SpComb: Validating that the `parent_name` stack does not exist and is the same as the `name` would prevent that.

ATM it looks like the server survives that kind of recursive stack, because it doesn't actually chase the parent/children stacks. The `Stack#parent_chain` method would get stuck in an infinite loop, but it isn't actually called anywhere outside of the specs (remove it?).

The `kontena stack ls` displays the tree wrong, and then `kontena stack upgrade` gets stuck in an infinite recursion... and so does `kontena stack rm`, so that kind of stack is tricky to remove :)

server/app/models/stack.rb
```diff
@@ -14,10 +15,34 @@ class Stack
 
   index({ grid_id: 1 })
   index({ name: 1 })
+  index({ parent_name: 1 })
```
SpComb: Needs a migration to actually create the index, it just needs to call `Stack.create_indexes`.

server/app/models/stack.rb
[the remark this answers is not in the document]
SpComb: Right, there's a fundamental choice here between letting the parent stack link to child stack services, or letting the child stack services link to the parent stack. You can't have both.

The current order of creating the child stacks before the parent stacks is probably the more natural order, because that means that something like an application service in the parent stack can link to / connect to a db service in a child stack. The limitation is that you can't have an LB service in the parent stack which is used to expose other services in child stacks, because the child stack services will not be able to link to the LB in the parent stack. You can still have an LB as a dependent stack, and have other dependent stacks link to that... or you could, if inter-dependency variables worked :confounded:

It feels a little dirty to create child stacks with these dangling `parent` links... it'll show up as `null` in the API JSON, until some suitably named stack gets created, and then it will magically change to `{"name": "..."}` :confused: The child stacks will also be left orphaned if the `kontena stack install` fails halfway-through and the parent stack doesn't get created, which means that you have to clean up all the child stacks individually with `kontena stack rm` before you can re-install the parent stack... or you install some other stack with the same name.

However, the deferred `parent_name` linking is simpler all around than setting the child `parent` relations once the parent stack gets created... maybe this isn't worth fixing until it actually becomes a problem.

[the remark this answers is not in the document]
SpComb: And maybe the LB usecase is more an example of the other kind of "external" dependencies, where the required LB stack would be assumed to already exist when creating the stack that requires it, and multiple installed stacks could use the same external LB stack. In that kind of inversed scenario the relations will be much more naturally modeled as mongoid relations.

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -2,136 +2,146 @@
...
 
 module Kontena::Cli::Stacks
   module Common
     include Kontena::Cli::Services::ServicesHelper
 
-    module StackNameParam
-      attr_accessor :stack_version
+    # @return [StackFileLoader] a loader for the stack origin defined through command-line options
+    def loader
+      @loader ||= loader_class.for(source)
```
SpComb: `kontena stack build` is broken:

```
NameError: undefined local variable or method `source' for #<Kontena::Cli::Stacks::BuildCommand:0x0000000195faf0>
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:14:in `loader'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:19:in `reader'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:33:in `stack'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/build_command.rb:29:in `execute'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:213:in `run'
```

SpComb:
Summary of my review comments:

### Minor comments that MUST be fixed in this PR before merging

- [ ] Fix `kontena stack build`
- [ ] Fix the `kontena stack remove --keep-dependent`
- [ ] Add a migration for the `Stack.parent_name` index

### Minor comments that SHOULD be fixed in this PR before merging

- [ ] Validate server stack `parent_name` (does not exist, does not match `name`)
- [ ] Remove the `kontena stack remove --ignore-not-found` option
- [ ] Revert the `ServiceGenerator.parse_data` to return `{'env' => ... }` as per the server API and fix the specs accordingly

### Major comments that can be regarded as separate issues and fixed in later PRs

- [ ] Fix leaking of variable `env` defaults between stacks
- [ ] Allow using `$STACK` and `$GRID` in `depend` stack `variables
- [ ] Allow using other parent stack `variables` in the `depend` stack `variables`
- [ ] Allow using references to preceding `$depend` stack names in the `depend` stack `variables`

kke:
TODO:

- [X] Fix `kontena stack build`
- [ ] Fix the `kontena stack remove --keep-dependent`
- [ ] Add a migration for the `Stack.parent_name` index

### Minor comments that SHOULD be fixed in this PR before merging

- [ ] Validate server stack `parent_name` (does not exist, does not match `name`)
- [ ] Remove the `kontena stack remove --ignore-not-found` option
- [ ] Revert the `ServiceGenerator.parse_data` to return `{'env' => ... }` as per the server API and fix the specs accordingly

### Major comments that can be regarded as separate issues and fixed in later PRs

- [ ] Fix leaking of variable `env` defaults between stacks
- [ ] Allow using `$STACK` and `$GRID` in `depend` stack `variables
- [ ] Allow using other parent stack `variables` in the `depend` stack `variables`
- [ ] Allow using references to preceding `$depend` stack names in the `depend` stack `variables`

kke:
> Right, it's because of the default opto to/from resolvers being env, 

Perhaps the `to: env` default is worthless anyway. I don't see how you would use it since in interpolation: `${foo}` will be the value of `foo`, even if it has not been put into env.

Let's just remove it.

kke:
Ready for round 2

SpComb:
Removing the `status/design-review` label, I'm assuming everyone's okay with the design on this, and I don't have anything to change.

server/app/mutations/stacks/create.rb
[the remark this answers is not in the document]
SpComb: BTW, by `validate` I mean something in the mutation `validate` method, like

```ruby
class Stacks::Create
  def validate
    if self.parent_name && self.grid.stacks.find_by(name: self.parent_name)
      add_error(:parent_name, :exists, "Parent stack #{self.parent_name} already exists")
    end
  end
end
```

The model validation is a good place for the `name == parent_name` check because that should never ever be true, but it's probably more convenient to do it in the mutation, and the resulting error message is more user-friendly.

server/db/migrations/27_stack_parent_index.rb
```diff
@@ -0,0 +1,6 @@
+class StackParentIndex < Mongodb::Migration
```
SpComb: This needs to be `29_stack_parent_index.rb`:

```
/app/app/services/mongodb/migrator.rb:56:in `block in migrations': Multiple migrations have the version 27 (Mongodb::Migrator::DuplicateMigrationVersionError)
	from /app/app/services/mongodb/migrator.rb:49:in `each'
	from /app/app/services/mongodb/migrator.rb:49:in `migrations'
	from /app/app/services/mongodb/migrator.rb:87:in `migrate_without_lock'
	from /app/app/services/mongodb/migrator.rb:75:in `migrate'
	from config/puma.rb:11:in `block in _load_from'
```

cli/lib/kontena/cli/stacks/yaml/service_extender.rb
```diff
@@ -12,7 +12,7 @@ def initialize(service_config)
       # @param [Hash] from
       # @return [Hash]
       def extend_from(from)
-        service_config['environment'] = extend_env_vars(from['environment'], service_config['environment'])
+        service_config['environment'] = extend_env_vars(from['env'], service_config['environment'])
```
SpComb: Surprised that the `extend` mechanic uses the generated service form... does that mean that things like `cmd` vs `command`, `memory` vs `mem_limit` etc are currently broken for extended services?

Separate issue... it would already be broken in the existing code, just noticed it here.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -131,7 +131,7 @@ def default_envs_to_options
       def variables
         @variables ||= ::Opto::Group.new(
           internals_interpolated_yaml.fetch('variables', {}).merge(default_envs_to_options),
-          defaults: { from: :env, to: :env }
+          defaults: { from: :env }
```
SpComb: This is a really small change, but does it also have some other side-effects? I don't know myself... I'd rather have a separate one-liner PR for this change than having to think about it here and now before being able to merge this.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -77,92 +84,150 @@ def internals_interpolated_yaml
...
-        return @variables if @variables
-        @variables = ::Opto::Group.new(
-          (internals_interpolated_yaml['variables'] || {}).merge(default_envs.each_with_object({}) { |env, obj| obj[env[0]] = { type: :string, value: env[1] } }),
-          defaults: {
-            from: :env,
-            to: :env
-          }
+        @variables ||= ::Opto::Group.new(
+          internals_interpolated_yaml.fetch('variables', {}).merge(default_envs_to_options),
+          defaults: { from: :env }
```
SpComb: This is a really small change, but does it also have some other side-effects? I don't know myself... I'd rather have a separate one-liner PR for this change than having to think about it here and now before being able to merge this :confused:

SpComb: Confirmed to fix the issue, though:

```
$ redis | grep PEERS
    - PEERS=
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/

$ kontena stack upgrade deptest test/deptest.yaml 
...

$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=
```

kke: Previously it turned this:

```yaml
variables:
  foo:
    type: string
    from:
      prompt: Foo?
```

to this:

```yaml
variables:
  foo:
    type: string
    from:
      env:  foo
      prompt: Foo?
    to:
      env: foo
```

The `env: foo` has no effect. The interpolator does not even read env variables any more and everything has to be declared:

```ruby
                opt = variables.option(var)
                if opt.nil?
                  raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
                else
                  val = opt.value
                end
```

Someone could have this:

```yaml
variables:
  foo:
    type: string
    value: hello
  bar:
    type: string
    from:
      env:  foo
```

And it would break after this.

server/app/mutations/stacks/create.rb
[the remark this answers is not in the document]
SpComb: Okay, while the "validate not exists" would work for `kontena stack install`, it would break `kontena stack upgrade` - new child stacks can be created on upgrades, where the parent stack already exists:

```
$ kontena stack upgrade deptest test/deptest.yaml 
Value for peers is not set. Substituting with an empty string.
> Installing new dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar ..
 [fail] Creating stack deptest-bar      
	parent_name: Parent stack deptest already exists
```

cli/lib/kontena/cli/stacks/yaml/reader.rb
kke: Reverted, will pr it separately after this is merged.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -319,13 +368,12 @@ def interpolate(content, use_opto: true, substitutions: {}, raise_on_unknown: fa
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  if variables.find { |opt| opt.to[:env][var] }
```
SpComb: This seems like a regression, with a `to: env: FOO` variable, you can no longer interpolate `${FOO}`:

```
 [error] RuntimeError : Undeclared variable 'FOO' in /home/kontena/kontena/kontena/cli/test/from-variable.yaml:16 --       - FOO=${FOO}
```

PR #2713, opened 2017-08-21
Make stack logs accept service names as filters

### LINKED ISSUE #2712: CLI: stack logs by service
```
k stack logs --service a --service b
```

or

```
k stack logs <stack> <a> <b>
```

kke:
Fixes #2712

* Makes grid container logs api accept service id's and not just service names as filter
* Makes stack logs command accept service names as filter parameters, like `kontena stack logs stack_name <service_name> <service2_name>`

### DIFF
--- a/cli/lib/kontena/cli/stacks/logs_command.rb
+++ b/cli/lib/kontena/cli/stacks/logs_command.rb
@@ -9,13 +9,20 @@ module Kontena::Cli::Stacks
     banner "Shows logs from services in a stack"
 
     parameter "NAME", "Stack name"
+    parameter "[SERVICE] ...", "Service names"
 
     requires_current_master
     requires_current_master_token
 
     def execute
-      show_logs("stacks/#{current_grid}/#{name}/container_logs") do |log|
-        show_log(log)
+      if service_list.empty?
+        show_logs("stacks/#{current_grid}/#{name}/container_logs") do |log|
+          show_log(log)
+        end
+      else
+        show_logs("grids/#{current_grid}/container_logs", services: service_list.map {|s| [name, s].join('/')}.join(',')) do |log|
+          show_log(log)
+        end
       end
     end
 
--- a/server/app/routes/v1/grids_api.rb
+++ b/server/app/routes/v1/grids_api.rb
@@ -127,8 +127,8 @@ module V1
             end
             unless r['services'].nil?
               services = r['services'].split(',').map do |service|
-                @grid.grid_services.find_by(name: service).try(:id)
-              end.delete_if{|s| s.nil?}
+                service.include?('/') ? service : @grid.grid_services.find_by(name: service).try(:id)
+              end.compact
 
               scope = scope.where(grid_service_id: {:$in => services})
             end
--- a/server/app/routes/v1/grids_api.rb
+++ b/server/app/routes/v1/grids_api.rb
@@ -127,7 +127,15 @@ module V1
             end
             unless r['services'].nil?
               services = r['services'].split(',').map do |service|
-                service.include?('/') ? service : @grid.grid_services.find_by(name: service).try(:id)
+                if service.include?('/')
+                  stack_name, service_name = service.split('/', 2)
+                  stack_id = @grid.stacks.where(name: stack_name).first.try(:id)
+                  if stack_id && service_name
+                    @grid.grid_services.where(stack_id: stack_id, name: service_name).first.try(:id)
+                  end
+                else
+                  @grid.grid_services.find_by(name: service).try(:id)
+                end
               end.compact
 
               scope = scope.where(grid_service_id: {:$in => services})
--- a/server/spec/api/v1/grids_spec.rb
+++ b/server/spec/api/v1/grids_spec.rb
@@ -340,8 +340,16 @@ describe '/v1/grids', celluloid: true do
         expect(json_response['logs'].first['data']).to eq('foo-1 1')
       end
 
-      it 'returns grid container logs for multiple services' do
+      it 'returns grid container logs for multiple services by service name' do
         get "/v1/grids/#{@grid.to_path}/container_logs?services=foo,bar", nil, request_headers
+        expect(response.status).to eq(200)
+        expect(json_response['logs'].size).to eq(2)
+        expect(json_response['logs'][0]['data']).to eq('foo-1 1')
+        expect(json_response['logs'][1]['data']).to eq('bar-1 1')
+      end
+
+      it 'returns grid container logs for multiple services by service id' do
+        get "/v1/grids/#{@grid.to_path}/container_logs?services=#{@grid.to_path}/foo,#{@grid.to_path}/bar", nil, request_headers
         expect(response.status).to eq(200)
         expect(json_response['logs'].size).to eq(2)
         expect(json_response['logs'][0]['data']).to eq('foo-1 1')
--- a/server/spec/api/v1/grids_spec.rb
+++ b/server/spec/api/v1/grids_spec.rb
@@ -348,12 +348,40 @@ describe '/v1/grids', celluloid: true do
         expect(json_response['logs'][1]['data']).to eq('bar-1 1')
       end
 
-      it 'returns grid container logs for multiple services by service id' do
-        get "/v1/grids/#{@grid.to_path}/container_logs?services=#{@grid.to_path}/foo,#{@grid.to_path}/bar", nil, request_headers
-        expect(response.status).to eq(200)
-        expect(json_response['logs'].size).to eq(2)
-        expect(json_response['logs'][0]['data']).to eq('foo-1 1')
-        expect(json_response['logs'][1]['data']).to eq('bar-1 1')
+      context 'stack service' do
+        let(:stack) do
+          Stacks::Create.run(
+            current_user: david,
+            grid: @grid,
+            name: 'foostack',
+            stack: 'stack',
+            version: '0.1.1',
+            source: '...',
+            variables: { foo: 'bar' },
+            registry: 'file',
+            services: [
+              { name: 'app', image: 'my/app:latest', stateful: false },
+              { name: 'foo', image: 'my/app:latest', stateful: false },
+              { name: 'redis', image: 'redis:2.8', stateful: true }
+            ]
+          ).result
+        end
+
+        before do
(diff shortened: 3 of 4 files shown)


server/app/routes/v1/grids_api.rb
```diff
@@ -127,8 +127,8 @@ class GridsApi < Roda
             end
             unless r['services'].nil?
               services = r['services'].split(',').map do |service|
-                @grid.grid_services.find_by(name: service).try(:id)
-              end.delete_if{|s| s.nil?}
+                service.include?('/') ? service : @grid.grid_services.find_by(name: service).try(:id)
```
jakolehm: I don't get how this could work. `id` is internal id (not a string).

kke: Yes the response from `service ls` fooled me:

```json
{"services":[{"id":"test/twm/twemproxy" ..}]}
```

cli/lib/kontena/cli/stacks/logs_command.rb
```diff
@@ -9,13 +9,20 @@ class LogsCommand < Kontena::Command
     banner "Shows logs from services in a stack"
 
     parameter "NAME", "Stack name"
+    parameter "[SERVICE] ...", "Service names"
```
SpComb: I would suggest `--service` for consistency with `kontena grid logs`, and keep the additional arguments for possible multi-stack log tailing (`kontena stack logs stack-1 stack-2`)?

cli/lib/kontena/cli/stacks/logs_command.rb
```diff
@@ -9,13 +9,20 @@ class LogsCommand < Kontena::Command
...
 
     def execute
-      show_logs("stacks/#{current_grid}/#{name}/container_logs") do |log|
-        show_log(log)
+      if service_list.empty?
+        show_logs("stacks/#{current_grid}/#{name}/container_logs") do |log|
+          show_log(log)
+        end
+      else
+        show_logs("grids/#{current_grid}/container_logs", services: service_list.map {|s| [current_grid, name, s].join('/')}.join(',')) do |log|
```
SpComb: The `grids `API params probably should not need to include the actual grid name, so just `?service=stack/service`?

server/app/routes/v1/grids_api.rb
```diff
@@ -127,8 +127,16 @@ class GridsApi < Roda
             end
             unless r['services'].nil?
               services = r['services'].split(',').map do |service|
-                @grid.grid_services.find_by(name: service).try(:id)
-              end.delete_if{|s| s.nil?}
+                if service.include?('/')
+                  stack_name, service_name = service.split('/', 3).last(2)
```
SpComb: :-1: for just ignoring the `grid/` prefix, leave it out of the API instead?

server/app/routes/v1/grids_api.rb
```diff
@@ -127,8 +127,16 @@ class GridsApi < Roda
...
+                  stack_id = @grid.stacks.where(name: stack_name).first.try(:id)
+                  if stack_id && service_name
+                    @grid.grid_services.where(stack_id: stack_id, name: service_name).first.try(:id)
+                  end
```
SpComb: What does this block return in the `else` case... I assume it's supposed to be `nil`, that could be explicit... or better yet, return an actual HTTP error if you give a non-existant stack/service name?

server/spec/api/v1/grids_spec.rb
```diff
@@ -340,14 +340,22 @@
...
       end
 
+      it 'returns grid container logs for multiple services by service id' do
+        get "/v1/grids/#{@grid.to_path}/container_logs?services=#{@grid.to_path}/foo,#{@grid.to_path}/bar", nil, request_headers
+        expect(response.status).to eq(200)
+        expect(json_response['logs'].size).to eq(2)
+        expect(json_response['logs'][0]['data']).to eq('foo-1 1')
+        expect(json_response['logs'][1]['data']).to eq('bar-1 1')
+      end
+
```
SpComb: Please also spec the non-existent service case.

server/app/routes/v1/grids_api.rb
kke: It ignores unknown stuff in service names too. Followed that practice.

cli/lib/kontena/cli/stacks/logs_command.rb
kke: That's quite difficult to use in scripts and tedious to type. 

Stack monitor is consistent with this:
```
Usage:
    kontena stack monitor [OPTIONS] NAME [SERVICES] ...

  Monitor services in a stack

Parameters:
    NAME                          Stack name
    [SERVICES] ...                Stack services to monitor
```

SpComb: OK

server/app/routes/v1/grids_api.rb
SpComb: I wouldn't say it's a good practice, but okay, I'll make a separate issue for it:  #2714

server/app/routes/v1/grids_api.rb
```diff
@@ -127,8 +127,16 @@ class GridsApi < Roda
...
-                @grid.grid_services.find_by(name: service).try(:id)
-              end.delete_if{|s| s.nil?}
+                if service.include?('/')
+                  stack_name, service_name = service.split('/', 2)
+                  stack_id = @grid.stacks.where(name: stack_name).first.try(:id)
+                  if stack_id && service_name
+                    @grid.grid_services.where(stack_id: stack_id, name: service_name).first.try(:id)
+                  end
+                else
+                  @grid.grid_services.find_by(name: service).try(:id)
```
SpComb: If the service name does not include a `/`, then this should probably be fixed to only the matching stackless service, not the first service similarly named service in any stack? aka something like `@grid.stacks.find_by(name: Stack::NULL_STACK).grid_services.find_by(name: service)`

server/spec/api/v1/grids_spec.rb
```diff
@@ -340,14 +340,50 @@
...
+          foo_container = stack.grid_services.find_by(name: 'foo').containers.create!(name: 'foo-1', host_node: node, container_id: 'eee')
+          foo_container.container_logs.create!(name: "foo-1", data: 'foo-1 1', type: 'stdout', grid: @grid, host_node: node, grid_service: stack.grid_services[1])
+        end
+
+        it 'returns grid container logs for multiple services by service id' do
+          get "/v1/grids/#{@grid.to_path}/container_logs?services=foostack/app,foostack/foo,foostack/nonexist,nostack/app", nil, request_headers
+          expect(response.status).to eq(200)
+          expect(json_response['logs'].size).to eq(2)
+          expect(json_response['logs'][0]['data']).to eq('app-1 1')
+          expect(json_response['logs'][1]['data']).to eq('foo-1 1')
```
SpComb: BTW I see that this is based on the existing specs, but I suggest using a single `eq`, just to have more informative rspec failures if it fails, i.e. something like

```ruby
expect(json_response).to eq 'logs' => [
  hash_including('data' => 'app-1 1'),
  hash_including('data' => 'foo-1 1'),
]
```

server/app/routes/v1/grids_api.rb
SpComb: I think this is a bug that also needs to be fixed now that we're making the `?service=` stack-aware. The grid service `name` is not unique within a stack, and there can be multiple services with a matching name (in different stacks), so this `@grid.grid_services.find_by(name: ...)` leads to unpredictable/broken behavior:

```
$ kontena stack install test/test.yaml 
 [done] Creating stack test      
 [done] Triggering deployment of stack test     
```

With only a `test/redis` service, `--service redis` returns the logs for that stack service:

```
$ kontena grid logs --tail=1 --service redis
2017-08-22T08:43:00.000Z test.redis-1: 1:M 22 Aug 08:43:00.491 * Ready to accept connections
```

Create a new top-level `redis` service:
```
$ kontena service create redis redis
 [done] Creating redis service      
$ kontena service deploy redis
⊛ Deployed instance development/null/redis-1 to node core-01
```
```
$ kontena service ls
NAME               INSTANCES   STATEFUL   STATE     EXPOSED PORTS
⊝ redis            1 / 1       no         running
⊝ test/redis       1 / 1       no         running
```

The `grid logs` API is broken: you are unable to request the logs for the `redis` service outside of the `test` stack.

```
$ kontena grid logs --tail=1 --service redis
2017-08-22T08:43:00.000Z test.redis-1: 1:M 22 Aug 08:43:00.491 * Ready to accept connections
```

The actual result varies, and `grid logs --service redis` can also return the logs for the `redis` service, depending on the phase of the moon when the services were created.

kke: Maybe `kontena grid logs --service redis` should show logs for all redises in the grid?

SpComb: That's also an option, because at the API level, looking at how the `/v1/services` URLs work, the `redis` case should actually be `/v1/grids/:id/container_logs?service=null/redis` (with the explicit `null` stack prefix).

However, we don't want to expose those in the CLI, so `kontena grid logs --service null/redis` should be avoided, and `kontena grid logs --service redis` should IMO be normalized to `null/redis`, like it is for the `kontena service` commands...

SpComb: Separate issue:  #2718

server/spec/api/v1/grids_spec.rb
kke: I'm not sure if that works. In another branch I get this:

```
       expected: [hash_including("name"=>"child-stack-1"), hash_including("name"=>"child-stack-2")]
            got: [{"name"=>"child-stack-1"}, {"name"=>"child-stack-2"}]
```

https://stackoverflow.com/questions/23815944/rspec-match-array-of-hashes

SpComb: Oh right, it needs to be `expect(...).to match ...`, not `eq`

kke: http://rspec.info/blog/2014/01/new-in-rspec-3-composable-matchers/

```
    expect(worker.queue).to match [
      a_hash_including(:klass => "Class1", :id => 37),
      a_hash_including(:klass => "Class2", :id => 42)
    ]
```

Issue #2232, opened 2017-04-28
Agent LogWorker can run multiple `watch_queue` and `process_queue` tasks


SpComb:
This is because both `initialize` and `on_connect` call `async.start`, which both unconditionally call `async.watch_queue` and `async.process_queue`. Those tasks may quit if the actor is `!queue_processing?` which is the case in `initialize`, but the `initialize -> start` task will wait for `Actor[:etcd_launcher].running?` before it calls `async.process_queue`, at which point it seems possible for the actor to already be `queue_processing?`, so the task stays running.

Example from a `Celluloid.dump`:

```
kontena-agent | Celluloid::Actor 0x2abead8ce0d8
kontena-agent | Celluloid::Cell 0x2abead8ce59c: Kontena::Workers::LogWorker
kontena-agent | State: Running (executing tasks)
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/mailbox.rb:63:in `sleep'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/mailbox.rb:63:in `wait'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/mailbox.rb:63:in `block in check'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/timers-4.1.1/lib/timers/wait.rb:33:in `while_time_remaining'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/timers-4.1.1/lib/timers/wait.rb:11:in `for'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/mailbox.rb:58:in `check'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:155:in `block in run'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/timers-4.1.1/lib/timers/group.rb:66:in `wait'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:152:in `run'
kontena-agent | 	/usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:131:in `block in start'
[...]

PR #2724, opened 2017-08-24
[WIP] Stream service logs from docker via fluentd

jakolehm:


agent/lib/kontena/fluentd/message_processor.rb
```diff
@@ -0,0 +1,85 @@
...
+        @messages = 0
+        @running = false
+        subscribe('websocket:connected', :on_connect) # from master_info RPC
+        subscribe('websocket:disconnect', :on_disconnect)
+        async.report
+      end
+
+      def on_connect(_, _)
+        info "websocket connected"
+        async.start unless @running
```
SpComb: If this actor crashes and restarts, it will not `start` again until the websocket reconnects. TODO: websocket client observable :)

agent/lib/kontena/fluentd/message_processor.rb
```diff
@@ -0,0 +1,85 @@
...
+          data: record['log']
+        }
+        @messages += 1
+        rpc_client.async.notification('/containers/log', [msg])
+        Actor[:fluentd_worker].async.on_log_event(msg)
+      rescue NoMethodError
+        stop
+      rescue => exc
+        error "#{exc.class.name}: #{exc.message}"
+        error exc.backtrace.join("\n") if exc.backtrace
```
SpComb: Just use `error exc`

agent/lib/kontena/fluentd/message_processor.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+
+      def stop
+        info "stopped processing"
+        @running = false
```
SpComb: The `defer { process_messages }` thread will remain blocked on the `queue.pop`, and might not notice the `!@running` unless a new log record gets pushed to the queue before the agent reconnects. If the queue remains idle between the `start` and `stop`, then you might end up with multiple `process_messages` threads running: #2232

agent/lib/kontena/fluentd/message_processor.rb
```diff
@@ -0,0 +1,85 @@
...
+          stack: record['io.kontena.stack.name'],
+          instance: record['io.kontena.service.instance_number'],
+          time: Time.at(timestamp).utc.xmlschema,
+          type: record['source'],
+          data: record['log']
+        }
+        @messages += 1
+        rpc_client.async.notification('/containers/log', [msg])
+        Actor[:fluentd_worker].async.on_log_event(msg)
+      rescue NoMethodError
```
SpComb: Wut

agent/lib/kontena/fluentd/server_connection.rb
```diff
@@ -0,0 +1,34 @@
...
+      def initialize(connection, queue)
+        @peeraddr = connection.peeraddr
+        info "connection opened from #{@peeraddr[1]}:#{@peeraddr[2]}"
+        @connection = connection
+        @queue = queue
+        async.receive
+      end
+
+      def receive
+        unpacker = MessagePack::Unpacker.new(@connection)
```
SpComb: Needs to be in a separate `defer { ... }` thread or this will block the actor, preventing the server from terminating it?

agent/lib/kontena/fluentd/server.rb
```diff
@@ -0,0 +1,44 @@
...
+      include Celluloid
+      include Kontena::Logging
+
+      finalizer :finalize
+
+      def initialize(autostart = true)
+        info 'initialized'
+        @connections = []
+        @queue = SizedQueue.new(1000)
+        @message_processor = Kontena::Fluentd::MessageProcessor.new(@queue)
```
SpComb: Separate supervised actor, or this will leave the old message processors actors hanging around if this actor crashes and restarts, spawning a new actor?

agent/lib/kontena/fluentd/message_processor.rb
jakolehm: WIP code, it will catch situation where `fluentd_worker` has gone and does not come back (shutdown).

agent/lib/kontena/fluentd/server_connection.rb
jakolehm: Yes.

agent/lib/kontena/fluentd/server.rb
jakolehm: Yes.

PR #2732, opened 2017-08-24
Initial tls-sni support and new api for domain authorizations

jnummelin:
replaces #2565 

This makes it possible to build better support for certs and automated renewal on top of these.

## TODO

- [ ] API docs
- [x] moar specs

PR #2736, opened 2017-08-25
New certificate model, API, service YAML

jnummelin:
follows #2732 

Certificates are "promoted" as first-class citizens in both API and in model level also.

Certificates can be bundled to a service much like secrets:
```
stack: jussi/certs
version: 0.1.0
services:
  certs:
    image: nginx:alpine
    certificates:
      - subject: kontena.io
        name: CERTIFICATE
        type: env
```

By doing so, during the next deployment the certificate will be "injected" into container envs:
```
$ docker exec certs.certs-1 env | grep CERTIFICATE
CERTIFICATE=certificateprivate_key
```

The actual injection happens through secrets in server-agent RPC API.

One of the goals of this PR is to also enable building automatic-renewal of certs that are for domain that are authorized with tls-sni-01

## TODO

- [x] fix specs
- [x]  Finalize CLI side for highlighting certs that are to expire soon
- [ ]  API docs
- [ ]  usage docs
- [x]  change merge-to branch when #2732 merged

### DIFF
--- a/cli/lib/kontena/cli/certificate/show_command.rb
+++ b/cli/lib/kontena/cli/certificate/show_command.rb
@@ -1,0 +1,19 @@
+require_relative '../services/services_helper'
+
+module Kontena::Cli::Certificate
+  class ShowCommand < Kontena::Command
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+
+    parameter "SUBJECT", "Certificate subject"
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    def execute
+      certificate = client.get("certificates/#{current_grid}/#{self.subject}")
+      puts YAML.dump(certificate)
+    end
+  end
+end
--- a/cli/lib/kontena/cli/certificate_command.rb
+++ b/cli/lib/kontena/cli/certificate_command.rb
@@ -1,10 +1,12 @@
 
 class Kontena::Cli::CertificateCommand < Kontena::Command
 
-
+  subcommand ["list", "ls"], "List certificates", load_subcommand('certificate/list_command')
+  subcommand "show", "Show certificate details", load_subcommand('certificate/show_command')
   subcommand "register", "Register to LetsEncrypt", load_subcommand('certificate/register_command')
   subcommand "authorize", "Create DNS authorization for domain", load_subcommand('certificate/authorize_command')
   subcommand "get", "Get certificate for domain", load_subcommand('certificate/get_command')
+
 
   def execute
   end
--- a/cli/lib/kontena/cli/certificate_command.rb
+++ b/cli/lib/kontena/cli/certificate_command.rb
@@ -5,6 +5,7 @@ class Kontena::Cli::CertificateCommand < Kontena::Command
   subcommand "show", "Show certificate details", load_subcommand('certificate/show_command')
   subcommand "register", "Register to LetsEncrypt", load_subcommand('certificate/register_command')
   subcommand "authorize", "Create DNS authorization for domain", load_subcommand('certificate/authorize_command')
+  subcommand "request", "Request certificate for domain", load_subcommand('certificate/request_command')
   subcommand "get", "Get certificate for domain", load_subcommand('certificate/get_command')
 
 
--- a/cli/lib/kontena/cli/stacks/yaml/custom_validators/certificates_validator.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/custom_validators/certificates_validator.rb
@@ -1,0 +1,22 @@
+module Kontena::Cli::Stacks::YAML::Validations::CustomValidators
+  class CertificatesValidator < HashValidator::Validator::Base
+    def initialize
+      super('stacks_valid_certificates')
+    end
+
+    def validate(key, value, validations, errors)
+      unless value.is_a?(Array)
+        errors[key] = 'certificates must be array'
+        return
+      end
+      certificate_item_validation = {
+        'subject' => 'string',
+        'name' => 'string',
+        'type' => 'string'
+      }
+      value.each do |certificate|
+        HashValidator.validator_for(certificate_item_validation).validate(key, certificate, certificate_item_validation, errors)
+      end
+    end
+  end
+end
--- a/server/app/models/certificate.rb
+++ b/server/app/models/certificate.rb
@@ -1,0 +1,24 @@
+class Certificate
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :subject, type: String
+  field :valid_until, type: DateTime
+  field :alt_names, type: Array
+  field :encrypted_private_key, type: String, encrypted: true # PEM encoded
+  field :certificate, type: String # PEM encoded certificate, no need to encrypt
+
+  belongs_to :grid
+
+  validates_presence_of :subject, :valid_until
+  validates_uniqueness_of :subject, scope: [:grid_id]
+
+  index({ grid_id: 1 })
+  index({ subject: 1 })
+  index({ grid_id: 1, subject: 1 }, {unique: true})
+
+  def to_path
+    "#{self.grid.name}/#{self.subject}"
+  end
+
+end
--- a/server/app/models/certificate.rb
+++ b/server/app/models/certificate.rb
@@ -7,6 +7,8 @@ class Certificate
   field :alt_names, type: Array
   field :encrypted_private_key, type: String, encrypted: true # PEM encoded
   field :certificate, type: String # PEM encoded certificate, no need to encrypt
+  field :chain, type: String
+  field :full_chain, type: String
 
   belongs_to :grid
 
--- a/server/app/models/certificate.rb
+++ b/server/app/models/certificate.rb
@@ -7,8 +7,7 @@ class Certificate
   field :alt_names, type: Array
   field :encrypted_private_key, type: String, encrypted: true # PEM encoded
   field :certificate, type: String # PEM encoded certificate, no need to encrypt
-  field :chain, type: String
-  field :full_chain, type: String
+  field :chain, type: String # Trust chain
 
   belongs_to :grid
 
@@ -21,6 +20,14 @@ class Certificate
 
   def to_path
     "#{self.grid.name}/#{self.subject}"
+  end
+
+  def full_chain
+    self.certificate + self.chain
+  end
+
+  def bundle
+    self.full_chain + self.private_key
   end
 
 end
--- a/server/app/models/grid_service_certificate.rb
+++ b/server/app/models/grid_service_certificate.rb
@@ -1,0 +1,9 @@
+class GridServiceCertificate
+  include Mongoid::Document
+
+  field :subject, type: String # To identify the cert object
+  field :name, type: String, default: 'SSL_CERTS'
+  field :type, type: String, default: 'env' # For future use where we could expose certs as files
+
+  embedded_in :grid_service
+end
--- a/server/app/mutations/grid_certificates/request_certificate.rb
+++ b/server/app/mutations/grid_certificates/request_certificate.rb
@@ -93,16 +93,29 @@ module GridCertificates
       csr = Acme::Client::CertificateRequest.new(names: self.domains)
       certificate = le_client.new_certificate(csr)
 
-      certificate_model = Certificate.create!(
-        grid: self.grid,
-        subject: self.domains[0],
-        alt_names: self.domains[1..-1],
-        valid_until: certificate.x509.not_after,
-        private_key: certificate.request.private_key.to_pem,
-        certificate: certificate.to_pem,
-        chain: certificate.chain_to_pem,
-        full_chain: certificate.fullchain_to_pem
-      )
+      certificate_model = self.grid.certificates.find_by(subject: self.domains[0])
+
+      if certificate_model
+        certificate_model.alt_names = self.domains[1..-1]
+        certificate_model.valid_until = certificate.x509.not_after
+        certificate_model.private_key = certificate.request.private_key.to_pem
+        certificate_model.certificate = certificate.to_pem
+        certificate_model.chain = certificate.chain_to_pem
+        certificate_model.full_chain = certificate.fullchain_to_pem
+
+        certificate_model.save
+      else
+        certificate_model = Certificate.create!(
+          grid: self.grid,
+          subject: self.domains[0],
+          alt_names: self.domains[1..-1],
+          valid_until: certificate.x509.not_after,
+          private_key: certificate.request.private_key.to_pem,
+          certificate: certificate.to_pem,
+          chain: certificate.chain_to_pem,
+          full_chain: certificate.fullchain_to_pem
+        )
+      end
 
       certificate_model
 
@@ -110,7 +123,6 @@ module GridCertificates
       error exc
       add_error(:acme_client, :error, exc.message)
     end
-
 
     def le_client
       @le_client ||= acme_client(self.grid)
--- a/server/app/mutations/grid_certificates/request_certificate.rb
+++ b/server/app/mutations/grid_certificates/request_certificate.rb
@@ -94,7 +94,6 @@ module GridCertificates
         certificate_model.private_key = certificate.request.private_key.to_pem
         certificate_model.certificate = certificate.to_pem
         certificate_model.chain = certificate.chain_to_pem
-        certificate_model.full_chain = certificate.fullchain_to_pem
 
         certificate_model.save
       else
@@ -105,8 +104,7 @@ module GridCertificates
           valid_until: certificate.x509.not_after,
           private_key: certificate.request.private_key.to_pem,
           certificate: certificate.to_pem,
-          chain: certificate.chain_to_pem,
-          full_chain: certificate.fullchain_to_pem
+          chain: certificate.chain_to_pem
         )
       end
 
--- a/server/app/mutations/grid_certificates/request_certificate.rb
+++ b/server/app/mutations/grid_certificates/request_certificate.rb
@@ -15,7 +15,6 @@ module GridCertificates
       array :domains do
         string
       end
-      string :cert_type, in: ['cert', 'fullchain'], default: 'fullchain'
     end
 
     optional do
@@ -93,22 +92,16 @@ module GridCertificates
 
       csr = Acme::Client::CertificateRequest.new(names: self.domains)
       certificate = le_client.new_certificate(csr)
-      cert_priv_key = certificate.request.private_key.to_pem
-      certificate_pem = nil
-      case self.cert_type
-        when 'fullchain'
-          certificate_pem = certificate.fullchain_to_pem
-        when 'cert'
-          certificate_pem = certificate.to_pem
-      end
 
       certificate_model = Certificate.create!(
         grid: self.grid,
         subject: self.domains[0],
         alt_names: self.domains[1..-1],
         valid_until: certificate.x509.not_after,
-        private_key: cert_priv_key,
-        certificate: certificate_pem
+        private_key: certificate.request.private_key.to_pem,
+        certificate: certificate.to_pem,
+        chain: certificate.chain_to_pem,
+        full_chain: certificate.fullchain_to_pem
       )
 
       certificate_model
--- a/server/app/mutations/grid_certificates/request_certificate.rb
+++ b/server/app/mutations/grid_certificates/request_certificate.rb
@@ -1,0 +1,128 @@
+require 'timeout'
+
+require_relative 'common'
+require_relative '../../services/logging'
+
+module GridCertificates
+  class RequestCertificate < Mutations::Command
+    include Common
+    include Logging
+    include WaitHelper
+
+    required do
+      model :grid, class: Grid
+
+      array :domains do
+        string
+      end
+      string :cert_type, in: ['cert', 'fullchain'], default: 'fullchain'
+    end
+
+    optional do
+      array :linked_services do
+        string
+      end
+    end
+
+    def validate
+      self.domains.each do |domain|
+        domain_authz = get_authz_for_domain(self.grid, domain)
+
+        if domain_authz
+          if domain_authz.authorization_type == 'dns-01'
+            # Check that the expected DNS record is already in place
+            unless validate_dns_record(domain, domain_authz.challenge_opts['record_content'])
+              add_error(:dns_record, :invalid, "Expected DNS record not present for domain #{domain}")
+            end
+          end
+        else
+          add_error(:authorization, :not_found, "Domain authorization not found for domain #{domain}")
+        end
+
+      end
+    end
+
+    def verify_domain(domain)
+      domain_authorization = get_authz_for_domain(self.grid, domain)
+      challenge = le_client.challenge_from_hash(domain_authorization.challenge)
+      if domain_authorization.state == :created
+        info 'requesting verification for domain #{domain}'
+        success = challenge.request_verification
+        if success
+          domain_authorization.state = :requested
+          domain_authorization.save
+        else
+          add_error(:request_verification, :failed, "Requesting verification failed")
+        end
+      end
+
+      wait_until!("domain verification for #{domain} is valid", interval: 1, timeout: 30, threshold: 10) {
+        challenge.verify_status != 'pending'
+      }
+
+      case challenge.verify_status
+      when 'valid'
+        domain_authorization.state = :validated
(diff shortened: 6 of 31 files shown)


cli/lib/kontena/cli/certificate/show_command.rb
```diff
@@ -0,0 +1,23 @@
...
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+
+    parameter "SUBJECT", "Certificate subject"
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    def show_certificate(certificate)
```
jakolehm: This is not used?

cli/lib/kontena/cli/certificate/show_command.rb
```diff
@@ -0,0 +1,23 @@
...
+      YAML.dump(certificate)
+    end
+
+    def execute
+      certificate = client.get("certificates/#{current_grid}/#{self.subject}")
+      puts YAML.dump(certificate)
```
jakolehm: Not sure if we want to dump raw yaml here, do we?

cli/lib/kontena/cli/stacks/yaml/custom_validators/certificates_validator.rb
```diff
@@ -0,0 +1,22 @@
+module Kontena::Cli::Stacks::YAML::Validations::CustomValidators
+  class CertificatesValidator < HashValidator::Validator::Base
```
jakolehm: Needs specs.

docs/using-kontena/certificates.md
```diff
@@ -27,7 +27,23 @@ The email is needed for Let's Encrypt to notify when certificates are about to e
 
 ## Create domain authorization
 
-To be able to request certificates for a domain you must first prove that you are in charge of that domain. For this, Kontena certificate management supports DNS-based authorization.
+To be able to request certificates for a domain you must first prove that you are in charge of that domain. For this, Kontena certificate management supports DNS-based authorization as well as newer TLS-SNI based verification.
```
jakolehm: Docs should be refactored to docs repo.

server/app/mutations/domain_authorizations/authorize.rb
```diff
@@ -0,0 +1,77 @@
...
+    def validate
+      if self.authorization_type == 'tls-sni-01'
+        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type") unless self.linked_service
+      end
+      if self.linked_service
+        add_error(:linked_service, :not_found, "Linked service needs to point to existing service") unless @linked_service = resolve_service(self.grid, linked_service)
+      end
+    end
+
+    def execute
```
jakolehm: Consider refactoring this to 2 methods (one for each type).

server/app/models/grid_domain_authorization.rb
```diff
@@ -1,14 +1,23 @@
...
+
 class GridDomainAuthorization
   include Mongoid::Document
   include Mongoid::Timestamps
   include Mongoid::Enum
 
   belongs_to :grid
 
-  enum :state, [:created, :requested, :validated], default: :created
+  enum :state, [:created, :requested, :validated, :error], default: :created
```
jakolehm: enum 😰

server/app/views/v1/domain_authorizations/_domain_authorization.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.domain authorization.domain
+json.challenge authorization.challenge
+json.challenge_opts authorization.challenge_opts
+json.authorization_type authorization.authorization_type
+json.linked_service authorization.grid_service.to_path if authorization.grid_service
```
jakolehm: `linked_service` vs `service_id` ?

cli/lib/kontena/cli/certificate/show_command.rb
jnummelin: most of our `show` commands are outputting pretty yamlish output. hence I thought it might make sense to do it with `Yaml.dump` instead of multiple `puts`. 

In this case there's no need to hide or modify anything from what the API gives.

SpComb: Agreed. I think the formatting of most show commands is mostly identical, and the bits where it differs from the `YAML.dump` are a bit of a pain (like `initial node` vs `initial_member`)

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      csr = Acme::Client::CertificateRequest.new(names: self.domains)
+      certificate = acme_client.new_certificate(csr)
+      cert_priv_key = certificate.request.private_key.to_pem
+      certificate_pem = nil
+      case self.cert_type
+        when 'fullchain'
+          certificate_pem = certificate.fullchain_to_pem
+        when 'chain'
+          certificate_pem = certificate.chain_to_pem
```
SpComb: What does this do... a `Certificate` that doesn't include the actual certificate, just the chain?

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,132 @@
...
+        valid_until: certificate.x509.not_after,
+        private_key: cert_priv_key,
+        certificate: certificate_pem
+      )
+
+      certificate_model
+
+    rescue Acme::Client::Error => exc
+      error "#{exc.class.name}: #{exc.message}"
+      error exc.backtrace.join("\n") if exc.backtrace
```
SpComb: `error exc`

server/app/models/grid_service_certificate.rb
```diff
@@ -0,0 +1,9 @@
+class GridServiceCertificate
+  include Mongoid::Document
+
+  field :subject, type: String # To identify the cert object
+  field :name, type: String, default: 'SSL_CERTS'
+  field :type, type: String, default: 'env' # For future use where we could expose certs as files
```
SpComb: BTW I dislike this `type` vs `name` model - the `name` would either be an environ name or a filesystem path? Why not just have an `env` field, and then perhaps add a `filepath` field in the future?

Not worth breaking consistency over the existing secret API/model, but just IMO.

```yaml
- certificate: foo.com
  env: SSL_CERTS
```

server/app/views/v1/certificates/_certificate.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.subject certificate.subject
```
SpComb: Missing `id`

server/app/models/grid_service_certificate.rb
jnummelin: yeah, wanted to keep it fully consistent with secrets model

cli/lib/kontena/cli/certificate/show_command.rb
jnummelin: > Agreed.

Not sure with what. :) To use `YAML.dump` or NOT to use it?

cli/lib/kontena/cli/stacks/yaml/custom_validators/certificates_validator.rb
jnummelin: done

cli/lib/kontena/cli/certificate/show_command.rb
jnummelin: yep, dead code

server/app/views/v1/certificates/_certificate.json.jbuilder
jnummelin: added

server/app/mutations/grid_certificates/request_certificate.rb
jnummelin: Apparently, yes. 🙄 doesn't make much sense, copied that from the previous API functionality. I remember thinking back then that `chain` would return the actual cert plus the one above in the chain but not the full chain. No idea where I got that into my head.

I'll remove that brainfart from this new API

SpComb:
Needs update for #2732 in master.

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,128 @@
...
+      # some domain verifications has failed, errors already added
+      return if has_errors?
+
+      csr = Acme::Client::CertificateRequest.new(names: self.domains)
+      certificate = le_client.new_certificate(csr)
+      cert_priv_key = certificate.request.private_key.to_pem
+      certificate_pem = nil
+      case self.cert_type
+        when 'fullchain'
+          certificate_pem = certificate.fullchain_to_pem
```
SpComb: I don't think it's logical to determine the `fullchain` vs `cert` here when requesting/storing the cert. The `Certificate` model should just include both the cert and the chain.

Then the cert/chain can be selected when using the certificate in a service. Consider some service that wants both the cert and the chain separately?

server/app/mutations/grid_services/common.rb
```diff
@@ -91,6 +91,25 @@ def build_grid_service_secrets(existing_secrets)
       service_secrets
     end
 
+    # @return [Array<GridServiceCertificate>]
+    def build_grid_service_certificates(existing_certificates)
+      service_certificates = []
+      self.certificates.each do |certificate|
+        service_certificate = existing_certificates.find{ |c|
+          c.subject == secret['subject'] && c.name == secret['name']
```
SpComb: `certificate` vs `secret`?

Missing specs?

server/app/serializers/rpc/service_pod_serializer.rb
```diff
@@ -72,6 +72,16 @@ def build_secrets
         secrets << {name: "SSL_CERTS", type: 'env', value: domain_auth.tls_sni_certificate}
       end
 
+      # Inject certificates as secrets
+      service.certificates.each do |certificate|
+        grid_cert = grid.certificates.find_by(subject: certificate.subject)
+        item = {name: certificate.name, type: certificate.type, value: nil}
+        if grid_cert
+          item[:value] = [grid_cert.certificate, grid_cert.private_key].join
```
SpComb: The current `LE_CERTIFICATE_example.com_PRIVATE_KEY/CERTIFICATE/BUNDLE` allows using the key and cert separately, and I imagine some applications expect to be configured with the key/cert as separate files/params, and possibly even the chain as a third parameter (like [Apache < 2.4.8](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcertificatechainfile)). Is this kind of cert+key bundle specific to haproxy?

I'm thinking this should have some kind of parametrization to allow deploying the cert/key/chain separately, in order to support a variety of applications, not just `kontena/lb`. I don't know what that YAML would look like... 

```yaml
certificates:
  - subject: example.com
    cert_env: SSL_CERT
    key_env: SSL_KEY
    chain_env: SSL_CHAIN
```

server/spec/api/v1/certificates_spec.rb
```diff
@@ -0,0 +1,90 @@
+describe '/v1/containers' do
```
SpComb: `certificates`

jnummelin:
> Missing GridServices::Update support?
Yes, wanted to first get overall design level acceptance before finalising all the details

server/app/serializers/rpc/service_pod_serializer.rb
jnummelin: > Is this kind of cert+key bundle specific to haproxy?

Yes, of course there might be other systems using the same mechanism too but HAProxy has worked like this as long as I remember using it

> I'm thinking this should have some kind of parametrization to allow deploying the cert/key/chain separately

Not a bad idea at all. It just needs option in yaml for all four:
- private key
- cert
- chain
- bundle (key + cert)


IMO the yaml you posted looks pretty ok, maybe not good to have the option called `xyz_env` as we might have someday support for injecting these through files also. Maybe `cert_var` would be more future proof? 

The typical use case when using Kontena LB would then be:
```
certificates:
  - subject: example.com
    bundle_var: SSL_CERT
    type: env
```

@jakolehm Any thoughts if we'd go to this direction?

server/spec/api/v1/certificates_spec.rb
jnummelin: copy-pasta strikes again :D

server/app/serializers/rpc/service_pod_serializer.rb
SpComb: Important point is that for the separate cert+key case, you need both from the same cert, and I reckon you should be able to have both of those in the same item.

```yaml
certificates:
  - subject: example.com
    type: env
    key_name: SSL_KEY
    cert_name: SSL_CERT
```

jakolehm: I think current implementation (bundle) is ok default.. we can improve this later in a separate PR.

jnummelin:
@SpComb @jakolehm PTAL

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      certificate_model = Certificate.create!(
+        grid: self.grid,
+        subject: self.domains[0],
+        alt_names: self.domains[1..-1],
+        valid_until: certificate.x509.not_after,
+        private_key: certificate.request.private_key.to_pem,
+        certificate: certificate.to_pem,
+        chain: certificate.chain_to_pem,
+        full_chain: certificate.fullchain_to_pem
```
SpComb: [`fullchain_to_pem`](https://github.com/unixcharles/acme-client/blob/0.3.7/lib/acme/client/certificate.rb#L20) is exactly the same as `to_pem + chain_to_pem` (I think, the git repo tags don't go back as far as the ancient 0.3.4 version we're using...)

Don't necessarily need to store this separately, it's duplication.

server/app/serializers/rpc/service_pod_serializer.rb
```diff
@@ -73,6 +73,16 @@ def build_secrets
         secrets << {name: "SSL_CERTS", type: 'env', value: domain_auth.tls_sni_certificate}
       end
 
+      # Inject certificates as secrets
+      service.certificates.each do |certificate|
+        grid_cert = grid.certificates.find_by(subject: certificate.subject)
+        item = {name: certificate.name, type: certificate.type, value: nil}
+        if grid_cert
+          item[:value] = [grid_cert.full_chain, grid_cert.private_key].join
```
SpComb: I suggest some model method for this, like `grid_cert.to_bundle`?

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,121 @@
...
+        end
+
+      end
+    end
+
+    def verify_domain(domain)
+      domain_authorization = get_authz_for_domain(self.grid, domain)
+      challenge = le_client.challenge_from_hash(domain_authorization.challenge)
+      if domain_authorization.state == :created
+        info 'requesting verification for domain #{domain}'
```
SpComb: Needs `"..."` quotes

```
I, [2017-09-13T09:05:08.475867 #13]  INFO -- GridCertificates::RequestCertificate: requesting verification for domain #{domain}
```

cli/lib/kontena/cli/certificate/show_command.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    parameter "SUBJECT", "Certificate subject"
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    def execute
+      certificate = client.get("certificates/#{current_grid}/#{self.subject}")
+      puts YAML.dump(certificate)
```
SpComb: It would be nice if this also supported some option like `kontena certificate show --cert test.188.226.131.23.nip.io | openssl x509 -noout -text`, equivalent to `kontena vault read --value LE_CERTIFICATE_test_188_226_131_23_nip_io_CERTIFICATE | openssl x509 -noout -text`:
```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            03:e9:1d:72:49:86:e0:a3:7f:21:73:be:d0:0c:a7:48:a8:b2
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=US, O=Let's Encrypt, CN=Let's Encrypt Authority X3
    ...
```

May also want to be able to export the private key too?

cli/lib/kontena/cli/certificate_command.rb
```diff
@@ -1,11 +1,13 @@
 
 class Kontena::Cli::CertificateCommand < Kontena::Command
 
-
+  subcommand ["list", "ls"], "List certificates", load_subcommand('certificate/list_command')
+  subcommand "show", "Show certificate details", load_subcommand('certificate/show_command')
   subcommand "register", "Register to LetsEncrypt", load_subcommand('certificate/register_command')
   subcommand "authorize", "Create DNS authorization for domain", load_subcommand('certificate/authorize_command')
   subcommand "get", "Get certificate for domain", load_subcommand('certificate/get_command')
```
SpComb: How many :crab: would I need to sacrifice to get this renamed to `kontena certificate request`?

We're changing the CLI anyways... need to also update the docs, and something for the changelog to explain how to move from the current secrets stuff to the new `certificates`?

SpComb:
The `vault_cert_prompt` resolver needs replacing (?) to use the certificates API instead... the `lb` stacks that include support for `SSL_CERTS` `secret`s need to be rewritten to use `certificates` instead.

cli/lib/kontena/cli/certificate_command.rb
jnummelin: We could do that. And that way we could still use `kontena certificate get` for the old way of handling certs as secrets, with a deprecation warning of course. WDYT?

cli/lib/kontena/cli/certificate/show_command.rb
jnummelin: Can be added, IMO better to do in separate PR

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,121 @@
...
+    required do
+      model :grid, class: Grid
+
+      array :domains do
+        string
+      end
+    end
+
+    optional do
+      array :linked_services do
```
SpComb: Unused?

cli/lib/kontena/cli/certificate_command.rb
SpComb: I think that might be a good idea... with the domain authz changes it doesn't matter because the command still behaves the same for the user, but this removes all the `kontena certificate get` options (`--secret-name` `--cert-type` ) and also changes the way the resulting certificates are used... it might as well be a completely new command because it has very little in common with the old command.

If we're keeping the old API and mutation, then it wouldn't cost us anything to hang on to and just deprecate the old `kontena certificate get` command... and users might still have a legit need for managing certs using old secrets-based certificates, because support for those is hardcoded into the various LB stacks. It's not nice to force users to rewrite their stacks immediately across an upgrade by breaking the existing stacks, without giving them a chance to migrate gradually by being able to use both the old and new syntax?

SpComb: BTW it might also make it easier for users to upgrade if we picked up any existing `LE_CERTIFICATE_*` secrets in a migration, and created `Certificate` models for them?

jnummelin: Added `request` command and made `get` use the old API.

jnummelin:
Both old and new world now co-existing peacefully. :) Fixed also updating a cert.

server/app/models/certificate.rb
```diff
@@ -0,0 +1,26 @@
...
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :subject, type: String
+  field :valid_until, type: DateTime
+  field :alt_names, type: Array
+  field :encrypted_private_key, type: String, encrypted: true # PEM encoded
+  field :certificate, type: String # PEM encoded certificate, no need to encrypt
+  field :chain, type: String
+  field :full_chain, type: String
```
SpComb: The `full_chain` is still redundant: 

```
irb(main):006:0> cert = Certificate.first
irb(main):012:0> cert.certificate + cert.chain == cert.full_chain
=> true
```

server/app/mutations/grid_certificates/request_certificate.rb
SpComb: Still dead code.

server/app/mutations/grid_certificates/request_certificate.rb
SpComb: The log message is still broken (`#{domain}` does not get interpolated)

server/app/mutations/grid_certificates/request_certificate.rb
```diff
@@ -0,0 +1,133 @@
...
+        verify_domain(domain)
+      end
+
+      # some domain verifications has failed, errors already added
+      return if has_errors?
+
+      csr = Acme::Client::CertificateRequest.new(names: self.domains)
+      certificate = le_client.new_certificate(csr)
+
+      certificate_model = self.grid.certificates.find_by(subject: self.domains[0])
```
SpComb: :+1: would renew certificate again

jnummelin:
Did all the cleanup

server/spec/serializers/rpc/service_pod_serializer_spec.rb
```diff
@@ -175,6 +175,24 @@
...
     end
 
+    describe '[:secrets]' do
+      it 'includes certificates as secrets' do
+        Certificate.create!(grid: grid,
+          subject: 'kontena.io',
+          valid_until: Time.now + 90.days,
+          private_key: 'private_key',
+          certificate: 'certificate',
+          full_chain: 'full_chain')
```
SpComb:
```
  1) Rpc::ServicePodSerializer#to_hash [:secrets] includes certificates as secrets

     Failure/Error: Certificate.create!(grid: grid,

     Mongoid::Errors::UnknownAttribute:

       

       message:

         Attempted to set a value for 'full_chain' which is not allowed on the model Certificate.

       summary:

         Without including Mongoid::Attributes::Dynamic in your model and the attribute does not already exist in the attributes hash, attempting to call Certificate#full_chain= for it is not allowed. This is also triggered by passing the attribute to any method that accepts an attributes hash, and is raised instead of getting a NoMethodError.

       resolution:

         You can include Mongoid::Attributes::Dynamic if you expect to be writing values for undefined fields often.

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:99:in `process_attribute'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:25:in `block in process_attributes'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:23:in `each_pair'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:23:in `process_attributes'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/document.rb:111:in `block in initialize'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/threaded/lifecycle.rb:89:in `_building'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/document.rb:105:in `initialize'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:179:in `new'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:179:in `block in create!'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/threaded/lifecycle.rb:161:in `_creating'

     # ./vendor/bundle/ruby/2.4.0/gems/mongoid-5.2.1/lib/mongoid/persistable/creatable.rb:175:in `create!'

     # ./spec/serializers/rpc/service_pod_serializer_spec.rb:180:in `block (4 levels) in <top (required)>'
```

SpComb:
There might be an issue with `kontena certificate request` -> `kontena service deploy` for the LB service using that cert not actually updating, because the service doesn't notice that the cert has changed... fix later?

jnummelin:
yes, let's fix in another PR

PR #2738, opened 2017-08-28
Add metadata to fluent event hash

### LINKED ISSUE #2735: Make fluentd forwarded logs more structured by default
Currently when `fluentd` log forwarding is active, the logs contain only these fields

* `source` (`stderr`/`stdout`)
* `log` (the log message)
* `@log_name` or the _tag_ (`node.grid.stack.service.container_number`)

However, `fluentd` is best served structured logs, especially when it is combined with ElasticSearch. The current format makes it a bit cumbersome and inaccurate to filter by stack (`@log_name:*.mystack.*`).

I propose adding `node`, `grid`, `stack`, `service` and `cnt_no` as fields in `fluentd` log forwarding so that we could make concise queries `stack:mystack` etc. Some of these fields could then be dropped from `@log_name`, shortening it and reducing it to eg. `stack.service` or even just `kontena` (remember syslog inputs are often tagged `syslog` or `system`).

This can of course be accomplished with a `filter_parser` in `fluentd`, but the logs should be as useful as possible by default.

jnummelin:
fixes #2735

### DIFF
--- a/agent/spec/lib/kontena/workers/fluentd_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/fluentd_worker_spec.rb
@@ -60,7 +60,9 @@ describe Kontena::Workers::FluentdWorker do
       fluentd = instance_double(Fluent::Logger::FluentLogger)
       subject.wrapped_object.instance_variable_set('@fluentd', fluentd)
       subject.wrapped_object.instance_variable_set('@forwarding', true)
-      expect(fluentd).to receive(:post).with('web.nginx.1', {log: 'foo bar', source: 'stdout'})
+      subject.wrapped_object.instance_variable_set('@node', 'node-1')
+      subject.wrapped_object.instance_variable_set('@grid', 'terminal-a')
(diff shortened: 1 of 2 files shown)


agent/spec/lib/kontena/workers/fluentd_worker_spec.rb
```diff
@@ -60,7 +60,9 @@
       fluentd = instance_double(Fluent::Logger::FluentLogger)
       subject.wrapped_object.instance_variable_set('@fluentd', fluentd)
       subject.wrapped_object.instance_variable_set('@forwarding', true)
-      expect(fluentd).to receive(:post).with('web.nginx.1', {log: 'foo bar', source: 'stdout'})
+      subject.wrapped_object.instance_variable_set('@node', 'node-1')
```
jakolehm: Just a reminder that `instance_variable_set` is not very good practice.

jnummelin: yes, not the best way to do things. I'll see if I can make it work some other way around.

PR #2753, opened 2017-08-31
Add Agent reconnect backoff

### LINKED ISSUE #2027: Agent reconnect exponential backoff
When a Kontena master is down but the HAproxy instance is not, the agents try to reconnect once per second indefinitely. It would be nice if this was configurable or if there was a back-off algorithm in place so that a master that is down for a long time is not causing unnecessary resource utilization. For example, with 7 nodes when my master is down there is a constant ~68KiB/s in and out. With 100 nodes that would be about 7.5Mbit/s just for reconnection attempts from all of the nodes and 

I propose an exponential backoff to a random number between 45 and 75 seconds between attempts (random number chosen again after each attempt).

jnummelin:
This PR adds increasing backoff for agent reconnects.

Didn't implement it as exponential as exponential would make the backoff time quite long with only few attempts.

fixes #2027

### DIFF
--- a/agent/lib/kontena/websocket_client.rb
+++ b/agent/lib/kontena/websocket_client.rb
@@ -24,6 +24,8 @@ module Kontena
     CLOSE_TIMEOUT = 10.0
     WRITE_TIMEOUT = 10.0 # this one is a little odd
 
+
+
     # @param [String] api_uri
     # @param [String] node_id
     # @param [String] node_name
@@ -32,7 +34,7 @@ module Kontena
     # @param [Array<String>] node_labels
     # @param [Hash] ssl_params
     # @param [String] ssl_hostname
-      def initialize(api_uri, node_id, node_name:, grid_token: nil, node_token: nil, node_labels: [], ssl_params: {}, ssl_hostname: nil, autostart: true)
+    def initialize(api_uri, node_id, node_name:, grid_token: nil, node_token: nil, node_labels: [], ssl_params: {}, ssl_hostname: nil, autostart: true)
       @api_uri = api_uri
       @node_id = node_id
       @node_name = node_name
@@ -44,6 +46,7 @@ module Kontena
 
       @connected = false
       @connecting = false
+      @backoff_interval = 0
 
       if @node_token
         info "initialized with node token #{@node_token[0..8]}..., node ID #{@node_id}"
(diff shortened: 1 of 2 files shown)


agent/lib/kontena/websocket_client.rb
```diff
@@ -168,6 +179,14 @@ def connect_client(ws)
       ws.disconnect # close socket
     end
 
+    def backoff_reconnect!
+      @backoff_interval += 2 # exponential would make backoff really long with only few attempts
```
SpComb: You generally want to have some maximum cap on the backoff interval... with a lot of clients, it also helps to randomize the actual backoff.

agent/lib/kontena/websocket_client.rb
```diff
@@ -74,8 +77,16 @@ def rpc_client
...
 
     def start
-      every(CONNECT_INTERVAL) do
-        connect! if !connected? unless connecting?
+      loop do
+        if !connected? && !connecting?
+          connect!
+          # Backoff connect attempts between un-succesfull trys
+          sleep @backoff_interval
+          backoff_reconnect!
```
SpComb: This feels a bit wrong, it's hard to understand exactly how this reconnect backoff will behave.

The `connect!` -> `connected!` -> `reset_reconnect_backoff` can happen before the sleep completes and this calls `backoff_reconnect!`... It'll probably still mostly work, but I have no idea as to exactly how it will behave.

jnummelin: yep, it's really hard to make it 100% accurate as the `connected!` happens "async" from this loops point of view. Based on my local testing this seems to work nicely, but I agree that there is some randomy feel to it.

agent/lib/kontena/websocket_client.rb
jnummelin: good point, 60s max?

colinmollenhour: Yes, definitely needs to be a max. 60s would seem reasonable to me. 

Also this might not be a concern but introducing some randomness might be a good idea. E.g. if there is a large number of nodes and they are all disconnected simultaneously it might be beneficial to stagger their reconnection attempts by throwing in some randomness.

SpComb: https://www.awsarchitectureblog.com/2015/03/backoff.html

https://github.com/kontena/kontena-ipam/blob/accff08e6033436aaee8738d078b1354662e7e64/lib/helpers.rb#L24

colinmollenhour: Well there ya go.. :)  👍

SpComb:
Replaced by  #2916

PR #2754, opened 2017-09-01
Add LE TOS agreement on register command

jnummelin:
Previously there was no explicit TOS agreement asked from user.

image

### DIFF
--- a/cli/lib/kontena/cli/certificate/register_command.rb
+++ b/cli/lib/kontena/cli/certificate/register_command.rb
@@ -12,8 +12,14 @@ module Kontena::Cli::Certificate
       token = require_token
 
       data = {email: email}
-      response = client(token).post("certificates/#{current_grid}/register", data)
-      puts 'Email registered to LetsEncrypt'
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/certificate/register_command.rb
```diff
@@ -12,8 +12,14 @@ def execute
       token = require_token
 
       data = {email: email}
-      response = client(token).post("certificates/#{current_grid}/register", data)
-      puts 'Email registered to LetsEncrypt'
+      puts "By registering, you agree on Let's Encrypt Terms of Service: https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf"
+      agree_tos = prompt.yes?("Continue?")
```
SpComb: Should there also be an `--agree-tos` flag to use for scripting?

jnummelin: yes, I'll add that

PR #2755, opened 2017-09-01
Return related services in secret json

### LINKED ISSUE #2545: Return related services in a secret JSON
It would be nice to see what services are consuming the secret.

nevalla:
This PR will return related services in secret json
```
{
  "secrets":[
    {  
      "id":"big-one/foo", 
      "name":"foo", 
      "created_at":"2017-09-01T10:28:27.414Z", 
      "updated_at":"2017-09-01T10:28:27.414Z", 
      "services":[
        {
          "id":"big-one/null/app", 
          "name":"app"
        }
      ]
    }
  ]
}
```

Fixes #2545

### DIFF
--- a/server/app/views/v1/grid_secrets/index.json.jbuilder
+++ b/server/app/views/v1/grid_secrets/index.json.jbuilder
@@ -3,7 +3,10 @@ json.secrets @grid_secrets do |grid_secret|
   json.name grid_secret.name
   json.created_at grid_secret.created_at
   json.updated_at grid_secret.updated_at
-  json.services grid_secret.services do |grid_service|
+  services_with_secret = @grid_services.find_all do |s|
+    s.secrets.any?{|secret| secret.secret == grid_secret.name}
+  end
+  json.services services_with_secret.each do |grid_service|
     json.id grid_service.to_path
     json.name grid_service.name
   end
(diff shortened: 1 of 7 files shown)


server/app/views/v1/grid_secrets/index.json.jbuilder
```diff
@@ -3,4 +3,8 @@ json.secrets @grid_secrets do |grid_secret|
   json.name grid_secret.name
   json.created_at grid_secret.created_at
   json.updated_at grid_secret.updated_at
+  json.services grid_secret.services do |grid_service|
```
SpComb: This smells like it will kill `kontena secret list` performance again (#2234), with a large number of secrets, because it's O(N) `grid_services` queries? Also hard to fix with `includes`... do these have to be included in the `index` view, or is it enough to have them in in the show view?

nevalla: Good point. Related services would be nice to see without revealing the secret value. Any workarounds?

SpComb: The docs on mongoid `includes` say:

>  Note: This will work for embedded relations that reference another collection via belongs_to as well.

But the `GridServiceSecret` is just `field :name, type: String`... if that embedded document was changed to a `belongs_to :grid_secret` relation on the `has_many :grid_service_secrets`, then would `@grid.grid_secrets.includes(:grid_service_secrets)` work?

PR #2758, opened 2017-09-01
Add ability do not wait deployment to complete. Just trigger master

### LINKED ISSUE #2757: Support async stack deployments

dzubchik:
This fixes #2757

### DIFF
--- a/cli/lib/kontena/cli/stacks/deploy_command.rb
+++ b/cli/lib/kontena/cli/stacks/deploy_command.rb
@@ -10,7 +10,7 @@ module Kontena::Cli::Stacks
 
     parameter "NAME", "Stack name"
 
-    option '--async', :flag, 'Do not wait service deployment', default: false
+    option '--[no-]wait', :flag, 'Do not wait service deployment', default: true
 
     requires_current_master
     requires_current_master_token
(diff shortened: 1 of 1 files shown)


cli/lib/kontena/cli/stacks/deploy_command.rb
```diff
@@ -10,6 +10,8 @@ class DeployCommand < Kontena::Command
 
     parameter "NAME", "Stack name"
 
+    option '--async', :flag, 'Do not wait service deployment', default: false
```
jnummelin: This is human-facing "interface", maybe `--no-wait` would make more sense?

cli/lib/kontena/cli/stacks/deploy_command.rb
```diff
@@ -21,7 +23,9 @@ def execute
       spinner "Waiting for deployment to start" do
         wait_for_deployment_to_start(deployment)
```
jnummelin: I don't see why we would want to wait for the deployment to start if we're not gonna wait for it to finish either. So I'd wrap this also within  `if !async`

cli/lib/kontena/cli/stacks/deploy_command.rb
```diff
@@ -21,7 +23,9 @@ def execute
...
       end
-      wait_for_deploy_to_finish(deployment)
+      if !async?
```
jnummelin: I think we're using more `unless self.async?` style for these kind of checks, reads a bit better :)

cli/lib/kontena/cli/stacks/deploy_command.rb
```diff
@@ -18,10 +20,12 @@ def execute
       spinner "Triggering deployment of stack #{pastel.cyan(name)}" do
         deployment = deploy_stack(name)
       end
-      spinner "Waiting for deployment to start" do
-        wait_for_deployment_to_start(deployment)
+      unless :'no-wait'?
```
SpComb: `no_wait?`

dzubchik:
@jnummelin , updated PR according to comments. I've done --no-wait by analogy from upgrade command. Is it ok?

PR #2761, opened 2017-09-04
Don't fetch too many log items in stream loop

jakolehm:
Fetching max 10k items in a tight stream loop will exhaust server too easily.


SpComb:
Specs need to be fixed to mock the `LOGS_STREAM_CHUNK` instead of `LOGS_LIMIT_MAX`

PR #2771, opened 2017-09-06
CLI: Enable anchor/aliases support in YAML parsing

### LINKED ISSUE #2769: Enable anchors in CLI YAML parsing
`YAML.safe_load` does not enable anchors by default

```
safe_load(yaml, whitelist_classes = [], whitelist_symbols = [], aliases = false, filename = nil)
```

kke:
Fixes #2769 by calling `safe_load` with `content, [], [], true` to set the "aliases" option to true.

This is actually what the allowed stack top level keyword `data` was added for.

This enables something like:

```yaml
stack: foofoo/foo
variables:
  stack_affinity:
    type: string
    required: false
    from: prompt
data:
  with_affinity: &with_affinity
    affinity: $stack_affinity
services:
  foo:
    <<: *with_affinity
    image: foofoo
  bar:
    <<: *with_affinity
    image: barbar
    environment:
      - foofoo=foo
```

Also adds a filename to Psych error message output when applicable

### DIFF
--- a/cli/lib/kontena/cli/master/config/import_command.rb
+++ b/cli/lib/kontena/cli/master/config/import_command.rb
@@ -40,7 +40,7 @@ module Kontena::Cli::Master::Config
         require 'json'
         JSON.parse(data)
       when 'yaml', 'yml'
-        YAML.safe_load(data)
+        ::YAML.safe_load(data, [], [], true)
       else
         exit_with_error "Unknown input format '#{self.format}'"
       end
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -95,7 +95,7 @@ module Kontena::Cli::Stacks
         where.prepend InstanceMethods
 
         where.option '--values-from', '[FILE]', 'Read variable values from YAML' do |filename|
-          values_from_file.merge!(::YAML.safe_load(File.read(filename)))
+          values_from_file.merge!(::YAML.safe_load(File.read(filename)), [], [], true, filename)
           true
         end
 
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -94,14 +94,15 @@ module Kontena::Cli::Stacks
       def self.included(where)
         where.prepend InstanceMethods
 
-        where.option '--values-from', '[FILE]', 'Read variable values from YAML' do |filename|
-          values_from_file.merge!(::YAML.safe_load(File.read(filename)), [], [], true, filename)
-          true
+        where.option '--values-from', '[FILE]', 'Read variable values from YAML', multivalued: true do |filename|
+          values_from_file.merge!(::YAML.safe_load(File.read(filename), [], [], true, filename))
+          filename
         end
 
         where.option '-v', "VARIABLE=VALUE", "Set stack variable values, example: -v domain=example.com. Can be used multiple times.", multivalued: true, attribute_name: :var_option do |var_pair|
           var_name, var_value = var_pair.split('=', 2)
           values_from_value_options.merge!(::YAML.safe_load(::YAML.dump(var_name => var_value)))
+          var_pair
         end
       end
 
--- a/cli/lib/kontena/cli/stacks/registry/show_command.rb
+++ b/cli/lib/kontena/cli/stacks/registry/show_command.rb
@@ -15,7 +15,7 @@ module Kontena::Cli::Stacks::Registry
     def execute
       require 'semantic'
       unless versions?
-        stack = ::YAML.safe_load(stacks_client.show(stack_name.stack_name, stack_name.version))
+        stack = ::YAML.safe_load(stacks_client.show(stack_name.stack_name, stack_name.version), [], [], true)
         puts "#{stack['stack']}:"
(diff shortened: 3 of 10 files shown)


SpComb:
Did this already happen in #2707?

kke:
Not for all parts

cli/lib/kontena/cli/master/config/import_command.rb
```diff
@@ -40,7 +40,7 @@ def convert(data)
         require 'json'
         JSON.parse(data)
       when 'yaml', 'yml'
-        YAML.safe_load(data)
+        ::YAML.safe_load(data, [], [], true)
```
kke: no coverage

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -95,7 +95,7 @@ def self.included(where)
         where.prepend InstanceMethods
 
         where.option '--values-from', '[FILE]', 'Read variable values from a YAML file', multivalued: true do |filename|
-          values_from_file.merge!(::YAML.safe_load(File.read(filename)))
+          values_from_file.merge!(::YAML.safe_load(File.read(filename), [], [], true, filename))
```
kke: good coverage

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -110,6 +110,7 @@ def self.included(where)
         where.option '-v', "VARIABLE=VALUE", "Set stack variable values, example: -v domain=example.com. Can be used multiple times.", multivalued: true, attribute_name: :var_option do |var_pair|
           var_name, var_value = var_pair.split('=', 2)
           values_from_value_options.merge!(::YAML.safe_load(::YAML.dump(var_name => var_value)))
+          var_pair
```
kke: just to return the correct thing so that `var_option_list` is an array of `-v` values.

cli/lib/kontena/cli/stacks/registry/show_command.rb
```diff
@@ -15,7 +15,7 @@ class ShowCommand < Kontena::Command
     def execute
       require 'semantic'
       unless versions?
-        stack = ::YAML.safe_load(stacks_client.show(stack_name.stack_name, stack_name.version))
+        stack = ::YAML.safe_load(stacks_client.show(stack_name.stack_name, stack_name.version), [], [], true)
```
kke: no coverage

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -83,7 +83,7 @@ def internals_interpolated_yaml
               substitutions: default_envs,
               warnings: false
             )
-          )
+          ), [], [], true, file
```
kke: good coverage

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -104,7 +104,7 @@ def fully_interpolated_yaml
               use_opto: true,
               raise_on_unknown: true
             )
-          )
+          ), [], [], true, file
```
kke: good coverage

cli/lib/kontena/cli/vault/import_command.rb
```diff
@@ -16,7 +16,7 @@ class ImportCommand < Kontena::Command
 
     def parsed_input
       require "json"
-      json? ? JSON.load(input) : YAML.safe_load(input)
+      json? ? JSON.load(input) : YAML.safe_load(input, [], [], true, path)
```
kke: good coverage for YAML, no coverage for JSON

cli/lib/kontena/scripts/completer.rb
```diff
@@ -90,7 +90,7 @@ def containers
   def yml_services
     require 'yaml'
     if File.exist?('kontena.yml')
-      yaml = YAML.safe_load(File.read('kontena.yml'))
+      yaml = YAML.safe_load(File.read('kontena.yml'), [], [], true, 'kontena.yml')
```
kke: perhaps not very necessary, not very well covered, but i see no way to fail either.

cli/lib/kontena/stacks_cache.rb
```diff
@@ -20,7 +20,7 @@ def read
       end
 
       def load
-        YAML.safe_load(read)
+        ::YAML.safe_load(read, [], [], true, path)
```
kke: no coverage

cli/lib/kontena/cli/master/config/import_command.rb
kke: but should not fail either.

kke:
> NameError: undefined local variable or method 'path'

I think a candidate for this would the stacks cache, which is used when something is pulled from the registry. 

And the path in that case comes from a `path` method which is also called before the `YAML.safe_load`.

There are no specs for any of the registry commands 🥇

matti:
please get this merged, already patched in our growing list of monkey patches asdfasdf

cli/lib/kontena/stacks_cache.rb
SpComb: Is this `load` method even called somewhere? All the users of `Kontena::StacksCache` that I can find use `pull`, and that's a separate code path with it's own `::YAML.safe_load`. I suspect this is actually dead code?

PR #2785, opened 2017-09-07
Fix server MongoPubsub to rescue block errors, and not use async_thread

SpComb:
* Errors raised from `MongoPubsub.subscribe(...) do ...` blocks get logged, and the subscription continues processing messages, rather dying and having messages accumulate in the queue
* Move all `@subscriptions` handling into the `MongoPubsub` actor, don't access the `subscribers` array from outside of the actor thread
* Use a threadsafe `Queue` instead of an `Array`, and `close` it to `stop`

### DIFF
--- a/server/app/services/mongo_pubsub.rb
+++ b/server/app/services/mongo_pubsub.rb
@@ -160,7 +160,7 @@ class MongoPubsub
           channel = item['channel']
           data = item['data']
 
-          actor.queue_message(channel, data)
+          actor.async.queue_message(channel, data)
         end
       rescue => exc
         error "error while tailing: #{exc.message}"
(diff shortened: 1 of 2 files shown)


server/app/services/mongo_pubsub.rb
```diff
@@ -158,14 +159,8 @@ def tail!
...
-          
-          subscribers = self.subscriptions.select{|s| s.channel == channel }
-          subscribers.each do |subscription|
-            begin
-              subscription.queue_message(data.dup) if subscription
-            rescue Celluloid::DeadActorError
-            end
-          end
+
+          actor.async.queue_message(channel, data)
```
SpComb: Still sneaking this in, because I forgot to `git push` it yesterday.

Thinking this is better for perf, and it shouldn't violate ordering.

PR #2780, opened 2017-09-07
Improve agent ServicePodWorker container restart handling

SpComb:
Fixes #2781 to update the `@container.started_at` timestamps on restarts, causing the the `die` events from the restart to get ignored
Fixes #2782 to not block the RPC server actor on the `ServicePods::Restarter`, but just notify the `ServicePodWorker` actor instead
Fixes #3260 by having the `ServicePodWorker` process both the `ServicePods::Restarter` and `ServicePods::Terminator` actions exclusively, preventing any races
Unblocks  #2786

Rationale: this PR is important to get all of the container `ServicePods::*` lifecycle actions under the direct control of the `ServicePodWorker`. That will allow us to replace the container event based mechanisms for things like DNS registration and weave attach with actions triggered directly by the `ServicePodWorker` from the exclusive `apply`. Currently we can't do that yet, because things like `kontena service restart` and container healthcheck restarts happen outside the control of the `ServicePodWorker`... it might be able to recover from those via the `handle_restart_on_die`, but... let's not make that any more confusing?

### Server
* Change the server `GridServices::Restart` mutation to notify `/service_pods/restart`, instead of blocking on the RPC request
* Fix the server `GridServices::Restart` to not use `async_thread`

### Agent
* Change the agent  `/service_pods/restart` RPC handler to notify `service_pod:restart`  instead of blocking the `RpcServer` thread on the `Kontena::ServicePods::Restarter#perform` -> `Docker::Container#restart` call
* Change the agent `ContainerHealthCheckWorker` to also notify  `service_pod:restart`  instead of performing `Kontena::ServicePods::Restarter` directly
* The `ServicePodManager` handles the notifications and calls `ServicePodWorker#restart`
* The `ServicePodWorker` sets a `@restarting_at` timestamp, and performs the `Kontena::ServicePods::Restarter` via the exclusive `apply` mechanism if the container was  `started_at` before the `@restarting_at`

PR #2786, opened 2017-09-07
Remove server AsyncHelper#async_thread

SpComb:
- [x] Requires #2778
- [x] Requires #2780
- [x] Requires #2783
- [x] Requires #2784
- [x] Requires #2785


SpComb:
This should be ready to go once the specs pass, no more `async_thread` usages in the server.

PR #2787, opened 2017-09-07
[WIP] Separate infra container

jakolehm:


agent/lib/kontena/models/service_pod.rb
```diff
@@ -249,6 +248,34 @@ def data_volume_config
         }
       end
 
+      def infra_config
+        config = {
+          'Image' => 'kubernetes/pause:latest',
```
SpComb: :wink:

agent/lib/kontena/service_pods/infra_manager.rb
```diff
@@ -0,0 +1,76 @@
...
+      def start_infra(infra_container)
+        infra_container.start!
+        attach_network(infra_container) if infra_container.overlay_network && infra_container.overlay_cidr
+      end
+
+      def terminate
+        infra_container = get_container(service_pod.service_id, service_pod.instance_number, CONTAINER_TYPE)
+        if infra_container
+          detach_network(infra_container) if infra_container.overlay_network && infra_container.overlay_cidr
+          remove_container(infra_container)
```
SpComb: Safer to first `remove_container` and then `detach_network`. Avoid ending up with the IPAM address released but the container still alive. The other way around is handled by the routine IPAM cleanup.

agent/lib/kontena/service_pods/infra_manager.rb
```diff
@@ -0,0 +1,76 @@
+module Kontena
+  module ServicePods
+    class InfraManager
```
SpComb: The `Manager` somehow made me assume this was an Actor.

agent/lib/kontena/models/service_pod.rb
```diff
@@ -206,8 +202,11 @@ def service_host_config
           host_config['PortBindings'] = self.build_port_bindings
         end
 
-        host_config['NetworkMode'] = self.net
-        host_config['DnsSearch'] = [self.domainname, self.domainname.split('.', 2)[1]]
```
SpComb: Doesn't the service container still need the `DnsSearch`?

agent/lib/kontena/network_adapters/dns_manager.rb
```diff
@@ -2,177 +2,99 @@
...
-
-    OVERLAY_SUFFIX = '16'
-
-    # @param [String] container_id
-    # @param [String] overlay_cidr
-    def attach_overlay(container)
-      if container.overlay_network.nil?
-        # overlay network migration for 0.16 compat
-        # override overlay network /19 -> /16 suffix for existing containers that may need to be migrated
-        overlay_cidr = "#{container.overlay_ip}/#{OVERLAY_SUFFIX}"
```
SpComb: Time to drop the 0.16 compatibility?

agent/lib/kontena/models/service_pod.rb
jakolehm: Didn't want to write own yet ...

agent/lib/kontena/service_pods/infra_manager.rb
jakolehm: Yeah, probably bad name.

agent/lib/kontena/network_adapters/dns_manager.rb
jakolehm: We need to migrate containers anyway so no need to keep legacy.

SpComb: :+1: perfectly okay with this, being able to migrate containers across upgrades would spare us a lot of backwards-compatibility pain. But it will need something to coordinate the container upgrades to not be too disruptive...

agent/lib/kontena/models/service_pod.rb
jakolehm: I thought it will get this from infra container.. need to check this.

PR #2791, opened 2017-09-08
Migrate restart-policy containers

### LINKED ISSUE #2790: Remove restart policy from old containers
It would be safer to remove restart policy from old containers (although they seem to work). This can be done via docker update api.

jakolehm:
Fixes #2790

### DIFF
--- a/agent/lib/kontena/service_pods/migrator.rb
+++ b/agent/lib/kontena/service_pods/migrator.rb
@@ -1,0 +1,23 @@
+module Kontena
+  module ServicePods
+    class Migrator
+
+      # @param service_container [Docker::Container]
+      # @return [Docker::Container]
+      def self.migrate_container(service_container)
+        service_container.update({
+          'RestartPolicy' => {}
+        })
+
+        service_container
+      end
+
+      # @param service_container [Docker::Container]
+      def self.legacy_container?(service_container)
+        return true if service_container.host_config != {}
+
+        false
+      end
+    end
+  end
+end
--- a/agent/lib/kontena/service_pods/migrator.rb
+++ b/agent/lib/kontena/service_pods/migrator.rb
@@ -1,20 +1,31 @@
 module Kontena
   module ServicePods
     class Migrator
+      include Kontena::Logging
 
       # @param service_container [Docker::Container]
-      # @return [Docker::Container]
-      def self.migrate_container(service_container)
-        service_container.update({
-          'RestartPolicy' => {}
-        })
+      def initialize(service_container)
+        @service_container = service_container
+      end
 
-        service_container
+      # @return [Docker::Container]
+      def migrate
+        remove_restart_policy if @service_container.autostart?
+      end
+
+      def remove_restart_policy
+        info "removing restart policy from container #{@service_container.name_for_humans}"
+        @service_container.update({
+          'RestartPolicy' => {
+            'Name' => 'no',
+            'MaximumRetryCount' => 0
+          }
+        })
       end
 
       # @param service_container [Docker::Container]
       def self.legacy_container?(service_container)
-        return true if service_container.host_config != {}
+        return true if service_container.autostart?
 
         false
       end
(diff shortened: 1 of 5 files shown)


agent/lib/kontena/service_pods/migrator.rb
```diff
@@ -0,0 +1,23 @@
+module Kontena
+  module ServicePods
+    class Migrator
+
+      # @param service_container [Docker::Container]
+      # @return [Docker::Container]
+      def self.migrate_container(service_container)
+        service_container.update({
+          'RestartPolicy' => {}
```
SpComb: It would be nicer to have some kind of explicit `migrate_restart_policy` method with a comment saying that version 1.4 etc - these are things we're going to need to keep track of in the future as well.

Should also get logged.

agent/lib/kontena/service_pods/migrator.rb
```diff
@@ -0,0 +1,23 @@
...
+        })
+
+        service_container
+      end
+
+      # @param service_container [Docker::Container]
+      def self.legacy_container?(service_container)
+        return true if service_container.host_config != {}
```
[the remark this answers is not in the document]
SpComb: No, the `service_container.host_config` will never be empty. Fresh containers deployed by the agent will have something like this: `{ "RestartPolicy"=>{"Name"=>"", "MaximumRetryCount"=>0} }`

This needs to check for `service_Container.host_config['RestartPolicy']['Name']`... probably worth restoring the `Docker::Container#autostart?` method from #2689

jakolehm:
Should work now.

agent/lib/kontena/service_pods/migrator.rb
```diff
@@ -0,0 +1,34 @@
...
+        @service_container.update({
+          'RestartPolicy' => {
+            'Name' => 'no',
+            'MaximumRetryCount' => 0
+          }
+        })
+      end
+
+      # @param service_container [Docker::Container]
+      def self.legacy_container?(service_container)
```
jnummelin: Is this really needed? it already checks the same thing in `migrate` anyway.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -192,6 +193,9 @@ def wait_for_port(service_pod, container, timeout: 300.0)
     def ensure_desired_state
       debug "state of #{service_pod.name}: #{service_pod.desired_state}"
       service_container = get_container(service_pod.service_id, service_pod.instance_number)
+
+      migrate_container(service_container) if service_container && legacy_container?(service_container)
```
jnummelin: Does the `PodWorker` really need to care/understand about `legacy_container?`? IMO it should just call `Migrator` to do it's magic and `Migrator` decides if things should be done or not. This feels a bit like leaking details. :)

SpComb: Yeah agreed, with the `remove_restart_policy if @service_container.autostart?` the `legacy_container?` isn't really needed anymore.

agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -393,5 +397,18 @@ def sync_state_to_master(service_pod, service_container, error = nil)
...
+    # @return [Boolean]
+    def legacy_container?(container)
+      Kontena::ServicePods::Migrator.legacy_container?(container)
+    end
+
+    # @param container [Docker::Container]
+    # @return [Docker::Container]
+    def migrate_container(container)
+      Kontena::ServicePods::Migrator.new(container).migrate
+      container.reload
```
SpComb: ...except for this `container.reload`... might be better to do in the `remove_restart_policy`, immediately after the `update`. It's also easier to follow like that.

SpComb: TBH this kind of mutation of the container object's own state is a little nasty in general... you'd like to see something like `service_container = migrate_container(service_container)` to make it clear that the container might change because it got migrated somehow.

Migrations like "the container labels are outdated" would imply that the migrator will terminate the container, and then `ensure_desired_state` would need to continue with `!service_container`. That requires having `migrate_container` return `nil`?

jakolehm:
Made changes based on review comments.

PR #2795, opened 2017-09-12
CLI: Add --values-from-stack option to stack commands for reading values from an installed stack

### LINKED ISSUE #2515: CLI: kontena stack build to read values from installed stack
Currently need to --values-to && --values-from or I have to enter values every time. This command should read them from an installed stack.

### LINKED ISSUE #2739: kontena stack upgrade --values-from-installed-stack
related https://github.com/kontena/kontena/issues/2515

kke:
Fixes #2739
Fixes #2515

Adds `--values-from-stack` to stack install/upgrade/show/validate that reads stack variable values from an installed stack from the master. It's also possible to give the same name as the current stack during upgrade to avoid answering any questions.

It also makes `--values-from` be multivalued, meaning that you can read from multiple files.

### DIFF
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -100,11 +100,10 @@ module Kontena::Cli::Stacks
         end
 
         where.option '--values-from-stack', '[STACK_NAME]', 'Read variable values from an installed stack', multivalued: true do |stackname|
-          response = client.get("stacks/#{current_grid}/#{stackname}")
-          variables = response['variables']
+          variables = read_values_from_stacks(stackname)
           Kontena.logger.debug { "Received variables from stack #{stackname} on Master: #{variables.inspect}" }
           warn "Stack #{stackname} does not have any values for variables" if variables.empty?
-          values_from_installed_stack.merge!(variables)
+          values_from_installed_stacks.merge!(variables)
           stackname
         end
 
@@ -115,6 +114,27 @@ module Kontena::Cli::Stacks
       end
 
       module InstanceMethods
+        def read_values_from_stacks(stackname)
+          result = {}
+          response = client.get("stacks/#{current_grid}/#{stackname}")
+          result.merge!(response['variables']) if response['variables']
+          if response['children']
+            response['children'].each do |child_info|
+              result.merge!(
+                read_values_from_stacks(child_info['name']).tap do |child_result|
+                  child_result.keys.each do |key|
+                    new_key = child_info['name'].dup # foofoo-redis-monitor
+                    new_key.sub!("#{stackname}-", '') # monitor
+                    new_key.concat ".#{key}" # monitor.foovariable
+                    child_result[new_key] = child_result.delete(key)
+                  end
+                end
+              )
+            end
+          end
+          result
+        end
+
         def values_from_file
           @values_from_file ||= {}
         end
@@ -123,12 +143,12 @@ module Kontena::Cli::Stacks
           @values_from_value_options ||= {}
         end
 
-        def values_from_installed_stack
-          @values_from_installed_stack ||= {}
+        def values_from_installed_stacks
+          @values_from_installed_stacks ||= {}
         end
 
         def values_from_options
-          @values_from_options ||= values_from_file.merge(values_from_installed_stack).merge(values_from_value_options)
+          @values_from_options ||= values_from_installed_stacks.merge(values_from_file).merge(values_from_value_options)
         end
 
         # Transforms a hash
--- a/cli/lib/kontena/cli/stacks/common.rb
+++ b/cli/lib/kontena/cli/stacks/common.rb
@@ -94,9 +94,24 @@ module Kontena::Cli::Stacks
       def self.included(where)
         where.prepend InstanceMethods
 
-        where.option '--values-from', '[FILE]', 'Read variable values from YAML' do |filename|
-          values_from_file.merge!(::YAML.safe_load(File.read(filename)))
-          true
+        where.option '--values-from', '[FILE|STACK_NAME]', 'Read variable values from a YAML file or an installed stack', multivalued: true do |filename|
+          if File.exist?(filename)
+            values_from_file.merge!(::YAML.safe_load(File.read(filename)))
+          else
+            begin
+              response = client.get("stacks/#{current_grid}/#{filename}")
+              variables = response['variables']
+              Kontena.logger.debug { "Received variables from stack #{filename} on Master: #{variables.inspect}" }
+              warn "Stack #{filename} does not have any values for variables" if variables.empty?
+              values_from_installed_stack.merge!(variables)
+            rescue Kontena::Errors::StandardError => ex
+              if ex.status == 404
+                exit_with_error "File or stack not found: #{filename}"
+              end
+              raise ex
+            end
+          end
+          filename
         end
 
         where.option '-v', "VARIABLE=VALUE", "Set stack variable values, example: -v domain=example.com. Can be used multiple times.", multivalued: true, attribute_name: :var_option do |var_pair|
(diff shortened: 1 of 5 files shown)


cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -94,9 +94,24 @@ module StackValuesFromOption
       def self.included(where)
         where.prepend InstanceMethods
 
-        where.option '--values-from', '[FILE]', 'Read variable values from YAML' do |filename|
-          values_from_file.merge!(::YAML.safe_load(File.read(filename)))
-          true
+        where.option '--values-from', '[FILE|STACK_NAME]', 'Read variable values from a YAML file or an installed stack', multivalued: true do |filename|
+          if File.exist?(filename)
```
SpComb: `kontena stack install --values-from ../some/path test/redis.yaml` => `DEBUG [Request]: GET http://localhost:9292/v1/stacks/development/../some/path`

Having this option either take a YAML file or a stack name feels pretty dirty, and having the `File.exist?` check work this way around feels particularly awful, leaking garbage into  API requests:

`kontena stack install --values-from '../some/path?lolwut=true' test/redis.yaml` => `172.17.0.1 - - [15/Sep/2017:11:36:01 +0000] "GET /v1/stacks/development/../some/path?lolwut=true HTTP/1.1" 404 2 0.0743`

IMO have a separate `--values-from-stack` option, or at least flip the logic around to be like `if filename =~ /\A(?!-)(\w|-)+\z/ && !File.exist?(filename)`... then that just means that `--values-from redis` will fail if you e.g. happen to be in a path that has a directory named `redis` - not ideal.

kke:
Extracted the stack part from `--values-from` to `--values-from-stack`

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -45,6 +45,10 @@ def install_dependencies
         caret "Installing dependency #{pastel.cyan(dependency[:stack])} as #{pastel.cyan(target_name)}"
         cmd = ['stack', 'install', '-n', target_name, '--parent-name', stack_name]
 
+        values_from_list.each do |vals_from|
+          cmd.concat ['--values-from', vals_from + "-" + dependency['name']]
```
SpComb: Hmm... I think I would expect the `--values-from` YAML file to contain some structured `child: foo: "value"` values in this case.

It's not exactly intuitive to have `kontena stack install --values-from values.yaml` try and read the child stack values from `values.yaml-foo`?

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -139,6 +139,11 @@ def execute
         else
           cmd = ['stack', 'install', '--name', stackname]
           cmd.concat ['--parent-name', stack['parent_name']] if stack['parent_name']
+
+          values_from_list.each do |vals_from|
+            cmd.concat ['--values-from', stackname.sub(name, vals_from)]
```
SpComb: What is the `String#sub` doing?

cli/lib/kontena/cli/stacks/install_command.rb
kke: I think that was supposed  to read them from `values-foo.yaml` but there's something odd  going on.

cli/lib/kontena/cli/stacks/upgrade_command.rb
kke: I think that's actually supposed to work when using `--values-from-stack`. 

Replacing something like `foo-redis` with `bar-redis` when running with `--values-from bar`.

cli/lib/kontena/cli/stacks/install_command.rb
SpComb: That doesn't sound sensible to me either. I wouldn't expect `kontena stack install --values-from values.yaml` to try reading for a file called `values-foo.yaml`.

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: Specs would be nice.

cli/lib/kontena/cli/stacks/install_command.rb
SpComb: Should this also be `values_from_stack_list.each` -> `--values-from-stack`?

Looking at the code for `--values-from`, that uses the same code as `-v`, so it might even already be accepting YAML `foo.bar` keys?

kke: The `-v` injection to children that is done through `dependency_variables_from_options` should already do that. If you have something like `foo.bar=123` then child stack `foo` will get `-v bar=123`.

kke:
It would be possible to rewrite the whole install thing to be recursive inside the command instead of going through `Kontena.run!` in the spirit of `stack upgrade` (create an `InstallResolver` or something like the `ChangeResolver` in #2823 )

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -94,9 +94,18 @@ module StackValuesFromOption
...
           values_from_file.merge!(::YAML.safe_load(File.read(filename)))
-          true
+          filename
+        end
+
+        where.option '--values-from-stack', '[STACK_NAME]', 'Read variable values from an installed stack', multivalued: true do |stackname|
+          response = client.get("stacks/#{current_grid}/#{stackname}")
+          variables = response['variables']
+          Kontena.logger.debug { "Received variables from stack #{filename} on Master: #{variables.inspect}" }
+          warn "Stack #{filename} does not have any values for variables" if variables.empty?
```
SpComb: The `warn` breaks:

```
NameError: undefined local variable or method `filename' for #<Kontena::Cli::Stacks::InstallCommand:0x00000002512910>
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:105:in `block (2 levels) in included'
  /usr/lib/ruby/2.3.0/logger.rb:427:in `add'
  /usr/lib/ruby/2.3.0/logger.rb:455:in `debug'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:105:in `block in included'
```

cli/lib/kontena/cli/stacks/common.rb
```diff
@@ -94,9 +94,18 @@ module StackValuesFromOption
...
+        where.option '--values-from', '[FILE]', 'Read variable values from a YAML file', multivalued: true do |filename|
           values_from_file.merge!(::YAML.safe_load(File.read(filename)))
-          true
+          filename
+        end
+
+        where.option '--values-from-stack', '[STACK_NAME]', 'Read variable values from an installed stack', multivalued: true do |stackname|
+          response = client.get("stacks/#{current_grid}/#{stackname}")
+          variables = response['variables']
+          Kontena.logger.debug { "Received variables from stack #{filename} on Master: #{variables.inspect}" }
```
SpComb: This is also broken with `DEBUG=true`

```
NameError: undefined local variable or method `filename' for #<Kontena::Cli::Stacks::UpgradeCommand:0x0000000134a2f0>
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:105:in `block (2 levels) in included'
  /usr/lib/ruby/2.3.0/logger.rb:427:in `add'
  /usr/lib/ruby/2.3.0/logger.rb:455:in `debug'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/common.rb:105:in `block in included'
```

kke: doh

kke:
The upgrade part is a lot cleaner in the dry-run pr

We could still replace the mess in this PR with something like `exit_with_error "--values-from-stack can't currently be used together with stacks with dependencies" if ....`.

kke:
Actually I think it should just read all the child stacks at once starting from the initial `--values-from-stack`, add a `name.` prefix to them and let the `-v` thing pass them along when running sub-installs.

kke:
And now it does that. Also added specs for --values-from-stack with dependencies.

PR #2802, opened 2017-09-14
Fix stack variables to not write to env by default

### LINKED ISSUE #2798: Stack variable envs leak between dependent child stacks
From #2707: Installing a stack with multiple dependent child stacks that share the same stack variable names, values for those per-stack variables can leak between the child stacks when using the default `to: env` and `from: env` variable resolvers.

## Example

```
$ kontena stack install test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```

```
$ kontena stack upgrade deptest test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```

#### DEBUG `kontena stack upgrade`
```
DEBUG ENV['peers']: 
DEBUG execute deptest-bar with parent_name=deptest values={"peers"=>"foo=redis://-foo:6379/"} defaults={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}
DEBUG upgrade deptest-bar with variables={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}


DEBUG ENV['peers']: foo=redis://-foo:6379/
DEBUG execute deptest-foo with parent_name=deptest values={} defaults={"peers"=>nil, "PARENT_STACK"=>"deptest"}
DEBUG upgrade deptest-foo with variables={"peers"=>"foo=redis://-foo:6379/", "PARENT_STACK"=>"deptest"}

DEBUG ENV['peers']: foo=redis://-foo:6379/
DEBUG execute deptest with parent_name= values={} defaults={"foo"=>"deptest-foo", "bar"=>"deptest-bar"}
DEBUG upgrade deptest with variables={"foo"=>"deptest-foo", "bar"=>"deptest-bar"}
```

(slightly different stack variable values from the example)

#### `deptest.yaml`
```yaml
stack: test/deptest
depends:
  foo:
    stack: deptest-child.yaml
  bar:
    stack: deptest-child.yaml
    variables:
      peers: "foo=redis://$STACK-foo:6379/"
services:
  redis:
    image: redis
    environment:
      - PEERS=foo=redis://$foo:6379/
```

#### `deptest-child.yaml`
```yaml
stack: test/deptest-child
variables:
  peers:
    type: string
    required: false
services:
  redis:
    image: redis
    environment:
      - PEERS=${peers}
```

kke:
Fixes #2798

Before this PR stacks had these defaults applied to all variables:

```yaml
from:
  env: <same_as_variable_name>
to:
  env: <same_as_variable_name>
```

After this PR the variables are no longer sent to local env to avoid leaking stack variables from parent stacks to child stacks when using stack dependencies.

The default becomes:
```yaml
from:
  env: <same_as_variable_name>
```

### Breaking change

After this PR something like this no longer works:

```yaml
variables:
   db_host:
     type: string
     from:
       prompt: Enter DB_HOST
  backup_db_host:
     type: string
     from:
       prompt: Enter backup DB_HOST or leave blank to use the DB_HOST
       env: db_host
```

To fix, change to:

```diff
-       env: db_host
+       variable: db_host
```

or send the value to env manually:

```diff
     from:
       prompt: Enter DB_HOST
+     to:
+       env: db_host
```

Opto has been updated to 1.8.7 to bring in the `variable` resolver.

### DIFF
--- a/cli/kontena-cli.gemspec
+++ b/cli/kontena-cli.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "launchy", "~> 2.4.3"
   spec.add_runtime_dependency "hash_validator", "~> 0.7.1"
   spec.add_runtime_dependency "retriable", "~> 2.1.0"
-  spec.add_runtime_dependency "opto", "1.8.5"
+  spec.add_runtime_dependency "opto", "1.8.7"
   spec.add_runtime_dependency "semantic", "~> 1.5"
   spec.add_runtime_dependency "liquid", "~> 4.0.0"
   spec.add_runtime_dependency "tty-table", "~> 0.8.0"
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -131,7 +131,7 @@ module Kontena::Cli::Stacks
       def variables
         @variables ||= ::Opto::Group.new(
           internals_interpolated_yaml.fetch('variables', {}).merge(default_envs_to_options),
-          defaults: { from: :env, to: :env }
+          defaults: { from: :env }
         )
       end
 


SpComb:
Confirmed to fix #2798

```
$ kontena stack install test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```

```
$ kontena stack upgrade deptest test/deptest.yaml 
...
$ kontena service show deptest-foo/redis | grep PEERS
    - PEERS=
$ kontena service show deptest-bar/redis | grep PEERS
    - PEERS=foo=redis://$STACK-foo:6379/
```

SpComb:
Look a quick look through all the public stacks on the stack registry, and I couldn't find any `from: env: ...` usage that would be broken like this. All of the `from: env: ...` usages were like

```yaml
variables:
  version:
    from:
      env: VERSION
```

SpComb:
@jakolehm @jnummelin aware of any `from: env: ...` usages that would break here? I couldn't find any, so I think this is an easy call to fix the dependent stack variable leakage. I'll mention it in the release changelogs.

kke:
I don't think many people realized it was there.

jakolehm:
> aware of any from: env: ... usages that would break here?

Not aware, haven't really used/seen that anywhere.

matti:
I had no idea that you could do that. After seeing the usage example, I’m quite sure that I would have never done that.

kke:
There's a fixture btw that uses this, I don't know why specs are still passing.

https://github.com/kontena/kontena/blob/386dd763a6c242057410abbedbbcfe665f872165/cli/spec/fixtures/stack-with-variables.yml

SpComb:
> There's a fixture btw that uses this, I don't know why specs are still passing.

The `from: env` usages there are similar to what I found in the stacks, so stuff like variable `tag` being `from: env: TAG`. The `to: env: TAG ` is ATM broken per  #2815, but then it's not actually interpolated in that spec fixture. I have no idea what the `WORDPRESS_DB_PASSWORD=${STACK}_secret` and `type: string # has a callback that writes the value to vault` there are supposed to be, though.

superzadeh:
It seems this breaks the way I was using variables to define tags on my docker images. I did not have to check this during the RC, but still wanted to share it here in case anyone else runs into it (which I doubt).

The problem seems to be that if you used an `to: env` variable in a `kontena.yml` that has the same name as an `ENV` variable in your `docker-compose` file, then the `kontena stack` validation fails. Looks like this is a bug in the YML reader when it tries to interpolate a variable.

Here is config to reproduce this:

```yml
# kontena.yml
stack: demo
version: 1.0.0

variables:
  myvar:
    type: string
    default: 'test'
    from:
      env: ENV_VAR
    to:
      env: ENV_VAR

services:
  node:
    extends:
      file: docker-compose.yml
      service: node

# docker-compose.yml
version: '2.1'
services:
  node:
    image: node:alpine
    environment:
      - ENV_VAR=$ENV_VAR
```

then run

```sh
$ ENV_VAR=whatever
$ kontena stack validate
 [error] RuntimeError : Undeclared variable 'ENV_VAR' in c:/repos/demo/docker-compose.yml:5 --       - ENV_VAR=$ENV_VAR
         See C:/Users/superzadeh/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception
```


*******
For more context, here is how I was using this in my setup and ran into the error:

Below are the files that I use as an example. The way it works is:
- I am using Travis as CI. Travis sets `TRAVIS_BRANCH` when it builds
- If the command `kontena stack build` is run in a CI environment, then the tag is the branch-name. This allows me to easily have find images related to a branch. This is especially useful because it lets me tag images automatically with `:staging`, `:v1.3.0` based on my branch or git tags.
- If the `TRAVIS_BRANCH` is not set, then the current kontena `${GRID}` will be used as tag. This provides a nice default, since the image built/pushed has the same name as the current grid.
- Finally, in `docker-compose.yml`, the tag used when you build the image (using `docker-compose build` for example) is `local`. If for some reason `docker-compose build` is used in a CI, then the image is tagged with the branch name (there are some case where we are not using `docker-compose build` in the CI instead of `kontena-stack build`).

```yml
# kontena.yml
stack: mystack
version: 1.0.0

variables:
  app_branch:
    type: string
    default: ${GRID}
    from:
      env: TRAVIS_BRANCH
    to:
      env: TRAVIS_BRANCH

services:
  myservice:
    extends:
      file: docker-compose.yml
      service: myservice
```

```yml
# docker-compose.yml
version: '2.1'
services:
  myservice:
    image: bunch/myservice:${TRAVIS_BRANCH-local}
    build:
      context: .
      dockerfile: Dockerfile
```

The full exception stack trace:

```
E, [2017-10-16T15:18:56.884759 #9176] ERROR -- CLI: Undeclared variable 'TRAVIS_BRANCH' in c:/repos/akula/docker-compose.yml:3 --     image: bunch/akula:${TRAVIS_BRANCH-local} (RuntimeError)
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:379:in `block (2 levels) in interpolate'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:369:in `gsub'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:369:in `block in interpolate'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:364:in `map'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:364:in `with_index'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:364:in `interpolate'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:99:in `fully_interpolated_yaml'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:249:in `validate'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:202:in `execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:353:in `from_external_file'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:444:in `extend_config'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:321:in `process_config'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:278:in `block in parse_services'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:277:in `each'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:277:in `parse_services'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:211:in `block in execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:205:in `chdir'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/yaml/reader.rb:205:in `execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/common.rb:33:in `stack'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/cli/stacks/validate_command.rb:55:in `execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/command.rb:213:in `run'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/command.rb:213:in `run'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/lib/kontena/command.rb:213:in `run'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/clamp-1.1.2/lib/clamp/command.rb:132:in `run'
C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/kontena-cli-1.4.0/bin/kontena:18:in `<top (required)>'
C:/Ruby24-x64/bin/kontena:22:in `load'
C:/Ruby24-x64/bin/kontena:22:in `<main>'
```

I made the following change to fix this. It now means that `docker-compose build` should only be used locally/in development, whereas `kontena stack build` should be used for everything else.

```diff
# docker-compose.yml
version: '2.1'
services:
  myservice:
-    image: bunch/myservice:${TRAVIS_BRANCH-latest}
+    image: bunch/myservice:local
    build:
      context: .
      dockerfile: Dockerfile

# kontena.yml
variables:
  app_branch:
    type: string
    default: ${GRID}
    from:
      env: TRAVIS_BRANCH
    to:
-      env: TRAVIS_BRANCH
+      env: TAG

services:
  myservice:
    extends:
      file: docker-compose.yml
      service: myservice
    instances: 2
-    image: "bunch/myservice:${TRAVIS_BRANCH}"
+    image: "bunch/myservice:${TAG}"
```

SpComb:
> It seems this breaks the way I was using variables to define tags on my docker images. I did not have to check this during the RC, but still wanted to share it here in case anyone else runs into it (which I doubt).

Thanks for reporting this regression as #2943. It's not an intended effect of this "breaking change", whichjust removed the default `to:` `env: ...`. If you're explicitly using to-env in your YAML, then it should behave the same as before.

Issue #2801, opened 2017-09-14
Updating a certificate does not update services using that certificate


SpComb:
From #2736: Renewing an existing certificate using `kontena certificate request` does not update services that are using that certificate:

```
$ curl -v https://test.188.226.131.23.nip.io/ > /dev/null
* 	 server certificate verification OK
* 	 server certificate status verification SKIPPED
* 	 common name: test.188.226.131.23.nip.io (matched)
* 	 server certificate expiration date OK
* 	 server certificate activation date OK
* 	 certificate public key: RSA
* 	 certificate version: #3
* 	 subject: CN=test.188.226.131.23.nip.io
* 	 start date: Thu, 14 Sep 2017 09:35:00 GMT
* 	 expire date: Wed, 13 Dec 2017 09:35:00 GMT
* 	 issuer: C=US,O=Let's Encrypt,CN=Let's Encrypt Authority X3
* 	 compression: NULL
```

```
$ bundle exec bin/kontena certificate request test.188.226.131.23.nip.io
 [done] Requesting certificate for test.188.226.131.23.nip.io      
$ bundle exec bin/kontena service deploy ingress-lb/lb
 [done] Deploying service ingress-lb/lb      
⊛ Deployed instance development/ingress-lb/lb-1 to node core-01
$ curl -v https://test.188.226.131.23.nip.io/ > /dev/null
* 	 start date: Thu, 14 Sep 2017 09:35:00 GMT
* 	 expire date: Wed, 13 Dec 2017 09:35:00 GMT
```

This is because the `GridService.updated_at` timestamp has not changed, so the agent does not re-create the container on the deploy.

The existing `GridCertificates::GetCertificate` mutation uses the `GridSecrets::Update/Create` mutations, and those trigger `refresh_grid_services` to update any services that are using those certificate secrets.

Workaround is to `kontena service deploy --force`:

```
$ kontena service deploy --force ingress-lb/lb
 [done] Deploying service ingress-lb/lb      
⊛ Deployed instance development/ingress-lb/lb-1 to node core-01
$ curl -v https://test.188.226.131.23.nip.io/ > /dev/null
* 	 start date: Thu, 14 Sep 2017 09:42:00 GMT
* 	 expire date: Wed, 13 Dec 2017 09:42:00 GMT
```

jnummelin:
I think the "natural" fix is to trigger similar `refresh_grid_services` for all the services using a given cert when it's updated.

PR #2816, opened 2017-09-14
Certificate auto renewal

jnummelin:
Now with TLS-SNI domain verifications it's possible to renew certificates fully automatically.

Tested to actually work locally with some SSH tunnel magic, tunneling the LE tls-sni request from DO ubuntu to local dev box. :)

Maybe some of the logic like `should_renew?` should move to `Certificate` model?

In for 1.4 still?

### DIFF
--- a/server/app/jobs/certificate_renew_job.rb
+++ b/server/app/jobs/certificate_renew_job.rb
@@ -5,7 +5,7 @@ class CertificateRenewJob
   include WaitHelper
   include DistributedLocks
 
-  RENEW_INTERVAL = 1.day.to_i
+  RENEW_INTERVAL = 1.hour.to_i
 
   RENEW_TRESHOLD = 7.days
 
--- a/server/app/jobs/certificate_renew_job.rb
+++ b/server/app/jobs/certificate_renew_job.rb
@@ -4,7 +4,7 @@ class CertificateRenewJob
   include CurrentLeader
   include WaitHelper
 
-  RENEW_INTERVAL = 5 * 60 # Run the renew check every 5 mins
+  RENEW_INTERVAL = 1.day.to_i
 
   RENEW_TRESHOLD = 7.days
 
@@ -69,6 +69,7 @@ class CertificateRenewJob
         wait_until!("deployment of tls-sni secret is finished", timeout: 300, threshold: 20) {
           domain_auth.reload.status != :deploying
         }
+        raise "Deployment of tls-sni secret failed" if domain_auth.reload.status == :deploy_error
       else
         # No point to continue, cert renewal not gonna succeed
         raise "Domain authorization failed: #{outcome.errors.message}"
--- a/server/app/jobs/certificate_renew_job.rb
+++ b/server/app/jobs/certificate_renew_job.rb
@@ -3,6 +3,7 @@ class CertificateRenewJob
   include Logging
   include CurrentLeader
   include WaitHelper
+  include DistributedLocks
 
   RENEW_INTERVAL = 1.day.to_i
 
@@ -13,12 +14,16 @@ class CertificateRenewJob
   end
 
   def perform
-    # TODO Maybe sleep a while first to allow everything else to start up before running this loop?
+    # sleep a while first to allow everything to settle down before running this loop
+    sleep 5.minutes
     info 'starting to watch certificate renewals'
-    every(RENEW_INTERVAL) do
+    loop do
       if leader?
-        renew_certificates
+        with_dlock('cert_renewal_job') {
+          renew_certificates
+        }
       end
+      sleep RENEW_INTERVAL
     end
   end
 
@@ -31,14 +36,12 @@ class CertificateRenewJob
   def renew_certificate(certificate)
     if can_renew?(certificate)
       info "certificate renewal needed for #{certificate.subject}"
-      begin
-        authorize_domains(certificate)
-        request_new_cert(certificate)
-      rescue => exc
-        error "Failed to renew certificate for #{certificate.subject}"
-        error exc
-      end
+      authorize_domains(certificate)
+      request_new_cert(certificate)
     end
+  rescue => exc
+      error "Failed to renew certificate for #{certificate.subject}"
+      error exc
   end
 
   # Checks if all domains are authorized with tls-sni, we can't automate anything else for now
--- a/server/app/jobs/certificate_renew_job.rb
+++ b/server/app/jobs/certificate_renew_job.rb
@@ -1,0 +1,83 @@
+class CertificateRenewJob
+  include Celluloid
+  include Logging
+  include CurrentLeader
+
+  RENEW_INTERVAL = 5 * 60 # Run the renew check every 5 mins
+
+  def initialize(perform = true)
+    async.perform if perform
+  end
+
+  def perform
+    # TODO Maybe sleep a while first to allow everything else to start up before running this loop?
+    info 'starting to watch certificate renewals'
+    every(RENEW_INTERVAL) do
+      if leader?
+        renew_certificates
+      end
+    end
+  end
+
+  def renew_certificates
+    Grid.each do |grid|
+      grid.certificates.each do |cert|
+        renew_certificate(cert)
+      end
+    end
+  end
+
+  def renew_certificate(certificate)
+    if should_renew?(certificate) && can_renew?(certificate)
+      info "certificate renewal needed for #{certificate.subject}"
+      begin
+        authorize_domains(certificate)
+        request_new_cert(certificate)
+      rescue => exc
+        error "Failed to renew certificate for #{certificate.subject}"
+        error exc
+      end
+    end
+  end
+
+  def should_renew?(certificate)
+    certificate.valid_until > Time.now + 7.days
+  end
+
+  # Checks if all domain are authorized with tls-sni, we can't automate anything else for now
+  def can_renew?(certificate)
+    certificate.all_domains.each do |domain|
+      domain_auth = certificate.grid.grid_domain_authorizations.find_by(domain: domain)
+      unless domain_auth && domain_auth.authorization_type == 'tls-sni-01'
+        return false
+      end
+    end
+
+    true
+  end
+
+  def authorize_domains(certificate)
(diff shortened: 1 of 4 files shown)


server/app/jobs/certificate_renew_job.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def initialize(perform = true)
+    async.perform if perform
+  end
+
+  def perform
+    # TODO Maybe sleep a while first to allow everything else to start up before running this loop?
+    info 'starting to watch certificate renewals'
+    every(RENEW_INTERVAL) do
+      if leader?
```
jnummelin: In theory this has a possibility for a race. As the renewal process is bit slow and if one has many certs, leader can change while processing all certs. That would mean that the new leader might process same certs and thus make "double-renewal"

jakolehm:
> In for 1.4 still?

Yes please.

server/app/models/certificate.rb
```diff
@@ -30,4 +30,8 @@ def bundle
     self.full_chain + self.private_key
   end
 
+  def all_domains
```
SpComb: `# @return [Array<String>]`

server/app/jobs/certificate_renew_job.rb
SpComb: Maybe better to use `with_dlock` instead of `leader?` for this. It doesn't necessarily have to be the leader that handles it?

Although in that case, the job would need to scale/randomize the `every` loop to spread the intervals across all of the workers, or it would happen every `RENEW_INTERVAL`/number-of-workers...

IMO this doesn't need to happen as frequently as once every 5 minutes... perhaps only once a day... there just needs to be some room for retrying.

server/app/jobs/certificate_renew_job.rb
```diff
@@ -0,0 +1,91 @@
...
+        renew_certificates
+      end
+    end
+  end
+
+  def renew_certificates
+    Grid.each do |grid|
+      grid.certificates.each do |cert|
```
SpComb: `Certificate.where(valid_until < Time.now + RENEW_TRESHOLD)` - cheap optimization

The `can_renew?` probably still needs to be in the loop, because you can't query across relations... should maybe raise an error, although there's nothing useful the server can do if it's unable to renew an expiring cert.

server/app/jobs/certificate_renew_job.rb
```diff
@@ -0,0 +1,91 @@
...
+        grid: certificate.grid,
+        domain: domain,
+        authorization_type: 'tls-sni-01',
+        linked_service: "#{domain_auth.grid_service.stack.name}/#{domain_auth.grid_service.name}"
+      )
+      if outcome.success?
+        domain_auth = outcome.result
+        wait_until!("deployment of tls-sni secret is finished", timeout: 300, threshold: 20) {
+          domain_auth.reload.status != :deploying
+        }
```
SpComb: Also  `raise if domain_auth.reload.status == :deploy_error` after the wait loop?

server/app/jobs/certificate_renew_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def request_new_cert(certificate)
+    info "requesting certificate for #{certificate.subject}"
+    outcome = GridCertificates::RequestCertificate.run(grid: certificate.grid, domains: certificate.all_domains)
+    unless outcome.success?
+      raise "Certificate request failed: #{outcome.errors.message}"
+    end
+    info "certificate for #{certificate.subject} renewed succesfully"
+  end
```
SpComb: re #2801 this also doesn't re-deploy the services using the certificate, in order to actually update the renewed cert into usage... although if the `GridCertificates::RequestCertificate` is fixed to mark the services as updated, then the `GridSchedulerJob` will trigger deploys for those services.

jnummelin: Yes, let's fix this in separate PR. (which is already cooking...)

server/app/jobs/certificate_renew_job.rb
jnummelin: good idea

server/app/jobs/certificate_renew_job.rb
jnummelin: 👍

server/app/jobs/certificate_renew_job.rb
jnummelin: It doesn't matter whether it is the leader or not, just needs to be ONE worker that handles it. Thus I thought it would be most simple to handle this only on leader. IMO running this only on leader makes it super simple and the possible race should be somewhat theoretical. Putting in loop randomization and dlocks makes it unnecessarily complex IMO.

jnummelin: Yes, the renewal loop should be ok to run once a day. Now it starts renewals 7 days before expiration, so it should have at least 6 times to retry.

server/app/jobs/certificate_renew_job.rb
jnummelin: done

server/app/jobs/certificate_renew_job.rb
```diff
@@ -0,0 +1,88 @@
...
+  RENEW_INTERVAL = 1.day.to_i
+
+  RENEW_TRESHOLD = 7.days
+
+  def initialize(perform = true)
+    async.perform if perform
+  end
+
+  def perform
+    # TODO Maybe sleep a while first to allow everything else to start up before running this loop?
```
SpComb: Hmm... the way that Celluloid `every` works is that it only runs the block after the interval has expired, not right at the start.

Wondering if the `RENEW_THRESHOLD = 1.day` is now even too long... it means that if the server restarts every night, it might never actually run this job.

server/app/jobs/certificate_renew_job.rb
SpComb: > Putting in loop randomization and dlocks makes it unnecessarily complex IMO.

Yes, the dlocks are not really that suitable for trying to schedule tasks across all the workers.

> IMO running this only on leader makes it super simple and the possible race should be somewhat theoretical.

Hmm... the leader period is ~60s, I think it's entirely normal to see the leader change as often as once a minute?

```
I, [2017-09-15T10:15:53.434178 #8]  INFO -- LeaderElectorJob: participating leader elections
I, [2017-09-15T10:15:53.456462 #8]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:15:53.511674 #18]  INFO -- LeaderElectorJob: participating leader elections
I, [2017-09-15T10:15:54.078487 #16]  INFO -- LeaderElectorJob: participating leader elections
I, [2017-09-15T10:15:54.502236 #12]  INFO -- LeaderElectorJob: participating leader elections
I, [2017-09-15T10:16:51.483532 #16]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:17:49.505809 #18]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:18:47.527563 #8]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:19:45.553094 #18]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:20:43.578673 #16]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:21:41.591414 #18]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:22:39.621962 #12]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:23:37.643350 #16]  INFO -- LeaderElectorJob: won election ♚
I, [2017-09-15T10:25:33.672318 #12]  INFO -- LeaderElectorJob: won election ♚
```

SpComb: If the different `CertificateRenewJob` every loops start drifting out of phase, then I imagine it's entirely possible to hit a race here. It just needs the `every` loop on the current leader to be the worker where the `every` loop is firing the earliest, and then for a routine leader re-election to happen, which means that the `every` loop might fire again very soon on the new leader.

Maybe keep the `leader?` for the scheduling, but then stick a `with_dlock` around the call to `renew_certificates`?

jnummelin: yeah, makes sense

server/app/jobs/certificate_renew_job.rb
SpComb: I think I'd stick to maximum delays on the order of an hour for the `every` loops, so maybe drop the `RENEW_INTERVAL` down to like `1h`?

Longer intervals would really need some kind of scheduling system that manages the timestamps in mongo.

jnummelin: oh, very true. :) Maybe something like:
```
sleep 5.minutes
loop do
  do_the_magic
  sleep RENEW_THRESHOLD
end
````
??

File: server/app/jobs/certificate_renew_job.rb (hunk 1 above)
Comment (jnummelin) [reply server/app/jobs/certificate_renew_job.rb]: done

File: server/app/jobs/certificate_renew_job.rb [hunk 7]
```diff
@@ -0,0 +1,91 @@
+class CertificateRenewJob
+  include Celluloid
+  include Logging
+  include CurrentLeader
+  include WaitHelper
+  include DistributedLocks
+
+  RENEW_INTERVAL = 1.day.to_i
```
SpComb: Re-commenting on this because I wrote it in the wrong place, and it got lost: Wondering if the `RENEW_INTERVAL = 1.day` is now even too long... it means that if the server gets restarted every day, this job might never actually run?

I think I'd stick to maximum delays on the order of an hour for timers like `every`/`sleep`... longer intervals would really need some kind of scheduling system that manages the timestamps in mongo?

So maybe drop the RENEW_INTERVAL down to like 1h?

SpComb: Correction, this is less of a concern now with the initial 5-minute run. I'm a little wary of doing a 24h `Celluloid.sleep`, but I don't see why it wouldn't work.

Long-term we might want to have some kind of mongo-based scheduled tasks thing.

PR #2822, opened 2017-09-15
Fix interpolation of stack variables declared by `to: env`

### LINKED ISSUE #2815: [1.4] Stack reader does not allow interpolating `to: env: FOO` using `${FOO}`
Regression from https://github.com/kontena/kontena/pull/2707#discussion_r138916318: with a stack variable `foo` using `to: env: FOO`, you can no longer interpolate `${FOO}`:

```
 [error] RuntimeError : Undeclared variable 'FOO' in /kontena/test/test/to-env.yaml:12 --       - FOO=${FOO}
```

This pattern is widely used in existing stacks, so this is a serious regression. There should be an e2e spec for it.

#### `to-env.yaml`
```yaml
stack: test/to-env
variables:
  foo:
    type: string
    from:
      prompt:
    to:
      env: FOO
services:
  redis:
    image: redis
    environment:
      - FOO=${FOO}
```

#### `kontena stack validate test/to-env.yaml `
```
> Enter foo : foo
 [error] RuntimeError : Undeclared variable 'FOO' in /kontena/test/test/to-env.yaml:12 --       - FOO=${FOO}
         See /root/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception
```

#### `kontena stack validate test/to-env.yaml `
```
> Enter foo : foo
---
stack: test/to-env
variables:
  foo:
    label: foo
    type: string
    from:
      prompt: 
      env: foo
    to:
      env: FOO
    errors: {}
    value: foo
  STACK:
    label: STACK
    type: string
    from:
      env: STACK
    to:
      env: STACK
    errors: {}
    value: to-env
  GRID:
    label: GRID
    type: string
    from:
      env: GRID
    to:
      env: GRID
    errors: {}
    value: validate
services:
  redis:
    image: redis
    environment:
    - FOO=foo
```

kke:
Fixes #2815

#2707 removed this functionality but seems like it is still needed.

When you have something like:

```yaml
stack: test/to-env
variables:
  foo:
    type: string
    value: hello
    to:
      env: FOO
services:
  redis:
    image: redis
    environment:
      - FOO=${FOO}
```

Then without this PR you would get "Undeclared variable FOO". This PR makes any environment variables that are listed in "to: env: " to be considered "declared".

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -368,7 +368,11 @@ module Kontena::Cli::Stacks
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
+                  if variables.find { |opt| opt.to[:env][var] }
+                    val = env[var]
+                  else
+                    raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
+                  end
                 else
                   val = opt.value
                   if opt.to['env']
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -368,8 +368,9 @@ module Kontena::Cli::Stacks
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  if variables.find { |opt| opt.to[:env][var] }
-                    val = env[var]
+                  to_env = variables.find { |opt| opt.to[:env][var] }
+                  if to_env
+                    val = to_env.value
                   else
                     raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
                   end
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -376,9 +376,6 @@ module Kontena::Cli::Stacks
                   end
                 else
                   val = opt.value
-                  if opt.to['env']
-                    Array(opt.to['env']).each { |env_var| env[env_var] = val.to_s }
-                  end
                 end
               else
                 val = substitutions[var]
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -368,7 +368,7 @@ module Kontena::Cli::Stacks
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  to_env = variables.find { |opt| opt.to[:env][var] }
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -368,7 +368,12 @@ def interpolate(content, use_opto: true, substitutions: {}, raise_on_unknown: fa
...
-                  raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
+                  to_env = variables.find { |opt| opt.to[:env][var] }
+                  if to_env
+                    val = to_env.value
+                  else
+                    raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
+                  end
                 else
                   val = opt.value
                   if opt.to['env']
```
SpComb: What does this `if opt.to['env']` => `Array(opt.to['env']).each { |env_var| env[env_var] = val.to_s }` do? Doesn't the `to: env` resolver itself already do that?

kke: It indeed should. I don't know why it's there

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -368,7 +368,7 @@ def interpolate(content, use_opto: true, substitutions: {}, raise_on_unknown: fa
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  to_env = variables.find { |opt| opt.to[:env][var] }
```
kke: I think this used to mostly work accidentally when `to: env: string` because of `String#[]` but could have produced some invalid results:

```ruby
"foofoo"["oof"]
=> "oof"
```

kke: And it also produced no such method `[]` for NilClass when the option had no `to[:env]`

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -368,12 +368,14 @@ def interpolate(content, use_opto: true, substitutions: {}, raise_on_unknown: fa
               if use_opto
                 opt = variables.option(var)
                 if opt.nil?
-                  raise RuntimeError, "Undeclared variable '#{var}' in #{file}:#{line_num} -- #{row}" if raise_on_unknown
+                  to_env = variables.find { |opt| Array(opt.to[:env]).include?(var) }
```
SpComb: Might want to consider a `Hash` for variable lookups at some point, this is `O(N*M)` on the number of variables/interpolations. With a complex enough stack you might notice it... needs profiling before it's worth considering trying to optimize it though.

BTW: Looks like the same also applies to `Opto::Group#option`...

kke: Well opto could just use [sets](http://ruby-doc.org/stdlib-2.4.1/libdoc/set/rdoc/Set.html) (which i believe to use hash internally). They're much faster than arrays and mostly behave in a similar fashion.

Issue #2311, opened 2017-05-12
Stack upgrade should also update services


SpComb:
The current semantics of the `kontena stack upgrade --no-deploy` (`PUT /v1/stacks/:grid/:stack`) and `kontena stack deploy` (`POST /v1/stacks/:grid/:stack/deploy`) are weird, because the upgrade itself does not have any visible effect on the `kontena service` state, nor can the upgraded services be deployed using `kontena service deploy`.

The `Stacks::Create` mutation will create the initial `StackRevision`, and run `GridServices::Create` for each service. This works as expected, and the services will be seen in an initialized state with their initial configuration after a `kontena stack install --no-deploy`.

The `Stacks::Update` mutation will validate any created, updated or deleted services before creating a new `StackRevision`, but it will only run `GridServices::Create` for any newly added services. This isn't that great, because it does not update the actual `GridService` configuration, and does not allow deploying the updated services individually.

The `Stacks::Deploy` mutation will then actually run the `GridServices::Update` mutations, in addition to re-running `GridServices::Create` for any spuriously removed (?) services after the create/update. It will then proceed to run the `StackDeployWorker`, which will run `GridServices::Delete` for any removed services before deploying the configured stack services.

It seems like it would be better to call the `GridServices::Update` from the `Stacks::Update` mutation:

* Being able to `kontena stack upgrade --no-deploy foo ...` and then `kontena service deploy foo/bar` would serve as a good workaround for the kind of manual stack service management in #1767 

* Not seeing any changes in the `kontena service` configuration after `kontena stack upgrade --no-deploy` is confusing
  * #1659 Upgrading a stack does not change values 
  * #1670 Stack upgrade did not update image url  
  * Support cases on the community slack

Not sure how the `GridServices::Delete` should be handled, though. The `Stacks::Deploy` mutation is probably still the correct place for it.

SpComb:
OTOH the `GridSchedulerJob` currently implements a [`lagging_behind?`](https://github.com/kontena/kontena/blob/v1.2.2/server/app/services/grid_scheduler.rb#L75) logic that triggers deploys for services that have been `updated` without being `deployed`.

I'm not sure that's a good idea? It would break these kinds of manual deployments.

PR #2823, opened 2017-09-15
Add --dry-run to simulate kontena stack upgrade

### LINKED ISSUE #2819: Having a dry-run option for stack upgrade would be great
We've got a Continuous Deployment system which acts on certain changes, but we'd also like to verify on Pull Requests that changes don't break deployment.

For this a dry-run option would be awesome, which would talk to the stack and even verify the `kontena.yml` parameters but not make any changes.

kke:
Fixes #2819

Adds `--dry-run` option to stack upgrade and also gives better output before asking for confirmation when running without `--dry-run`.

In the future the report could show some finer diff of the changes to upgraded services.

### DIFF
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -1,4 +1,5 @@
 require_relative 'common'
+require 'json'
 
 module Kontena::Cli::Stacks
   class UpgradeCommand < Kontena::Command
@@ -60,6 +61,10 @@ module Kontena::Cli::Stacks
       normalized_data
     end
 
+    def dry_msg(message)
+      puts pastel.red('DRY RUN: ') + message
+    end
+
     def merge_data(local_data, remote_data)
       merged = {}
       unless local_data.nil? || local_data.empty?
@@ -75,6 +80,66 @@ module Kontena::Cli::Stacks
         end
       end
       merged
+    end
+
+    def dry_report(merged)
+      added_services = []
+      removed_services = []
+      kept_services = {}
+      merged.each do |stackname, data|
+        if data[:remote].nil?
+          dry_msg "New stack #{stackname} would be installed to master"
+          added_services = data[:local][:stack]['services'].map { |svc| "#{stackname}/#{svc['name']}" }
+        elsif data[:local]
+          data[:local][:stack]['services'].each do |svc|
+            remote_svc = data[:remote]['services'].find { |s| s['name'] == svc['name'] }
+            if remote_svc
+              kept_services["#{stackname}/#{svc['name']}"] ||= {}
+              #kept_services["#{stackname}/#{svc['name']}"][:before] = JSON.load(JSON.dump(remote_svc)) # naughty deep clone/dup
+              #kept_services["#{stackname}/#{svc['name']}"][:after] = JSON.load(JSON.dump(svc))
+            else
+              added_services << "#{stackname}/#{svc['name']}"
+            end
+          end
+        end
+      end
+
+      merged.each do |stackname, data|
+        if data[:local].nil?
+          dry_msg "Stack  #{stackname} would be removed from master"
+          removed_services = data[:remote]['services'].map { |svc| "#{stackname}/#{svc['name']}" }
+        else
+          data[:local][:stack]['services'].each do |svc|
+            unless data[:remote]['services'].find { |s| s['name'] == svc['name'] }
+              added_services << "#{stackname}/#{svc['name']}"
+            end
+          end
+          data[:remote]['services'].each do |svc|
+            unless data[:local][:stack]['services'].find { |s| s['name'] == svc['name'] }
+              removed_services << "#{stackname}/#{svc['name']}"
+            end
+          end
+        end
+      end
+
+      unless added_services.empty?
+        dry_msg "The following services would be added to master"
+        added_services.each { |svc| puts " - #{svc}" }
+      end
+
+      unless removed_services.empty?
+        dry_msg "The following services would be removed from master"
+        removed_services.each { |svc| puts " - #{svc}" }
+      end
+
+      unless kept_services.empty?
+        dry_msg "The following services would be kept:"
+        kept_services.each do |svc_name, before_after|
+          #before = before_after[:before]
+          #after = before_after[:after]
+          puts "- #{svc_name}"
+        end
+      end
     end
 
     def execute
@@ -104,11 +169,7 @@ module Kontena::Cli::Stacks
         end
         confirm unless force?
         removes.reverse_each do |removed_stack|
-          if dry_run?
-            caret "Would remove stack #{removed_stack} without --dry-run", dots: false
-          else
-            Kontena.run!('stack', 'remove', '--force', '--keep-dependencies', removed_stack)
-          end
+          Kontena.run!('stack', 'remove', '--force', '--keep-dependencies', removed_stack) unless dry_run?
           merged.delete(removed_stack)
         end
       end
@@ -134,6 +195,11 @@ module Kontena::Cli::Stacks
         abort_on_validation_errors(data[:local][:loader].reader.errors, data[:local][:loader].source)
       end
 
+      if dry_run?
+        dry_report(merged)
+        return
+      end
+
       merged.reverse_each do |stackname, data|
         stack = data[:local][:stack]
         if data[:remote]
@@ -150,30 +216,16 @@ module Kontena::Cli::Stacks
 
           cmd << '--no-deploy'
           cmd << data[:local][:loader].source
-          if dry_run?
-            caret "Would install new dependency #{cmd.last} as #{stackname} without --dry-run", dots: false
-          else
-            caret "Installing new dependency #{cmd.last} as #{stackname}"
-            Kontena.run!(cmd)
-          end
+          caret "Installing new dependency #{cmd.last} as #{stackname}"
+          Kontena.run!(cmd)
         end
 
-        if deploy?
-          if dry_run?
-            caret "Would run stack deploy for #{stackname} without --dry-run", dots: false
-          else
-            Kontena.run!(['stack', 'deploy', stackname])
-          end
-        end
+        Kontena.run!(['stack', 'deploy', stackname]) if deploy?
       end
     end
 
     def update_stack(name, data)
-      if dry_run?
-        caret "Would send upgraded data to master for stack #{name} without --dry-run", dots: false
-      else
-        client.put(stack_url(name), data)
-      end
+      client.put(stack_url(name), data)
     end
 
     def stack_url(name)
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -80,8 +80,8 @@ module Kontena::Cli::Stacks
     # @param new_data [Hash] data from files
     # @return [Kontena::Cli::Stacks::ChangeRsolver]
     def process_data(old_data, new_data)
-      caret "Analyzing upgrade" # not a spinner, because the stack executes can create prompts
-      # Execute the local stacks
+      logger.debug { "Master stacks: #{old_data.keys.join(",")} YAML stacks: #{new_data.keys.join(",")}" }
+
       new_data.reverse_each do |stackname, data|
         reader = data[:loader].reader
         set_env_variables(stackname, current_grid) # set envs for execution time
@@ -97,7 +97,9 @@ module Kontena::Cli::Stacks
 
       set_env_variables(stack_name, current_grid) # restore envs
 
-      Kontena::Cli::Stacks::ChangeResolver.new(old_data, new_data)
+      spinner "Analyzing upgrade" do
+        Kontena::Cli::Stacks::ChangeResolver.new(old_data, new_data)
+      end
     end
 
     def display_report(changes)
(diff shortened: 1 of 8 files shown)


SpComb:
I think a `kontena stack upgrade --dry-run` would be a great idea, but as per the discussion in #2819 I think it would also need some kind of service-level diff. That just needs to take #2311 into account in terms of the subtleties around how `kontena stack upgrade --no-deploy` behaves.

As is, this is only really be useful for seeing how the dependent child stacks change, because for normal stacks, this would pretty much just always output `Would send upgraded data to master for stack #{name} without --dry-run`, which isn't that useful.

kke:
Now it just compares added/removed services and dependent stacks.

kke:
The `--dry-run` option in current master can't stay as it will actually do stuff, it was for debugging purposes and should have been removed.

kke:
@SpComb We need this or something else or the existing `--dry-run` will be dangerous if used accidentally.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -77,6 +82,66 @@ def merge_data(local_data, remote_data)
...
+        added_services.each { |svc| puts " - #{svc}" }
+      end
+
+      unless removed_services.empty?
+        dry_msg "The following services would be removed from master"
+        removed_services.each { |svc| puts " - #{svc}" }
+      end
+
+      unless kept_services.empty?
+        dry_msg "The following services would be kept:"
```
SpComb: Better to say `The following services would be upgraded`, because the services are not just kept, they can change

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -130,6 +195,11 @@ def execute
         abort_on_validation_errors(data[:local][:loader].reader.errors, data[:local][:loader].source)
       end
 
+      if dry_run?
+        dry_report(merged)
```
SpComb: Should this happen before the `unless removes.empty?` -> `merged.delete(removed_stack)`? ATM removing a child stack and doing `--dry-run` prompts for the stack removal, and then the `dry_report` never hits the `Stack #{stackname} would be removed from master` case:

```
kontena stack upgrade -v foo=asdf --dry-run deptest test/deptest.yaml 

Stacks to be removed because they are no longer depended on:
- deptest-foo

Warning: This can not be undone, data will be lost.
> Destructive command. You can skip this prompt by running this command with --force option. Are you sure? Yes
DRY RUN: The following services would be kept:
- deptest/redis
- deptest-bar/redis
```

I think it would be more friendly if `--dry-run` didn't prompt about destructive commands that will lose data...

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -77,6 +82,66 @@ def merge_data(local_data, remote_data)
...
+      end
+
+      merged.each do |stackname, data|
+        if data[:local].nil?
+          dry_msg "Stack  #{stackname} would be removed from master"
+          removed_services = data[:remote]['services'].map { |svc| "#{stackname}/#{svc['name']}" }
+        else
+          data[:local][:stack]['services'].each do |svc|
+            unless data[:remote]['services'].find { |s| s['name'] == svc['name'] }
+              added_services << "#{stackname}/#{svc['name']}"
```
SpComb: New services get reported twice, both in the first each's `elsif data[:local]` and this each's `else`:

```
DRY RUN: The following services would be added to master
 - deptest-foo/extra
 - deptest-foo/extra
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -77,6 +82,66 @@ def merge_data(local_data, remote_data)
...
+      merged.each do |stackname, data|
+        if data[:remote].nil?
+          dry_msg "New stack #{stackname} would be installed to master"
+          added_services = data[:local][:stack]['services'].map { |svc| "#{stackname}/#{svc['name']}" }
+        elsif data[:local]
+          data[:local][:stack]['services'].each do |svc|
+            remote_svc = data[:remote]['services'].find { |s| s['name'] == svc['name'] }
+            if remote_svc
+              kept_services["#{stackname}/#{svc['name']}"] ||= {}
+              #kept_services["#{stackname}/#{svc['name']}"][:before] = JSON.load(JSON.dump(remote_svc)) # naughty deep clone/dup
```
SpComb: Dead code?

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -77,6 +82,66 @@ def merge_data(local_data, remote_data)
...
+        end
+      end
+
+      merged.each do |stackname, data|
+        if data[:local].nil?
+          dry_msg "Stack  #{stackname} would be removed from master"
+          removed_services = data[:remote]['services'].map { |svc| "#{stackname}/#{svc['name']}" }
+        else
+          data[:local][:stack]['services'].each do |svc|
+            unless data[:remote]['services'].find { |s| s['name'] == svc['name'] }
```
SpComb: Broken for new stacks:

```
DRY RUN: New stack deptest-quux would be installed to master
NoMethodError: undefined method `[]' for nil:NilClass
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:113:in `block (2 levels) in dry_report'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:112:in `each'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:112:in `block in dry_report'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:107:in `each'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:107:in `dry_report'
  /home/kontena/kontena/kontena/cli/lib/kontena/cli/stacks/upgrade_command.rb:199:in `execute'
```

I suspect this block isn't even needed, because the `added_services` have already been collected earlier?

cli/lib/kontena/cli/stacks/upgrade_command.rb
kke: Yep it was there for simulation purposes. Needs some reworking.

kke:
@SpComb Complete reimplementation, ptal.

kke:
Btw, now asks for confirmation always, not sure if good.

kke:
> No, I don't think it makes sense to ask for confirmation if the stacks/services haven't changed at all, and are just getting upgraded. Particularly not with Destructive command.

The heavier `confirm_command` is used only when services or stacks are are going to be destroyed. If you explicitly `service rm` or `stack rm` you get the same heavy confirm too.

Maybe the `--dry-run` is actually not necessary if the command prompts every time if this was what you wanted, you can just ctrl-c your way out of it.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -19,152 +22,197 @@ class UpgradeCommand < Kontena::Command
...
+    def execute
+      set_env_variables(stack_name, current_grid)
 
-      depends.each do |stack|
-        key = "#{parent_name}-#{stack['name']}"
-        normalized_data.merge!(normalize_local_data(stack.merge('parent_name' => parent_name), key))
-      end
-      normalized_data
-    end
+      gather_data
```
SpComb: The use of `@old_data` and `@new_data` as pseudo-global variables bothers me. Why can't this method return something? It's so hard to follow the flow of data across these functions :confused:

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -19,152 +22,197 @@ class UpgradeCommand < Kontena::Command
...
+      return if dry_run?
 
-      normalized_data = { stack_name => data }
+      get_confirmation
 
-      return normalized_data if skip_dependencies?
+      run_removes
+      run_installs
+      run_upgrades
+      run_deploys if deploy?
```
SpComb: I wonder what the resulting order of parent/child stack remove/install/upgrade/deploy operations here is now, because it's slightly different than the current situation... also not entirely sure what it should be, or if there even is any easily defined correct ordering. If the parent stack services link to any of the child stack services, then the ordering will actually matter, like:

* removing a child stack + the link from the parent stack service to the child stack service => the child stack remove will fail until the parent stack is updated *and* deployed to remove the linking
* adding a child stack + linking to it from a parent service => the child stack install must run before the parent stack upgrade

SpComb:
>> No, I don't think it makes sense to ask for confirmation if the stacks/services haven't changed at all, and are just getting upgraded. Particularly not with Destructive command.
> 
> The heavier confirm_command is used only when services or stacks are are going to be destroyed. If you explicitly service rm or stack rm you get the same heavy confirm too.

It makes sense to prompt when removing a child stack, and it might even make sense to start prompting when removing services, but the way the `get_confirmation` is right now, it always prompts every single time you use the `upgrade` command. That just deflates the value of the confirm prompt, everyone will learn to just automatically hit enter without even reading it, and it becomes pointless. 

```
$ bin/kontena stack install test/redis.yaml 
> Redis version : 3.0
 [done] Creating stack redis      
 [done] Triggering deployment of stack redis     
$ bin/kontena stack upgrade redis test/redis.yaml 
 [done] Reading stack redis from master     
 [done] Parsing test/redis.yaml     
> Analyzing upgrade ..
> Redis version : 3.0
SERVICES:
These services will be upgraded:
- redis/redis

STACKS:
These stacks will be upgraded and deployed:
- redis


> Destructive command. You can skip this prompt by running this command with --force option. Are you sure? (Y/n) 
```

The `Destructive command` prompt isn't right.

cli/lib/kontena/cli/stacks/upgrade_command.rb
kke: That's kind of what instance variables are for. Methods that do nothing to/with instance state are marked by [Reek](https://github.com/troessner/reek) as [Utility Functions](https://github.com/troessner/reek/blob/master/docs/Utility-Function.md) and are considered a code smell.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -19,152 +22,211 @@ class UpgradeCommand < Kontena::Command
...
+
+      get_confirmation(changes)
+
+      run_removes(changes.removed_stacks)
+
+      deployable_stacks = []
+      deployable_stacks.concat run_installs(changes)
+      deployable_stacks.concat run_upgrades(changes)
+
+      run_deploys(deployable_stacks) if deploy?
```
SpComb: I suspect that the new ordering of the remove -> install -> upgrade -> deploy operations here isn't as good as the current order of operations. Running the installs and upgrades separately before running all of the deploys as a separate step at the end means that upgrades of each stack might not actually see the installed/upgraded versions of the earlier stacks.

I think the best way would be to install/upgrade + deploy each child stack in the same order as defined in the YAML, before upgrading the parent stack. Any old stacks should then probably be removed at the end.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -19,152 +22,211 @@ class UpgradeCommand < Kontena::Command
...
-        end
+      result
+    end
+
+    # Preprocess data and return a ChangeResolver
+    # @param old_data [Hash] data from master
+    # @param new_data [Hash] data from files
+    # @return [Kontena::Cli::Stacks::ChangeRsolver]
+    def process_data(old_data, new_data)
+      caret "Analyzing upgrade" # not a spinner, because the stack executes can create prompts
```
SpComb: Surprising ordering here, because the `Analyzing upgrade` comes before the stack variable prompts. It can't analyze the upgrade if it doesn't have the new stack variable values yet?

```
> Analyzing upgrade ..
> Redis version : (3.0) 
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -19,152 +22,211 @@ class UpgradeCommand < Kontena::Command
...
+      end
+
+      unless changes.added_stacks.empty?
+        puts pastel.red("These new stack dependencies #{will} be installed:")
+        changes.added_stacks.each { |stack| puts pastel.red("- #{stack}") }
+        puts
+      end
+
+      unless changes.upgraded_stacks.empty?
+        puts pastel.cyan("These stacks #{will} be upgraded#{' and deployed' if deploy?}:")
```
SpComb: Should this be skipped if it's a solo-stack without any child stacks dependencies? It's maybe extra noise in that case.

```
$ kontena stack upgrade redis test/redis.yaml 
 [done] Reading stack redis from master     
 [done] Parsing test/redis.yaml     
> Analyzing upgrade ..
> Redis version : 3.0
SERVICES:
These services will be upgraded:
- redis/redis

STACKS:
These stacks will be upgraded and deployed:
- redis


 [done] Upgrading stack redis     
 [done] Triggering deployment of stack redis     
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
kke: Could be. I separated the deploy because it seemed like a good idea to install everything properly first and then trigger the deploys, but I'm not any sort of wizard with the server/agent stuff.

kke: Running the removes last certainly sounds like a good idea.

PR #2831, opened 2017-09-15
Add cloud plugin to installers

jakolehm:


cli/omnibus/config/software/kontena-cli.rb
```diff
@@ -11,5 +11,6 @@
   gem "install rb-readline -v 0.5.4 --no-ri --no-doc"
   gem "install nokogiri -v 1.6.8 --no-ri --no-doc"
   gem "install kontena-cli -v #{default_version} --no-ri --no-doc"
+  gem "install kontena-plugin-cloud -v #{default_version} --no-ri --no-doc"
```
SpComb: Not sure about the `-v #{default_version}`... the `kontena-plugin-cloud` gem version does not match the Kontena version?

jakolehm: Fixed

PR #2833, opened 2017-09-17
Redis as an alternative pubsub backend

jakolehm:
This PR refactors `MongoPubsub` to two separate classes, `Pubsub::Base` (backend agnostic logic) and `Pubsub::Mongo`. Also adds Redis support via `Pubsub::Redis`. Redis is used automatically if `REDIS_URL` environment variable is set.


server/app/services/pubsub/base.rb
```diff
@@ -0,0 +1,139 @@
...
+        while data = @queue.shift
+          send_message(data)
+        end
+      end
+
+      private
+
+      # @param [Hash] data
+      def send_message(data)
+        payload = HashWithIndifferentAccess.new(data)
```
jakolehm: Changed this a bit. Previously the passed data was not actually a hash (it was `BSON::Binary`)... it seemed like a bug because it also unpacked binary for each subscriber. Now unpack happens in mongo/redis backends.

server/app/jobs/leader_elector_job.rb
```diff
@@ -43,11 +41,12 @@ def elect
   end
 
   def announce_election
-    MongoPubsub.publish(PUBSUB_KEY, {})
+    self.elect
```
SpComb: Subtle changes to the `LeaderElectorJob` are a little scary to review :fearful:

jakolehm: Indeed, this should not be in this pr.. will cleanup.

jakolehm:
Story continues in #3178

PR #2850, opened 2017-09-21
Add cert delete API and CLI command

### LINKED ISSUE #2849: Missing kontena certificate delete
There is no `kontena certificate delete` command to have the server forget about a certificate that is no longer in use by any services.

This is actually important, because the server certificate autorenewal could end up chewing through the LetsEncrypt rate limits with a lot of old, unused certificates.

jnummelin:
fixes #2849

### DIFF
--- a/server/app/mutations/grid_certificates/remove_certificate.rb
+++ b/server/app/mutations/grid_certificates/remove_certificate.rb
@@ -1,0 +1,23 @@
+
+require_relative 'common'
+require_relative '../../services/logging'
+
+module GridCertificates
+  class RemoveCertificate < Mutations::Command
+    include Common
+    include Logging
+
+    required do
+      model :certificate, class: Certificate
+    end
+
+    def validate
+      services_using = certificate.grid.grid_services.where(:'certificates.subject' => certificate.subject).map { |s| s.to_path}
+      add_error(self.certificate.subject, :certificate_in_use, "Certificate still in use in services: #{services_using.join(',')}") if services_using.size > 0
+    end
+
+    def execute
+      self.certificate.destroy
+    end
+  end
+end
--- a/server/app/mutations/grid_certificates/remove_certificate.rb
+++ b/server/app/mutations/grid_certificates/remove_certificate.rb
@@ -13,7 +13,7 @@ module GridCertificates
 
     def validate
       services_using = certificate.grid.grid_services.where(:'certificates.subject' => certificate.subject).map { |s| s.to_path}
-      add_error(self.certificate.subject, :certificate_in_use, "Certificate still in use in services: #{services_using.join(',')}") if services_using.size > 0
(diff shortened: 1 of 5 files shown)


server/app/routes/v1/certificates_api.rb
```diff
@@ -73,6 +73,20 @@ def register(data)
           end
         end
 
+        r.delete do
+          r.on ':subject' do |subject|
+            @certificate = @grid.certificates.find_by(subject: subject)
+            if @certificate
+              @certificate.destroy
```
SpComb: This needs to be a mutation that checks that the certificate is not in use by any services.

jnummelin: true.

jnummelin: fixed

server/app/mutations/grid_certificates/remove_certificate.rb
```diff
@@ -0,0 +1,23 @@
...
+    include Common
+    include Logging
+
+    required do
+      model :certificate, class: Certificate
+    end
+
+    def validate
+      services_using = certificate.grid.grid_services.where(:'certificates.subject' => certificate.subject).map { |s| s.to_path}
+      add_error(self.certificate.subject, :certificate_in_use, "Certificate still in use in services: #{services_using.join(',')}") if services_using.size > 0
```
SpComb: Bad idea to use the actual subject as the error key, it should be something constant like `:certificate` instead.

```
 [fail] Removing certificate for test.188.226.131.23.xip.io from development grid      
	test:
	  '188':
	    '226':
	      '131':
	        '23':
	          xip:
	            io: 'Certificate still in use in services: development/ingress-lb/lb'
```

jnummelin: oh, totally forgot the dotted functionality there.

SpComb: Replace `add_error(self.certificate.subject, ...)` with just `add_error(:certificate, ...)`

server/spec/api/v1/certificates_spec.rb
```diff
@@ -86,5 +84,30 @@
...
+        expect(response.status).to eq(200)
+      }.to change{Certificate.count}.by (-1)
+
+    end
+
+    it 'fails deleting certificate as it\'s in use' do
+      GridService.create!(grid: grid, name: 'redis', image_name: 'redis', certificates: [GridServiceCertificate.new(subject: 'kontena.io', name: 'SSL_CERT')])
+      expect {
+        delete "/v1/certificates/#{grid.name}/kontena.io", nil, request_headers
+        expect(response.status).to eq(422)
```
SpComb: Wouldn't hurt to also `expect(json_response['error']).to eq ....` here

jnummelin:
fixed the error key

PR #2856, opened 2017-09-22
CLI: Allow referencing child stack variables from parent

kke:
Makes it possible to reference child stack variable values from a parent stack.

After this PR, something like this can be done:

```yaml
stack: user/stack
version: 0.1.0
depends:
  mysql:
    stack: zontena/mysql

services:
  app:
     image: myawesomeapp
     environment:
       - "DB_USERNAME=${mysql.username}" # <--- !!!
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -19,6 +19,7 @@ module Kontena::Cli::Stacks
 
     option '--parent-name', '[PARENT_NAME]', "Set parent stack name", hidden: true
     option '--skip-dependencies', :flag, "Do not install any stack dependencies"
+    option '--dry-run', :flag, "Simulate install"
 
     requires_current_master
     requires_current_master_token
@@ -27,6 +28,7 @@ module Kontena::Cli::Stacks
       set_env_variables(stack_name, current_grid)
 
       values_from_installed_stacks.merge!(install_dependencies) unless skip_dependencies?
+
 
       stack # runs validations
 
@@ -55,6 +57,8 @@ module Kontena::Cli::Stacks
           cmd.concat ['-v', "#{key}=#{value}"]
         end
 
+        cmd << '--dry-run' if dry_run?
+
         cmd << '--no-deploy' unless deploy?
 
         cmd << dependency['stack']
@@ -68,14 +72,18 @@ module Kontena::Cli::Stacks
     end
 
     def create_stack
+      return if dry_run?
       spinner "Creating stack #{pastel.cyan(stack['name'])} " do
         client.post("grids/#{current_grid}/stacks", stack)
       end
-      stack
     end
 
     def deploy_stack
-      Kontena.run!(['stack', 'deploy', stack['name']])
+      if dry_run?
+        caret "Stack #{stack['name']} deploy would be triggered", dots: false
+      else
+        Kontena.run!(['stack', 'deploy', stack['name']])
+      end
     end
   end
 end
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -26,7 +26,7 @@ module Kontena::Cli::Stacks
     def execute
       set_env_variables(stack_name, current_grid)
 
-      install_dependencies unless skip_dependencies?
+      values_from_installed_stacks.merge!(install_dependencies) unless skip_dependencies?
 
       stack # runs validations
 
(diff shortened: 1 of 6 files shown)


cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -36,12 +38,16 @@ def execute
       dump_variables if values_to
 
       create_stack
+
       deploy_stack if deploy?
+
+      stack
     end
 
     def install_dependencies
```
SpComb: `@return [Hash{String => Hash}] child stack variables`

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -19,14 +19,16 @@ class InstallCommand < Kontena::Command
...
+    option '--dry-run', :flag, "Simulate install"
 
     requires_current_master
     requires_current_master_token
 
     def execute
       set_env_variables(stack_name, current_grid)
 
-      install_dependencies unless skip_dependencies?
+      values_from_installed_stacks.merge!(install_dependencies) unless skip_dependencies?
```
SpComb: This would be easier to read with a named local variable. Something like:

```
unless skip_dependencies?
  values_from_dependencies = install_depenencies
  values_from_installed_stacks.merge!(values_from_dependencies)
end
```

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -51,21 +57,33 @@ def install_dependencies
...
         end
 
+        cmd << '--dry-run' if dry_run?
+
         cmd << '--no-deploy' unless deploy?
 
         cmd << dependency['stack']
-        Kontena.run!(cmd)
+        dependency_result = Kontena.run!(cmd)
+        dependency_result_variables = dependency_result['variables'] ||= {}
```
SpComb: The `foo_bar = foo['bar'] ||= {}` is a little surprising, not sure how the `||=` assignment actually behaves there.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -366,7 +378,7 @@ def interpolate(content, use_opto: true, substitutions: {}, raise_on_unknown: fa
           if row.strip.start_with?('interpolate:') || row.strip.start_with?('evaluate:')
             row
           else
-            row.gsub(/(?<!\$)\$(?!\$)\{?\w+\}?/) do |v| # searches $VAR and ${VAR} and not $$VAR
+            row.gsub(/(?<!\$)\$(?!\$)\{?(?:\w|\.)+\}?/) do |v| # searches $VAR and ${VAR} and not $$VAR
```
SpComb: What happens to `foo=$GRID.kontena.local`? Would these interpolations have worked earlier, but would now fail?

cli/lib/kontena/cli/stacks/yaml/stack_file_loader/raw_content_loader.rb
```diff
@@ -0,0 +1,32 @@
+module Kontena::Cli::Stacks
+  module YAML
+    class RawContentLoader < StackFileLoader
+      # For loading from a string directly, probably mostly useful in specs
```
SpComb: Not entirely sure if this is a good idea.

SpComb:
Not sure if creating opto variables that are literally called `foo.bar` is the right design here.

ATM all of the child stack variables even end up getting POST'd to the server as `variables` for the top-level stack, which fails:

```
BSON::String::IllegalKey ('foo.peers' is an illegal key in MongoDB. Keys may not start with '$' or contain a '.'.):
  /usr/lib/ruby/gems/2.4.0/gems/bson-4.2.1/lib/bson/string.rb:69:in `to_bson_key'
...
  /app/app/mutations/stacks/create.rb:64:in `execute'
```

cli/lib/kontena/cli/stacks/install_command.rb
kke: It's a typo but it does work, it sets both variables unless the middle one is already defined.

```
> foo = bar ||= 1
> foo 
=> 1
> bar 
=> 1
```

cli/lib/kontena/cli/stacks/yaml/reader.rb
kke: Yep, those would have worked but after this will not. Perhaps `$foo[foo]` / `${foo[foo]}` would be safer. Also not entirely sure what happens in liquid with this dot notation.

Implementing groups in opto would probably help a bit. https://github.com/kontena/opto/issues/39

kke: One option could be to only allow this in Liquid.

PR #2862, opened 2017-09-26
Docs: Add stack depends parent/children fields to API documentation

kke:
Adds the new parent and children fields to the stack JSON response documentation.

### DIFF
--- a/server/docs/source/index.html.md
+++ b/server/docs/source/index.html.md
@@ -688,7 +688,8 @@ Accept: application/json
     "stack": "my/redis",
     "version": "0.1.0",
     "registry": "file://",
-    "services": []
+    "services": [],
+    "parent_name": "parent-stack-name"
 }
 ```
 
--- a/server/docs/source/index.html.md
+++ b/server/docs/source/index.html.md
@@ -582,6 +582,10 @@ to | The end date and time (example: `?to=2017-01-01T13:15:00.00Z`) | now
   "version": "0.1.0",
   "registry": "https://stack-registry.kontena.io",
   "expose": "peer",
+  "parent": { "name": "parent-stack-name" },
+  "children": [
+    { "name": "child-stack-name" }
+  ],
   "services": [
       {
           "name": "arbiter",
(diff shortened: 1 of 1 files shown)


server/docs/source/index.html.md
```diff
@@ -582,6 +582,10 @@ to | The end date and time (example: `?to=2017-01-01T13:15:00.00Z`) | now
   "version": "0.1.0",
   "registry": "https://stack-registry.kontena.io",
   "expose": "peer",
+  "parent": { "name": "parent-stack-name" },
+  "children": [
+    { "name": "child-stack-name" }
```
jakolehm: Should parent/children objects also have a stack id?

kke: Currently it goes like:

```ruby
json.parent stack.has_parent? ? { name: stack.parent_name } : nil
json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
```

Does the id have any significance? I don't think you can't use it for anything in the API, the `stack.to_path` which is used as "id" in the docs comes from `"#{self.grid.try(:name)}/#{self.name}"`, so the JSON could just become:

```ruby
json.parent stack.has_parent? ? { name: stack.parent_name, id: "#{stack.grid.try(:name)}/#{stack.parent_name}" } : nil
json.children stack.children.pluck(:name).map { |child_name| { id: "#{stack.grid.try(:name)}/#{child_name}", name: child_name } }
```

SpComb:
Waiting for #2866 (not RC-critical)

kke:
Updated for #2866

server/docs/source/index.html.md
```diff
@@ -582,6 +582,16 @@ to | The end date and time (example: `?to=2017-01-01T13:15:00.00Z`) | now
   "version": "0.1.0",
   "registry": "https://stack-registry.kontena.io",
   "expose": "peer",
+  "parent": {
+    "name": "parent-stack-name",
+    "id": "my-grid/parent-stack-name"
```
SpComb: The `id` can also be `null` if it's an orphaned child stack, and the parent is missing. It will still have the `name` in that case.

This can easily happen as a result of failed `kontena stack install` commands, where the child stack is fine, but the parent stack fails to validate/install.

PR #2864, opened 2017-09-26
Remove experimental status from volumes

### LINKED ISSUE #2857: Remove experimental status from volume commands

jakolehm:
Fixes #2857


SpComb:
This mainly means that we must implement any future volume changes like automatic affinities in a backwards-compatible way re the current scopes.

jnummelin:
> This mainly means that we must implement any future volume changes like automatic affinities in a backwards-compatible way re the current scopes.

Yes.

PR #2866, opened 2017-09-27
Add ID field to stack dependency relations in JSON response

kke:
Ref #2862

Adds the "id" field to stack parent/child hash in stacks api JSON response.

### DIFF
--- a/server/app/views/v1/stacks/_stack.json.jbuilder
+++ b/server/app/views/v1/stacks/_stack.json.jbuilder
@@ -16,11 +16,13 @@ json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
 
-json.parent do
-  if stack.has_parent?
+if stack.has_parent?
+  json.parent do
     json.id stack.parent.to_path
     json.name stack.parent.name
   end
+else
+  json.parent nil
 end
 
 json.children(stack.children) do |child_stack|
--- a/server/app/views/v1/stacks/_stack.json.jbuilder
+++ b/server/app/views/v1/stacks/_stack.json.jbuilder
@@ -15,5 +15,15 @@ json.variables latest_rev.variables
 json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
-json.parent stack.has_parent? ? { name: stack.parent_name } : nil
-json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
+
+json.parent do
+  if stack.has_parent?
+    json.id stack.parent.to_path
+    json.name stack.parent.name
+  end
+end
+
+json.children(stack.children) do |child_stack|
+  json.id child_stack.to_path
+  json.name child_stack.name
+end
--- a/server/spec/api/v1/stacks_spec.rb
+++ b/server/spec/api/v1/stacks_spec.rb
@@ -89,7 +89,7 @@ describe '/v1/stacks', celluloid: true do
 
     context 'nested stacks' do
 
-      let(:child_stack_1) do
+      let!(:child_stack_1) do
         outcome = Stacks::Create.run(
           current_user: david,
           grid: grid,
@@ -108,7 +108,7 @@ describe '/v1/stacks', celluloid: true do
         outcome.result
       end
 
-      let(:child_stack_1_2) do
+      let!(:child_stack_1_2) do
         outcome = Stacks::Create.run(
           current_user: david,
           grid: grid,
@@ -127,7 +127,7 @@ describe '/v1/stacks', celluloid: true do
         outcome.result
       end
 
-      let(:child_stack_2) do
+      let!(:child_stack_2) do
         outcome = Stacks::Create.run(
           current_user: david,
           grid: grid,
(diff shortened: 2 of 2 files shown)


server/app/views/v1/stacks/_stack.json.jbuilder
```diff
@@ -15,5 +15,5 @@ json.variables latest_rev.variables
 json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
-json.parent stack.has_parent? ? { name: stack.parent_name } : nil
-json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
+json.parent stack.has_parent? ? { name: stack.parent_name, id: "#{stack.grid.try(:name)}/#{stack.parent_name}" } : nil
+json.children stack.children.pluck(:name).map { |child_name| { name: child_name, id: "#{stack.grid.try(:name)}/#{child_name}" } }
```
SpComb: Too much oneliner. Look at how the other structured JSON fields are implemented, these should be something like

```ruby
json.parent do
  if parent_stack = stack.parent_stack
    json.id parent_stack.to_path
    json.name parent_stack.name
  end
end
json.children stack.children do |child_stack|
  json.id child.to_path
  json.name child.name
end
```

kke: Seems wrong to create so many extra queries

1) find and load the parent stack to get its name (which is already known at this point)
2) find and load the parent stack grid to get its name (which is already known at this point)
3) find and load the complete child stacks (while we only want the name)
4) find and load the grid for each child to get the same name several times (which is already known at this point)

server/app/views/v1/stacks/_stack.json.jbuilder
```diff
@@ -15,5 +15,11 @@ json.variables latest_rev.variables
 json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
-json.parent stack.has_parent? ? { name: stack.parent_name } : nil
-json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
+
+json.parent do
+  json.partial! 'app/views/v1/stacks/parent', stack: stack
```
SpComb: I don't think this particularly needs to be a `partial!`, it's only ever used in one place, so this just makes this harder to read because you have to open up the separate partial files as well now.

server/app/views/v1/stacks/_stack.json.jbuilder
SpComb: It's an optimization for this specific case, yes... in general, the relations should really be actual mongoid relations, because then `@grid.stacks.includes(:parent_stack, :child_stacks)` should AFAIK be enough to allow writing the straightforward code while avoiding most of the O(N) queries for the `/v1/grids/:grid/stacks` case. The view code would get really ugly if we started to optimize everything like this.

Although it looks like the current `includes(:grid, :grid_services)` isn't enough to optimize out the references in the included documents, as the current code leads to further `{"find"=>"grids", "filter"=>{"_id"=>BSON::ObjectId('580878e5de35781b2c000017')}, "limit"=>1, "singleBatch"=>true}` and `{"find"=>"stacks", "filter"=>{"_id"=>BSON::ObjectId('59c8ffb942132f000ff5fed5')}` queries for each stack. That's presumably for the `stack.grid_services[...].grid/stack` references, and I don't know if mongoid is smart enough to optimize for something like `includes(:grid_services.grid)`?

server/app/views/v1/stacks/_stack.json.jbuilder
```diff
@@ -15,5 +15,6 @@ json.variables latest_rev.variables
 json.services stack.grid_services.to_a do |grid_service|
   json.partial! 'app/views/v1/grid_services/grid_service', grid_service: grid_service
 end
-json.parent stack.has_parent? ? { name: stack.parent_name } : nil
-json.children stack.children.pluck(:name).map { |child_name| { name: child_name } }
+
+json.parent stack.parent.try(:to_relation)
+json.children stack.children.map(&:to_relation)
```
SpComb: I think we want to have the views match the resulting JSON API schema as explicitly as possible, and thus avoid defining parts of the JSON schema in the models. It's more important to be able to see what the JSON API schema is from reading the views, even at the cost of a couple lines of code duplication for the `name` and `id`. IOW this should be something like in https://github.com/kontena/kontena/pull/2866#discussion_r141277250

AFAIK there's only a couple places in the views where the structure of the JSON depends on what's in the models, like the grid `{ "stats": ... }`, and even those are already sometimes a pain.

SpComb:
BTW using a partial for the `parent` and `children` would also make sense, but then it should be something like `app/views/v1/services/relation`. The same partial should be used for both the parent and child stacks. There might even be other places in the views that could use some similar partial, like the domain authorization `linked_service`.

server/app/views/v1/stacks/_stack.json.jbuilder
```diff
@@ -15,5 +15,17 @@ json.variables latest_rev.variables
...
+
+json.parent do
+  if stack.has_parent?
+    json.id stack.parent.to_path
+    json.name stack.parent.name
+  end
+end
+
+json.children do
+  stack.children.map do |child_stack|
```
SpComb: You can't use `json.*` from within `map`, it's a DSL function with side-effects, this is results in either the `children` being completely missing from the JSON response, or a single JSON object like `{"children":{"id":"development/deptest-foo","name":"deptest-foo"}}`.

Look at how the jbuilder thing for arrays works for e.g. the similar `GridService.secrets` case: https://github.com/kontena/kontena/blob/11e97b7c6803510ae71c750807697beaa7ca8867/server/app/views/v1/grid_secrets/show.json.jbuilder#L5-L8

PR #2867, opened 2017-09-27
List auto-renewable certificates; Migrate existing LE cert secrets to new certificate models

### LINKED ISSUE #2848: kontena certificate list should indicate which certificates are eligible to be autorenewed by the server
Only certificates where all of the certificate domains have authorizations of type `tls-sni-01` can be autorenewed by the server. Certificates with `dns-01` authorizations cannot be auto-renewed, and this should be apparent in the `kontena certificate list` output.

https://github.com/kontena/kontena/blob/6b134c6f96d44ff05fe0056f188d8979f20a62f6/server/app/jobs/certificate_renew_job.rb#L47-L57

jnummelin:
Fixes  #2848

Pre 1.4 era, LE certs are stored as secrets. In 1.4 we introduce new model to store certs. This PR adds a migration that collects all LE certs from secrets and transforms those into new cert models.

Added also flag on API & CLI to show whether or not a given cert is automatically renewable. Without that people might get the impression that all LE certs could be automatically renewed.

The included fixture cert is a valid LE cert but the private key is mangled so should not pose any issues. Anyone should not be able to use it without the working key.


cli/lib/kontena/cli/certificate/list_command.rb
```diff
@@ -15,7 +15,7 @@ class ListCommand < Kontena::Command
     THREE_DAYS = 3 * 24 * 60 * 60
 
     def fields
-      quiet? ? ['subject'] : {subject: 'subject', "expiration" => 'expires_in'}
+      quiet? ? ['subject'] : {subject: 'subject', "expiration" => 'expires_in', auto_renewable?: 'auto_renewable'}
```
SpComb: The exact formatting here might not be the best, but it works. We can fix it later if it's too ugly:

```
SUBJECT                        EXPIRATION   AUTO_RENEWABLE?
⊛ test.188.226.131.23.xip.io   77 days      true
⊛ test.188.226.131.23.nip.io   67 days      false
```

jakolehm: `AUTO_RENEWABLE?` this looks a bit ugly but we can fix it later.

PR #2884, opened 2017-09-29
Add --use-defaults to stack install/upgrade/build

### LINKED ISSUE #2406: Add a switch to stack upgrade to not prompt values that have defaults

kke:
Fixes #2406
Replaces #2407

Adds `--use-defaults` flag to `kontena stack install`, `kontena stack upgrade` and `kontena stack build` commands.

When used, any variable default values (or previous values coming from the master when performing an upgrade) will be used as variable values. Can be used to avoid prompts when you have defined defaults for variables.

### DIFF
--- a/cli/lib/kontena/cli/stacks/build_command.rb
+++ b/cli/lib/kontena/cli/stacks/build_command.rb
@@ -16,6 +16,7 @@ module Kontena::Cli::Stacks
     option ['--[no-]sudo'], :flag, 'Run docker using sudo', hidden: Kontena.on_windows?, environment_variable: 'KONTENA_SUDO', default: false
 
     option ['-n', '--name'], 'NAME', 'Define stack name (by default comes from stack file)'
+    option '--use-defaults', :flag, "Use defaults for all variable prompts"
 
     include Common::StackValuesToOption
     include Common::StackValuesFromOption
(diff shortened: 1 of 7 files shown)


SpComb:
An e2e spec for `kontena stack upgrade` to not prompt for anything would be nice.

jakolehm:
Does this skip prompts for new variables (with defaults) during upgrade?

SpComb:
What if there's a chain of other resolvers, not just `from: prompt`? Picking the first example from our docs:

```yaml
variables:
  mysql_root_pw:
    type: string
    from:
      prompt: Enter a root password for MySQL or leave empty to auto generate
      random_string: 16
```

You don't want to use the "default" value in that case (there isn't a default password....); you want to skip the `prompt` resolver and use the `random_string` resolver.

Should this be `--no-prompt` instead of `--use-defaults`? For a variable with a `default` and `from: prompt`, wouldn't that also result in the variable using the default value?

kke:
We could add `exit_with_error 'Not a tty terminal' unless $stdin.tty?` to prompt_resolver

kke:
> Should this be --no-prompt instead of --use-defaults? For a variable with a default and from: prompt, wouldn't that also result in the variable using the default value?

Yes.

So `--no-prompt` could just disable prompt_resolver (or remove the `from: prompt` from the options).

SpComb:
Closing in favor of #2897, which is a better design for fixing this issue.

PR #2897, opened 2017-10-03
Add --no-prompt for disabling prompts in stack commands

### LINKED ISSUE #2406: Add a switch to stack upgrade to not prompt values that have defaults

kke:
Replaces #2884
Replaces #2407
Fixes #2406

Adds a `--no-prompt` option to `kontena stack [install|upgrade|validate|build]` which will disable any prompting. If there is a default value, it will be used instead of prompting, this happens when running an upgrade to a stack that has already been installed.

**Note:** If your stack relied on a default value coming from the prompt during an upgrade, after this it will break.

For example, something like:

```yaml
variables:
  pass:
    type: string
    from:
      prompt: Enter password
      random_string: 16
services:
  app:
    image: app
    environment:
      - "SECRET_PASSWORD=$pass"
```

will always generate a new random string during an upgrade when `--no-prompt` is used.

You should use vault instead:

```yaml
variables:
  pass:
    type: string
    from:
      vault: APP_PASSWORD
      prompt: Enter password
      random_string: 16
    to:
      vault: APP_PASSWORD

services:
  app:
    image: app
    secrets:
      - name: SECRET_PASSWORD
        secret: APP_PASSWORD
        type: env
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -180,8 +180,9 @@ module Kontena::Cli::Stacks
 
       # Removes any `from: prompt` entries from variables
       def disable_prompts
+        prompts = Set.new(%w(prompt service_link vault_cert_prompt))
         variables.each do |option|
-          option.from.delete_if { |k,_| k.to_s == 'prompt' }
+          option.from.delete_if { |k,_| prompts.include?(k.to_s) }
         end
       end
 
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -178,6 +178,13 @@ module Kontena::Cli::Stacks
         variables.build_option(name: 'PARENT_STACK', type: :string, value: parent_name)
       end
 
+      # Removes any `from: prompt` entries from variables
+      def disable_prompts
+        variables.each do |option|
+          option.from.delete_if { |k,_| k.to_s == 'prompt' }
+        end
+      end
+
       # @return [Boolean] did this stack come from a local file?
       def from_file?
         loader.origin == 'file'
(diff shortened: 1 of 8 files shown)


SpComb:
Doesn't work for `kontena stack upgrade`, the `UpgradeCommand#process_data` calls `reader.execute` directly, which still prompts.

SpComb:
Unfortunately this now breaks the very first example in our [Stack File Variables](https://www.kontena.io/docs/references/stack-file-variables.html) docs with `from: prompt -> random_string`, because it relies on the `prompt` resolver using the default value to keep the same password across stack upgrades:

```
$ kontena stack upgrade --no-prompt mysql test/prompts.yaml 
 [done] Reading stack mysql from master     
 [done] Parsing test/prompts.yaml     
> Enter a root password for MySQL or leave empty to auto generate : 4jhz9T0ayAH7jyDt
 [done] Analyzing upgrade     
 [done] Upgrading stack mysql     
 [done] Triggering deployment of stack mysql     
 [done] Deploying service mysql     
```

With `kontena stack upgrade --no-prompt`, the prompt with the default value gets skipped, and the password gets reset from `random_string` each time...

TBH I think this just means that the example is broken, it should really be using `to/from: vault` instead, with the `from: vault' before the `from: random_string`?

kke:
Yep the example is a bit stupid.

kke:
We could have a `default_resolver` which you could put before the random_string. :D 

```yaml
from:
  prompt: Enter stuff
  default: true
  random_string: 16
```

SpComb:
> Unfortunately this now breaks the very first example in our Stack File Variables docs with from: prompt -> random_string, because it relies on the prompt resolver using the default value to keep the same password across stack upgrades:

Looking at  #2884 vs #2897 (this), maybe the right way to handle this is to have `--no-prompt` preserve the "use default as value" behavior for just hitting enter at the prompt to accept the default value. For a stack install, the default value may be empty, and it will continue to the next resolver. For a stack upgrade, there will be a default value, so that will get used. That would best match the existing conventions for how `prompt` is used re defaults / chained resolvers?

> We could have a default_resolver which you could put before the random_string. :D

Doesn't that already exist? :P https://github.com/kontena/opto/blob/master/lib/opto/resolvers/default.rb

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -178,6 +178,14 @@ def create_parent_variable(parent_name)
         variables.build_option(name: 'PARENT_STACK', type: :string, value: parent_name)
       end
 
+      # Removes any `from: prompt` entries from variables
+      def disable_prompts
+        prompts = Set.new(%w(prompt service_link vault_cert_prompt))
+        variables.each do |option|
+          option.from.delete_if { |k,_| prompts.include?(k.to_s) }
```
SpComb: I think this counts as a hack.

I mean, something (?) like `option.from.each { |k, v| option.from[k] = DefaultResolver.new if prompts.include?(k.to_s) }` would work, but it's ugly...

kke: The defaults injection and value setting happen in the same manner. I don't see any better alternative except make some kind of global state to prompt resolver, which is not good either. Or implement this in opto like `group.new(disabled: [PrompResolver])`

kke:
Replaced by `--reuse-values` in #3185

PR #2899, opened 2017-10-03
More descriptive error message when stack YAML has no "stack:"

kke:
When the stack YAML has no `stack: stackname` the current error message is something like: `[error] RuntimeError : Variable validation failed: {"STACK"=>{:presence=>"Required value missing"}}`

This PR changes that to: `The file redis.yml does not have the required top level keyword 'stack:'`

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/stack_file_loader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/stack_file_loader.rb
@@ -64,7 +64,9 @@ module Kontena::Cli::Stacks
 
       # @return [StackName] an accessor to StackName for the target file
       def stack_name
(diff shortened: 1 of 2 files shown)


SpComb:
Shouldn't this be in something like the `Kontena::Cli::Stacks::YAML::ValidatorV3`, or is that too late?

kke:
Yep, too late. The `stack_name` is needed before validation.

SpComb:
Are there any side-effects of running the `#execute` method completely, rather than raising in the case of variable validation errors? Do things like vault setters happen before or after that?

kke:
I don't think it should change that, it's the same `valid` and `errors` calls to opto that happen here.

PR #2902, opened 2017-10-04
Make travis gem deploy idempotent

SpComb:
Allow restarting the travis CLI deploy job to retry omnibus builds by skipping the gem deploy if the gem has already been deployed.

### For an already released version
```
+ cat ./VERSION
+ VERSION=1.4.0.pre12
+ echo Checking for deployed kontena-cli gem version 1.4.0.pre12...
Checking for deployed kontena-cli gem version 1.4.0.pre12...
+ gem fetch -v 1.4.0.pre12 kontena-cli
+ tee /tmp/gem-fetch.log
+ grep -q ERROR:  Could not find a valid gem 'kontena-cli' (= 1.4.0.pre12) in any repository
+ test -e ./kontena-cli-1.4.0.pre12.gem
+ echo gem version 1.4.0.pre12 already deployed
gem version 1.4.0.pre12 already deployed
+ ls -l kontena-cli-1.4.0.pre12.gem
-rw-rw-r-- 1 kontena kontena 193536 loka   4 12:21 kontena-cli-1.4.0.pre12.gem
+ sha256sum kontena-cli-1.4.0.pre12.gem
c0e14ee7c822befde94fc8be1f3192f72b605f5814d9db328a6351eb53de370a  kontena-cli-1.4.0.pre12.gem
+ exit 0
```

### For an unreleased version
```
+ set -ue
+ cat ./VERSION
+ VERSION=1.4.0.pre13
+ echo Checking for deployed kontena-cli gem version 1.4.0.pre13...
Checking for deployed kontena-cli gem version 1.4.0.pre13...
+ gem fetch -v 1.4.0.pre13 kontena-cli
+ grep -q ERROR:  Could not find a valid gem 'kontena-cli' (= 1.4.0.pre13) in any repository
+ tee /tmp/gem-fetch.log
+ echo gem version 1.4.0.pre13 has not yet been deployed
gem version 1.4.0.pre13 has not yet been deployed
build/travis/deploy_gem.sh: 22: build/travis/deploy_gem.sh: RUBYGEMS_USER: parameter not set
```

### With random rubygems errors
```
+ set -ue
+ cat ./VERSION
+ VERSION=1.4.0.pre13
+ echo Checking for deployed kontena-cli gem version 1.4.0.pre13...
Checking for deployed kontena-cli gem version 1.4.0.pre13...
+ gem fetch -v 1.4.0.pre13 kontena-cli
+ + tee /tmp/gem-fetch.log
[...]

PR #2901, opened 2017-10-04
Release 1.4.0.pre12

SpComb:
Bump the v1.4 branch to include #2896, and also re-test the kontena-cli deb push to bintray after pre11 failed.

PR #2903, opened 2017-10-04
Fix CLI omnibus liblzma source

SpComb:
See https://github.com/chef/omnibus-software/issues/780

It looks like the liblzma upstream changed the http://tukaani.org/xz/... download location to now always redirect to sourceforge, even for the most recent version. Sourceforce downloads can redirect from https -> http (depending on the mirror, it seems), which gets rejected by omnibus's net_fetcher: `redirection forbidden: https://downloads.sourceforge.net/project/lzmautils/xz-5.2.3.tar.gz?r=&ts=1507104244&use_mirror=cfhcable -> http://cfhcable.dl.sourceforge.net/project/lzmautils/xz-5.2.3.tar.gz (RuntimeError)`

This causes the CLI deploy job to fail in travis (https://github.com/kontena/kontena/pull/2901#issuecomment-334090899), and it cannot be restarted to retry it (#2902).

Update to the most recent version of the [`software/liblzma.rb`](https://github.com/eoly/omnibus-software/blob/master/config/software/liblzma.rb) definition, change the source url to use sourceforge.net directly, and fix the sporadic sourceforge redirect failures by using the (undocumented) `unsafe: true` option to permit unsafe redirects: https://github.com/chef/omnibus/blob/52393d7cab443b61790f94c62775d9032d283497/lib/omnibus/fetchers/net_fetcher.rb#L165-L168

```ruby
      if source[:unsafe]
        log.warn(log_key) { "Permitting unsafe redirects!" }
        options[:allow_unsafe_redirects] = true
      end
```

The original download source was already using http://, and the release tarball is verified using the MD5 checksum, so it should be acceptable to follow http:// redirects.

The same issue may also apply to other projects hosted on sourceforge...


cli/omnibus/config/software/liblzma.rb
```diff
@@ -21,8 +21,11 @@
 license_file "COPYING"
 skip_transitive_dependency_licensing true
 
-source url: "http://tukaani.org/xz/xz-#{version}.tar.gz",
-       md5: "ef68674fb47a8b8e741b34e429d86e9d"
+version("5.2.2") { source md5: "7cf6a8544a7dae8e8106fdf7addfa28c" }
+version("5.2.3") { source md5: "ef68674fb47a8b8e741b34e429d86e9d" }
+
+source url: "https://sourceforge.net/projects/lzmautils/files/xz-#{version}.tar.gz/download",
+       unsafe: true, # accept https -> https redirects from sourceforge
```
SpComb: Oops, the trailing comma is a syntax error

```
cli/omnibus/config/software/liblzma.rb:30: syntax error, unexpected tSTRING_BEG, expecting keyword_do or '{' or '('
relative_path "xz-#{version}"
               ^
```

PR #2906, opened 2017-10-04
Fix CLI output hang after exceptions raised in spinners

kke:
Spinners may leave behind their spinner-threads making [`Kontena::Cli::Common#puts`](https://github.com/kontena/kontena/blob/master/cli/lib/kontena/cli/common.rb#L91) eat all the messages.

Normally this does not matter because usually the command exits and nobody notices. But when running through `Kontena.run!` or the way KOSH does it, the threads are left behind and prevent any further `puts`ing.

As a sidenote, the selfmade spinner implementation is far from excellent and should probably be replaced with something like `tty-spinner`

### DIFF
--- a/cli/lib/kontena/cli/spinner.rb
+++ b/cli/lib/kontena/cli/spinner.rb
@@ -168,6 +168,7 @@ module Kontena
               Kernel.puts msg
             end
           end
+          Thread.main['spinners'].delete(spin_thread)
         end
 
         exit(status) if status


SpComb:
Do we have an issue for `kontena shell freezes without any output if a command fails`? How to best to repro this for testing?

SpComb:
Looks like https://github.com/kontena/kontena-plugin-shell/issues/31, fixed in https://github.com/kontena/kontena-plugin-shell/pull/33 for each separate `kontena ...` command in the shell, but also broken for internal `Kontena.run!` commands that error out?

cli/lib/kontena/cli/spinner.rb
```diff
@@ -168,6 +168,7 @@ def self.spin(msg, &block)
               Kernel.puts msg
             end
           end
+          Thread.main['spinners'].delete(spin_thread)
```
SpComb: If we `delete` the `spin_thread` from `Thread.main['spinners']` here in the ensure, then what does the `Thread.main['spinners'].pop` later on after the `ensure` block do... because if nothing gets raised, then this will run both the `Thread.main['spinners'].delete(spin_thread)` and `Thread.main['spinners'].pop`?

SpComb: Yeah, it does both... the `pop` is a no-op if the `Thread.main['spinners']` is empty: 

```
 [done] Triggering deployment of stack stop-signal     
delete #<Thread:0x000000021da7c0@/home/kontena/kontena/kontena/cli/lib/kontena/cli/spinner.rb:96 dead> from Thread.main['spinners']: [#<Thread:0x000000021da7c0@/home/kontena/kontena/kontena/cli/lib/kontena/cli/spinner.rb:96 dead>]
pop from Thread.main['spinners']: []
```

I'm guessing this would be relevant for nested spinners, though... where in the CLI are those actually used? I couldn't find any with a quick look around.

kke: Hopefully not anywhere. There was one in the master deploy wizard but I think it has been removed.

SpComb: Could this code be simplified to not be broken in complex cases that are not actually used? :/

kke: The whole spinner.rb should be buried. Very deep.

SpComb: Nuke it from orbit, it's the only way to be sure

SpComb: Dunno how the nested spinners are supposed to work, but testing by changing the `kontena stack install ` -> `Installing dependency ...` into a spinner, it actually behaves alright with this buggy change, at least without errors:

```
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo .. \
 [done] Creating stack deptest-foo      
 [done] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo     
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar .. \
 [done] Creating stack deptest-bar      
 [done] Triggering deployment of stack deptest-bar     
 [done] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar     
 [done] Creating stack deptest      
```

In fact, I maybe like the "broken" output with the delete+pop in this PR more than the "working" output before this PR?

```
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo .. \
 [done] Creating stack deptest-foo      
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo .. |
 [done] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo     
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar .. \
 [done] Creating stack deptest-bar      
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar
 [done] Triggering deployment of stack deptest-bar     
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar
 [....] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar .. \
 [done] Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-bar     
 [done] Creating stack deptest      
```

PR #2910, opened 2017-10-05
Check ports from scheduled instances

jakolehm:
Currently there is a race condition with the port filter because filter checks reserved ports from actual containers. This PR improves port filtering logic so that ports are checked from scheduled service instances. There is still possibility to hit this race condition if two separate stacks (that have same port exposed) are deployed in parallel.

### DIFF
--- a/server/app/services/scheduler/filter/port.rb
+++ b/server/app/services/scheduler/filter/port.rb
@@ -9,14 +9,15 @@ module Scheduler
       # @return [Array<HostNode>]
       def for_service(service, instance_number, nodes)
         candidates = nodes.dup
-        ports = service.ports.map{|p| p['node_port']}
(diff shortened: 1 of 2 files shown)


server/app/services/scheduler/filter/port.rb
```diff
@@ -9,14 +9,15 @@ class Port
       # @return [Array<HostNode>]
       def for_service(service, instance_number, nodes)
         candidates = nodes.dup
-        ports = service.ports.map{|p| p['node_port']}
+        ports = service.ports.map { |p| p['node_port'] }
```
jnummelin: Not maybe part of this PR, but would it make sense to return early if no ports exposed or `service.net == 'host'`. :)

PR #2912, opened 2017-10-05
Fix server websocket exec error handling

### LINKED ISSUE #2605: Server container exec websocket API hangs connnection without any HTTP response on exceptions raised from the r.websocket block
Using low RPC timeouts + agent RPC handler sleeps to simulate a situation where the  [`/v1/containers/:id/exec` `r.websocket`](https://github.com/kontena/kontena/blob/c0e31da6a6d21172a37cd38d2bd2593a021e901f/server/app/routes/v1/containers_api.rb#L62) handler call to `Docker::StreamingExecutor#start` raises an `RpcClient::TimeoutError`, the puma server never sends any HTTP response at all. This causes the CLI websocket client to simply hang reading from the socket.

The server roda logs the HTTP 500 error:

```
172.17.0.1 - - [21/Jul/2017:09:47:52 +0000] "GET /v1/containers/development/core-01/weave/exec HTTP/1.1" 500 - 0.1077
```

And the error is dispatched to the [`RequestHelpers` `plugin :error_handler`](https://github.com/kontena/kontena/blob/c0e31da6a6d21172a37cd38d2bd2593a021e901f/server/app/helpers/request_helpers.rb#L10) block, which returns an `{code: ... ]` response. The [`Roda::Plugins::ErrorHandler`](https://github.com/jeremyevans/roda/blob/master/lib/roda/plugins/error_handler.rb#L69) is supposed to halt the request with a HTTP 500 error, but this does not seem to happen anymore after the [`Roda::Plugins::Websockets` `r.websocket`](https://github.com/jeremyevans/roda/blob/master/lib/roda/plugins/websockets.rb)  block is called.

Removing the roda error handler, and fixing the server `TokenAuthentication` middleware to propagate exceptions instead of returning HTTP 403 errors, allows the exception to propagate all the way up to puma, which then does return a HTTP 500 error response to the client.

SpComb:
Fixes #2605

Change the roda request handler to run the RPC `/containers/create_exec` request outside of the `r.websocket` block, allowing it to fail via the normal roda error handling. The `r.websocket` block only registers the websocket handlers, which is not expected to fail.

Rewrite the `Docker::StreamingExecutor`:

* Split up `setup` vs `start(ws)` for error handling purposes

* Deduplicate the `register_stdin_ws_events` vs `register_run_ws_events`

* Better debug logging

* Fix `MongoPubsub.subscribe do ...` to use `EventMachine.schedule { @ws.write(...) }`

    The socket is owned by EventMachine, and the pubsub subscription thread should not be doing concurrent writes to the websocket.

* Fix the `@ws.on :message` handler to abort the websocket on any errors, not just `JSON::ParserError`

* Add specs

### DIFF
--- a/server/app/services/docker/streaming_executor.rb
+++ b/server/app/services/docker/streaming_executor.rb
@@ -76,6 +76,8 @@ module Docker
       debug { "exec #{self.exec_id} run with shell=#{shell} tty=#{tty} stdin=#{stdin}: #{cmd.inspect}" }
 
       @rpc_client.notify('/containers/run_exec', self.exec_id, cmd, tty, stdin)
+
+      running!
     end
 
     # @param width [Integer]
@@ -172,9 +174,8 @@ module Docker
       debug { "websocket message: #{data.inspect}"}
 
       if data.has_key?('cmd')
-        fail "already running" if running?
+        fail "unexpected cmd: already running" if running?
         exec_run(data['cmd'], shell: @shell, tty: @tty, stdin: @interactive)
-        running!
       end
 
       if data.has_key?('stdin')
--- a/server/app/services/docker/streaming_executor.rb
+++ b/server/app/services/docker/streaming_executor.rb
@@ -186,14 +186,14 @@ module Docker
     #
     # Can be called multiple times (abort -> on_websocket_close)
     def teardown
-      if @exec_session
-        exec_terminate
-        @exec_session = nil
-      end
-
       if @subscription
         @subscription.terminate
         @subscription = nil
+      end
+
+      if @exec_session
+        exec_terminate
+        @exec_session = nil
       end
     end
   end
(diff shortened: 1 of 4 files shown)


SpComb:
Design review because I'm not sure if this is suitable as a bugfix for 1.4.

Still needs specs.

SpComb:
This shouldn't change the functionality of the code, and the websocket exec has fairly good tests. I think this would still be suitable for 1.4.1.

server/app/services/docker/streaming_executor.rb
```diff
@@ -3,111 +3,236 @@ class StreamingExecutor
...
+
+    # exec is running, and ready to accept input/tty_resize
+    #
+    # @return [Boolean]
+    def running?
+      @running
+    end
+
+    # Valid after setup()
+
```
jakolehm: Confusing newline.

PR #2767, opened 2017-09-05
Support shm_size

jakolehm:
Allows to define size of `/dev/shm`.

Fixes #2764

PR #2918, opened 2017-10-07
Support for stop_signal

### LINKED ISSUE #2878: Support for `stop-signal`
Docker supports giving a custom `stop-signal` during container creation. This is the signal Docker will send to the main process when container is stopped.

## Example use case:

Nginx uses `SIGTERM` as default stop signal. Unfortunately that does not allow graceful shutdown of nginx and ongoing requests will get dropped. Using `SIGQUIT` stops nginx processes gracefully and thus would allow true zero-downtime deployment with Kontenas default rolling deployment.

mikkopiu:
Closes #2878 
Based implementation on #2767

Added support for `stop_signal`, as in <https://docs.docker.com/compose/compose-file/#stop_signal>

Also fixed apparent typo in `agent/spec/lib/kontena/models/service_pod_spec.rb` `Entrypoint` test case name.

When I ran the specs, there were some errors in `agent` and `server`, but nothing that seemed to relate to this PR.

### DIFF
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -352,7 +352,7 @@ module GridServices
               integer :initial_delay, default: 10
             end
           end
-          string :stop_signal, matches: /^((SIG)([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)$/i
+          string :stop_signal, matches: /\A((SIG)([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)\z/i
           string :stop_grace_period, matches: Duration::VALIDATION_PATTERN
         end
       end
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -352,7 +352,7 @@ module GridServices
               integer :initial_delay, default: 10
             end
           end
-          string :stop_signal
+          string :stop_signal, matches: /^((SIG)([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)$/i
           string :stop_grace_period, matches: Duration::VALIDATION_PATTERN
         end
       end
--- a/server/app/mutations/grid_services/common.rb
+++ b/server/app/mutations/grid_services/common.rb
@@ -352,6 +352,7 @@ module GridServices
               integer :initial_delay, default: 10
             end
           end
+          string :stop_signal
           string :stop_grace_period, matches: Duration::VALIDATION_PATTERN
         end
       end
--- a/agent/lib/kontena/models/service_pod.rb
+++ b/agent/lib/kontena/models/service_pod.rb
@@ -43,6 +43,7 @@ module Kontena
                   :wait_for_port,
                   :volume_specs,
                   :read_only,
+                  :stop_signal,
                   :stop_grace_period
       attr_accessor :entrypoint, :cmd
 
@@ -88,6 +89,7 @@ module Kontena
         @networks = attrs['networks'] || []
         @wait_for_port = attrs['wait_for_port']
         @read_only = attrs['read_only']
+        @stop_signal = attrs['stop_signal']
         @stop_grace_period = attrs['stop_grace_period']
       end
 
@@ -172,6 +174,7 @@ module Kontena
         docker_opts['User'] = self.user if self.user
         docker_opts['Cmd'] = self.cmd if self.cmd
         docker_opts['Entrypoint'] = self.entrypoint if self.entrypoint
+        docker_opts['StopSignal'] = self.stop_signal if self.stop_signal
 
         if self.can_expose_ports? && self.ports
           docker_opts['ExposedPorts'] = self.build_exposed_ports
--- a/agent/spec/lib/kontena/models/service_pod_spec.rb
+++ b/agent/spec/lib/kontena/models/service_pod_spec.rb
@@ -217,9 +217,18 @@ describe Kontena::Models::ServicePod do
       expect(service_config['Entrypoint']).to be_nil
     end
 
-    it 'does not include Entrypoint if nil' do
+    it 'includes Entrypoint if set' do
       data['entrypoint'] = ['/bin/sh']
       expect(service_config['Entrypoint']).to eq(['/bin/sh'])
+    end
+    
+    it 'does not include StopSignal if nil' do
+      expect(service_config['StopSignal']).to be_nil
+    end
+
+    it 'includes StopSignal if set' do
+      data['stop_signal'] = 'SIGQUIT'
+      expect(service_config['StopSignal']).to eq('SIGQUIT')
     end
 
     it 'includes empty ExposedPorts if no ports are defined' do
--- a/cli/lib/kontena/cli/services/create_command.rb
+++ b/cli/lib/kontena/cli/services/create_command.rb
@@ -45,6 +45,7 @@ module Kontena::Cli::Services
     option "--health-check-initial-delay", "DELAY", "Initial delay for health check"
     option "--health-check-port", "PORT", "Port for health check"
     option "--health-check-protocol", "PROTOCOL", "Protocol of health check"
+    option "--stop-signal", "STOP_SIGNAL", "Alternative signal to stop container"
     option "--stop-timeout", "STOP_TIMEOUT", "Timeout (duration) to stop a container"
     option "--read-only", :flag, "Read-only root fs for the container", default: false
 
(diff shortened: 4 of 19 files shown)


server/app/mutations/grid_services/common.rb
```diff
@@ -352,6 +352,7 @@ def common_validations
               integer :initial_delay, default: 10
             end
           end
+          string :stop_signal
```
SpComb: Maybe also validate the signal here to catch mistakes earlier, rather than having to wait for the deploy fail - although looking at Docker's [`ParseSignal`](https://github.com/moby/moby/blob/a3efe9722f34af5cf4443fe3a5c4e4e3e0457b54/pkg/signal/signal.go#L31-L44), it's difficult to validate this meaningfully without encoding a complete list of all signals. I'm not sure if the regex for matching the supported values isn't just more trouble than it's worth:

```
string :stop_signal, matches: /\A((SIG)?([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)\z/i
```

mikkopiu: I thought about validating the input but could not (at least quickly) find conclusive documentation what signals are valid (over all possible OS's using signals).

I could try looking into docker-compose's implementation, whether they have any validation rules.

mikkopiu:
@SpComb : I tried to create a simple E2E test case but I'm having some difficulties:

* `kontena container inspect <id>` returns 404 on stopped containers (initial idea to get the `docker inspect` JSON)
* The `swipely/docker-api` library appears to only support Docker API version 1.16 => I can only get the exit _status_ (basically just `"exited"`), and not the full `State` object, which would contain the `ExitCode`

Here's my current progress (totally hacky): https://github.com/mikkopiu/kontena/pull/2/files . But that fails on account of `container.info` missing `ExitCode`

Unfortunately I'm not fluent enough in Ruby to pursue this further, unless you have some ideas?

The basic idea was to copy the test case that `docker-compose` is using: https://github.com/docker/compose/blob/master/tests/acceptance/cli_test.py#L1816-L1826

jnummelin:
You should be able to inspect the running container and look at `StopSignal` property.
```
kontena container inspect twilight-lake-92/nginx-1 | jq .Config.StopSignal
"SIGTERM"
```
In the specs you should be able to get it with something like:
```
k = run "kontena container inspect #{container}"
JSON.parse(k.out).dig('Config', 'StopSignal')
```

The problem you are facing in your hacky spec is that the data you get in `Container.all` and with `Container.get('foobar')` are different. So if you want the full data of container, always do `Container.get('foobar')`.

There's also one other issue in that approach. One should be able to run the E2E tests in any environment, so the created service containers are not always running in your local docker. Hence getting the container data through docker API is not really an option.

mikkopiu:
@jnummelin: Thanks for the help! That's probably the best option -- I went a bit too far on my "end", shouldn't test the Docker API but rather what I've implemented here.

I'll rewrite the test cases probably later today, and look at the validation a bit more.

SpComb:
Great to also see an e2e spec! This isn't the easiest thing to test for, but e2e specs are important.

> But that fails on account of container.info missing ExitCode

When you're inspect the container, it's under `State` in the JSON: `Docker::Container.get('signal-test').info['State']['ExitCode'] => 0`

> `kontena container inspect <id>` returns 404 on stopped containers (initial idea to get the docker inspect JSON)

Are you using the actual Kontena container ID? It's not immediately apparent from `kontena service show`, but the syntax is what `kontena container ls` uses: `kontena container inspect core-01/stop-signal.redis-1`

```json
{
  "Id": "0e886a275ce3a273817a88d6d5c6bb2a77e9125a522f761db8cb1d581fbb7d61",
  "Created": "2017-10-09T11:25:56.329328731Z",
  "Path": "/w/w",
  "Args": [
    "docker-entrypoint.sh",
    "redis-server"
  ],
  "State": {
    "Status": "exited",
    "Running": false,
    "Paused": false,
    "Restarting": false,
    "OOMKilled": false,
    "Dead": false,
    "Pid": 0,
    "ExitCode": 137,
    "Error": "",
    "StartedAt": "2017-10-09T15:01:49.230339568Z",
    "FinishedAt": "2017-10-10T08:37:27.810400236Z"
  },
  ...
}
```

That info should also contain the exit code.

The Kontena container shouldn't be cleaned up until it's actually destroyed in Docker. If `kontena container inspect` is giving 404 for a stopped container, then that would be a bug.

> The swipely/docker-api library appears to only support Docker API version 1.16 => I can only get the exit status (basically just "exited"), and not the full State object, which would contain the ExitCode

The Docker API will actually also return the newer fields when inspecting, even with the older API version:

```
irb(main):003:0> Docker::Container.get('signal-test').info['State']
=> {"Status"=>"exited", "Running"=>false, "Paused"=>false, "Restarting"=>false, "OOMKilled"=>false, "Dead"=>false, "Pid"=>0, "ExitCode"=>0, "Error"=>"", "StartedAt"=>"2017-10-10T08:34:33.008885561Z", "FinishedAt"=>"2017-10-10T08:34:33.054949522Z"}
irb(main):004:0> Docker::API_VERSION
=> "1.16"
```

(in the agent we [monkey-patch the `Docker::API_VERSION`](https://github.com/kontena/kontena/blob/master/agent/lib/docker/version.rb))

SpComb:
> @jnummelin: Thanks for the help! That's probably the best option -- I went a bit too far on my "end", shouldn't test the Docker API but rather what I've implemented here.

As for testing the `Config.StopSignal` vs the actual resulting exit code from the bash trap - I think it's worthwhile testing both if feasible, although it's good to test the `Config.StopSignal` first, because that makes any spec failures easier to pick up if there's a typo in the Kontena code, and it doesn't actually get set on the container.

TBH testing the actual exit code will also get a bit more complicated than what you have ATM, because `kontena stack|service stop` doesn't actually wait for the container to stop. The container might still be running at the point when `kontena service stop` returns, and you inspect the container, which would make the spec flaky without extra logic to actually wait for the container to be stopped. I'd be happy with an e2e spec to just test for the `Config.StopSignal`.

mikkopiu:
I had to clean my commits up a bit (sorry for the accidental extra PR) but now this PR has the simpler E2E spec (`Config.StopSignal` is set). Couldn't figure out a stable way to test the actual ExitCode.

@SpComb : About the regex validation:
At least looking through the `SignalMap`s Docker uses, that regex you showed could be a valid option to at least prevent totally incorrect signals. It won't be perfect, but it might be overkill to maintain lists similar to these ones Docker has:

* https://github.com/moby/moby/blob/master/pkg/signal/signal_darwin.go
* https://github.com/moby/moby/blob/master/pkg/signal/signal_freebsd.go
* https://github.com/moby/moby/blob/master/pkg/signal/signal_solaris.go
* https://github.com/moby/moby/blob/master/pkg/signal/signal_linux.go
* https://github.com/moby/moby/blob/master/pkg/signal/signal_windows.go

What do you think?

server/app/mutations/grid_services/common.rb
SpComb: > @SpComb : About the regex validation:
> At least looking through the SignalMaps Docker uses, that regex you showed could be a valid option to at least prevent totally incorrect signals. It won't be perfect, but it might be overkill to maintain lists similar to these ones Docker has: ...
> What do you think?

Yeah, definitely overkill to hardcode the list of valid signals. If the regex looks okay and doesn't end up accidentally excluding any valid signals, then it wouldn't hurt to include it... it will just also need `GridServices::Create` specs for it, not convinced if worth the effort. I'm ready to accept this without it, but great if you want to write those up.

mikkopiu:
@SpComb : I added a validation regex (modified yours to force non-numeric strings to start with `SIG`) and some rudimentary specs for that validation. Hopefully those look OK (and are in the correct place).

server/app/mutations/grid_services/common.rb
```diff
@@ -352,7 +352,7 @@ def common_validations
               integer :initial_delay, default: 10
             end
           end
-          string :stop_signal
+          string :stop_signal, matches: /^((SIG)([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)$/i
```
SpComb: See #2324 for the hilarious reasons for why these regexp must use `/\A...\z/` instead of `/^...$`

server/app/mutations/grid_services/common.rb
```diff
@@ -352,6 +352,7 @@ def common_validations
               integer :initial_delay, default: 10
             end
           end
+          string :stop_signal, matches: /^((SIG)([A-Z0-9]+|RTMIN\+\d+|RTMAX-\d+)|\d+)$/i
```
SpComb: Fix this to `/\A...\z/`

mikkopiu: Fixed, and also learned something new today -- didn't know there was a difference.

Issue #2114, opened 2017-04-07
CLI: kontena service, stack deploy progress, errors


SpComb:
Using the new `service_deploys -> instance_deploys` given by the server API.

SpComb:
I suggest something like this:

```
 [done] Triggering deployment of stack test     
 [done] Waiting for deployment to start     
 [done] Deploying service (1/3) foo:  ⊛
 *      Deploying service (2/3) bar: ⊛⊛⊙⊙⊝⊝⊝
```

For a stack with three services... the first service foo has only one instance, and it was deployed... the second service bar has 7 instances, of which two have been deployed, two are ongoing, and three are still waiting for their deployment to be created/started.

Then if once instance deploy fails, it would update to:

```
 [done] Triggering deployment of stack test     
 [done] Waiting for deployment to start     
 [done] Deploying service (1/3) foo:  ⊛
 [error] Failed to deploy service bar-2 to node asdf: Docker::Error::Foo: foobar
 *      Deploying service (2/3) bar: ⊛⊛⊛⊗⊝⊝⊝
```

And then once the stack deploy errors out because of that, it would end up with something like:

```
 [done] Triggering deployment of stack test     
 [done] Waiting for deployment to start     
 [done] Deploying service (1/3) foo:  ⊛
 [error] Failed to deploy service bar-2 to node asdf: Docker::Error::Foo: foobar
 [fail] Deploying service (2/3) bar: ⊛⊛⊛⊗⊝⊝⊝
 [error] Failed to deploy service bar: one or more instances failed
 [error] Failed to deploy stack: service bar deploy failed: one or more instances failed
```

SpComb:
The server `/v1/stacks/:grid/:stack/deploys/:deploy` API is missing the `services_count` information needed to display the `Deploying service (X/Y) ...` thing, but that's trivial to add:

```patch
diff --git a/server/app/views/v1/stack_deploys/_stack_deploy.json.jbuilder b/server/app/views/v1/stack_deploys/_stack_deploy.json.jbuilder
index bbad1c3..f07d449 100644
--- a/server/app/views/v1/stack_deploys/_stack_deploy.json.jbuilder
+++ b/server/app/views/v1/stack_deploys/_stack_deploy.json.jbuilder
@@ -2,6 +2,7 @@ json.id stack_deploy.id.to_s
 json.stack_id stack_deploy.stack.to_path
 json.created_at stack_deploy.created_at
 json.state stack_deploy.deploy_state
[...]

PR #2931, opened 2017-10-11
Fix stack deploy to not raise RuntimeError on deploy errors

SpComb:
See https://github.com/kontena/kontena/issues/2114#issuecomment-335724117

The instance errors in `wait_for_deployment_to_start` result in an unhandled `RuntimeError` exception, which results in `run the command again with environment DEBUG=true set to see the full exception`. 

Change the `wait_for_deployment_to_start` to just `puts` + `abort` instead, as the CLI handles the resulting `SystemExit` properly within the spinner and `Kontena.run!`. The `wait_for_deploy_to_finish` also aborts in a similar fashion, although that happens outside of any `spinner`.

### Before
```
 [done] Creating stack affinity-fail      
 [done] Triggering deployment of stack affinity-fail     
 [fail] Waiting for deployment to start     
        Cannot find applicable node for service instance development/affinity-fail/redis-1: Did not find any nodes for affinity filter: node==nonexist
 [error] RuntimeError : deploy failed
```

### After

The `puts` get queued up within the `spinner`, and thus get output after the `[fail] Waiting for deployment to start`

```
 [done] Creating stack affinity-fail      
 [done] Triggering deployment of stack affinity-fail     
 [fail] Waiting for deployment to start     
Stack deploy failed
 - Cannot find applicable node for service instance development/affinity-fail/redis-1: Did not find any nodes for affinity filter: node==nonexist
```

### DIFF
--- a/cli/spec/kontena/cli/stacks/deploy_command_spec.rb
+++ b/cli/spec/kontena/cli/stacks/deploy_command_spec.rb
@@ -43,6 +43,6 @@ describe Kontena::Cli::Stacks::DeployCommand do
     })
     expect(subject).to receive(:sleep).once
 
-    expect{subject.run(['test-stack'])}.to exit_with_error.and output(/deploy failed/).to_stderr
+    expect{subject.run(['test-stack'])}.to exit_with_error.and output(/Stack deploy failed/).to_stdout
   end
 end
--- a/cli/lib/kontena/cli/stacks/stacks_helper.rb
+++ b/cli/lib/kontena/cli/stacks/stacks_helper.rb
@@ -9,10 +9,10 @@ module Kontena::Cli::Stacks
           deployment = client.get("stacks/#{deployment['stack_id']}/deploys/#{deployment['id']}")
         end
         if deployment['state'] == 'error'
-          $stderr.puts "Stack deploy failed"
+          puts "Stack deploy failed"
           deployment['service_deploys'].each do |service_deploy|
             if service_deploy['state'] == 'error'
-              $stderr.puts " - #{service_deploy['reason']}"
+              puts " - #{service_deploy['reason']}"
             end
           end
 
--- a/cli/lib/kontena/cli/stacks/stacks_helper.rb
+++ b/cli/lib/kontena/cli/stacks/stacks_helper.rb
@@ -9,9 +9,10 @@ module Kontena::Cli::Stacks
           deployment = client.get("stacks/#{deployment['stack_id']}/deploys/#{deployment['id']}")
         end
         if deployment['state'] == 'error'
+          $stderr.puts "Stack deploy failed"
           deployment['service_deploys'].each do |service_deploy|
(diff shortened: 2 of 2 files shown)


cli/spec/kontena/cli/stacks/deploy_command_spec.rb
```diff
@@ -43,6 +43,6 @@
     })
     expect(subject).to receive(:sleep).once
 
-    expect{subject.run(['test-stack'])}.to exit_with_error.and output(/deploy failed/).to_stderr
+    expect{subject.run(['test-stack'])}.to exit_with_error.and output(/Stack deploy failed/).to_stdout
```
kke: Errors should probably go to `$stderr`

SpComb: The way that `$stderr.puts` behaves when called in a spinner results in semi-broken output... `puts` gets queued up.

```
 [done] Creating stack affinity-fail      
 [done] Triggering deployment of stack affinity-fail     
    *   Waiting for deployment to start .. -Stack deploy failed
 - Cannot find applicable node for service instance development/affinity-fail/redis-1: Did not find any nodes for affinity filter: node==nonexist
 [fail] Waiting for deployment to start     
```

kke: Hmm right. There's some `spin.fail!` variant that does  something differently.

SpComb: The `wait_for_deployment_to_start` is called within a spinner, so it would need to take the spinner as an argument to use `spin.fail!`... 

```ruby
        spinner "Waiting for deployment to start" do
          wait_for_deployment_to_start(deployment)
        end
```

It would be easiest to just raise an exception with the entire multi-line message, but there's only support for handling `Kontena::Errors::StandardError`, which doesn't really fit here.

kke: If there was an error class such as `Error` then `raise Error, "dang!"` would get output like `Error : dang!` :) 

I don't know.. if you don't want to begin..rescue inside the spinner block, I guess putsing to stdout is good enough.

Of course spinner could do something like
```ruby
stderr_orig = $stderr
stderr_new = StringIO.new
$stderr = stderr_new

..do the spinning stuff..

$stderr = stderr_orig
stderr_new.rewind
$stderr.print sderr_new.read
```

to delay output of any stderr putsing to happen after the spinning has finished.

SpComb: I think using `puts` for spinner-compatible output together with the `abort` => `raise SystemExit` is the easiest thing to do here, without starting to refactor our error-handling infrastructure in the CLI.

I don't think adding explicit `begin ... rescues` to all spinner blocks is the right answer, and also capturing `$stderr` would break the use of `DEBUG=api` to see what's going on during spinners.

Issue #2810, opened 2017-09-14
kontena plugin uninstall does not prompt or accept --force


SpComb:
Looks like there's some old `--force` flag being used in the spec that no longer exists:

```
ERROR: Unrecognised option '--force'

See: 'kontena plugin uninstall --help'
  removes installed plugin (FAILED - 1)

Failures:

  1) plugin uninstall removes installed plugin
     Failure/Error: expect(k.code).to eq(0)
     
       expected: 0
            got: 1
     
       (compared using ==)
     # ./spec/features/plugin/uninstall_spec.rb:10:in `block (2 levels) in <top (required)>'
```

SpComb:
The `kontena plugin uninstall` prompt/`--force` flag were removed in #2434. Was that by design, or should they be restored?

SpComb:
Let's call this a bug. AFAIK there's no reason why the `kontena plugin uninstall` command shouldn't continue to prompt for confirmation, or accept `--force`.

kke:
I don't see how removing a plugin would make you lose anything. You can just reinstall the plugin.

SpComb:
Either this is a cli bug that needs to be fixed to restore the prompt/`--force`, or then it's a test chore where the e2e spec needs to be fixed to not call `kontena plugin uninstall --force`. I'll accept a PR that does either of those as fixing this issue for 1.4.0...

SpComb:
BTW this is going to be annoying for anyone that's been scripting `kontena plugin uninstall`. They will have necessarily been using `--force` before, and anything like that is going to fail after 1.3 -> 1.4.

Still considering if we shouldn't just reconsider this and revert the `kontena plugin uninstall` change to restore `--force`....

jakolehm:
Could we add `--force` as hidden option just for backwards compatibility?

SpComb:
Didn't happen for 1.4.0, please fix any callers of `kontena plugin uninstall --force`.

Re-open and revive #2937 if this turns out to still be an issue.

PR #2935, opened 2017-10-12
Fix remaining plugin uninstall --force usages in tests

SpComb:
See #2810

There were still a couple places in the e2e specs using `kontena plugin uninstall --force`. They were in un-checked `around/before/after do run ...` calls, so they were only showing up in the `DEBUG_KOMMANDO=1` output. The plugin specs probably still ended up testing both installs and upgrades, but just not necessarily in the expected ways.

Someone really needs to fix the e2e specs to actually fail by default on any command failures..

### DIFF
--- a/test/spec/features/plugin/install_spec.rb
+++ b/test/spec/features/plugin/install_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe 'plugin install' do
   after(:each) do
-    run('kontena plugin uninstall --force aws')
+    run('kontena plugin uninstall aws')
   end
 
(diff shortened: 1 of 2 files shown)


matti:
Mr Kommando here, is there something Kommando could do or is the problem in kontena’s `run` wrapper? @spcomb

SpComb:
> Mr Kommando here, is there something Kommando could do or is the problem in kontena’s run wrapper? @SpComb

Nothing wrong in Kommando... the Kontena wrapper just needs a `run!` variant that raises an exception `if k.code != 0`.

PR #2937, opened 2017-10-13
Fix cli plugin uninstall to ignore --force

### LINKED ISSUE #2810: kontena plugin uninstall does not prompt or accept --force

SpComb:
Fixes #2810

Adds a hidden `--force` option to the `kontena plugin uninstall` command that just gets ignored.

Add specs.

### DIFF
--- a/cli/lib/kontena/cli/plugins/uninstall_command.rb
+++ b/cli/lib/kontena/cli/plugins/uninstall_command.rb
@@ -7,6 +7,7 @@ module Kontena::Cli::Plugins
     include Kontena::PluginManager::Common
 
     parameter 'NAME', 'Plugin name'
+    option "--force", :flag, "Unused", hidden: true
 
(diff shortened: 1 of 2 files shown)


SpComb:
I guess we can drop this, because we already released 1.4.0 without the `--force` flag, and nobody's complaining loudly enough yet.

PR #2945, opened 2017-10-17
Fix variable interpolation in extended stack files

### LINKED ISSUE #2943: kontena-cli tries to interpolate docker-compose variables as stack variables
## Steps to reproduce

Create two files in a folder:

```yml
# kontena.yml
stack: demo
version: 1.0.0

services:
  node:
    extends:
      file: docker-compose.yml
      service: node

# docker-compose.yml
version: '2.1'
services:
  node:
    image: node:alpine
    environment:
      - ENV_VAR=$ENV_VAR
```

then run:

```sh
$ ENV_VAR=whatever
$ kontena stack validate
 [error] RuntimeError : Undeclared variable 'ENV_VAR' in c:/repos/kontenabug/docker-compose.yml:5 --       - ENV_VAR=$ENV_VAR
         See C:/Users/charl/.kontena/kontena.log or run the command again with environment DEBUG=true set to see the full exception
```

Expected result:
- the stack should be validated

Actual result:
- the stack is not validated. The kontena-cli throws an error about `ENV_VAR` being undeclared. This should not be the case, since `ENV_VAR` is not a kontena variable, but a "regular" `docker-compose` variable.

**Note**: Adding a `kontena stack` variable definition in the `docker-compose.yml` file allows to validate the stack. This create a non-valid `docker-compose` file, but it makes me think that the `docker-compose` is treated as a `kontena stack` file because of the `extend`.

```diff
# docker-compose.yml
version: '2.1'

+variables:
+  myvar:
+    type: string
+    default: 'test'
+    from:
+      env: ENV_VAR
+    to:
+      env: ENV_VAR

services:
  node:
    image: node:alpine
    environment:
      - ENV_VAR=$ENV_VAR
```

kke:
Fixes #2943

Extending Docker compose files that include ENV-variables was broken in 1.4.0. This PR fixes that.

It's still required that you declare the env variables in the parent yaml:

```yaml
# kontena.yml
stack: user/stack
variables:
  ENV_VAR:
    type: string
    from:
      prompt: Enter env var value

services:
  mysql:
    extends:
      file: docker-compose.yml
      service: mysql

# docker-compose.yml
version: '2'
services:
  mysql:
    image: mysql:5.6
    environment:
      - ENV_VAR=$ENV_VAR
```

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/reader.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/reader.rb
@@ -46,10 +46,6 @@ module Kontena::Cli::Stacks
         @notifications    = []
       end
 
-      def is_compose?
-        internals_interpolated_yaml['version'] =~ /^[23](?:\.\d+)?$/
-      end
-
       # @param without_defaults [TrueClass,FalseClass] strip the GRID, STACK, etc from response
       # @param without_vault [TrueClass,FalseClass] strip out any values that are going to or coming from VAULT
       # @return [Hash] a hash of key value pairs representing the values of stack variables
@@ -204,10 +200,6 @@ module Kontena::Cli::Stacks
         raise RuntimeError, "Variable validation failed: #{variables.errors.inspect} in #{file}" unless variables.valid? || skip_validation
 
         validate unless skip_validation
-
-        if is_compose? && internals_interpolated_yaml['version'].split('.').first != '2'
-          notifications << { 'file' => "Docker compose yaml version #{internals_interpolated_yaml['version']} is not fully supported" }
-        end
 
         result = {}
         Dir.chdir(from_file? ? File.dirname(File.expand_path(file)) : Dir.pwd) do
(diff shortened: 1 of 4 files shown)


cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -201,6 +205,10 @@ def execute(service_name = nil, name: loader.stack_name.stack, parent_name: nil,
 
         validate unless skip_validation
 
+        if is_compose? && internals_interpolated_yaml['version'].split('.').first != '2'
+          notifications << { 'file' => "Docker compose yaml version #{internals_interpolated_yaml['version']} is not fully supported" }
```
SpComb: Warning about supported compose versions might be a good idea, but this is now confusing me because both `kontena.yaml` and `docker-compose.yaml` files have a version field. Do we really want to warn about unsupported docker compose YAML versions if you you `kontena stack install kontena.yaml` with something like `version: '3.1'`? Should this only warn if you're actually loading a compose file with `extends`?

PR #2950, opened 2017-10-20
Add support for entrypoint in stack files

### LINKED ISSUE #2949: Stack yml does not support entrypoint
Entrypoint is supported by api.

kke:
Fixes #2949

Adds / fixes support for the `entrypoint` keyword in stack/compose YAMLs.

### DIFF
--- a/test/spec/support/container_helper.rb
+++ b/test/spec/support/container_helper.rb
@@ -1,3 +1,5 @@
+require 'json'
+
 module ContainerHelper
 
   # Checks if a deployed container exists with given name
@@ -24,4 +26,10 @@ module ContainerHelper
     k.out.match("^(.+\/#{name})")[1]
   end
 
+  # @raise [RuntimeError]
+  # @return [Hash]
+  def inspect_container(container_id)
+    k = run("kontena container inspect #{container_id}")
+    ::JSON.parse(k.out)
+  end
 end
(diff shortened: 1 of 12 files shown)


kke:
For reference, https://github.com/kontena/kontena/blob/9defedc8bc82cea6eb48ee2708bd6a69e2424829/cli/lib/kontena/cli/stacks/service_generator.rb includes all the keys imported from compose yml's.

Should cross check with https://kontena.io/docs/references/stack-file-vs-compose.html

kke:
I think `depends_on`, `logging` and `env_file` are not supported when extending.

SpComb:
I think this should have an e2e test.

SpComb:
I don't mean a test spec that runs `kontena stack validate`, I mean an actual e2e spec that deploys the stack and looks at the deployed container entrypoint.

That kind of real e2e spec is needed to make sure that e.g. the CLI is sending the right form of `{ "entrypoint": ...}`  JSON that doesn't just get ignored/rejected by the server, or possibly even cause the container create to fail on the agent (is it an array or a string in the Docker API? Both, I think...). The current specs don't cover that.

SpComb:
Tested to work with both the `/w/w` entrypoint and in `network_mode: host`, I think I'll write up an e2e spec for this next.

test/spec/support/container_helper.rb
```diff
@@ -24,4 +26,10 @@ def container_id(name)
     k.out.match("^(.+\/#{name})")[1]
   end
 
-end
\ No newline at end of file
+  # @raise [RuntimeError]
+  # @return [Hash]
+  def inspect_container(container_id)
```
SpComb: This e2e spec helper is identical to what's also in #2951

Issue #2924, opened 2017-10-10
Too many (>25) TLS certificates causes the load balancer to fail


japsu:
There seems to be an implicit, undocumented limit on the amount of certificates that can be attached to a load balancer. We hit this limit at 25 certificates. When we try to add the 26th one, deployment fails with this error message:

```
[done] Reading stack platform metadata from Kontena Master
[done] Upgrading stack platform
[done] Triggering deployment of stack platform
[done] Waiting for deployment to start
[fail] Deploying service lb1
Deployment of service plat2-grid/platform/lb1 failed:
- halting deploy of plat2-grid/platform/lb1, one or more instances failed
- GridServiceInstanceDeployer::StateError: Service instance is not running, but restarting (on node autumn-river-15)
```

Digging deeper, `sudo docker logs platform.lb1-1` on `autumn-river-15` gives me this:

```
goroutine 1 [running, locked to thread]:
panic(0x7eb2e0, 0xc820238460)
    /usr/lib/go1.6/src/runtime/panic.go:481 +0x3e6
github.com/urfave/cli.HandleAction.func1(0xc8201bd2e8)
    /build/amd64-usr/var/tmp/portage/app-emulation/runc-1.0.0_rc2_p9/work/runc-1.0.0_rc2_p9/Godeps/_workspace/src/github.com/urfave/cli/app.go:478 +0x38e
panic(0x7eb2e0, 0xc820238460)
    /usr/lib/go1.6/src/runtime/panic.go:443 +0x4e9
github.com/opencontainers/runc/libcontainer.(*LinuxFactory).StartInitialization.func1(0xc8201bcbf8, 0xc82001a0a0, 0xc8201bcd08)
    /build/amd64-usr/var/tmp/portage/app-emulation/runc-1.0.0_rc2_p9/work/runc-1.0.0_rc2_p9/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:259 +0x136
github.com/opencontainers/runc/libcontainer.(*LinuxFactory).StartInitialization(0xc820058870, 0x7f81a4b55500, 0xc820238460)
    /build/amd64-usr/var/tmp/portage/app-emulation/runc-1.0.0_rc2_p9/work/runc-1.0.0_rc2_p9/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:277 +0x5b1
main.glob.func8(0xc820076780, 0x0, 0x0)
[...]

PR #2951, opened 2017-10-20
Validate service env size limits

SpComb:
See #2924 requires LB fix as workaround, this just mitigates the fallout

Deploying an LB service with additional SSL certificates such that the resulting `SSL_CERTS` env size grows over 128KB currently fails very badly: the existing LB service instances will get terminated, and replaced with new LB service instances that will start, but immediately panic: `standard_init_linux.go:178: exec user process caused "argument list too long" [recovered]

Mitigate this by pre-validating the env size, and having the agent deploy fail early **before** terminating the existing service container.

* Reject oversize service envs via the service create/update mutation validation
* Fail agent service deploys if the containers envs are oversize
  * Validate the resulting env size after evaluating any combining all env secrets
  * Change the agent pod creator to evaluate the service config earlier, before cleaning up any existing container

## Testing

```yaml
stack: test/secrets-oversize
variables:
  cert:
    type: string
  cert_count:
    type: integer
services:
  lb:
    image: kontena/lb
    secrets: []
    certificates:
    # {% for i in (1..cert_count) %}
      - subject: $cert
        type: env
        name: SSL_CERTS
    # {% endfor %}
```

```
bin/kontena stack upgrade -v cert=test.188.226.131.23.xip.io -v cert_count=32 secrets-oversize test/secrets-oversize.yaml
 [done] Reading stack secrets-oversize from master     
 [done] Parsing test/secrets-oversize.yaml     
 [done] Analyzing upgrade     
 [done] Upgrading stack secrets-oversize     
 [done] Triggering deployment of stack secrets-oversize     
 [fail] Deploying service lb     
Deployment of service development/secrets-oversize/lb failed:
  - halting deploy of development/secrets-oversize/lb, one or more instances failed
  - GridServiceInstanceDeployer::ServiceError: Kontena::Models::ServicePod::ConfigError: Env SSL_CERTS is too large at 165801 bytes (on node core-01)
```

After the deploy failure, the LB service instances are still reachable and working.

## TODO
- [x] e2e specs


agent/lib/kontena/models/service_pod.rb
```diff
@@ -349,6 +356,10 @@ def build_env
         secrets_hash.each do |name, value|
           env << "#{name}=#{value}"
         end
+        env.each do |envstr|
+          name, value = envstr.split('=', 2)
+          raise ConfigError, "Env #{name} is too large at #{envstr.bytesize} bytes" if envstr.bytesize > ENV_MAX_STRLEN
```
SpComb: No idea how `envstr.length` vs `envstr.bytesize` really works here in case of non-ASCII env values. Maybe if everything is in UTF-8, from the `envstr.encoding` through the docker-api gem client and Docker itself, then this will work correctly?

SpComb: There may also be some odd edge cases here for envs that are exactly 128KB in size - like does the kernel limit include the terminating NUL or some other overhead? Have not tested for those edgecases yet. Maybe the validation should just be a couple bytes stricter just to avoid those?

PR #2952, opened 2017-10-20
Add "kontena plugin upgrade" to upgrade all plugins

### LINKED ISSUE #2948: Auto-update plugins
It would be nice if plugins would auto-update to latest version (at least when they throw errors).

kke:
Fixes #2948 

Adds a `kontena plugin upgrade` command that can be used to upgrade plugins.

It has the option flag `--silent` for use in cronjobs.


kke:
Automatic updating would require some kind of "last_checked" timestamp to config and adding something like `Kontena.run('plugin update') if last_checked + interval > Time.now.to_i` to `bin/kontena`.

SpComb:
Isn't `kontena plugin update foo` basically equivalent to `kontena plugin install foo`? It's a little confusing having two ways to do the same thing, and also not that great in terms of code/test duplication.

The `kontena plugin update` functionality makes plenty of sense, need to test that. Although as an APT user, I would expect it to be `kontena plugin upgrade` ;)

kke:
Yep maybe it could be just for upgrading all.

kke:
ping @SpComb 

- Removed the option to `kontena plugin update xyz`, it always upgrades all. (use `kontena plugin install xyz` to upgrade individually)
- Renamed the sub-command from `kontena plugin update` to `kontena plugin upgrade`

test/spec/features/plugin/upgrade_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  before(:each) do
+    run('kontena plugin uninstall aws')
+    run('kontena plugin uninstall digitalocean')
+    run('kontena plugin install --version 0.2.7 aws')
+    run('kontena plugin install --version 0.2.5 digitalocean')
+  end
+
+  it 'upgrades all plugins' do
+    k = run('kontena plugin upgrade')
+    expect(k.code).to be_zero
```
SpComb: Consider `run!` for better error output on failures

test/spec/features/plugin/upgrade_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect(Gem::Version.new(version) > Gem::Version.new("0.2.7")).to be_truthy
+      elsif plugin == 'digitalocean'
+        expect(Gem::Version.new(version) > Gem::Version.new("0.2.5")).to be_truthy
+      end
+    end
+  end
+
+  it 'does nothing if no updates' do
+    k = run('kontena plugin upgrade')
+    expect(k.code).to be_zero
```
SpComb: Same here

test/spec/features/plugin/upgrade_spec.rb
SpComb: Doesn't need this `expect(k.code).to be_zero` anymore

PR #2960, opened 2017-10-24
Turn off debugging by using DEBUG=false

### LINKED ISSUE #2959: DEBUG=false should disable debug
Currently there's no other way to disable debug than `unset DEBUG`.

kke:
Fixes #2959

### DIFF
--- a/cli/lib/kontena_cli.rb
+++ b/cli/lib/kontena_cli.rb
@@ -1,4 +1,5 @@
 require 'kontena/autoload_core'
+require 'set'
 
 $KONTENA_START_TIME = Time.now.to_f
 at_exit do
@@ -147,12 +148,6 @@ module Kontena
     end
   end
 
-  DEBUG_KEYWORDS = Set.new(%w(true api websocket plugin))
-
-  def self.debug?
-    DEBUG_KEYWORDS.include?(ENV['DEBUG'].to_s)
-  end
-
   def self.logger
     return @logger if @logger
     if log_target.respond_to?(:tty?) && log_target.tty?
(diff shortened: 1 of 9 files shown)


cli/lib/kontena_cli.rb
```diff
@@ -158,11 +159,18 @@ def self.logger
...
-    logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+    logger.level = debug? ? Logger::DEBUG : Logger::INFO
     logger.progname = 'CLI'
     @logger = logger
   end
 
+  DEBUG_KEYWORDS = Set.new(%w(true api websocket plugin))
+
+  def self.debug?
+    DEBUG_KEYWORDS.include?(ENV['DEBUG'].to_s)
```
SpComb: Hmm... so something like `DEBUG=1 kontena ...` would have enabled debug output before, but now gets ignored... probably not worth worrying about this too much, although maybe consider special-casing `false` instead?

kke: heh, true. `== false` would've been so much simpler.

cli/lib/kontena_cli.rb
```diff
@@ -158,11 +159,14 @@ def self.logger
...
       logger.formatter = Kontena::Cli::LogFormatter::StripColor.new
     end
-    logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+    logger.level = debug? ? Logger::DEBUG : Logger::INFO
     logger.progname = 'CLI'
     @logger = logger
   end
 
+  def self.debug?
+    !ENV['DEBUG'].nil? && ENV['DEBUG'].to_s != 'false'
```
SpComb: `DEBUG=` -> `ENV['DEBUG'].empty?` should also be false.

kke: true

PR #2964, opened 2017-10-25
Add commands&apis to list and delete domain authorizations

### LINKED ISSUE #2963: No way to remove `tls-sni-01` authorizations from the load balancer?
Related to #2924.

I tried freeing up some space for certificates by combining multiple names into a single certificate authorized using the `tls-sni-01` method.

Unfortunately, I failed to realize each authorization adds one phantom certificate to the load balancer.

Even after `kontena certificate rm`ing all the names I added using `kontena certificate add --type=tls-sni-01`, the maximum amount of actual certificates my load balancer can handle before failing (as detailed in #2924) seems to have permanently decreased from 25 to 22.

(Unfortunately, I need 24 certificates to support all services I currently have :) )

There seems to be no obvious way to query the `tls-sni-01` authorizations present in a load balancer, let alone remove them.

jnummelin:
There's no way to delete a domain authorization at the moment. Now with the env size limits this leaves all tls-sni authorizations linked to a service to "linger" around for eternity. :( And with the Linux env size limit issue, things are even worse now when all the certs (including the tls-sni ones) are bundled to single `SSL_CERTS` env.

Deleting a tls-sni domain auth that's still "in use" will cause the cert renewal to fail as the renewal logic cannot see where the new tls-sni certs should be deployed to.

fixes #2963

### DIFF
--- a/cli/lib/kontena/cli/certificate/domain_authorization/list_command.rb
+++ b/cli/lib/kontena/cli/certificate/domain_authorization/list_command.rb
@@ -1,0 +1,24 @@
+
+module Kontena::Cli::Certificate::DomainAuthorization
+  class ListCommand < Kontena::Command
+    include Kontena::Cli::Common
+    include Kontena::Cli::GridOptions
+    include Kontena::Cli::TableGenerator::Helper
+    include Kontena::Util
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    def fields
+      quiet? ? ['domain'] : {domain: 'domain', authorization_type: 'authorization_type', status: 'status'}
+    end
+
+    def execute
+      authorizations = client.get("grids/#{current_grid}/domain_authorizations")['domain_authorizations']
+
+      print_table(authorizations)
+    end
+
+  end
+end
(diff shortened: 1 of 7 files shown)


SpComb:
See https://github.com/kontena/kontena/issues/2963#issuecomment-339299263 for my thoughts as to why this doesn't really fix #2963: the Kontena server should be smart enough to avoid polluting the LB service with unnecessary `tls-sni` challenge certs.

This kind of API may still be useful for cancelling domain authorizations before they get validated or expire.

BTW, the [ACME API](https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.7.5.2) also supports excplicitly deactivating authorizations to cancel the challenges. That could potentially be relevant because there's a [rate limit](https://letsencrypt.org/docs/rate-limits/) on 300 Pending Authorizations, but I doubt that would happen unless someone was trying to develop some automation against our certificate API.

SpComb:
Missing REST API docs.

SpComb:
e2e specs passing apart from #3333

Issue #2763, opened 2017-09-04
CLI show-command filtering and formatting


kke:
Most of the `kontena *** show` commands output something that is almost but not exactly YAML.

For example, `kontena stack show xyz` outputs something like this:

```yaml
twemproxy-redis2:
  created: 2017-09-04T12:55:32.287Z
  state: running
  stack: kke/redis
  expose: - 
  services:
    redis:
      created: 2017-09-04T12:55:32.307Z
...
...
```

If this service was exposing something, the stack show output would actually be valid YAML, but the unescaped `-` in `expose` makes it invalid.

`kontena node show xxx` outputs:

```yaml
foo/core-2:
  id: VPRT:GVF3:YCTO:CYV5:JRGD:432W:SBPS:7UDH:IELG:OUFO:VNHH:PKE5
  agent version: 1.4.0.dev
  docker version: 1.10.3
  connected: yes
  last connect: 2017-09-04T12:16:45.194Z
  os: CoreOS 1122.3.0 (MoreOS)
  drivers:
    storage: overlay
    volume: local
  initial node: no
  labels:
  stats:
    cpus: 1
    filesystem:
      - /var/lib/docker: 0.55 of 15.57 GB
```

This is valid YAML, as long as you use `YAML.load` instead of `safe_load`, which refuses because timestamp is recognized and Time is not by default in the safe classes list.

This all seems quite nice, until you look at the source:

```ruby
# node show
     puts "#{node['id']}:"
      puts "  id: #{node['node_id']}"
      puts "  agent version: #{node['agent_version']}"
      puts "  docker version: #{node['docker_version']}"
...
      puts "  drivers:"
      puts "    storage: #{node['driver']}"
      puts "    volume: #{node['volume_drivers'].to_a.map{|v| v['name']}.join(',')}"
...
      if fs
        puts "    filesystem:"
        fs.each do |filesystem|
          puts "      - #{filesystem['name']}: #{to_gigabytes(filesystem['used'], 2)} of #{to_gigabytes(filesystem['total'], 2)} GB"
        end
      end
```

```ruby
# stack show
      puts "#{stack['name']}:"
      puts "  created: #{stack['created_at']}"
      puts "  updated: #{stack['updated_at']}"
      puts "  services:"
      stack['services'].each do |service|
        show_service(service['id'])
      end

    def show_service(service_id)
      token = require_token
      service = get_service(token, service_id)
      pad = '    '.freeze
      puts "#{pad}#{service['name']}:"
[...]

PR #2966, opened 2017-10-26
Fix CLI certificate show output formatting

### LINKED ISSUE #2957: Kontena certificate show outputs raw yaml
Which is not really inline with other commands. 

Two options:

* fix `certificate show` to match other commands
* fix all other commands to output raw yaml

SpComb:
Fixes #2957 

Disclaimer: any variations in color, syntax and indentation are a result of the hand-crafted nature of this output and are not considered flaws in the product.

#### `kontena certificate show test.188.226.131.23.xip.io`
```yaml
development/test.188.226.131.23.xip.io:
  subject: test.188.226.131.23.xip.io
  valid until: 2017-12-14T13:34:00Z
  alt names:
    - test3.188.226.131.23.xip.io
  auto renewable: true
```

#### `bin/kontena certificate show test2.188.226.131.23.xip.io`
```yaml
development/test2.188.226.131.23.xip.io:
  subject: test2.188.226.131.23.xip.io
  valid until: 2018-02-06T13:47:20Z
  auto renewable: true
```

### DIFF
--- a/cli/lib/kontena/cli/certificate/show_command.rb
+++ b/cli/lib/kontena/cli/certificate/show_command.rb
@@ -25,7 +25,7 @@ module Kontena::Cli::Certificate
 
       show_yaml(cert['id'],
         'subject'         => cert['subject'],
-        'valid until'     => cert['valid_until'],
+        'valid until'     => Time.parse(cert['valid_until']),
         'alt names'       => (alt_names && !alt_names.empty?) ? alt_names : nil,
         'auto renewable'  => cert['auto_renewable'],
       )
--- a/cli/lib/kontena/cli/certificate/show_command.rb
+++ b/cli/lib/kontena/cli/certificate/show_command.rb
@@ -11,9 +11,14 @@ module Kontena::Cli::Certificate
     requires_current_master_token
     requires_current_grid
 
+    def print_yaml(object)
+      puts YAML.dump(object).sub("---\n", '')
+    end
+
     def execute
       certificate = client.get("certificates/#{current_grid}/#{self.subject}")
-      puts YAML.dump(certificate)
+      
+      print_yaml(certificate.delete('id') => certificate)
     end
   end
 end
--- a/cli/lib/kontena/cli/certificate/show_command.rb
+++ b/cli/lib/kontena/cli/certificate/show_command.rb
@@ -11,14 +11,24 @@ module Kontena::Cli::Certificate
     requires_current_master_token
     requires_current_grid
 
-    def print_yaml(object)
-      puts YAML.dump(object).sub("---\n", '')
+    # @param id [String]
+    # @param attrs [Hash{String => nil, Object}] elides nil values
+    def show_yaml(id, attrs)
+      puts YAML.dump(
+        id => Hash[attrs.select{|k, v| !!v}]
+      ).sub("---\n", '')
     end
 
     def execute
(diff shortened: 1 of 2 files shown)


jakolehm:
Output without alt_names?

SpComb:
> Output without `alt_names`?

The API returns `{"alt_names": []}`, so that shows up in the YAML as `alt_names: []`...

In the `service show` and `node show` commands, that would either omit the `alt_names` line completely, or render it as `alt_names:` without any list lines.

The generic `print_yaml` would need to know which fields are "optional", and omit those if they're `nil?` or `empty?`. It might also be appropriate to elide any nil/empty fields from the output in general?

SpComb:
I can write this out as hand-written `puts` if that's what we really want, or then I can change the way that the `print_yaml'`is used to elide the empty `alt_names` array... but ATM I'm not even sure how it *should* be output:

* left out entirely from the output?
* `alt_names: ` without any array lines (i.e. YAML `null`)
* `alt_names: []`

jakolehm:
We probably don't have strict guidelines for this... but going through other commands it seems that we usually hide empty values unless it's critical for a user to see that value is empty. What is the case here... should user always see that alt_names is there?

My biggest concern here is that we invent yet another output format. If we want to show everything as raw yaml, then we should change other commands also. The current/original cli output format is not yaml, it just happens to resemble yaml.

kke:
In some commands an empty array is replaced with a dash and that is the only thing tat makes the output from those commands not YAML. Reference: #2763

```
alt_names: -
```

If they all showed YAML, we could have something like the formatting option in docker cli, so that you could easily dig some value from the output in scripts.

SpComb:
> We probably don't have strict guidelines for this... but going through other commands it seems that we usually hide empty values unless it's critical for a user to see that value is empty. What is the case here... should user always see that alt_names is there?

Sounds like we should omit the `alt_names` if it's empty.

Next question... comparing `kontena node show` with `kontena node list`, should it be:

* `alt_names`
* `alt names`

#### `kontena node show`
```
development/core-01:
  agent version: 1.4.0.dev
  docker version: 1.12.6
  connected: yes
  last connect: 2017-11-14T12:48:20.806Z
  last seen: 2017-11-14T13:01:53.421Z
```

#### `kontena service show`
```
development/null/redis:
  created: 2017-08-22T08:43:07.895Z
  updated: 2017-09-06T11:17:07.107Z
  desired_state: running
  read_only: no
```

SpComb:
Output is now like this, with `show_yaml` doing the equivalent of `Hash.compact`:

```
$ bin/kontena certificate show test.188.226.131.23.xip.io
development/test.188.226.131.23.xip.io:
  subject: test.188.226.131.23.xip.io
  valid until: '2017-12-14T13:34:00.000+00:00'
  alt names:
  - test3.188.226.131.23.xip.io
  auto renewable: true
$ bin/kontena certificate show test2.188.226.131.23.xip.io
development/test2.188.226.131.23.xip.io:
  subject: test2.188.226.131.23.xip.io
  valid until: '2018-02-06T13:47:20.000+00:00'
  auto renewable: true
```

SpComb:
Trying the get the right output re the timestamps is going to be hard: https://github.com/kontena/kontena/issues/2763#issuecomment-344273241

SpComb:
Replaced the `YAML.dump` with hand-crafted `puts` to avoid the quoting around the formatted timestamp values in the output...

PR #2967, opened 2017-10-26
Fix service affinity filters to be stack-scoped

### LINKED ISSUE #2961: Service affinity filter is not stack-aware, and does not work correctly if multiple stacks use the same service names
When scheduling a service with a `service==foo` affinity, the bare service names are matched without considering their stack scope. If multiple stacks have identically named services that match the affinity filter, then all of those external services are considered as matching candidates.

This breaks expectations when installing a stack with services `foo` and `bar`, and using service affinity to schedule both services onto the same nodes: if the same stack is installed multiple times, or the service names are generic enough to also be used in other stacks, then the two stack services may get deployed onto different nodes.

The service affinity filters should probably be stack-scoped by default, and perhaps also support explicit `service==$STACK/foo` references. However, that breaks if trying to refer to a stack-less service, unless using `service==null/foo` - maybe `service==/foo`?

It is currently not possible to use explicit stack-scoped service affinities like `service==$STACK/foo` as a workaround.

Using `container==$STACK.service-%s` as a workaround kinda works, although it looks like there's an additional bug re the dependent service deploys:  #2962

SpComb:
Fixes #2961

Resolve `service==foo` affinites within the scope of the scheduled service's stack, and also support the use of explicitly scoped `service==stack/foo` affinities.

The design edgecase is stack services with affinities to stackless services... these will now fail to schedule until rewritten to `service==null/foo`, and it would be best to avoid any references to the `null` stack... That edgecase conflicts with the more common use of bare service names to refer to services within the same stack, which matches the semantics of e.g. service links. We either need to require the use of explicit `service==$STACK/foo` affinities, or break affinities with stackless services.

Also includes some drastic refactoring/simplification of the affinity filter code to minimize code duplication.


server/app/services/scheduler/filter/affinity.rb
```diff
@@ -87,70 +76,32 @@ def split_affinity(affinity)
...
-        container_names = node.containers.map{|c|
-          c.labels['io;kontena;container;name'].to_s
-        }
-        if compare == '=='
-          container_names.any?{|n| n == value}
-        elsif compare == '!='
-          !container_names.any?{|n| n == value}
-        end
+      def container_match?(node, value)
+        node.containers.where(name: value).exists?
```
SpComb: Simplified the `labels['io;kontena;container;name']` to just query by `name` for convenience, but that may not be correct... I'm not sure what the significance of matching by the label is, because AFAICT the two should always be the same for service containers?

The only difference is that matching by label prevents using affinities like `container==weave`, but is that intentional?

server/app/models/grid_service.rb
```diff
@@ -220,6 +220,20 @@ def overlay_network?
     self.net.to_s == 'bridge'
   end
 
+  # @param link [String] service, stack/service, /service
+  # @return [GridService, nil]
+  def resolve_service(link)
```
jakolehm: Maybe this method should be within affinity filter.. it's the only place where it's used?

server/app/services/scheduler/filter/affinity.rb
jakolehm: Simple is better.

server/app/models/grid_service.rb
SpComb: There's something very similar going on in [`GridCertificates::Common#resolve_service`](https://github.com/kontena/kontena/blob/eb00568ccd013406be354dd24f30b5c2a3453a9f/server/app/mutations/grid_certificates/common.rb#L50-L56) used for domain authorization linked services, the [grid `.../container_logs` API](https://github.com/kontena/kontena/blob/eb00568ccd013406be354dd24f30b5c2a3453a9f/server/app/routes/v1/grids_api.rb#L140-L148), the `GridServices::Common` mutation ([build](https://github.com/kontena/kontena/blob/eb00568ccd013406be354dd24f30b5c2a3453a9f/server/app/mutations/grid_services/common.rb#L36-L41) and [validate](https://github.com/kontena/kontena/blob/eb00568ccd013406be354dd24f30b5c2a3453a9f/server/app/mutations/grid_services/common.rb#L160-L168))... it would be nice to have a single place to handle those kind of `stack/service` references, and IMO that would belong in the models. Unfortunately each of those uses is also slightly different, so it's not appropriate to try and refactor all of them in this PR.

I guess I'll move this into the affinity filter for now, and leave it up to a later refactoring PR to try and consolidate all of those implementations of the same thing.

server/app/models/grid_service.rb
```diff
@@ -244,12 +244,18 @@ def dependant_services
...
+        {:volumes_from => {:$regex => /^#{self.name_with_stack}-(%s|\d+)/}},
+
+        # the agent will resolve volumes_from: foo-1 to container stack.foo-1 if container foo-1 does not exist
+        {:volumes_from => {:$regex => /^#{self.name}-(%s|\d+)/}, :stack_id => self.stack.id},
+
+        {:affinity => "service==#{self.name}", :stack_id => self.stack.id},
+        {:affinity => "service!=#{self.name}", :stack_id => self.stack.id},
+        {:affinity => "service==#{self.stack.name}/#{self.name}"},
+        {:affinity => "service!=#{self.stack.name}/#{self.name}"},
+        {:affinity => {:$regex => /^container(==|!=)#{self.name_with_stack}-(%s|\d+)$/}},
```
SpComb: The match for `container` affinities used to be missing completely, this now uses `name_with_stack` to get the `stack.service-X` naming.

## Testing

```yaml
stack: test/container-affinity
services:
  redis:
    image: redis
  bar:
    image: redis
    affinity:
      - container==$STACK.redis-1
```

```
I, [2017-10-31T15:11:48.622084 #12]  INFO -- StackDeployWorker: deploying service development/container-affinity/redis...
I, [2017-10-31T15:11:48.688947 #12]  INFO -- GridServiceSchedulerWorker: starting development/container-affinity/redis deploy
...
I, [2017-10-31T15:11:54.036320 #12]  INFO -- GridServiceDeployer: service development/container-affinity/redis has been deployed
I, [2017-10-31T15:11:54.053083 #12]  INFO -- GridServiceSchedulerWorker: deploying dependent service development/container-affinity/bar of deployed service development/container-affinity/redis
I, [2017-10-31T15:11:54.088635 #16]  INFO -- GridServiceSchedulerWorker: starting development/container-affinity/bar deploy
....
I, [2017-10-31T15:11:54.307250 #12]  INFO -- StackDeployWorker: skipping deployment of development/container-affinity/bar because it will be deployed by dependencies
...
I, [2017-10-31T15:11:59.849990 #16]  INFO -- GridServiceDeployer: service development/container-affinity/bar has been deployed
```

server/app/models/grid_service.rb
```diff
@@ -244,12 +244,18 @@ def dependant_services
...
-        {:volumes_from => {:$regex => /^#{self.name}-\d+/}},
-        {:affinity => "service==#{self.name}"},
-        {:affinity => "service!=#{self.name}"},
-        {:net => {:$regex => /^container:#{self.name}-%s/}},
-        {:net => {:$regex => /^container:#{self.name}-\d+/}}
+        # if this is a stackless service, this also includes bare volumes_from service names in other stacks
+        {:volumes_from => {:$regex => /^#{self.name_with_stack}-(%s|\d+)/}},
+
+        # the agent will resolve volumes_from: foo-1 to container stack.foo-1 if container foo-1 does not exist
+        {:volumes_from => {:$regex => /^#{self.name}-(%s|\d+)/}, :stack_id => self.stack.id},
```
SpComb: Only trigger dependent deploys for services within the same stack:

```
I, [2017-10-31T15:46:02.429285 #17]  INFO -- GridServiceDeployer: service development/volumes-from-1/redis has been deployed
I, [2017-10-31T15:46:02.484254 #17]  INFO -- GridServiceSchedulerWorker: deploying dependent service development/volumes-from-1/backup of deployed service development/volumes-from-1/redis
```

server/app/models/grid_service.rb
```diff
@@ -244,12 +244,18 @@ def dependant_services
...
     dependant = []
     dependant += grid.grid_services.where(:$or => [
-        {:volumes_from => {:$regex => /^#{self.name}-%s/}},
-        {:volumes_from => {:$regex => /^#{self.name}-\d+/}},
-        {:affinity => "service==#{self.name}"},
-        {:affinity => "service!=#{self.name}"},
-        {:net => {:$regex => /^container:#{self.name}-%s/}},
-        {:net => {:$regex => /^container:#{self.name}-\d+/}}
+        # if this is a stackless service, this also includes bare volumes_from service names in other stacks
+        {:volumes_from => {:$regex => /^#{self.name_with_stack}-(%s|\d+)/}},
```
SpComb: The `#{self.name_with_stack}` form is actually impossible to use for stack services because of how the [dependency filter](https://github.com/kontena/kontena/blob/3d608c58a0c0ce046e3a3f2065c39431b2fbc13f/server/app/services/scheduler/filter/dependency.rb#L33-L34) is written: `Cannot find applicable node for service instance development/volumes-from/backup-1: Did not find any nodes for service dependency volumes_from: ["volumes-from.redis-1"]`.

The way that this works for stackless services is also broken re #2982

SpComb:
Needs merge with #3100, this can be simplified now that the `dependent_services` logic is gone.

The stack-scoping of `volumes_from` is still very badly broken, but that should be a separate PR because it also needs agent fixes.

SpComb:
Merged with #3100 to remove the `GridService#dependant_services` changes. The `Scheduler::Filter::Affinity` parts should all still work the same.

PR #2988, opened 2017-10-31
Kontena certificate import/export

### LINKED ISSUE #2825: kontena certificates cannot be imported/exported
The new `kontena certificate` model only works for certificates requested from LE, and it is also very difficult to inspect the actual certificate, to see any relevant X509 details.

With the old `secret` based model, you could both inspect the managed LE certificate details (`kontena vault read --value LE_CERTIFICATE_test_188_226_131_23_nip_io_CERTIFICATE | openssl x509 -noout -text`) and also import your own certificates (`kontena vault write foo_CERTIFICATE < cert.pem`).

There should be some kind of `kontena certificate export` / `kontena certificate import` workflow too. As a bonus, the CLI/server could now also validate any imported certificates, avoiding the need to [debug invalid certificates](https://github.com/kontena/kontena-loadbalancer/blob/c8014e93c010b3f0fefa277f1708e7dbe8e413b4/entrypoint.sh#L45) imported as secrets...

SpComb:
Depends on #3011 :white_check_mark: 
Fixes #2825

Adds new `kontena certificate import/export` commands, uses those for a true e2e spec that deploys `kontena/lb` with a generated cert and curls it to check that the certificate actually gets deployed correctly :tada: 

* Add `kontena certificate import --private-key=key.pem --chain ca.pem cert.pem` -> `PUT /v1/certificates/:grid/:subject` `{"certificate": ..., "private_key": ..., "chain": [ ... ]}` -> `GridCertificates::Import`
* Add `kontena certificate export` -> `GET /v1/certificates/:grid/:subject/export` (with audit-logging)
* e2e specs for `kontena certificate import/export`
* e2e spec for `kontena/lb` with a generated and imported certificate

## TODO
- [x] API docs
- [ ] User docs


cli/lib/kontena/cli/certificate/import_command.rb
```diff
@@ -0,0 +1,26 @@
...
+    def execute
+      certificate = spinner "Importing certificate from #{cert_file}..." do
+        client.post("grids/#{current_grid}/certificates",
+          certificate: File.read(self.cert_file),
+          private_key: File.read(self.key_file),
+          chain: chain_file_list.map{|chain_file| File.read(chain_file)},
+        )
+      end
+
+      puts YAML.dump(certificate)
```
jakolehm: Why raw dump?

SpComb: The same code as in the `kontena certificate show' command... the fix for that is WiP over in #2966

server/app/routes/v1/grids/grid_certificates.rb
```diff
@@ -1,11 +1,18 @@
...
   # POST /v1/grids/:name/certificates
   r.post do
     data = parse_json_body
     data[:grid] = @grid
-    outcome = GridCertificates::RequestCertificate.run(data)
+
+    if data['domains']
+      outcome = GridCertificates::RequestCertificate.run(data)
+    elsif data['certificate']
+      outcome = GridCertificates::Import.run(data)
```
SpComb: Needs updates to API docs... and maybe these should be separate API endpoints rather than having two different JSON request payloads.

SpComb:
> Not the most best practice to add new commands in a patch release but maybe this is exception to the rule (because certificate is too broken)?

Yes... this is mostly just new functionality without changes to existing code/functionality, and it "fixes" a issues with the new certificates: you can't use it for non-LE certs, so you'd end up having to support both the old `secrets` certs and the new `certificates` in an LB stack... AFAIK with this + #2990 it should be possible to just completely replace the old `secrets` certs with the new `certificates` certs.

cli/lib/kontena/cli/certificate/import_command.rb
```diff
@@ -0,0 +1,26 @@
...
+    option ['--chain'], 'CHAIN_FILE', "Path to CA cert chain file", :multivalued => true, :attribute_name => :chain_file_list
+
+    requires_current_master
+    requires_current_master_token
+    requires_current_grid
+
+    def execute
+      certificate = spinner "Importing certificate from #{cert_file}..." do
+        client.post("grids/#{current_grid}/certificates",
+          certificate: File.read(self.cert_file),
```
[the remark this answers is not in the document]
SpComb: Using clamp usage errors:

```
$ bin/kontena certificate import --key test/ssl/asdf.pem test/ssl/cert.pem
ERROR: option '--key': No such file: test/ssl/asdf.pem

See: 'kontena certificate import --help'
```
```
$ bin/kontena certificate import --key test/ssl/key.pem test/ssl/asdf.pem
ERROR: parameter 'CERT_FILE': No such file: test/ssl/asdf.pem

See: 'kontena certificate import --help'
```

server/app/routes/v1/grids/grid_certificates.rb
SpComb: @jakolehm @jnummelin opinions? There's now two separate `POST /v1/grids/:grid/certificates` APIs with different JSON payloads: 

* https://github.com/kontena/kontena/blob/9a976607ac0a7ca28aa89ec2b16fb85c1dbe39f9/server/docs/source/index.html.md#request-lets-encrypt-certificate
* https://github.com/kontena/kontena/blob/9a976607ac0a7ca28aa89ec2b16fb85c1dbe39f9/server/docs/source/index.html.md#import-certificate

cli/lib/kontena/cli/certificate/import_command.rb
```diff
@@ -0,0 +1,39 @@
...
+    def execute
+      certificate = spinner "Importing certificate from #{cert_file}..." do
+        client.post("grids/#{current_grid}/certificates",
+          certificate: self.cert_file.read(),
+          private_key: self.key_file.read(),
+          chain: chain_file_list.map{|chain_file| chain_file.read() },
+        )
+      end
+
+      puts YAML.dump(certificate)
```
jakolehm: Pls, no yaml dumps :)

server/app/routes/v1/grids/grid_certificates.rb
jnummelin: IMO Two different payloads for same `POST` API sounds bad. Why not use `PUT /v1/certificates/my-grid/example.com` to import the cert?

cli/lib/kontena/cli/certificate/import_command.rb
SpComb: Can we decide how to show these over in #2966, and then I can use the same code here? I don't want to write out the `kontena certificate show` code here

server/app/routes/v1/grids/grid_certificates.rb
jakolehm: I agree with @jnummelin .

SpComb: I can change this to use `PUT /v1/certificates/:grid/:subject`, but then that also means that the `kontena certificate import` CLI command also needs to know the cert subject beforehand... I figure the CLI would take the expected subject CN as an argument, and then the server `GridCertificates::Import` mutation needs to validate that the certificate subject actually matches the expected subject?

jnummelin: hmm, why couldn't CLI look into the cert and pass the correct subject?

SpComb: I can also do that... then both the CLI and the server will need to have some code to load the cert, but that shouldn't be too bad.

cli/lib/kontena/cli/certificate/import_command.rb
SpComb: This is now using the `show_certificate` from #2966

```
$ bin/kontena certificate import --key test/ssl/key.pem --chain test/ssl/ca.pem test/ssl/cert.pem
 [done] Importing certificate from test/ssl/cert.pem...     
development/test:
  subject: test
  valid until: 2017-11-30T19:30:52Z
  alt names:
    - test-1
  auto renewable: false
```

server/app/routes/v1/grids/grid_certificates.rb
SpComb: Changed to use `PUT /v1/certificates/:grid/:subject` for the import. The server validates that the imported cert matches the subject. The CLI has an optional `--subject` param that defaults to the loaded cert's subject.

SpComb:
The new e2e specs in this PR are still passing (unmodified) with the changes.

```

certificate import
  exports certificate bundle
  exports certificate
  exports private key
  exports chain
  logs audit

certificate import
  imports certificates from file

kontena/lb certificates
  deploys the certificate to the LB for https

Finished in 37.34 seconds (files took 0.65096 seconds to load)
7 examples, 0 failures
```

cli/lib/kontena/cli/certificate/import_command.rb
```diff
@@ -0,0 +1,61 @@
...
+    include Common
+
+    # @raise [ArgumentError]
+    def open_file(path)
+      File.open(path)
+    rescue Errno::ENOENT
+      raise ArgumentError, "File not found: #{path}"
+    end
+
+    parameter 'CERT_FILE', "Path to PEM-encodede X.509 certificate file" do |path|
```
jnummelin: PEM-encodede -> PEM-encoded

Issue #1790, opened 2017-02-05
health_check to support other states than 200


matti:
currently `health_check:` expects endpoint to return 200.

There are some applications that just don't return 200 from any url (when they are initially deployed, at least)

jakolehm:
Related #1760

jnummelin:
#1760 would change < 400 statuses to be accepted as `healthy´. @matti Would that cover your requirement?

matti:
Yes, but we have other cases where 403 is to be considered healthy (http basic auth means healthy)

jnummelin:
Hmm, that would be problematic since defining http health_check also means that LB will use that same endpoint for checking the health. And HAProxy expects 2xx or 3xx for a healthy check.

matti:
I guess you can't have it all then.

jnummelin:
Well, HAProxy supports also customizing the expected status code, so maybe a regex based approach could be taken.

SpComb:
I think the expected health-check status codes really depend on the application, there is no one correct set of expected http status codes for all applications. Should we support a configurable health check status whitelist instead?

```yaml
health_check:
  protocol: http
  http_status:
  - 200
  - 403
```

This would be some kind of `io.kontena.health_check.http_status=200,403` label, used by the agent `ContainerHealthCheckWorker`, and compiled into a haproxy `http-check expect rstatus ^(200|403)$` config?

matti:
influxdb /ping endpoint returns 204, needs to use /debug/vars that returns 200, but also other stuff

SpComb:
Partially fixed in #3006 for 1.4.1 to accept any 2xx reponse as healthy.

SpComb:
Looks to me like most* other HTTP-based healthchecks are hardcoded/default to accepting 2xx and 3xx as healthy. Maybe we should just tweak the agent http healthcheck to match the behavior of e.g. haproxy, and call it a day so that we can close this issue.

Unclear if we really need/want the extra complexity of configurable healthcheck statuses...

PR #3006, opened 2017-11-05
Fix agent healthchecks to accept any HTTP 2xx response as healthy

jakolehm:
Fixes partially #1790

### DIFF
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -9,7 +9,7 @@ module Kontena::Workers
     include Kontena::Helpers::PortHelper
     include Kontena::Helpers::RpcHelper
 
-    HEALTHY_STATUSES = (200 .. 208).to_a.freeze
+    HEALTHY_STATUSES = (200 .. 299).to_a.freeze
 
     finalizer :log_exit
 
--- a/agent/lib/kontena/workers/container_health_check_worker.rb
+++ b/agent/lib/kontena/workers/container_health_check_worker.rb
@@ -9,7 +9,7 @@ module Kontena::Workers
     include Kontena::Helpers::PortHelper
     include Kontena::Helpers::RpcHelper
 
-    HEALTHY_STATUSES = [200]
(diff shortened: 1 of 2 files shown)


agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -9,7 +9,7 @@ class ContainerHealthCheckWorker
     include Kontena::Helpers::PortHelper
     include Kontena::Helpers::RpcHelper
 
-    HEALTHY_STATUSES = [200]
+    HEALTHY_STATUSES = (200 .. 208).to_a.freeze
```
SpComb: Should this just be `200...300` to also cover any custom codes up to 299?

The `to_a` should be unnecessary?

```
irb(main):005:0> (200...300).include? 206
=> true
```

SpComb:
Accepting any 2xx status as healthy sounds reasonable to me from a design perspective. Complete fix for #1790 needs some kind of configuration/label, but this is a good first step IMO.

agent/lib/kontena/workers/container_health_check_worker.rb
SpComb: Should be `HEALTHY_STATUSES = (200...300)` to actually match any 2xx status codes, not just up to 208. There are [standardized](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success) HTTP response codes beyond 208.

jakolehm:
@SpComb PTAL.

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -9,7 +9,7 @@ class ContainerHealthCheckWorker
     include Kontena::Helpers::PortHelper
     include Kontena::Helpers::RpcHelper
 
-    HEALTHY_STATUSES = [200]
+    HEALTHY_STATUSES = (200 .. 299).to_a.freeze
```
SpComb: Any particular reason to use `to_a` here? I just imagine that `Range#include?` is going to be more efficient than `Array#include?` with 100 elements, but it doesn't really matter.

PR #3009, opened 2017-11-06
CLI: Set master name from KONTENA_MASTER when configuring from ENV

kke:
When configuring the CLI from ENV variables, this PR makes it possible to set the current master name using the `KONTENA_MASTER` environment variable (which is also used to select a master from the `.kontena_client.json` by name). 

Before this PR the ENV-configured `KONTENA_URL` master was always named "default".


kke:
Fixed in #3090

PR #3016, opened 2017-11-07
CLI: Upgrade excon dependency to 0.59

kke:
Plugins do not seem to use any of the `Excon::Errors` classes, just `Excon.new` and  `Excon.defaults`.


SpComb:
Are there any known issues fixed by the newer version of excon? Should this go in 1.4.1, or 1.5?

kke:
No, maybe 1.5. Just trying to keep up with the dependencies, it's less work than trying to upgrade from legacy versions.

PR #3021, opened 2017-11-07
Fix CLI specs to not give false positives on SystemExit

SpComb:
The CLI specs were trapping `SystemExit`, and allowing any specs that hit `exit_with_error` to pass: https://github.com/kontena/kontena/blob/d2a8ef67cffbe62805f0b61419f9a74df84c3ae8/cli/spec/spec_helper.rb#L47-L55

Rescuing the `SystemExit` also bypassed any `except` assertions after/around the failing call, which includes `expect{...}.to output(...).to_stdout` assertions:

```ruby
    it 'should abort with error message if master is not configured' do
      expect { subject.run(['not_existing']) }.to output(/something that the command does not output/).to_stderr
      expect(true).to be false
    end
```

```
$ rspec --format doc --color spec/kontena/cli/master/use_command_spec.rb -e 'if master is not configured'

Kontena::Cli::Master::UseCommand
  #use
Got SystemExit: exit - Exit code: 1
    should abort with error message if master is not configured

Finished in 0.0288 seconds (files took 0.40903 seconds to load)
1 example, 0 failures
```

## Broken specs

Some specs were buggy but still passing because of this bug. Not aware of any bugs in the actual CLI code that were hidden by this bug, though:

```
  1) Kontena::Cli::Nodes::RemoveCommand for an online node without a node token does not remove the node
     Failure/Error: expect{subject.run(['node-1'])}.to exit_with_error.and output(" [error] Node node-1 is still online. You must terminate the node before removing it.\n").to_stderr
       expected block to output " [error] Node node-1 is still online. You must terminate the node before removing it.\n" to stderr, but output " [error] Node node-1 is still connected using a grid token. You must terminate the node before removing it.\n"
     # ./spec/kontena/cli/nodes/remove_command_spec.rb:52:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:52:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:48:in `catch'
     # ./spec/spec_helper.rb:48:in `block (2 levels) in <top (required)>'

  2) Kontena::Cli::Stacks::UpgradeCommand#execute --no-deploy option does not trigger deploy
     Failure/Error: fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
     
     RuntimeError:
       SystemExit with code 1: 
       * Reading stack stack-a from master.. 
       * Reading stack stack-a from master.. done
       * Parsing /home/kontena/kontena/kontena/cli/spec/fixtures/kontena_v3.yml.. 
       * Parsing /home/kontena/kontena/kontena/cli/spec/fixtures/kontena_v3.yml.. done
       * Analyzing upgrade.. 
       * Analyzing upgrade.. done
       SERVICES:
       These services will be removed from master:
        - stack-a/foo
     
       These new services will be created to master:
        - stack-a/wordpress
        - stack-a/mysql
     
       STACKS:
       These stacks will be upgraded:
       - stack-a
     
     
       Warning: This can not be undone, data will be lost.
        [error] Command requires --force
     # ./spec/spec_helper.rb:54:in `rescue in block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:58:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:48:in `catch'
     # ./spec/spec_helper.rb:48:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # SystemExit:
     #   exit
     #   ./lib/kontena/command.rb:222:in `exit'

  3) Kontena::Cli::Master::UseCommand#use should abort with error message if master is not configured
     Failure/Error: expect { subject.run(['not_existing']) }.to exit_with_error.and output(/Could not resolve master with name: 'not_existing'/).to_stderr
       expected block to output /Could not resolve master with name: 'not_existing'/ to stderr, but output " [error] Could not resolve master by name 'not_existing'. For a list of known masters please run: kontena master list\n"
     # ./spec/kontena/cli/master/use_command_spec.rb:19:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:52:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:48:in `catch'
     # ./spec/spec_helper.rb:48:in `block (2 levels) in <top (required)>'

Finished in 0.28216 seconds (files took 0.62986 seconds to load)
18 examples, 3 failures

Failed examples:

rspec ./spec/kontena/cli/nodes/remove_command_spec.rb:49 # Kontena::Cli::Nodes::RemoveCommand for an online node without a node token does not remove the node
rspec ./spec/kontena/cli/stacks/upgrade_command_spec.rb:73 # Kontena::Cli::Stacks::UpgradeCommand#execute --no-deploy option does not trigger deploy
rspec ./spec/kontena/cli/master/use_command_spec.rb:18 # Kontena::Cli::Master::UseCommand#use should abort with error message if master is not configured
```

### DIFF
--- a/cli/spec/spec_helper.rb
+++ b/cli/spec/spec_helper.rb
@@ -48,8 +48,8 @@ RSpec.configure do |config|
     catch :exit_with_error do
       begin
         example.run
-      rescue SystemExit
-        puts "Got SystemExit: #{$!.message} - Exit code: #{$!.status}"
+      rescue SystemExit => exc
+        fail "SystemExit with code #{exc.status}: #{exc.message}"
       end
     end
   end
--- a/cli/spec/spec_helper.rb
+++ b/cli/spec/spec_helper.rb
@@ -45,6 +45,8 @@ RSpec.configure do |config|
   end
 
   config.around(:each) do |example|
+    stdout = $stdout
+    stderr = $stderr
     $stdout = $stderr = StringIO.new
 
     begin
@@ -52,8 +54,8 @@ RSpec.configure do |config|
     rescue SystemExit => exc
       fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
     ensure
-      $stdout = STDOUT
-      $stderr = STDERR
+      $stdout = stdout
+      $stderr = stderr
     end
   end
 end
--- a/cli/spec/spec_helper.rb
+++ b/cli/spec/spec_helper.rb
@@ -46,23 +46,16 @@ RSpec.configure do |config|
 
   config.around(:each) do |example|
     catch :exit_with_error do
+      $stdout = $stderr = StringIO.new
+
       begin
         example.run
       rescue SystemExit => exc
-        fail "SystemExit with code #{exc.status}: #{exc.message}"
+        fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
+      ensure
+        $stdout = STDOUT
+        $stderr = STDERR
       end
-    end
-  end
-
-  unless ENV["DEBUG"]
-    config.before(:each) do
-      $stdout = StringIO.new
-      $stderr = StringIO.new
-    end
-
-    config.after(:each) do
-      $stdout = STDOUT
-      $stderr = STDERR
     end
   end
 end
--- a/cli/spec/spec_helper.rb
+++ b/cli/spec/spec_helper.rb
@@ -45,17 +45,15 @@ RSpec.configure do |config|
   end
 
   config.around(:each) do |example|
-    catch :exit_with_error do
-      $stdout = $stderr = StringIO.new
+    $stdout = $stderr = StringIO.new
 
-      begin
-        example.run
-      rescue SystemExit => exc
-        fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
-      ensure
-        $stdout = STDOUT
-        $stderr = STDERR
-      end
+    begin
+      example.run
+    rescue SystemExit => exc
+      fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
+    ensure
+      $stdout = STDOUT
+      $stderr = STDERR
     end
   end
 end
--- a/cli/spec/kontena/cli/cloud/login_command_spec.rb
+++ b/cli/spec/kontena/cli/cloud/login_command_spec.rb
@@ -20,18 +20,15 @@ describe Kontena::Cli::Cloud::LoginCommand do
   end
 
   it 'should give error if trying to use --code and --force' do
-    expect(subject).to receive(:exit_with_error).and_throw(:exit_with_error)
-    subject.run(['--code', 'abcd', '--force'])
+    expect{subject.run(['--code', 'abcd', '--force'])}.to exit_with_error
   end
 
   it 'should give error if trying to use --token and --force' do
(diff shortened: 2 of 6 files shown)


cli/spec/spec_helper.rb
```diff
@@ -46,25 +46,18 @@
...
   config.around(:each) do |example|
     catch :exit_with_error do
+      $stdout = $stderr = StringIO.new
+
       begin
         example.run
-      rescue SystemExit
-        puts "Got SystemExit: #{$!.message} - Exit code: #{$!.status}"
+      rescue SystemExit => exc
+        fail "SystemExit with code #{exc.status}: \n#{$stderr.string}"
```
SpComb: This needs something to include the output from `exit_with_error`, or specs that fail with `SystemExit` are impossible to figure out, because the actual `exit` call is so obfuscated: e.g. this case is failing in [`Kontena::Cli::Common#confirm`](https://github.com/kontena/kontena/blob/d2a8ef67cffbe62805f0b61419f9a74df84c3ae8/cli/lib/kontena/cli/common.rb#L270) because of the lack of `--force`, but it's next to impossible to figure that out:

```
  3) Kontena::Cli::Stacks::UpgradeCommand#execute --no-deploy option does not trigger deploy
     Failure/Error: fail "SystemExit with code #{exc.status}: #{exc.message}"
     
     RuntimeError:
       SystemExit with code 1: exit
     # ./spec/spec_helper.rb:52:in `rescue in block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:49:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:48:in `catch'
     # ./spec/spec_helper.rb:48:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # SystemExit:
     #   exit
     #   ./lib/kontena/cli/common.rb:168:in `exit'
```

Running the specs with `DEBUG=true` to not capture `$stdout`/`$stderr` does not help at all, because it also changes the behavior of `confirm`, and also the error handling for exceptions within the command themselves...

cli/spec/spec_helper.rb
```diff
@@ -45,22 +45,13 @@
   end
 
   config.around(:each) do |example|
-    catch :exit_with_error do
```
SpComb: Replaced specs using `expect(subject).to receive(:exit_with_error).and_throw(:exit_with_error)` with `expect{...}.to exit_with_error`, because I suspect this `catch :exit_with_error do ...` would also behave badly as regards `except` asssertions in the specs.

cli/spec/spec_helper.rb
```diff
@@ -45,22 +45,13 @@
...
-      end
-    end
-  end
-
-  unless ENV["DEBUG"]
-    config.before(:each) do
-      $stdout = StringIO.new
-      $stderr = StringIO.new
-    end
+    $stdout = $stderr = StringIO.new
```
kke: This should store the original `$stdout` / `$stderr` values and restore those in ensure instead of `$stdout = STDOUT`.

SpComb: That would probably be better than what the the original `before/after` wrapper did, yes.

cli/spec/spec_helper.rb
kke: Yeah it was some leftover from times before `expect{}.to exit_with_error`

PR #3023, opened 2017-11-07
Fix server omit container io.kontena.health_check.uri label for non-http health-check protocols

### LINKED ISSUE #3020: Agent configures LB HTTP health checks for a HTTP LB service with health_check.protocol=tcp
If I have a service with http health check and change that to tcp check, agent still goes and configures `.../health_check_uri` on Etcd. This makes LB config to use that endpoint as the health check for the upstreams.

LB config `health_check_uri`  should only be set if the health check protocol is `http`

## Workaround:

```
kontena etcd rm /kontena/haproxy/lb-stack/lb/services/service-a/health_check_uri
```

jnummelin:
fixes #3020 

Server will no longer send the URI for TCP checks. So the combination of `tcp` check and `http` mode LB link works. Even changing from `http` to `tcp` makes the agent to remove the URI from etcd as it gets the `io.kontena.health_check.uri` label as `nil`. 

I think it makes sense to have this kind of logic only on the server side.

# Testing

```
$ kontena service create --health-check-protocol http --health-check-port 80 nginx-health nginx:alpine
 [done] Creating nginx-health service      

$ kontena service deploy nginx-health
 [done] Deploying service nginx-health      
⊛ Deployed instance e2e/null/nginx-health-1 to node foo

$ docker inspect nginx-health-1 | jq .[0].Config.Labels | grep uri
  "io.kontena.health_check.uri": "/",
```
Now to change health check to `tcp` mode:
```
$ kontena service update --health-check-protocol tcp  --health-check-port 80 nginx-health
 [done] Updating nginx-health service      

$ kontena service deploy nginx-health
 [done] Deploying service nginx-health      
⊛ Deployed instance e2e/null/nginx-health-1 to node foo

$ docker inspect nginx-health-1 | jq .[0].Config.Labels | grep uri
<empty>
```

### DIFF
--- a/server/app/serializers/rpc/service_pod_serializer.rb
+++ b/server/app/serializers/rpc/service_pod_serializer.rb
@@ -118,7 +118,7 @@ module Rpc
         labels['io.kontena.load_balancer.mode'] = mode
       end
       if service.health_check && service.health_check.protocol
-        labels['io.kontena.health_check.uri'] = service.health_check.uri
+        labels['io.kontena.health_check.uri'] = service.health_check.uri if service.health_check.protocol == 'http'
         labels['io.kontena.health_check.protocol'] = service.health_check.protocol
         labels['io.kontena.health_check.interval'] = service.health_check.interval.to_s
         labels['io.kontena.health_check.timeout'] = service.health_check.timeout.to_s
--- a/server/spec/serializers/rpc/service_pod_serializer_spec.rb
+++ b/server/spec/serializers/rpc/service_pod_serializer_spec.rb
@@ -242,6 +242,16 @@ describe Rpc::ServicePodSerializer do
         expect(labels).to include('io.kontena.health_check.initial_delay' => '10')
       end
 
+      it 'does not include health uri if tco check' do
+        service.health_check = GridServiceHealthCheck.new(port: 80, protocol: 'tcp')
+        expect(labels).to include('io.kontena.health_check.protocol' => 'tcp')
+        expect(labels).not_to include('io.kontena.health_check.uri' => '/')
+        expect(labels).to include('io.kontena.health_check.port' => '80')
+        expect(labels).to include('io.kontena.health_check.interval' => '60')
+        expect(labels).to include('io.kontena.health_check.timeout' => '10')
+        expect(labels).to include('io.kontena.health_check.initial_delay' => '10')
+      end
+
       it 'includes no health check labels if protocol nil' do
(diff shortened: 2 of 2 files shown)


SpComb:
Confirm that without the `io.kontena.health_check.uri` label, the `/kontena/haproxy/.../services/.../health_check_uri` node is empty, and `kontena/lb` no longer configures the `option httpchk GET /`.

The haproxy backend servers are always [generated](https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#check) with [`server ... check`](https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#check), so in this scenario, haproxy will still keep doing TCP healthchecks on the HTTP service.

#### Before:
```
Nov  8 09:20:35 localhost haproxy[69]: Server healthcheck-tcp-whoami/healthcheck-tcp.whoami-1 is UP, reason: Layer7 check passed, code: 200, info: "OK", check duration: 0ms. 1 active and 0 backup servers online. 0 sessions requeued, 0 total in queue.
```

### After
```
Nov  8 10:37:24 localhost haproxy[93]: Server healthcheck-tcp-whoami/healthcheck-tcp.whoami-1 is UP, reason: Layer4 check passed, check duration: 0ms. 1 active and 0 backup servers online. 0 sessions requeued, 0 total in queue.
```

jnummelin:
IMO it's perfectly valid to always use TCP checks, if haproxy cannot connect to the proxied port, no point to try to send anything to there

SpComb:
> IMO it's perfectly valid to always default to TCP checks, if haproxy cannot connect to the proxied port, no point to try to send anything to there

Yeah, agreed. There's still a little bit of confusion still going on here because the LB will use the `KONTENA_LB_INTERNAL_PORT` for the TCP healthchecks instead of the `io.kontena.health_check.port`, but continuing to do TCP healthchecks for the HTTP service makes sense regardless.

SpComb:
It would be nice to clarify how the `health_check.protocol/uri` works re the LB healthchecks in the docs, but I don't know where in the docs that would happen - the [LB](https://kontena.io/docs/using-kontena/loadbalancer) docs don't have any mention of health checks, there should probably be a new section there to cover this + the #3006 equivalent for [haproxy `httpchk`](https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#4-option%20httpchk) ("responses 2xx and 3xx are
considered valid, while all other ones indicate a server failure")

PR #3024, opened 2017-11-08
Fix cli DEBUG=true with Kontena::Errors::StandardError

### LINKED ISSUE #2985: CLI NoMethodError: undefined method `debug?' for #<Kontena::MainCommand:0x000000026c6450>
CLI fails on unhandled exceptions like internal server errors in DEBUG mode:

```
DEBUG Request GET /v1/certificates/development/test.188.226.131.23.xip.io/export: 500 Internal Server Error: {"message":"Internal server error"}
NoMethodError: undefined method `debug?' for #<Kontena::MainCommand:0x000000026c6450>
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:244:in `rescue in run'
  /home/kontena/kontena/kontena/cli/lib/kontena/command.rb:200:in `run'
  bin/kontena:19:in `<top (required)>'
```

https://github.com/kontena/kontena/blob/3d608c58a0c0ce046e3a3f2065c39431b2fbc13f/cli/lib/kontena/command.rb#L244

Looks like the `Kontena::Command#run` rescue handler fails for anything that doesn't include `Kontena::Cli::Common` with `debug?` - that includes the top-level `Kontena::MainCommand`?

SpComb:
Really fixes #2985

The specs are a little annoying with the extra subcommands being defined, because those subcommands become global across all specs... but this bug is also annoying, because the behavior of the inherited `Kontena::Command` class depends on what mixin modules were included... This would have been avoided by #2243


kke:
You don't have to register the subcommands to `MainCommand`, just make `subject { TestClass.new('x') }`.

SpComb:
> Don't register test-subcommands to maincommand

That doesn't actually cover the affected code in the `MainCommand` class which is what really explodes for the case of errors raised from a subcommand that *does*  include `Kontena::Cli::Common`... but for this case it's okay, the test command classes are affected by the same bug too.

That does mean that the specs are now in the wrong place, they have nothing to do with `Kontena::MainCommand` anymore..

cli/spec/kontena/cli/main_command_spec.rb
```diff
@@ -1,45 +0,0 @@
-require "kontena/main_command"
-
-describe Kontena::MainCommand do
-
-  let(:subject) { described_class.new(File.basename($0)) }
-  describe '#subcommand_missing' do
```
SpComb: There were two different spec files for `Kontena::MainCommand`:

* `cli/spec/kontena/cli/main_command_spec.rb`
* `cli/spec/kontena/main_command_spec.rb`

Also moved the remaining specs in this file to `cli/spec/kontena/main_command_spec.rb`

Issue #2622, opened 2017-07-27
cli: external-registry rm is without https, while add is with


matti:
kontena external-registry rm registry.heroku.com
kontena external-registry add https://registry.heroku.com 

related: #2400

PR #3034, opened 2017-11-13
Fix CLI kontena external-registry add to use https:// by default

### LINKED ISSUE #2400: cli: external-registry without https says unprocessable entity
```
$ k external-registry add --options registry.something.com
 [fail] Adding registry.heroku.com to external registries
 [error] 422 : Unprocessable Entity (/v1/grids/mygrid/external_registries)
```

(requires https://, error could be nicer)

jakolehm:
Fixes #2400

### DIFF
--- a/cli/lib/kontena/cli/external_registries/add_command.rb
+++ b/cli/lib/kontena/cli/external_registries/add_command.rb
@@ -14,6 +14,8 @@ module Kontena::Cli::ExternalRegistries
       require_current_grid
       token = require_token
 
+      self.url = 'https://' + self.url unless self.url.start_with?('http')
+
       data = { username: username, password: password, email: email, url: url }
       spinner "Adding #{url.colorize(:cyan)} to external registries " do
         client(token).post("grids/#{current_grid}/external_registries", data)
--- a/cli/lib/kontena/cli/external_registries/add_command.rb
+++ b/cli/lib/kontena/cli/external_registries/add_command.rb
@@ -14,7 +14,7 @@ module Kontena::Cli::ExternalRegistries
       require_current_grid
       token = require_token
 
-      self.url = 'https://' + self.url unless self.url.start_with?('http')
+      self.url = "https://#{self.url}" unless self.url.start_with?('http')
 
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/external_registries/add_command.rb
```diff
@@ -14,6 +14,8 @@ def execute
       require_current_grid
       token = require_token
 
+      self.url = "https://#{self.url}" unless self.url.start_with?('http')
```
kke: Could be: `self.url.prepend('https://') unless self.url.start_with?('http')`

kke: (and it could be in the param definition `do` block)

```ruby
parameter '[URL]', 'Docker Registry url', default: 'https://index.docker.io/v2/' do |url|
  url.prepend('https://') unless url.start_with?('http')
end
```

jakolehm: I like `do` block but it needs to return always something.. and prepend modifies a string which is not very good with ruby 2.4 and later.

kke: Right, the block return is actually used, modifying the param does nothing.

SpComb:
This is a little simplistic, but not really a problem... TBH only the `URI.parse(...)` `host` and `port` are used, the schema and path gets ignored by the server... server might want some better validation of these though.

```
$ bin/kontena external-registry add -u test -e test@test -p 'test' htto://images.example.com
 [done] Adding https://htto://images.example.com to external registries      
$ bin/kontena external-registry add -u test -e test@test -p 'test' http.images.example.com
 [fail] Adding http.images.example.com to external registries      
 [error] 422 : Unprocessable Entity (/v1/grids/development/external_registries)
$ bin/kontena external-registry list
NAME                 USERNAME   EMAIL
htto                 test       test@test
```

SpComb:
This fixes #2622 (cli: external-registry rm is without https, while add is with )...  #2400 is fixed by #3043

PR #3036, opened 2017-11-13
Fix ping api to check that process is healthy

jakolehm:
Currently `/v1/ping` is mostly used for health checks. Problem is that it will always respond `pong` even if server process mostly dead (because of db connections etc). This PR fixes this situation by adding few critical checks to ping api call.

### DIFF
--- a/server/app/routes/v1/ping_api.rb
+++ b/server/app/routes/v1/ping_api.rb
@@ -1,5 +1,6 @@
 module V1
   class PingApi < Roda
+    include RequestHelpers
 
     plugin :json
     plugin :render, engine: 'json.jbuilder', views: 'app/views/v1/ping'
@@ -7,7 +8,13 @@ module V1
     route do |r|
       r.is do
         r.get do
-          render('show')
+          begin
+            Grid.count # test db connection
+            MongoPubsub.started? # test pubsub
+            render('show')
+          rescue
+            halt_request(500, {error: 'Internal server error'})
+          end
         end
       end
     end
--- a/server/app/routes/v1/ping_api.rb
+++ b/server/app/routes/v1/ping_api.rb
@@ -10,7 +10,7 @@ module V1
         r.get do
           begin
             Grid.count # test db connection
-            MongoPubsub.started? # test pubsub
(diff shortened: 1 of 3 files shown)


server/app/routes/v1/ping_api.rb
```diff
@@ -1,13 +1,20 @@
...
     plugin :json
     plugin :render, engine: 'json.jbuilder', views: 'app/views/v1/ping'
 
     route do |r|
       r.is do
         r.get do
-          render('show')
+          begin
+            Grid.count # test db connection
+            MongoPubsub.started? # test pubsub
```
SpComb: [`MongoPubsub.started?`](https://github.com/kontena/kontena/blob/dd68aadceb3f8d0cc6e447b54095ae60ea2f1e37/server/app/services/mongo_pubsub.rb#L137) doesn't actually check that the `MongoPubsub` actor is running and is still alive: it just checks that the default celluloid supervision container has been created during `Celluloid.boot`... this is not a good health-check, it's only meant for skipping stuff that happens early during server bootup before `Celluloid.boot`.

jakolehm: Good point, fixed.

server/app/routes/v1/ping_api.rb
```diff
@@ -1,13 +1,20 @@
...
     route do |r|
       r.is do
         r.get do
-          render('show')
+          begin
+            Grid.count # test db connection
+            MongoPubsub.actor.alive? # test pubsub
+            render('show')
+          rescue
+            halt_request(500, {error: 'Internal server error'})
```
SpComb: Maybe also log the exception for diagnostic purposes? Then it would show up in the logs if the healthcheck fails and the service gets restarted.

Not entirely sure how to include `Kontena::Logging` for `error exc` in the roda routes, though.

SpComb: `rescue => exc` and something to also pass it to `Logger#error` so that you get the error class + message + backtrace

jakolehm:
@SpComb PTAL

PR #3052, opened 2017-11-16
Travis e2e

jakolehm:


build/travis/test_e2e.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -ue
+
+cd test && \
+  bundle install --path vendor/bundle && \
+  bundle exec rake compose:setup && \
+  bundle exec rake
```
SpComb: The `bundle exec` here currently causes the specs relying on `kontena plugin install` to fail (vaugely related to #3148)

SpComb: To clarify, I don't think this is a problem, the e2e specs should be fixed to work when run with `bundle exec` instead: #3173

PR #3055, opened 2017-11-17
Make --email optional in external-registry add

### LINKED ISSUE #3054: kontena external-registry add requires deprecated -e option
Docker registry does not require that option on login anymore.

kke:
Fixes #3054

### DIFF
--- a/cli/lib/kontena/cli/external_registries/add_command.rb
+++ b/cli/lib/kontena/cli/external_registries/add_command.rb
@@ -6,8 +6,8 @@ module Kontena::Cli::ExternalRegistries
     parameter '[URL]', 'Docker Registry url', default: 'https://index.docker.io/v2/'
 
     option ['-u', '--username'], 'USERNAME', 'Username', required: true
-    option ['-e', '--email'], 'EMAIL', 'Email', required: true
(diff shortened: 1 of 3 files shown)


cli/lib/kontena/cli/external_registries/add_command.rb
```diff
@@ -6,8 +6,8 @@ class AddCommand < Kontena::Command
     parameter '[URL]', 'Docker Registry url', default: 'https://index.docker.io/v2/'
 
     option ['-u', '--username'], 'USERNAME', 'Username', required: true
-    option ['-e', '--email'], 'EMAIL', 'Email', required: true
     option ['-p', '--password'], 'PASSWORD', 'Password', required: true
+    option ['-e', '--email'], 'EMAIL', 'Email', required: false
```
jakolehm: Server mutation still requires it.

SpComb:
Has someone tested that the resulting `Docker::Image.create({...}, {..., email: nil})` actually works? It looks like that gets serialized to `X-Registry-Auth` with the base64-encoded `{"email": null}` JSON - does the Docker API ignore or reject that?

Not necessarily appropriate for 1.4.x if changing server API params to be optional.

PR #3065, opened 2017-11-20
Require --force if some items in node label rm list are missing

### LINKED ISSUE #3064: kontena node label rm exits with code 0 if label not found

kke:
Fixes #3064

Makes the `kontena node label remove` command exit with error when trying to remove a nonexisting label / labels.

You can force the removal of the ones that do exist by using `--force`.

Not sure if 1.4.3 or 1.5.0. In a way it's a breaking change. On the other hand, the wrong exit code could be considered a bug.

### DIFF
--- a/cli/lib/kontena/cli/nodes/labels/remove_command.rb
+++ b/cli/lib/kontena/cli/nodes/labels/remove_command.rb
@@ -5,14 +5,28 @@ module Kontena::Cli::Nodes::Labels
     parameter "NODE", "Node name"
     parameter "LABEL ...", "Labels"
 
+    option '--force', :flag, "Do not abort if items in label list are not found"
+
     requires_current_master
     requires_current_master_token
     requires_current_grid
 
     def execute
-      node = client.get("nodes/#{current_grid}/#{self.node}")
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/nodes/labels/remove_command.rb
```diff
@@ -5,14 +5,28 @@ class RemoveCommand < Kontena::Command
...
+      found_labels = label_list.uniq & node_data['labels']
+
+      if !force? && found_labels.size != label_list.uniq.size
+        missing = label_list - found_labels
+        exit_with_error "Label#{'s' if missing.size > 1} #{pastel.cyan(missing.join(', '))} not found on node #{pastel.cyan(node)}"
+      end
+
+      return nil if found_labels.empty?
+
+      data = { labels: node_data['labels'] - found_labels }
```
SpComb: This could probably be simplified to just `{ labels: node_data['labels'] - label_list }`, but that /should/ be equivalent to `found_labels`.

kke:
1.4.3 or 1.5.0?

SpComb:
Probably a normal bugfix for 1.5... it's not a regression in 1.4.

Issue #2733, opened 2017-08-24
Agent can hang with all supervised actors having been terminated but the websocket still connected and the node online


matti:
background: One node started having elevated CPU/IO

![slack_-_kontena_community](https://user-images.githubusercontent.com/3893/29668742-efa3b538-88e9-11e7-9123-ca1ece0d22e2.png)

When I was finally able to get in with SSH `docker logs --tail 100 -f kontena-agent` showed:

```
E, [2017-08-23T01:33:48.907969 #1] ERROR -- : Actor crashed!
Docker::Error::TimeoutError: read timeout reached
        /usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:52:in `rescue in request'
        /usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:38:in `request'
        /usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
        /usr/lib/ruby/gems/2.3.0/gems/docker-api-1.32.1/lib/docker/container.rb:223:in `block (2 levels) in <class:Container>'
        /app/lib/kontena/service_pods/restarter.rb:25:in `perform'
        /app/lib/kontena/workers/container_health_check_worker.rb:110:in `restart_container'
        /app/lib/kontena/workers/container_health_check_worker.rb:59:in `handle_action'
        /app/lib/kontena/workers/container_health_check_worker.rb:49:in `check_status'
        /app/lib/kontena/workers/container_health_check_worker.rb:31:in `block in start'
        /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/actor.rb:339:in `block in task'
        /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task.rb:44:in `block in initialize'
        /usr/lib/ruby/gems/2.3.0/gems/celluloid-0.17.3/lib/celluloid/task/fibered.rb:14:in `block in create'
W, [2017-08-23T04:18:49.722194 #1]  WARN -- Kontena::WebsocketClient: connection closed with code 1006: 
I, [2017-08-23T04:18:49.975822 #1]  INFO -- Kontena::WebsocketClient: connecting to master at wss://sensored
I, [2017-08-23T04:18:50.277794 #1]  INFO -- Kontena::WebsocketClient: connection established
[...]

Issue #3067, opened 2017-11-21
Agent: shutdown deadlock on LogWorker exclusive sleep affects automatic agent version upgrades


SpComb:
There is a deadlock in the `kontena-agent` 1.4.0 `Agent.shutdown` logic used to handle automatic agent version upgrades, which means that the agent will keep reconnecting to the server instead of shutting down. It seems like the agent will eventually shutdown once a 300s timeout expires, which should allow it to get re-started with the new version?

The effective deadlock is in `Kontena::WebsocketClient#on_close` => `Kontena::Agent.shutdown` => `Kontena::Workers::LogWorker#terminate` => `Kontena::Launchers::Etcd#running?`, because the `LogWorker` will be stuck in an exclusive sleep waiting for etcd actor to be ready, which will not happen until the websocket is connected and the server sends the node info.

It looks like this was introduced in #2728 for 1.4.0, because it wrapped the entire `LogWorker#start` method in `exclusive { ... }`, including the[ `wait_until!("etcd running") { Actor[:etcd_launcher].running? }`](https://github.com/kontena/kontena/blob/98dacbfc75006e73c361d94210ba1377e7da9302/agent/lib/kontena/workers/log_worker.rb#L78). It looks like `exclusive { sleep }` will also block actor termination, causing the `@supervisor.shutdown` to block.

---

[Tracing `Celluloid.shutdown`](https://gist.github.com/SpComb/1700fd30fca8f8f90bd52d5217c5c777), this exclusive sleep prevents `Kontena::Agent.shutdown` from completing if the etcd launcher is not yet running:

```
I, [2017-11-21T12:54:50.071965 #1]  INFO -- Kontena::WebsocketClient: connecting to master at ws://192.168.66.1:9292
I, [2017-11-21T12:54:50.099709 #1]  INFO -- Kontena::WebsocketClient: unsecure connection established without SSL
D, [2017-11-21T12:54:50.113744 #1] DEBUG -- Kontena::RpcServer: rpc notification: Kontena::Rpc::AgentApi#master_info [{"version"=>"1.4.0.dev"}]
D, [2017-11-21T12:54:50.117461 #1] DEBUG -- Kontena::WebsocketClient: Server closed connection with code 4010: agent version 1.5.0.dev is not compatible with server version 1.4.0.dev
[...]

PR #3068, opened 2017-11-21
Agent: Close websocket early during agent shutdown, and on actor termination

SpComb:
Mitigation for #2733 to have the agent disconnect from the master faster when shutting down or terminating itself

If the `Kontena::WebsocketClient` was terminated by Celluloid, the websocket may remain connected for a brief period of time, because the `connect_client` => `Celluloid.defer` thread (running `Kontena::Websocket::Client#read`) will still remain alive and responding to websocket pings: only the `connect_client` task/fiber gets terminated (`Celluloid.defer` raises `Celluloid::TaskTerminated`). This means means that the server may still see the node as online, even though `Celluloid.shutdown` was triggered or the supervisor actor crashed/terminated and terminated all supervised actors. 

As an implementation detail of the new `Kontena::WebsocketClient` actor design in 1.4, the `read` thread will crash on the first received websocket message, because the [`actor.on_message`](https://github.com/kontena/kontena/blob/v1.4.0/agent/lib/kontena/websocket_client.rb#L146) call will raise `Celluloid::DeadActorError` and crash the `defer` thread, dropping the last reference to the `Kontena::Websocket::Client` and closing the socket.

Fixes the `Kontena::WebsocketClient` to explicitly close the websocket connection early in the `Kontena::Agent` shutdown handler, as well as in the actor finalizer.

### DIFF
--- a/agent/lib/kontena/agent.rb
+++ b/agent/lib/kontena/agent.rb
@@ -112,8 +112,21 @@ module Kontena
       end
     end
 
+    # yield websocket client actor if exists and alive
+    #
+    def close_websocket_client(**options)
+      if websocket_client = Celluloid::Actor[:websocket_client]
+        websocket_client.close!(**options)
+      end
+    rescue => exc
+      error exc
+    end
+
     def handle_shutdown
       info "Shutting down..."
+
+      close_websocket_client reason: "Agent shutting down..."
+
       @supervisor.shutdown # shutdown all actors
       @write_pipe.close # let run! break and return
     end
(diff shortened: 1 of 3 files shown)


SpComb:
Because this PR introduces the `Kontena::WebsocketClient#closed?` state, I'm considering also fixing the reconnect-after-server-close bug in https://github.com/kontena/kontena/issues/3067#issuecomment-346029191... the `on_close` => `Kontena::Agent.shutdown` handlers just need to also set `@closed = true`?

jakolehm:
Should we move this to 1.5.0?

SpComb:
> Should we move this to 1.5.0?

I'm not aware of any specific 1.4 issues fixed by this PR, so doing it as a general cleanup in 1.5 sounds good.

SpComb:
@jakolehm still planning on taking a look at this for 1.5?

SpComb:
Ping @jnummelin not strictly required for 1.5, but it would be nice.

PR #2994, opened 2017-11-02
Fix server to only deploy pending/active domain authorization challenge certs

SpComb:
Fixes #2963 to only have the server deploy pending domain authorization challenge certs to the LB service
Also fixes #3025 to order the TLS-SNI challenge certs after any other `secrets`/`certificates`
Also fixes #3044 to mark invalid challenges as `error` and no longer deploy them after verification fails

Domain authorizations that have been validated, rejected or expired will be omitted from the next LB service deploy. Note that this doesn't actually trigger a deploy or service update to perturb the LB service: any deployed challenge certs will remain deployed, and get lazily cleaned up on the next actual deploy.

This should be enough for the Kontena master to automatically reduce the TLS-SNI-01 challenge cert bloat that causes issues re #2924 `SSL_CERTS` env size limits.

The domain authorization list/delete API/commands in #2964 may still be relevant in some more specialized cases, like accidentally trying to authorize the wrong domain, and wanting to cancel the domain authorization.

Issue #3025, opened 2017-11-08
TLS-SNI challenge certificates get deployed as the default LB certificate, breaking non-SNI connections


SpComb:
The SSL cert bundles in the `kontena/lb` `SSL_CERTS` env get split up into  `/etc/haproxy/certs/cert*_gen.pem` files, preserving the ordering. If haproxy receives a SSL connection without a SNI hostname, or with a SNI hostname that does not match any certificate, then the first `/etc/haproxy/certs/cert00_gen.pem` certificate will be loaded: https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#5.1-crt

> If no SNI is provided by the client or if the SSL library does not support TLS extensions, or if the client provides an SNI hostname which does not match any certificate, then the first loaded certificate will be presented. This means that when loading certificates from a directory, it is highly recommended to load the default one first as a file or to ensure that it will always be the first one in the directory.

I think Kontena should preserve the ordering of the service `certificates`, so placing the default certificate first should be enough to make sure it gets handled as the default certificate:

```yaml
  certificates: 
    - subject: test.188.226.131.23.xip.io
      name: SSL_CERTS
      type: env
    - subject: test2.188.226.131.23.xip.io
      name: SSL_CERTS
      type: env
```

```
$ curl -vk https://test.188.226.131.23.xip.io |& grep 'subject'
* 	 subject: CN=test.188.226.131.23.xip.io
$ curl -vk https://test2.188.226.131.23.xip.io |& grep 'subject'
* 	 subject: CN=test2.188.226.131.23.xip.io
$ curl -vk https://test4.188.226.131.23.xip.io |& grep 'subject'
* 	 subject: CN=test.188.226.131.23.xip.io
$ curl -vk https://188.226.131.23.nip.io |& grep 'subject'
* 	 subject: CN=test.188.226.131.23.xip.io
```

However, this goes horribly wrong if you use `kontena certificate authorize --type tls-sni-01`, because it turns out that the internal `SSL_CERT` secrets for the challenge certificates get generated before the `certificate` secrets: https://github.com/kontena/kontena/blob/724c94f779ed86c220e02db66cf60510ce21037c/server/app/serializers/rpc/service_pod_serializer.rb#L72-L86
```
$ bin/kontena certificate authorize --type tls-sni-01 --linked-service ingress-lb/lb test4.188.226.131.23.xip.io
 [done] Waiting for tls-sni-01 certificate to be deployed into development/ingress-lb/lb      
[...]

PR #3076, opened 2017-11-22
Deploy tls-sni challenge certs as separate SSL_CERT_acme_challenge_* envs

SpComb:
See #2963 and #2924 

Avoid consuming `SSL_CERTS` env size for TLS-SNI-01 domain authorizations by deploying the challenge certs using `SSL_CERT_acme_challenge_*`envs. This requires the service to be running an newer version of `kontena/lb` that includes https://github.com/kontena/kontena-loadbalancer/pull/28 - so I don't think we can actually merge and release this yet?

The Let's Encrypt challenge certs seem to be about 2.7KB each, so that's slightly over 2% of the total `SSL_CERTS` capacity per domain authorization challenge... #2994 should reduce the number of  authz challenges that get deployed, but without something like #2964 it would still be possible to fill up the LB service `SSL_CERTS`, and then be unable to deploy the service until the challenges expire or you manually remove them using `bin/kontena-console`.

### DIFF
--- a/server/app/serializers/rpc/service_pod_serializer.rb
+++ b/server/app/serializers/rpc/service_pod_serializer.rb
@@ -83,7 +83,9 @@ module Rpc
       # Why secrets? Well, secrets are already handled in a way they can be concatenated with same env names
       # Must be injected after any secrets/certificates, because the first certificate in `SSL_CERTS` is special
       service.grid_domain_authorizations.select{|d| d.deployable?}.each do |domain_auth|
-        secrets << {name: "SSL_CERTS", type: 'env', value: domain_auth.tls_sni_certificate}
+        env = "SSL_CERT_acme_challenge_#{domain_auth.domain.gsub(/[^a-z0-9]/, '_')}"
+
+        secrets << {name: env, type: 'env', value: domain_auth.tls_sni_certificate}
       end
 
       secrets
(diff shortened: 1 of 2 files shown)


SpComb:
The `SSL_CERT_acme_challenge_*` naming here also changes the sorting, as these challenge certs will now again sort before any `SSL_CERTS` => `certXX_gen.pem` files, and get used as the default cert (see #3025):

```
$ docker exec -it ingress-lb.lb-1 ls -la /etc/haproxy/certs
total 52
drwxr-xr-x. 2 root root 4096 Nov 22 15:15 .
drwxr-xr-x. 1 root root 4096 Nov 22 15:15 ..
-rw-r--r--. 1 root root 2722 Nov 22 15:15 SSL_CERT_acme_challenge_test3_188_226_131_23_xip_io.pem
-rw-r--r--. 1 root root 2718 Nov 22 15:15 SSL_CERT_acme_challenge_test5_188_226_131_23_xip_io.pem
-rw-r--r--. 1 root root 5181 Nov 22 15:15 SSL_CERT_test_188_226_131_23_xip_io.pem
-rw-r--r--. 1 root root 1500 Nov 22 15:15 cert00_gen.pem
```

That usecase would now need something specialized like `SSL_CERT__default` to sort before all the challenge certs.

PR #3082, opened 2017-11-23
Remove repository root Rakefile gem dependencies

### LINKED ISSUE #3073: Project root Rakefile requires colorize and dotenv gems
And there's no Gemfile.

kke:
Fixes #3073

Removes dependencies to Dotfile and Colorize gems which have to be manually installed as there's no Gemfile in the root.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,7 @@ begin
   require 'dotenv'
   Dotenv.load
 rescue LoadError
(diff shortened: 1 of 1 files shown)


SpComb:
I'm not entirely sure what the `.env` variables are used for in the rake tasks... if they're used in e.g. the release builds then just warning when unable to load might end up causing some damage?

PR #3096, opened 2017-12-04
Add --id to "kontena master token current"

### LINKED ISSUE #3072: kontena master token current --id
currently only `--token` and `--refresh-token`, `--id` would be helpful to remove all BUT current token

kke:
Fixes #3072

(Perhaps could wait for #2763)


SpComb:
Maybe also write some specs for the `kontena master token` commands?

kke:
> Maybe also write some specs for the `kontena master token` commands?

Issue: #3151

PR #3098, opened 2017-12-04
Suggest plugin install when using known cloud plugin subcommands

### LINKED ISSUE #3097: CLI: cloud subcommand error message unclear without plugin
When you don't have the cloud plugin installed, you get misleading error messages:

image

It should probably hint that you don't have the plugin installed.

kke:
Fixes #3097

Gives a hint about installing the cloud plugin when trying to use plugin subcommands without the plugin.

### DIFF
--- a/cli/lib/kontena/cli/cloud_command.rb
+++ b/cli/lib/kontena/cli/cloud_command.rb
@@ -1,9 +1,13 @@
 
 class Kontena::Cli::CloudCommand < Kontena::Command
+  include Kontena::Cli::Common
+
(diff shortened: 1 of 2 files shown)


kke:
> Although I'd also consider if it would make sense to just move all the kontena cloud ... commands into the plugin?

Yep it's a bit confusing, but the cloud plugin stuff mostly include things that are needed for hosted platforms. The non-plugin subcommands are for things like stack registry and cloud-authing selfhosted masters.

PR #3113, opened 2017-12-08
Add health check port to lb configs

### LINKED ISSUE #1709: LB health check port needs to be configurable
I have services that have their health checks on a separate port than the "regular" endpoints. This leads to problems because haproxy tries to query the configured health check path on the regular port instead of the management port and as far as I have seen, there is no way to configure this port yet.

It would be nice to be able to configure this.

jnummelin:
fixes #1709 

needs https://github.com/kontena/kontena-loadbalancer/pull/40 on LB side


jakolehm:
LB pr is now merged.

PR #3123, opened 2017-12-09
Stack inspect command

jakolehm:
Outputs stack yaml that is saved to the master api.


SpComb:
This command is useful, but it currently only outputs the raw stack YAML in human-readable source form, without any of the variable values or interpolations etc.

What about something that outputs the same kind of interpolated stack YAML as `kontena stack validate` does?

kke:
Maybe this should be `kontena stack pull` like the `kontena stack registry pull`.

PR #3124, opened 2017-12-10
Less shuffling when nodes missing with daemon strategy

jakolehm:
Example:

All nodes are online and service has been deployed with daemon strategy:

```
node-1: instance-1
node-2: instance-2
node-3: instance-3
node-4: instance-4
node-5: instance-5
```

node-2 goes to offline (or is terminated/removed):

old behaviour:

```
node-1: instance-1
node-2: 
node-3: instance-2
node-4: instance-3
node-5: instance-4
```

new behaviour:

```
node-1: instance-1
node-2: 
node-3: instance-3
node-4: instance-4
node-5: instance-2
```

### DIFF
--- a/server/app/services/scheduler/strategy/daemon.rb
+++ b/server/app/services/scheduler/strategy/daemon.rb
@@ -15,11 +15,33 @@ module Scheduler
         10.minutes
       end
 
+      # @param [GridService] grid_service
+      # @param [Integer] instance_number
+      # @param [Array<Scheduler::Node>] nodes
+      # @return [Scheduler::Node,NilClass]
+      def find_stateless_node(grid_service, instance_number, nodes)
+        prev_instance = grid_service.grid_service_instances.has_node.find_by(
+          instance_number: instance_number
+        )
+        if prev_instance
+          node = nodes.find { |n| n.node == prev_instance.host_node }
+          unless node
+            selector = (instance_number.to_f / nodes.size.to_f).floor
+            candidates = self.sort_candidates(nodes, grid_service, instance_number)
+            node = candidates.select { |c| c.schedule_counter <= selector }.last
+          end
+          node
+        else
+          candidates = self.sort_candidates(nodes, grid_service, instance_number)
+          candidates.first
+        end
+      end
+
       # @param [Array<Scheduler::Node>] nodes
       # @param [GridService] grid_service
       # @param [Integer] instance_number
       def sort_candidates(nodes, grid_service, instance_number)
-        nodes.sort_by{|node|
+        nodes.sort_by { |node|
           [node.schedule_counter, node.node_number]
         }
       end
--- a/server/spec/services/scheduler/strategy/daemon_spec.rb
+++ b/server/spec/services/scheduler/strategy/daemon_spec.rb
@@ -4,24 +4,114 @@ describe Scheduler::Strategy::Daemon do
   let(:grid) { Grid.create(name: 'test') }
 
   let(:host_nodes) do
-    [
+    10.times.map do |i|
+      n = i + 1
       HostNode.create!(
-      node_id: 'node1', name: 'node-1', connected: true, grid: grid, node_number: 1
-      ),
-      HostNode.create!(
-        node_id: 'node2', name: 'node-2', connected: true, grid: grid, node_number: 2
-      ),
-      HostNode.create!(
-        node_id: 'node3', name: 'node-3', connected: true, grid: grid, node_number: 3
-      ),
-    ]
+        node_id: "node#{n}", name: "node-#{n}", connected: true, grid: grid, node_number: n
+      )
+    end
   end
+
   let(:scheduler_nodes) do
     host_nodes.map{|n| Scheduler::Node.new(n)}
   end
 
   let(:stateless_service) do
     GridService.create!(name: 'test', grid: grid, image_name: 'foo/bar:latest', stateful: false)
+  end
+
+  describe '#find_node' do
+    it 'finds matching nodes' do
+      nodes = scheduler_nodes
+      nodes.each do |n|
+        node = subject.find_node(stateless_service, n.node_number, nodes)
+        node.schedule_counter += 1
+        expect(node).to eq(nodes[n.node_number - 1])
+      end
+    end
+
+    context 'with existing nodes' do
+      before(:each) do
+        host_nodes.each do |n|
+          stateless_service.grid_service_instances.create!(
+            host_node: n,
+            instance_number: n.node_number
+          )
+        end
+      end
+
+      it 'minimizes shuffle when nodes 2 & 3 are missing' do
+        nodes = scheduler_nodes.tap do |n|
+          n.delete_at(1)
+          n.delete_at(1)
+        end
+
+        scheduled = []
+        nodes.each_with_index do |n, i|
+          node = subject.find_node(stateless_service, i + 1, nodes)
+          node.schedule_counter += 1
+          scheduled << node
+        end
+        expect(scheduled.map {|s| s.node_number}).to eq([
+          1, 10, 9, 4, 5, 6, 7, 8
+        ])
+      end
+
+      it 'minimizes shuffle when nodes 2 & 3 are missing (2 instances per node)' do
+        nodes = scheduler_nodes.tap do |n|
+          n.delete_at(1)
+          n.delete_at(1)
+        end
+
+        scheduled = []
+        16.times do |i| # 2 instances per node
+          node = subject.find_node(stateless_service, i + 1, nodes)
+          node.schedule_counter += 1
+          scheduled << node
+        end
+        expect(scheduled.map {|s| s.node_number}).to eq([
+          1, 10, 9, 4, 5, 6, 7, 8,
+          9, 10, 1, 4, 5, 6, 7, 8
+        ])
+      end
+
+      it 'minimizes shuffle when last two nodes are missing' do
+        nodes = scheduler_nodes[0..-3]
+        scheduled = []
+        nodes.each_with_index do |n, i|
+          node = subject.find_node(stateless_service, i + 1, nodes)
+          node.schedule_counter += 1
+          scheduled << node
+        end
+        expect(scheduled.map {|s| s.node_number}).to eq([
+          1, 2, 3, 4, 5, 6, 7, 8
+        ])
+      end
+    end
+
+    context 'with newly added nodes' do
+      before(:each) do
+        host_nodes[0..-3].each do |n|
+          stateless_service.grid_service_instances.create!(
+            host_node: n,
+            instance_number: n.node_number
+          )
+        end
+      end
+
+      it 'finds correct nodes' do
+        nodes = scheduler_nodes
+        scheduled = []
+        nodes.each_with_index do |n, i|
+          node = subject.find_node(stateless_service, i + 1, nodes)
+          node.schedule_counter += 1
+          scheduled << node
+        end
+        expect(scheduled.map { |s| s.node_number}).to eq([
+          1, 2, 3, 4, 5, 6, 7, 8, 9, 10
+        ])
+      end
+    end
   end
 
   describe '#instance_count' do


server/app/services/scheduler/strategy/daemon.rb
```diff
@@ -15,11 +15,33 @@ def host_grace_period
...
+      def find_stateless_node(grid_service, instance_number, nodes)
+        prev_instance = grid_service.grid_service_instances.has_node.find_by(
+          instance_number: instance_number
+        )
+        if prev_instance
+          node = nodes.find { |n| n.node == prev_instance.host_node }
+          unless node
+            selector = (instance_number.to_f / nodes.size.to_f).floor
+            candidates = self.sort_candidates(nodes, grid_service, instance_number)
+            node = candidates.select { |c| c.schedule_counter <= selector }.last
```
SpComb: I'm not sure I understand what this is doing vs the `self.sort_candidates(nodes, grid_service, instance_number).first` - the `sort_candidates` already takes the `node.schedule_counter` into account, so that should end up picking the node with the fewest scheduled instances?

jakolehm: Updated description.. also specs should demonstrate how this behaves.

SpComb: Right, so the `.last` here is to effectively select those nodes that were running the high-numbered instances that are no longer part of the service after some hosts went offline. However, I'm a little concerned about this being a bit fragile... I think this assumes that e.g. the `sort_candidates` ranking of the nodes is stable across the scheduler runs... if the ranks change to put some host node with a lower-numbered instance at the end, then could you actually end up scheduling multiple instances onto that same node, while leaving the node that used to be running the high-numbered instances without any scheduled instances? It would need some actual testing to figure out if there are some kind of nasty corner cases like that...

I'm not sure how to make it less fragile either... perhaps it should explicitly select the `grid_service_instances.where(:instance_number.gt => total_instances)` nodes, but that's hard to fit in with the actual scheduler interface and the way it's used, particularly because the `GridServiceDeployer` deletes the high-numbered instances before actually scheduling the nodes.

And to answer my own question: if this was just `self.sort_candidates(...).first`, then it would end up scheduling multiple instances onto the same low-numbered nodes that already have instances running:

```
  1) Scheduler::Strategy::Daemon#find_node with existing nodes minimizes shuffle when nodes 2 & 3 are missing
     Failure/Error: expect(scheduled.map {|s| s.node_number}).to eq([
       
       expected: [1, 10, 9, 4, 5, 6, 7, 8]
            got: [1, 4, 5, 4, 5, 6, 7, 8]
```

SpComb: Oh NVM, the `daemon` strategy has its own `sort_candidates` which looks like it should be perfectly stable. I was looking at the `ha` strategy's `sort_candidates`.

server/app/services/scheduler/strategy/daemon.rb
```diff
@@ -15,11 +15,33 @@ def host_grace_period
         10.minutes
       end
 
+      # @param [GridService] grid_service
+      # @param [Integer] instance_number
+      # @param [Array<Scheduler::Node>] nodes
+      # @return [Scheduler::Node,NilClass]
+      def find_stateless_node(grid_service, instance_number, nodes)
+        prev_instance = grid_service.grid_service_instances.has_node.find_by(
+          instance_number: instance_number
```
SpComb: This results in O(N) queries again, per the number of daemon service instances?

```
D, [2017-12-14T10:14:08.879672 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | STARTED | {"find"=>"grid_service_instances", "filter"=>{"grid_service_id"=>BSON::ObjectId('58ad6ad2de3578692900007a'), "host_node_id"=>{"$ne"=>nil}, "instance_number"=>1}}
D, [2017-12-14T10:14:08.880512 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | SUCCEEDED | 0.000737478s
D, [2017-12-14T10:14:08.881683 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | STARTED | {"find"=>"host_nodes", "filter"=>{"_id"=>BSON::ObjectId('5811b7a3de3578398e000000')}, "limit"=>1, "singleBatch"=>true}
D, [2017-12-14T10:14:08.882532 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | SUCCEEDED | 0.000706158s
D, [2017-12-14T10:14:08.889784 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | STARTED | {"find"=>"grid_service_instances", "filter"=>{"grid_service_id"=>BSON::ObjectId('58ad6ad2de3578692900007a'), "host_node_id"=>{"$ne"=>nil}, "instance_number"=>2}}
D, [2017-12-14T10:14:08.890686 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | SUCCEEDED | 0.0007790340000000001s
D, [2017-12-14T10:14:08.891922 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | STARTED | {"find"=>"host_nodes", "filter"=>{"_id"=>BSON::ObjectId('59f9894f16b53b0014006aa4')}, "limit"=>1, "singleBatch"=>true}
D, [2017-12-14T10:14:08.892846 #9] DEBUG -- : MONGODB | mongo:27017 | kontena_development.find | SUCCEEDED | 0.00082344s
```

SpComb:
Experimenting with the algorithm here I came up with this version of `Scheduler::Strategy::Daemon#sort_candidates` which seems to give similar results without needing any modifications to `find_stateless_node`:

```ruby
      # @param [Array<Scheduler::Node>] nodes
      # @param [GridService] grid_service
      # @param [Integer] instance_number
      def sort_candidates(nodes, grid_service, instance_number)
        total_instances = nodes.size
        service_instances = grid_service.grid_service_instances.to_a
        service_instance = service_instances.find { |i| i.instance_number == instance_number }

        nodes.sort_by { |node|
          if service_instance && service_instance.host_node_id == node.id
            instance_rank = -1
          elsif service_instances.any? { |i| i.host_node_id == node.id && i.instance_number <= total_instances }
            instance_rank = 1
          else
            instance_rank = 0
          end

          [node.schedule_counter, instance_rank, node.node_number]
        }
      end
```

I think it's pretty clear for the case of `instances: 1`, although I'm not 100% sure why it even works for `instances: 2`...

The only difference in the spec cases is the order of the replacement nodes:

```
  1) Scheduler::Strategy::Daemon#find_node with existing nodes minimizes shuffle when nodes 2 & 3 are missing
     Failure/Error: expect(scheduled.map {|s| s.node_number}).to eq([
       
       expected: [1, 10, 9, 4, 5, 6, 7, 8]
            got: [1, 9, 10, 4, 5, 6, 7, 8]
       
       (compared using ==)
     # ./spec/services/scheduler/strategy/daemon_spec.rb:55:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'

  2) Scheduler::Strategy::Daemon#find_node with existing nodes minimizes shuffle when nodes 2 & 3 are missing (2 instances per node)
     Failure/Error: expect(scheduled.map {|s| s.node_number}).to eq([
       
       expected: [1, 10, 9, 4, 5, 6, 7, 8, 9, 10, 1, 4, 5, 6, 7, 8]
            got: [1, 9, 10, 4, 5, 6, 7, 8, 9, 10, 1, 4, 5, 6, 7, 8]
       
       (compared using ==)
     # ./spec/services/scheduler/strategy/daemon_spec.rb:72:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'
```

server/spec/services/scheduler/strategy/daemon_spec.rb
```diff
@@ -24,6 +20,100 @@
...
+
+        scheduled = []
+        16.times do |i| # 2 instances per node
+          node = subject.find_node(stateless_service, i + 1, nodes)
+          node.schedule_counter += 1
+          scheduled << node
+        end
+        expect(scheduled.map {|s| s.node_number}).to eq([
+          1, 10, 9, 4, 5, 6, 7, 8,
+          9, 10, 1, 4, 5, 6, 7, 8
```
SpComb: Hmm... strictly speaking shouldn't this case be like:

```ruby
        [
          1, 9, 10, 4, 5, 6, 7, 8, 9, 10,
          1, 7, 8, 4, 5, 6
        ]
```

Now it's still shuffling three instances unnecessarily:
* 14 from node 4 -> 6
* 15 from node 5 -> 7
* 16 from node 6 -> 8

SpComb: Oh... the "2 instances per node" spec only has 10 of the `GridServiceInstances`, not 20?

SpComb:
I did some more experimentation for the instances > 1 case, and had to also fix the specs for that, so here's my suggested algorithm as a branch: https://github.com/kontena/kontena/compare/feature/improve-stateless-daemon-bis

I haven't tested it yet with actual nodes, but it passes the specs here.

SpComb:
To be clear, the implementation here is also passing the improved `instances > 2` specs in that branch, the only difference being the order of the chosen replacement nodes. So I figure the two implementations are equivalent and the implementation here is probably also correct... I just find the `instance_rank` -based implementation to be a bit clearer and easier to understand, because it's more explicit about how the instances are matched/compared?

```
  1) Scheduler::Strategy::Daemon#find_node with existing nodes with 3 instances per node minimizes shuffle when nodes 2 & 3 are missing
     Failure/Error: expect(scheduled.map {|s| s.node_number}).to eq([
       
       expected: [1, 5, 6, 4, 5, 6, 7, 8, 9, 10, 1, 7, 8, 4, 5, 6, 7, 8, 9, 10, 1, 9, 10, 4]
            got: [1, 10, 9, 4, 5, 6, 7, 8, 9, 10, 1, 8, 7, 4, 5, 6, 7, 8, 9, 10, 1, 6, 5, 4]
       
       (compared using ==)
     # ./spec/services/scheduler/strategy/daemon_spec.rb:98:in `block (5 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'

  2) Scheduler::Strategy::Daemon#find_node with existing nodes with 2 instances per node minimizes shuffle when nodes 2 & 3 are missing
     Failure/Error: expect(scheduled.map {|s| s.node_number}).to eq([
       
       expected: [1, 7, 8, 4, 5, 6, 7, 8, 9, 10, 1, 9, 10, 4, 5, 6]
            got: [1, 10, 9, 4, 5, 6, 7, 8, 9, 10, 1, 8, 7, 4, 5, 6]
       
       (compared using ==)
     # ./spec/services/scheduler/strategy/daemon_spec.rb:76:in `block (5 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'
```

jakolehm:
Instance rank based implementation has this:

```
total_instances = nodes.size * grid_service.container_count
```

which might not be accurate if service has affinity rules etc?

SpComb:
Things like affinity rules affect the `nodes.size`... the implementation here also uses `nodes.size`.

There might be something weird going on during the `GridServiceDeployer#instance_count' phase, because the scheduler will get called with all connected nodes. The `node.size` is only correct during the second `GridServiceDeployer#select_nodes/deploy_service_instance` => `GridServiceScheduler#select_node` phase?

jakolehm:
Yes, you are right. Somehow I just saw red alert for instance calculation because it's usually wrong assumption 😄  But here nodes are already calculated correctly so it's safe in that way.

SpComb:
Found a case that this implementation seems to handle correctly, but where `instance_rank + node.schedule_counter` is too much of a simplification:

```
$ bin/kontena service deploy daemon-strategy/redis
 [done] Deploying service daemon-strategy/redis      
⊛ Deployed instance development/daemon-strategy/redis-1 to node core-03
⊛ Deployed instance development/daemon-strategy/redis-2 to node core-02
⊛ Deployed instance development/daemon-strategy/redis-3 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-4 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-5 to node core-02
⊛ Deployed instance development/daemon-strategy/redis-6 to node core-03
```

If you take core-02 offline and re-deploy, you end up with three instances on `core-01`:

```
$ bin/kontena service deploy daemon-strategy/redis
 [done] Deploying service daemon-strategy/redis      
⊛ Deployed instance development/daemon-strategy/redis-1 to node core-03
⊛ Deployed instance development/daemon-strategy/redis-2 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-3 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-4 to node core-01
```

SpComb:
Testing with some more weird and wonderful edge cases (pushed as specs to the `-bis` branch), which also needed some extra complications for the  `instance_rank` in terms of how the `node.schedule_counter` is taken into account... However, those edge cases do actually work in this branch's implementation, although I can't claim to understand how :)

e.g. having a daemon service with two instances per node, and adding a new node:

```
$ bin/kontena service deploy daemon-strategy/redis
 [done] Deploying service daemon-strategy/redis      
⊛ Deployed instance development/daemon-strategy/redis-1 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-2 to node core-02
⊛ Deployed instance development/daemon-strategy/redis-3 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-4 to node core-02
$ bin/kontena service deploy daemon-strategy/redis
 [done] Deploying service daemon-strategy/redis      
⊛ Deployed instance development/daemon-strategy/redis-1 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-2 to node core-02
⊛ Deployed instance development/daemon-strategy/redis-3 to node core-01
⊛ Deployed instance development/daemon-strategy/redis-4 to node core-02
⊛ Deployed instance development/daemon-strategy/redis-5 to node core-03
⊛ Deployed instance development/daemon-strategy/redis-6 to node core-03
```

SpComb:
Posted my alternative implementation as a separate PR, because it also turned out to need changes to how the `Scheduler::Node#scheduled_instances` are tracked: #3137

AFAIK based on my testing both of these PRs are working approaches, but I still don't understand how the logic in this PR actually works :)

jakolehm:
Let's close this in favor of #3137

PR #3132, opened 2017-12-12
Validate tls-sni domain authorization linked service port

### LINKED ISSUE #3128: Tls-sni authorize should error out if linked lb does not expose port 443
Because it cannot work.

jakolehm:
Fixes #3128

### DIFF
--- a/server/app/mutations/domain_authorizations/authorize.rb
+++ b/server/app/mutations/domain_authorizations/authorize.rb
@@ -30,7 +30,7 @@ module GridDomainAuthorizations
         @lb_service = resolve_service(self.grid, linked_service)
         if @lb_service.nil?
           add_error(:linked_service, :not_found, "Linked service needs to point to existing service")
-        elsif @lb_service && !https_open?(@lb_service)
+        elsif !https_open?(@lb_service)
           add_error(:linked_service, :invalid, "Linked service does not have port 443 open")
         end
       end
--- a/server/app/mutations/domain_authorizations/authorize.rb
+++ b/server/app/mutations/domain_authorizations/authorize.rb
@@ -23,11 +23,16 @@ module GridDomainAuthorizations
       unless grid.grid_secrets.find_by(name: LE_PRIVATE_KEY)
         add_error(:le_registration, :missing, "Let's Encrypt registration missing")
       end
-      if self.authorization_type == 'tls-sni-01'
-        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type") unless self.linked_service
+      if self.authorization_type == 'tls-sni-01' && self.linked_service.nil?
+        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type")
       end
       if self.linked_service
-        add_error(:linked_service, :not_found, "Linked service needs to point to existing service") unless @linked_service = resolve_service(self.grid, linked_service)
+        @lb_service = resolve_service(self.grid, linked_service)
+        if @lb_service.nil?
+          add_error(:linked_service, :not_found, "Linked service needs to point to existing service")
+        elsif @lb_service && !@lb_service.ports.include?(443)
+          add_error(:linked_service, :invalid, "Linked service does not have port 443 open")
+        end
       end
     end
 
@@ -69,11 +74,12 @@ module GridDomainAuthorizations
         challenge: challenge.to_h,
         challenge_opts: challenge_opts,
         tls_sni_certificate: verification_cert,
-        grid_service: @linked_service)
+        grid_service: @lb_service
+      )
 
       if self.authorization_type == 'tls-sni-01'
         # We need to deploy the linked service to get the certs in place
-        outcome = GridServices::Deploy.run(grid_service: @linked_service, force: true)
+        outcome = GridServices::Deploy.run(grid_service: @lb_service, force: true)
         if outcome.success?
           authz.grid_service_deploy = outcome.result
           authz.save
@@ -86,6 +92,5 @@ module GridDomainAuthorizations
     rescue Acme::Client::Error::Unauthorized
       add_error(:acme_client, :unauthorized, "Registration probably missing for LE")
     end
-
   end
 end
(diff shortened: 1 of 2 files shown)


server/app/mutations/domain_authorizations/authorize.rb
```diff
@@ -23,11 +23,16 @@ def validate
...
-        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type") unless self.linked_service
+      if self.authorization_type == 'tls-sni-01' && self.linked_service.nil?
+        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type")
       end
       if self.linked_service
-        add_error(:linked_service, :not_found, "Linked service needs to point to existing service") unless @linked_service = resolve_service(self.grid, linked_service)
+        @lb_service = resolve_service(self.grid, linked_service)
+        if @lb_service.nil?
+          add_error(:linked_service, :not_found, "Linked service needs to point to existing service")
+        elsif @lb_service && !@lb_service.ports.include?(443)
```
jnummelin: or if the service is in `net: host` mode

server/app/mutations/domain_authorizations/authorize.rb
```diff
@@ -23,11 +23,16 @@ def validate
...
         add_error(:le_registration, :missing, "Let's Encrypt registration missing")
       end
-      if self.authorization_type == 'tls-sni-01'
-        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type") unless self.linked_service
+      if self.authorization_type == 'tls-sni-01' && self.linked_service.nil?
+        add_error(:linked_service, :missing, "Service link needs to be given for tls-sni-01 authorization type")
       end
       if self.linked_service
-        add_error(:linked_service, :not_found, "Linked service needs to point to existing service") unless @linked_service = resolve_service(self.grid, linked_service)
+        @lb_service = resolve_service(self.grid, linked_service)
```
jnummelin: Originally, I tried to avoid using `lb_*` as variable names :), technically one can map the tls-sni cert into any service. Yes, in 99.99% of the cases it will be a LB

server/app/mutations/domain_authorizations/authorize.rb
SpComb: The `GridService.ports` is not an array of integers:

```
irb(main):002:0> GridService.find_by(name: 'lb').ports
=> [{"ip"=>"0.0.0.0", "protocol"=>"tcp", "node_port"=>80, "container_port"=>80}, {"ip"=>"0.0.0.0", "protocol"=>"tcp", "node_port"=>443, "container_port"=>443}]
```

SpComb:
And technically you could also be exposing the TLS service on a different non-443 port, if you have some external loadbalancer that's mapping it to port :443.

Dunno if this needs some kind of override flag for the validation check?

SpComb:
Maybe a better way to validate this would be in the `GridCertificates::RequestCertificate`, similarly to how it's done for `dns-01` challenges: have the server connect to the authorized domain on port :443

server/app/mutations/domain_authorizations/authorize.rb
jakolehm: Ok, the reason for variable name change was that it did override `linked_service` from mutation (that is a string).

server/app/mutations/domain_authorizations/authorize.rb
jakolehm: Fixed.

jakolehm: Fixed.

jakolehm:
@jnummelin PTAL

SpComb:
> Maybe a better way to validate this would be in the GridCertificates::RequestCertificate, similarly to how it's done for dns-01 challenges: have the server connect to the authorized domain on port :443

I can also take a go at implementing this method of external validation... the only issue would be if the server is unable to connect to the external domain for whatever reason.

server/app/mutations/domain_authorizations/authorize.rb
```diff
@@ -23,11 +23,16 @@ def validate
...
+        if @lb_service.nil?
+          add_error(:linked_service, :not_found, "Linked service needs to point to existing service")
+        elsif @lb_service && @lb_service.net != 'host' && !@lb_service.ports.any? { |p| p['node_port'] == 443 }
```
jnummelin: would it makes sense to have this kind of check on the `GridService` model? I mean something like `@lb_service.https_open?`

jnummelin:
> I can also take a go at implementing this method of external validation... the only issue would be if the server is unable to connect to the external domain for whatever reason.

I don't think we can ever make it 100% reliable given how external things can influence, so maybe some override/force flag would make sense.

server/app/mutations/domain_authorizations/authorize.rb
jakolehm: Imho does not make really sense to move this logic into the `GridService` model. I can refactor this code if necessary ...

jnummelin: for me having many conditionals and some of those negated, makes it hard to read. :) maybe simplify it by just removing the first `@lb_service` "condition", it shouldn't be falsey in the `elsif` branch?

jakolehm: Refactored.

SpComb:
>> Maybe a better way to validate this would be in the GridCertificates::RequestCertificate, similarly to how it's done for dns-01 challenges: have the server connect to the authorized domain on port :443
> 
> I can also take a go at implementing this method of external validation... the only issue would be if the server is unable to connect to the external domain for whatever reason.

Implemented in #3152... OTOH that might be a useless approach, because the ACME API will already return similar errors... they're maybe just a bit terse.

kke:
@SpComb is this still going to 1.4.3?

SpComb:
Bumping from 1.4.3 to 1.5.0 unless someone convinces me that we can't release 1.4.3 without this.

I haven't had a chance to actually test the validation here yet.

kke:
@SpComb still going to test?

PR #3087, opened 2017-11-24
Sync 1.4.1 to master

kke:
Brings 1.4 changelog to master and updates VERSION to 1.5.0.dev

PR #3142, opened 2017-12-15
Sync master VERSION to 1.5.0.dev and import missing 1.3/1.4 changelogs

SpComb:
Cleaner version of #3087 without the git merges

Changelogs imported using something along the lines of:
* `git diff  c7034f302acdf5bc406387bf78fea9f7a6a283b6^1..v1.4 -- CHANGELOG.md | patch -p1`
*  `git diff  v1.3.1..v1.3 -- CHANGELOG.md | patch -p1`


SpComb:
@kke PTAL, I'm getting confused by the `1.4.0.dev` when debugging the plugin issues :smiling_imp:

VERSION
```diff
@@ -1 +1 @@
-1.4.0.dev
+1.5.0.dev
```
kke: No `cli/` `agent/` `server/` VERSION?

SpComb: Oops, well spotted.

PR #3146, opened 2017-12-18
Fix stack commands not to modify STACK/GRID env of parent when installing child stacks

### LINKED ISSUE #3144: Stack depends mutates ${STACK}
Example:

```
stack: jakolehm/fn
depends:
  redis:
    stack: kontena/redis-sentinel
  mariadb:
    stack: kontena/mariadb
    variables:
      user: fn
      db: fn
      mem_limit: 512
variables:
  mariadb_pwd:
    type: string
    from:
      vault: ${STACK}-mariadb-password
  mq_url:
    type: string
    from:
      vault: ${STACK}-mq-url
      interpolate: mysql://fn:${mariadb_pwd}@tcp(db.${mariadb}.${GRID}.kontena.local:3306)/fn
    to:
      vault: ${STACK}-mq-url
```

`${STACK}` should be `fn` in variables section but it actually has mutated to `fn-mariadb`.

kke:
Fixes #3144 

Child stack installations mutated the $STACK variable. This PR makes the dependency installation restore the original STACK environment variable when returning from the child installation.

### DIFF
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -43,8 +43,6 @@ module Kontena::Cli::Stacks
       dependencies = loader.dependencies
       return if dependencies.nil?
 
-      original_env = current_stack_env
-
       dependencies.each do |dependency|
         target_name = "#{stack_name}-#{dependency['name']}"
         caret "Installing dependency #{pastel.cyan(dependency['stack'])} as #{pastel.cyan(target_name)}"
@@ -58,19 +56,10 @@ module Kontena::Cli::Stacks
 
         cmd << dependency['stack']
         Kontena.run!(cmd)
-        original_env.each { |k,v| ENV[k] = v }
+        set_env_variables(stack_name, current_grid)
       end
 
     end
-
-    def current_stack_env
-      {
-        'STACK' => ENV['STACK'],
-        'GRID' => ENV['GRID'],
-        'PLATFORM' => ENV['PLATFORM']
-      }
-    end
-
 
     def create_stack
       spinner "Creating stack #{pastel.cyan(stack['name'])} " do
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -42,6 +42,9 @@ module Kontena::Cli::Stacks
     def install_dependencies
       dependencies = loader.dependencies
       return if dependencies.nil?
+
+      original_env = current_stack_env
+
       dependencies.each do |dependency|
         target_name = "#{stack_name}-#{dependency['name']}"
         caret "Installing dependency #{pastel.cyan(dependency['stack'])} as #{pastel.cyan(target_name)}"
@@ -55,8 +58,19 @@ module Kontena::Cli::Stacks
 
         cmd << dependency['stack']
         Kontena.run!(cmd)
+        original_env.each { |k,v| ENV[k] = v }
       end
+
     end
+
+    def current_stack_env
+      {
+        'STACK' => ENV['STACK'],
+        'GRID' => ENV['GRID'],
+        'PLATFORM' => ENV['PLATFORM']
+      }
+    end
+
 
     def create_stack
       spinner "Creating stack #{pastel.cyan(stack['name'])} " do
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -24,9 +24,9 @@ module Kontena::Cli::Stacks
     requires_current_master_token
 
     def execute
-      set_env_variables(stack_name, current_grid)
-
       install_dependencies unless skip_dependencies?
+
+      set_env_variables(stack_name, current_grid)
 
       stack # runs validations
 
@@ -56,7 +56,6 @@ module Kontena::Cli::Stacks
 
         cmd << dependency['stack']
         Kontena.run!(cmd)
-        set_env_variables(stack_name, current_grid)
       end
 
     end
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -198,6 +198,7 @@ module Kontena::Cli::Stacks
         caret "Installing new dependency #{cmd.last} as #{added_stack}"
         deployable_stacks << added_stack
         Kontena.run!(cmd)
+        set_env_variables(stack_name, current_grid)
       end
       deployable_stacks
     end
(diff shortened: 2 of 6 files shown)


cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -198,6 +198,7 @@ def run_installs(changes)
         caret "Installing new dependency #{cmd.last} as #{added_stack}"
         deployable_stacks << added_stack
         Kontena.run!(cmd)
+        set_env_variables(stack_name, current_grid)
```
SpComb: This shouldn't strictly be necessary, I think? Yes, the `ENV['stack']` will leak from the `Kontena.run(['kontena', ',stack', 'install', ..])`, but OTOH all of the stacks reader executes happen before that. After the `run_installs`, I don't think the `run_upgrades`, `run_deploys` and `run_removes` would care about that `ENV['stack']` anymore?

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -219,6 +220,7 @@ def run_upgrades(changes)
     def run_deploys(deployable_stacks)
       deployable_stacks.each do |deployable_stack|
         Kontena.run!(['stack', 'deploy', deployable_stack])
+        set_env_variables(stack_name, current_grid)
```
SpComb: Probably also not necessary? I don't think the `kontena stack deploy` command mutates the `ENV`?

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -55,7 +56,9 @@ def install_dependencies
 
         cmd << dependency['stack']
         Kontena.run!(cmd)
+        set_env_variables(stack_name, current_grid)
```
SpComb: While this probably fixes the issue, it might be clearer to just change the `#execute` method to look like:
```ruby
    def execute
      install_dependencies unless skip_dependencies?

      set_env_variables(stack_name, current_grid)
      stack # runs validations
```

But note that this is assuming that:
* `loader.dependencies` doesn't touch the `self.stack` -> `@stack ||= self.reader.execute`
* `install_dependencies` doesn't need the `ENV['STACK']` to be set...?

jakolehm: Not sure if this is actually enough. Should we protect each dep install so that we reset whole ENV back to original? Now each dep can introduce new environment variables that can mess up next dependency (if it resolves variables from env).

SpComb: Right, there are also other envs than just `STACK` and `GRID` that can leak between stacks, if you're explicitly using `to: env: ...` in stack variables.

Not entirely sure if this is necessarily part of this issue though.... this sounds more like #2798, which was fixed before the 1.4 release by not exporting the stack variables to `ENV` by default ( #2802), as well as interpolating values directly from the opto variables instead of from `ENV` (related to #2822).

kke: That could be intentional and act as a way to pass stuff from children to parent in some cases. Setting values to local env rarely makes sense otherwise.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -220,6 +218,7 @@ def run_deploys(deployable_stacks)
       deployable_stacks.each do |deployable_stack|
         Kontena.run!(['stack', 'deploy', deployable_stack])
       end
+      set_env_variables(stack_name, current_grid)
```
SpComb: This probably doesn't hurt here, but I'm also not entirely sure what it's doing... Voodoo code with spooky action at a distance, or a talisman better just left out?

kke: image

SpComb:
The PR title is a bit misleading, though... What about something like e.g.

> Fix stack commands to not leak STACK/GRID env values up from dependent stacks

SpComb:
...or even more accurately:

> Fix stack commands to not use STACK/GRID end values leaked down from dependent stacks

test/spec/features/stack/install_spec.rb
```diff
@@ -142,5 +142,14 @@
...
       expect(k.out.split(/[\r\n]/)).to match array_including('twemproxy', 'twemproxy-redis_from_registry', 'twemproxy-redis_from_yml')
     end
+
+    it 'does not mutate the $STACK variable' do
+      with_fixture_dir("stack/depends") do
+        k = run 'kontena stack install'
+        expect(k.code).to eq (0)
+      end
+      k = run 'kontena service show twemproxy/twemproxy'
+      expect(k.out).to match(/STACKNAME=twemproxy$/m)
```
SpComb: Something odd with the  `/...$/m` regexp?

```
  1) stack install For a stack with dependencies does not mutate the $STACK variable
     Failure/Error: expect(k.out).to match(/STACKNAME=twemproxy$/m)
     
       expected "e2e/twemproxy/twemproxy:\r\n  created: 2017-12-18T14:23:19.371Z\r\n  updated: 2017-12-18T14:23:19.66...r\n          ip: 10.81.128.66\r\n          public ip: 82.181.8.229\r\n          status: running\r\n" to match /STACKNAME=twemproxy$/m
       Diff:
       @@ -1,2 +1,34 @@
       -/STACKNAME=twemproxy$/m
       +e2e/twemproxy/twemproxy:
       +  created: 2017-12-18T14:23:19.371Z
       +  updated: 2017-12-18T14:23:19.666Z
       +  stack: e2e/twemproxy
       +  desired_state: running
       +  image: ianblenke/twemproxy:latest
       +  revision: 1
       +  stack_revision: 1
       +  stateful: no
       +  scaling: 1
       +  strategy: ha
       +  read_only: no
       +  stop_grace_period: 10s
       +  deploy_opts:
       +  dns: twemproxy.twemproxy.e2e.kontena.local
       +  env: 
       +    - REDIS1_PORT_6379_TCP_ADDR=twemproxy-redis_from_registry
       +    - REDIS2_PORT_6379_TCP_ADDR=twemproxy-redis_from_yml
       +    - STACKNAME=twemproxy
       +  ports:
       +    - 6000:6000/tcp
       +  instances:
       +    twemproxy/twemproxy/1:
       +      scheduled_to: localhost
       +      deploy_rev: 2017-12-18 14:23:20 UTC
       +      rev: 2017-12-18 14:23:20 UTC
       +      state: running
       +      containers:
       +        twemproxy.twemproxy-1 (on localhost):
       +          dns: twemproxy-1.twemproxy.e2e.kontena.local
       +          ip: 10.81.128.66
       +          public ip: 82.181.8.229
       +          status: running
       
     # ./spec/features/stack/install_spec.rb:152:in `block (3 levels) in <top (required)>'
```

kke: `expect(k.out.lines.map(&:strip)).to include "STACKNAME=twemproxy"`? :)

SpComb: The `/.../m` flag doesn't seem to to matter, but the `\r` in the output seems to make the `$` not match:

```
irb(main):014:0> "   - FOO=bar\r\n".match /FOO=bar$/
=> nil
irb(main):015:0> "   - FOO=bar\n".match /FOO=bar$/
=> #<MatchData "FOO=bar">
```

PR #3158, opened 2017-12-19
Remove --values-from and prompts from kontena stack registry push

### LINKED ISSUE #3102: Kontena stack registry push has values-from options
A side effect of the stack file parser that is invoked for validation before pushing.

```
$ kontena stack registry push --help
Usage:
    kontena stack registry push [OPTIONS] [FILE]

  Pushes (uploads) a stack to the stack registry

Parameters:
    [FILE]                        Kontena stack file, registry stack name (user/stack or user/stack:version) or URL (default: "kontena.yml")

Options:
    --values-from [FILE]          Read variable values from a YAML file
    --values-from-stack [STACK_NAME] Read variable values from an installed stack
    -v VARIABLE=VALUE             Set stack variable values, example: -v domain=example.com. Can be used multiple times.
    -D, --[no-]debug              Enable debug (default: $DEBUG)
```

Not sure if bug or feature.

kke:
Fixes #3102

`kontena stack registry push xyz.yml` was prompting for variable values to validate/load the stack file, after this PR it no longer does that.

### DIFF
--- a/cli/lib/kontena/cli/stacks/registry/push_command.rb
+++ b/cli/lib/kontena/cli/stacks/registry/push_command.rb
@@ -7,8 +7,7 @@ module Kontena::Cli::Stacks::Registry
 
     banner "Pushes (uploads) a stack to the stack registry"
 
-    include Kontena::Cli::Stacks::Common::StackFileOrNameParam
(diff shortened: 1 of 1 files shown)


kke:
It wasn't even actually "validating" on purpose, the `stack` method just happens to run the interpolations.

PR #3175, opened 2017-12-22
Test cli with ruby-2.5.0

jakolehm:


kke:
2.1 is already EOL since 2017-03-31, maybe drop that while at it?

jakolehm:
Cannot drop that yet, omnibus is still using 2.1.

PR #3176, opened 2017-12-22
Fix server to validate certificates PEM trailing newlines

### LINKED ISSUE #3171: `tls-sni-01` certificate gets injected into LB without newlines between chain parts, causing HAproxy to ignore it
See this gist, from `kontena container inspect`ing one of my LB containers:

https://gist.github.com/japsu/a9120c9aea1b47d893ae535b0b26c3d0

`SSL_CERT_wiki.leonidasoy.fi` is `dns-01`, `SSL_CERT_grafana.plat2.leonidasoy.fi` is `tls-sni-01`.

In the latter, notice the absence of line breaks between the server certificate, intermediate CA certificate and private key. This apparently causes HAproxy to ignore the certificate, giving a wrong certificate when I try to navigate to `grafana.plat2.leonidasoy.fi`.

Both certificates were first `kontena certificate request`ed using 1.4.1 server. Having observed this bug, I updated to 1.4.3 server and re-requested the latter to no avail.

When I `kontena certificate export` the certificate, it shows newlines as expected.

SpComb:
See #3171

Prevents `Certificate` models from being created with PEM cert/key strings missing their trailing newlines, which causes the `bundle` to be invalid: `...0yDYg==\n-----END CERTIFICATE----------BEGIN CERTIFICATE-----\nMIIE...`

While this PR fixes the 1.4 `30_certificates_from_secrets.rb` migration to create the migrated certificates with the trailing PEM newline intact... this fix is too late for masters that have already been upgraded to 1.4. To actually fix #3171, would need to either:

* release a new migration to fix those migrated certificates post-facto
* add some kind of post-validation to fail the update/deploy if those invalid certs get used
* have the server add in the missing newlines when serializing the certificate as a secret

ATM this PR is mostly preventative, improving the specs and validation to avoid ending up in situations like #3171 again.


server/app/models/certificate.rb
```diff
@@ -22,6 +25,27 @@ def to_path
     "#{self.grid.name}/#{self.subject}"
   end
 
+  # @return [String] with trailing newline
+  def certificate
+    pem = super
+    pem += "\n" unless pem.end_with? "\n"
```
SpComb: TBH not entirely sure if we really need these workarounds to declare #3171 fixed: https://github.com/kontena/kontena/issues/3171#issuecomment-355297618

This extra complexity here is probably useless in practice?

server/app/models/certificate.rb
```diff
@@ -22,6 +25,27 @@ def to_path
     "#{self.grid.name}/#{self.subject}"
   end
 
+  # @return [String] with trailing newline
+  def certificate
+    pem = super
+    pem += "\n" unless pem.end_with? "\n" if pem
```
SpComb: I would still advocate just not bothering with these fixups for invalid certs, and just assuming that all certs stored in mongo are valid per the validation rules...

SpComb:
This PR is 98% just fixing the server specs to use realistic PEM values instead of invalid placeholders.

PR #3178, opened 2017-12-27
Pluggable master pubsub

jakolehm:


jakolehm:
@SpComb @jnummelin PTAL

server/app/initializers/pubsub.rb
```diff
@@ -0,0 +1,7 @@
+if ENV['NATS_SERVERS']
+  require_relative '../services/pubsub/nats'
+  Kernel.const_set('MasterPubsub', Pubsub::Nats)
+else
+  require_relative '../services/pubsub/mongo'
+  Kernel.const_set('MasterPubsub', Pubsub::Mongo)
```
SpComb: Could the `const_set` be avoided if the classmethods in `Pubsub::Base` were moved to an actual `class MasterPubsub`, and it just varied the `Celluloid.supervise type: ...`?

I'd imagine that this kind of meta-programming magic makes the code a little hard to follow, because the `MasterPubsub` class is used everywhere, but never actually defined... feels unnecessarily complicated.

SpComb: Also, :-1: for only requiring the implementation conditionally.... the `nats` dependencies are currently going to be there anyways, and it just makes the code harder to test. There could be a syntax error in the `pubsub/nats.rb` and you wouldn't even notice unless you run the server with `NATS_SERVERS` set...

server/app/boot_jobs.rb
```diff
@@ -2,9 +2,12 @@
 require_relative 'services/worker_supervisor'
 require_relative 'services/mongodb/migrator'
 
-unless ENV['RACK_ENV'] == 'test' || ENV['NO_MONGO_PUBSUB']
-  MongoPubsub.start!(PubsubChannel)
-
+unless ENV['RACK_ENV'] == 'test'
+  if servers = ENV['NATS_SERVERS']
+    MasterPubsub.start!(servers.split(','))
```
SpComb: The type confusion going on between the two `MasterPubsub.start!` method calls here isn't very nice. I can see what's going on now (the `MasterPubsub` class is defined programatically), but...

jakolehm: I agree, will fix.

server/app/services/pubsub/nats.rb
```diff
@@ -0,0 +1,49 @@
...
+        warning "connection to NATS closed"
+      end
+      async.stream!
+    end
+
+    # @param channel [String]
+    # @param data [Hash]
+    def publish(channel, data)
+      msg = { channel: channel, data: data }
+      @nats.publish(CHANNEL, MessagePack.pack(msg))
```
SpComb: I'm not sure if multiplexing all pubsub channels over a single backend channel is the optimal implementation in general, but I don't know how NATS behaves specifically.

I think the relevant question is whether our pubsub interface needs to preserve message ordering across different pubsub channels, and I don't think that's the case... pubsub messages should be allowed to be re-ordered across different channels, and I think that might also be a performance win.

SpComb:
To be clear, this is :+1: from a design point of view, it makes plenty of sense to have pluggable pubsub backends.

The actual implementation could just be a little less magical.

server/app/services/pubsub/nats.rb
jakolehm: > I'm not sure if multiplexing all pubsub channels over a single backend channel is the optimal implementation in general, but I don't know how NATS behaves specifically.

Single backend channel is much more simpler to implement vs multiple backend channels.. with multiple channels we get channel discovery complexity etc.

kke:
In conflict now

PR #3184, opened 2017-12-29
Use random initialization vector

### LINKED ISSUE #3183: Encrypted fields should use random iv
Currently all symmetrically encrypted fields use static iv. It would be better to switch to use random iv.

jakolehm:
Fixes #3183

### DIFF
--- a/server/app/models/certificate.rb
+++ b/server/app/models/certificate.rb
@@ -5,7 +5,7 @@ class Certificate
   field :subject, type: String
   field :valid_until, type: DateTime
   field :alt_names, type: Array
-  field :encrypted_private_key, type: String, encrypted: true # PEM encoded
+  field :encrypted_private_key, type: String, encrypted: { random_iv: true } # PEM encoded
   field :certificate, type: String # PEM encoded certificate, no need to encrypt
   field :chain, type: String # Trust chain
 
--- a/server/app/models/grid_domain_authorization.rb
+++ b/server/app/models/grid_domain_authorization.rb
@@ -17,7 +17,7 @@ class GridDomainAuthorization
 
   belongs_to :grid_service_deploy
 
-  field :encrypted_tls_sni_certificate, type: String, encrypted: true
+  field :encrypted_tls_sni_certificate, type: String, encrypted: { random_iv: true }
 
   belongs_to :grid_service # Usually a LB service
 
--- a/server/app/models/grid_secret.rb
+++ b/server/app/models/grid_secret.rb
@@ -5,7 +5,7 @@ class GridSecret
   belongs_to :grid
 
   field :name, type: String
-  field :encrypted_value, type: String, encrypted: true
+  field :encrypted_value, type: String, encrypted: { random_iv: true }
 
   validates_presence_of :name, :encrypted_value
   validates_uniqueness_of :name, scope: [:grid_id]
--- a/server/app/mutations/grid_secrets/update.rb
+++ b/server/app/mutations/grid_secrets/update.rb
@@ -10,9 +10,9 @@ module GridSecrets
     end
 
     def execute
+      return grid_secret if grid_secret.value == value
+
       grid_secret.value = value
-      return grid_secret unless grid_secret.changed?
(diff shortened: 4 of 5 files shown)


SpComb:
Backwards compatibility considerations with existing vault secrets in the DB?

jakolehm:
> Backwards compatibility considerations with existing vault secrets in the DB?

They seem to work just fine (tested manually).

SpComb:
The `VAULT_IV` now gets ignored for any new deployments, and it only needed for backwards-compatibility with existing secrets... but the server still requires it to work?

We should support deployments without `VAULT_IV`, either making it optional, or adding migration to re-encrypt all the secrets and then just forgetting the `VAULT_IV`.

SpComb:
The server [`Configuration`](https://github.com/kontena/kontena/blob/a099c4a238b0cc63498a323dcafeecdb48695e48/server/app/models/configuration.rb#L32)  model also uses `SymmetricEncryption.encrypt`, but it passes `true` for `random_iv`, so it's already using random IVs with the `@Enc` header.

SpComb:
The server [`DigestHelper`](https://github.com/kontena/kontena/blob/a099c4a238b0cc63498a323dcafeecdb48695e48/server/app/helpers/digest_helper.rb#L26) is using `SymmetricEncryption.encrypt` without `random_iv: true`, but OTOH it looks like it's actually re-encrypting the value already encrypted by the `Configuration` model itself... so it probably doesn't really matter, it's just confusing.

```
irb(main):019:0> Configuration.find_by(key: 'server.salt').value['v']
=> "QEVu...D0m2...VGV\n"
irb(main):020:0> Configuration['server.salt']
=> "4Og...0Q=\n"
irb(main):021:0> AccessToken.salt
=> "$2a$10$nisf...gQe"
irb(main):022:0> SymmetricEncryption.decrypt('4Og...0Q=\n')
=> "$2a$10$nis...Qe"
```

SpComb:
Testing shows that `symmetric-encrpytion` 3.x is indeed able to handle the existing ciphertexts that do not have any header and use a fixed IV: https://github.com/rocketjob/symmetric-encryption/blob/v3.9.1/lib/symmetric_encryption/symmetric_encryption.rb#L120-L126

This means that existing values without the `@Enc` header in mongo will get decrypted using the fixed `VAULT_IV`, and any new encrypted values will be stored with a random IV and the `@Enc` header. Testing with `kontena vault update test ...` before/after an upgrade:

```
> db.grid_secrets.find({"name": "test"})
{ "_id" : ObjectId("58e3c49cde35780bc6000000"), "name" : "test", "encrypted_value" : "dbGEfjVNCKDa2Jtwsq2N+w==\n", "grid_id" : ObjectId("580878e5de35781b2c000017"), "updated_at" : ISODate("2017-04-04T16:06:52.921Z"), "created_at" : ISODate("2017-04-04T16:06:52.921Z") }
> db.grid_secrets.find({"name": "test"})
{ "_id" : ObjectId("58e3c49cde35780bc6000000"), "name" : "test", "encrypted_value" : "QEVuQwBAEACAQvVtbI9Vk5YVmC8+rZGxRTRiwSBhDRvXazWJp6MV6g==\n", "grid_id" : ObjectId("580878e5de35781b2c000017"), "updated_at" : ISODate("2018-02-05T13:26:01.125Z"), "created_at" : ISODate("2017-04-04T16:06:52.921Z") }
```

```
irb(main):005:0> Base64.decode64("QEVuQwBAEACAQvVtbI9Vk5YVmC8+rZGxRTRiwSBhDRvXazWJp6MV6g==\n")
=> "@EnC\x00@\x10\x00\x80B\xF5ml\x8FU\x93\x96\x15\x98/>\xAD\x91\xB1E4b\xC1 a\r\e\xD7k5\x89\xA7\xA3\x15\xEA"
```

SpComb:
It wouldn't hurt to have a migration to re-encrpyt all the encrypted values in mongo using a random IV, but that doesn't necessarily need to be part of this PR.

PR #3185, opened 2018-01-02
Refactor stack change resolving

### LINKED ISSUE #3215: CLI stack upgrade --skip-dependencies is broken
The flag is not actually used anywhere in the code, so all dependent child stacks will still end up getting installed/upgraded.

```
$ bundle exec bin/kontena stack upgrade --skip-dependencies deptest test/deptest.yaml 
Value for peers is not set. Substituting with an empty string.
 [done] Analyzing upgrade     
SERVICES:
These services will be upgraded:
- deptest/redis
- deptest-foo/redis
- deptest-bar/redis

STACKS:
These stacks will be upgraded and deployed:
- deptest
- deptest-foo
- deptest-bar


 [done] Upgrading dependency deptest-bar     
 [done] Upgrading dependency deptest-foo     
 [done] Upgrading stack deptest     
 [done] Triggering deployment of stack deptest-bar     
```

jakolehm:
- refactor logic to separate classes
- fix stack api parent inconsistency (+ keep old behaviour for backwards compat)
- improve stack upgrade analyze output
- improve change resolving
- introduce `--reuse-values` flag

Fixes #3215

### DIFF
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -96,8 +96,8 @@ module Kontena::Cli::Stacks
 
       set_env_variables(stack_name, current_grid) # restore envs
 
-      old_set = Kontena::Stacks::DataSet.new(old_data)
-      new_set = Kontena::Stacks::DataSet.new(new_data)
+      old_set = Kontena::Stacks::StackDataSet.new(old_data)
+      new_set = Kontena::Stacks::StackDataSet.new(new_data)
       if skip_dependencies?
         [old_set, new_set].each(&:remove_dependencies)
       end
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -115,7 +115,6 @@ module Kontena::Cli::Stacks
         parent_name: data[:parent_name],
         name: data[:name]
       )
-      parsed_stack["parent"] = { "name" => parsed_stack.delete("parent_name") }
       data[:stack_data] = parsed_stack
     ensure
       prev_env.each { |k, v| ENV[k] = v }
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -82,12 +82,9 @@ module Kontena::Cli::Stacks
       new_data.reverse_each do |stackname, data|
         reader = data[:loader].reader
         set_env_variables(stackname, current_grid) # set envs for execution time
-        values = data[:variables]
-        if old_stack = old_data[stackname]
-          values = values.merge(old_stack[:stack_data]['variables'])
-        end
         parsed_stack = reader.execute(
-          values: values,
+          values: data[:variables],
+          defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
           parent_name: data[:parent_name],
           name: data[:name]
         )
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -20,7 +20,7 @@ module Kontena::Cli::Stacks
     option '--[no-]deploy', :flag, 'Trigger deploy after upgrade', default: true
 
     option '--force', :flag, 'Force upgrade'
-    option '--skip-dependencies', :flag, "Do not install any stack dependencies"
+    option '--skip-dependencies', :flag, "Do not upgrade any stack dependencies", default: false
     option '--dry-run', :flag, "Simulate upgrade"
 
     requires_current_master
@@ -80,21 +80,11 @@ module Kontena::Cli::Stacks
       logger.debug { "Master stacks: #{old_data.keys.join(",")} YAML stacks: #{new_data.keys.join(",")}" }
 
       new_data.reverse_each do |stackname, data|
-        reader = data[:loader].reader
-        set_env_variables(stackname, current_grid) # set envs for execution time
-        parsed_stack = reader.execute(
-          values: data[:variables],
-          defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
-          parent_name: data[:parent_name],
-          name: data[:name]
-        )
-        parsed_stack["parent"] = { "name" => parsed_stack.delete("parent_name") }
-        data[:stack_data] = parsed_stack
+        spinner "Processing stack #{pastel.cyan(stackname)}"
+        process_stack_data(stackname, data, old_data)
         hint_on_validation_notifications(reader.notifications, reader.loader.source)
         abort_on_validation_errors(reader.errors, reader.loader.source)
       end
-
-      set_env_variables(stack_name, current_grid) # restore envs
 
       old_set = Kontena::Stacks::StackDataSet.new(old_data)
       new_set = Kontena::Stacks::StackDataSet.new(new_data)
@@ -106,48 +96,52 @@ module Kontena::Cli::Stacks
       end
     end
 
+    # @param stackname [String]
+    # @param data [Hash]
+    # @param old_data [Hash]
+    def process_stack_data(stackname, data, old_data)
+      prev_env = ENV.clone
+      reader = data[:loader].reader
+      set_env_variables(stackname, current_grid) # set envs for execution time
+      parsed_stack = reader.execute(
+        values: data[:variables],
+        defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
+        parent_name: data[:parent_name],
+        name: data[:name]
+      )
+      parsed_stack["parent"] = { "name" => parsed_stack.delete("parent_name") }
+      data[:stack_data] = parsed_stack
+    ensure
+      prev_env.each { |k, v| ENV[k] = v }
+    end
+
+    # @param changes [Kontena::Stacks::ChangeResolver]
     def display_report(changes)
-      if !dry_run? && changes.removed_stacks.empty? && changes.replaced_stacks.empty? && changes.upgraded_stacks.size == 1 && changes.removed_services.empty?
-        return
-      end
-
-      will = dry_run? ? "would" : "will"
-
       puts
-      puts "SERVICES:"
-      puts "-" * 40
-
-      unless changes.removed_services.empty?
-        puts pastel.yellow("These services #{will} be removed from master:")
-        changes.removed_services.each { |svc| puts pastel.yellow("- #{svc}") }
-        puts
-      end
-
-      unless changes.added_services.empty?
-        puts pastel.green("These new services #{will} be created to master:")
-        changes.added_services.each { |svc| puts pastel.green("- #{svc}") }
-        puts
-      end
-
-      unless changes.upgraded_services.empty?
-        puts pastel.cyan("These services #{will} be upgraded:")
-        changes.upgraded_services.each do |svc|
-          puts pastel.cyan("- #{svc}")
-        end
-        puts
-      end
-
-      puts "STACKS:"
-      puts "-" * 40
+      caret "Calculated changes:", dots: false
 
       unless changes.removed_stacks.empty?
-        puts pastel.red("These stacks #{will} be removed because they are no longer depended on:")
-        changes.removed_stacks.each { |stack| puts pastel.red("- #{stack}") }
-        puts
+        changes.removed_stacks.each { |stack|
+          puts "  #{pastel.red('-')} #{stack}"
+          changes.removed_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.red('-')} #{s}"
+          end
+        }
       end
 
       unless changes.replaced_stacks.empty?
-        puts pastel.yellow("These stacks #{will} be replaced by other stacks:")
+        changes.replaced_stacks.each { |stack|
+          puts "  #{pastel.yellow('-/+')} #{stack}"
+          changes.upgraded_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.cyan('~')} #{s}"
+          end
+          changes.added_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.green('+')} #{s}"
+          end
+          changes.removed_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.red('-')} #{s}"
+          end
+        }
         changes.replaced_stacks.each do |installed_name, data|
           puts "- #{pastel.yellow(installed_name)} from #{pastel.cyan(data[:from])} to #{pastel.cyan(data[:to])}"
         end
@@ -155,16 +149,29 @@ module Kontena::Cli::Stacks
       end
 
       unless changes.added_stacks.empty?
-        puts pastel.green("These new stack dependencies #{will} be installed:")
-        changes.added_stacks.each { |stack| puts pastel.green("- #{stack}") }
-        puts
+        changes.added_stacks.each { |stack|
+          puts "  #{pastel.green('+')} #{stack}"
+          changes.added_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.green('+')} #{s}"
+          end
+        }
       end
 
       unless changes.upgraded_stacks.empty?
-        puts pastel.cyan("These stacks #{will} be upgraded#{' and deployed' if deploy?}:")
-        changes.upgraded_stacks.each { |stack| puts pastel.cyan("- #{stack}") }
-        puts
+        changes.upgraded_stacks.each { |stack|
+          puts "  #{pastel.cyan('~')} #{stack}"
+          changes.upgraded_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.cyan('~')} #{s}"
+          end
+          changes.added_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.green('+')} #{s}"
+          end
+          changes.removed_services.select { |s| s.split('/')[0] == stack }.each do |s|
+            puts "    #{pastel.red('-')} #{s}"
+          end
+        }
       end
+      puts
     end
 
     # requires heavier confirmation when something very dangerous is going to happen
@@ -181,12 +188,14 @@ module Kontena::Cli::Stacks
       @deployable_stacks ||= []
     end
 
+    # @param removed_stacks [Array<String>]
     def run_removes(removed_stacks)
       removed_stacks.reverse_each do |removed_stack|
         Kontena.run!('stack', 'remove', '--force', '--keep-dependencies', removed_stack)
       end
     end
 
+    # @param changes [Kontena::Stacks::ChangeResolver]
     # @return [Array] an array of stack names that have been installed, but not yet deployed
     def run_installs(changes)
       deployable_stacks = []
@@ -205,6 +214,7 @@ module Kontena::Cli::Stacks
       deployable_stacks
     end
 
+    # @param changes [Kontena::Stacks::ChangeResolver]
     # @return [Array] an array of stack names that have been upgraded, but not yet deployed
     def run_upgrades(changes)
       deployable_stacks = []
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -103,12 +103,12 @@ module Kontena::Cli::Stacks
     def process_stack_data(stackname, data, old_data)
       prev_env = ENV.clone
       reader = data[:loader].reader
-      set_env_variables(stackname, current_grid) # set envs for execution time
       values = data[:variables]
       if reuse_values? && old_data[stackname]
         old_vars = old_data[stackname][:stack_data]['variables']
         values = old_vars.merge(values)
       end
+      set_env_variables(stackname, current_grid) # set envs for execution time
       parsed_stack = reader.execute(
         values: values,
         defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
--- a/cli/lib/kontena/cli/stacks/upgrade_command.rb
+++ b/cli/lib/kontena/cli/stacks/upgrade_command.rb
@@ -1,7 +1,6 @@
-require_relative 'common'
-require_relative 'change_resolver'
-
 require 'json'
+require 'kontena/stacks/change_resolver'
+require_relative 'common'
 
 module Kontena::Cli::Stacks
   class UpgradeCommand < Kontena::Command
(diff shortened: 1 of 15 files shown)


cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -83,20 +82,30 @@ def process_data(old_data, new_data)
...
+        values = data[:variables]
+        if old_stack = old_data[stackname]
+          values = values.merge(old_stack[:stack_data]['variables'])
+        end
+        parsed_stack = reader.execute(
+          values: values,
           parent_name: data[:parent_name],
           name: data[:name]
         )
+        parsed_stack["parent"] = { "name" => parsed_stack.delete("parent_name") }
```
jakolehm: This is here so that old and new data are somewhat similar.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -83,20 +82,30 @@ def process_data(old_data, new_data)
       new_data.reverse_each do |stackname, data|
         reader = data[:loader].reader
         set_env_variables(stackname, current_grid) # set envs for execution time
```
SpComb: Best to keep the `set_env_variables`  as close to the `reader.execute` as possible. Probably doesn't matter for the `values.merge` bit, but still, don't want this drifting any further.

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: This changes what the server `PUT /v1/stacks/...` API sees from `{"parent_name": ...}` to `{"parent": {"name": ...}}`, including also `{"parent":{"name":null}}` for the top-level stack.

It works because the server `Stacks::Update` isn't expecting any `parent` or `parent_name` field, so both of them get ignored. Still feels a little wrong somehow.

Doesn't break the actual `POST /v1/grids/.../stacks` => `Stacks::Create` API, because that's from `kontena stack install --parent-name=...`

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -83,20 +82,30 @@ def process_data(old_data, new_data)
...
-        data[:stack_data] = reader.execute(
-          values: data[:variables],
-          defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
+        values = data[:variables]
+        if old_stack = old_data[stackname]
+          values = values.merge(old_stack[:stack_data]['variables'])
+        end
+        parsed_stack = reader.execute(
+          values: values,
```
SpComb: What happened to `values` vs `defaults`? Without the use of `defaults`, `kontena stack upgrade` is no longer prompting for variables at all, since it gets all of them directly in `values`.... that's only meant for overrides like `-v foo=bar`.

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: It feels a little wrong that post accepts `parent_name` but response contains `parent` hash. I can remove whole parent thing from `PUT` because it's not used there at all.

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: This was just an experiment (probably topic of another issue/pr)... for usability it would probably make sense to prompt only new variables by default (with option to force prompt everything).

cli/lib/kontena/stacks/stack_data.rb
```diff
@@ -0,0 +1,48 @@
+module Kontena::Stacks
+  class StackData
```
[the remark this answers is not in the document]
kke: I think we should spare the "Models" keyword for some future API object models that can be `.created`/`.saved` etc.

kke:
> introduce --reuse-values flag

I have a hunch that I made a `--values-from-stack` option that tries to guess names for each dependency too. 

Also there was `--use-defaults` at some point, previous answers will be set as default values to variables, so that should work also for this.

server/app/mutations/stacks/create.rb
```diff
@@ -12,7 +12,12 @@ class Create < Mutations::Command
     end
 
     optional do
-      string :parent_name
+      hash :parent do
+        required do
+          string :id
```
kke: The reason why this wasn't an id in the first place was that children are created before parents, so you don't have an ID (but i suppose names and id's are interchangeable in a lot of models)

SpComb:
>> introduce --reuse-values flag
> 
> I have a hunch that I made a --values-from-stack option that tries to guess names for each dependency too.
> 
> Also there was --use-defaults at some point, previous answers will be set as default values to variables, so that should work also for this.

Is this about avoiding prompts when upgrading? We've already discussed this extensively in #2897, but that PR got stuck.

kke:
Yes, right, that one and the one that was supposed to replace it.

jakolehm:
> I have a hunch that I made a --values-from-stack option that tries to guess names for each dependency too.

Dependencies are not needed here because we don't actually execute deps unless they have changed (version or name).

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -21,13 +20,14 @@ class UpgradeCommand < Kontena::Command
     option '--[no-]deploy', :flag, 'Trigger deploy after upgrade', default: true
 
     option '--force', :flag, 'Force upgrade'
-    option '--skip-dependencies', :flag, "Do not install any stack dependencies"
+    option '--skip-dependencies', :flag, "Do not upgrade any stack dependencies", default: false
+    option '--reuse-values', :flag, "Reuse existing values"
```
SpComb: This looks like the `--use-defaults` flag in #2884

IIRC the situation was that the logic behaved correctly for `kontena stack upgrade`, but it went badly for `kontena stack install`.

The proposed logic in #2897 for `--no-prompt` would work for both `kontena stack upgrade` and also `kontena stack install`, including complex cases with dynamic resolvers as default values...

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -81,91 +81,109 @@ def process_data(old_data, new_data)
...
         hint_on_validation_notifications(reader.notifications, reader.loader.source)
         abort_on_validation_errors(reader.errors, reader.loader.source)
       end
 
-      set_env_variables(stack_name, current_grid) # restore envs
-
+      old_set = Kontena::Stacks::StackDataSet.new(old_data)
+      new_set = Kontena::Stacks::StackDataSet.new(new_data)
+      if skip_dependencies?
+        [old_set, new_set].each(&:remove_dependencies)
```
SpComb: Huh... wouldn't it be easier to just change `gather_master_data` to not recurse instead, and then use some variant of `loader.to_h` instead of `loader.flat_dependencies`? It looks like the later might need some slight refactoring of `flat_dependencies` to split out the `{ basename => { ...., variables: ... } }` logic, but that's what this PR is for...?

Loading everything recursively only to just strip them out again does not really count as "refactoring" to me ^___^

cli/lib/kontena/stacks/change_resolver.rb
```diff
@@ -56,41 +56,63 @@ def remaining_stacks
...
-      old_names = old_data.keys
-      new_names = new_data.keys
+      old_names = old_data.stack_names
+      new_names = new_data.stack_names
 
       removed_stacks.concat(old_names - new_names)
       added_stacks.concat(new_names - old_names)
-      upgraded_stacks.concat(new_names & old_names)
+      (new_names & old_names).each do |candidate|
+        upgraded_stacks << candidate if stack_upgraded?(candidate)
```
SpComb: Could this be written as something like `upgraded_stacks = (new_names & old_names).select{|x| stack_upgraded?(x)`

server/app/mutations/stacks/create.rb
```diff
@@ -45,11 +51,28 @@ def validate_services
...
+      _, parent_name = if self.inputs.has_key?(:parent)
+        self.inputs.delete(:parent)[:id].split('/')
+      else
+        [self.grid.name, self.inputs.delete(:parent_name)]
+      end
+      return unless parent_name
+
+      parent_stack = self.grid.stacks.find_by(name: parent_name)
+      if parent_stack.nil?
+        add_error(:parent, :not_found, "parent stack #{parent_name} does not exist")
```
SpComb: No. This fundamentally breaks the parent-child dependent stack relationship, because as @kke said, the child stacks are installed before the parent stacks.

```
> Installing dependency /home/kontena/kontena/kontena/cli/test/deptest-child.yaml as deptest-foo ..
Value for peers is not set. Substituting with an empty string.
 [fail] Creating stack deptest-foo      
	parent: parent stack deptest does not exist
```

server/app/mutations/stacks/create.rb
```diff
@@ -45,11 +51,28 @@ def validate_services
...
+      else
+        @parent = parent_name
+      end
+    end
+
     def execute
       attributes = self.inputs.clone
       grid = attributes.delete(:grid)
-      parent = attributes.delete(:parent_name)
```
SpComb:
```

Mongoid::Errors::UnknownAttribute (
message:
  Attempted to set a value for 'parent_name' which is not allowed on the model StackRevision.
summary:
  Without including Mongoid::Attributes::Dynamic in your model and the attribute does not already exist in the attributes hash, attempting to call StackRevision#parent_name= for it is not allowed. This is also triggered by passing the attribute to any method that accepts an attributes hash, and is raised instead of getting a NoMethodError.
resolution:
  You can include Mongoid::Attributes::Dynamic if you expect to be writing values for undefined fields often.):
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:99:in `process_attribute'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:25:in `block in process_attributes'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:23:in `each_pair'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/attributes/processing.rb:23:in `process_attributes'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/document.rb:111:in `block in initialize'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/threaded/lifecycle.rb:89:in `_building'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/document.rb:105:in `initialize'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/factory.rb:25:in `new'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/factory.rb:25:in `build'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/relations/referenced/many.rb:82:in `build'
  /usr/lib/ruby/gems/2.4.0/gems/mongoid-5.2.1/lib/mongoid/relations/many.rb:77:in `create!'
  /app/app/mutations/stacks/create.rb:87:in `execute'
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -81,91 +81,109 @@ def process_data(old_data, new_data)
...
+            puts "    #{pastel.green('+')} #{s}"
+          end
+        }
       end
 
       unless changes.upgraded_stacks.empty?
-        puts pastel.cyan("These stacks #{will} be upgraded#{' and deployed' if deploy?}:")
-        changes.upgraded_stacks.each { |stack| puts pastel.cyan("- #{stack}") }
-        puts
+        changes.upgraded_stacks.each { |stack|
```
SpComb: Could this also explicitly list the dependent stacks that are *NOT* getting upgraded?

Right now it looks a little broken to me, because it will only list the top-level stack:

```
DEBUG Master stacks: deptest,deptest-bar,deptest-foo YAML stacks: deptest,deptest-foo,deptest-bar
 [done] Processing stack deptest-bar
 [done] Processing stack deptest-foo
Value for peers is not set. Substituting with an empty string.
 [done] Processing stack deptest
 [done] Analyzing upgrade     

> Calculated changes:
  ~ deptest
    ~ deptest/redis

    *   Upgrading stack deptest .. |
```

server/app/mutations/stacks/create.rb
SpComb: The `kontena stack install` command is still POST'ing `{"parent_name": ...}`, whereas `kontena stack upgrade` is now PUT'ing `{"parent": {"name": ... } }`.

Not sure what the meaning of the `id` is...

server/spec/mutations/stacks/create_spec.rb
```diff
@@ -23,6 +23,31 @@
...
+
+        expect(outcome).to be_success
+        expect(outcome.result.parent_name).to eq 'stack-parent'
+        expect(outcome.result.has_parent?).to be_truthy
+      }.to change{ Stack.count }.by(1)
+    end
+
+    it 'creates a new grid stack with deprecated parent_name' do
+      grid
+      grid.stacks.create!(name: 'stack-parent')
```
SpComb: Creating the parent stack as part of the spec breaks the specs (false-positive).

server/app/mutations/stacks/create.rb
SpComb: I see, this didn't show up in the specs because it only happened after I commented out the `#validate_parent`, which does `self.inputs.delete(:parent_name)` as a side-effect, preventing this.

server/app/mutations/stacks/create.rb
```diff
@@ -45,11 +51,28 @@ def validate_services
       add_error("services.#{error.service}.links", :recursive, error.message)
     end
 
+    def validate_parent
+      _, parent_name = if self.inputs.has_key?(:parent)
+        self.inputs.delete(:parent)[:id].split('/')
+      else
+        [self.grid.name, self.inputs.delete(:parent_name)]
```
SpComb: :-1: for mutating `self.inputs` in the validate methods

SpComb:
Needs merge with master to bring in the `travis.yml` changes to also run the e2e specs here, those are particularly important for these kinds of refactoring PRs.

I was a little concerned that the server specs didn't catch the `kontena stack install` breakage, but thankfully it does show up in the e2e specs:

```
rspec ./spec/features/stack/install_spec.rb:134 # stack install For a stack with dependencies installs all dependencies
rspec ./spec/features/stack/install_spec.rb:142 # stack install For a stack with dependencies does not mutate the $STACK variable
rspec ./spec/features/stack/remove_spec.rb:53 # stack remove for a stack that has dependencies removes all the dependencies
rspec ./spec/features/stack/upgrade_spec.rb:98 # stack upgrade for a stack that has dependencies when a new dependency is added installs the added stack
rspec ./spec/features/stack/upgrade_spec.rb:115 # stack upgrade for a stack that has dependencies when a dependency is removed removes the stack
rspec ./spec/features/stack/upgrade_spec.rb:135 # stack upgrade for a stack that has dependencies when a dependency is replaced removes the stack
```

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: We need to load everything recursively so that variable referencing works.

server/app/mutations/stacks/create.rb
jakolehm: Easy to fix.

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: By variable referencing, do you mean the way that [`StackFileLoader#flat_dependencies`](https://github.com/kontena/kontena/blob/c5fecb2b9370f2742285a0f1bf7842b01b9cef62/cli/lib/kontena/cli/stacks/yaml/stack_file_loader.rb#L114-L119) manipulates the `values_from_options` to handle args like `-v deptest-foo.url=bar`?

I figure the `flat_dependencies` method just needs to refactored to split out the initial non-recursive part, and then just use that instead of `flat_dependencies` if `--skip-dependencies`:

```
      def flat_parent(basename, opts = {})
        opt_variables = opts[:variables] || {}

        return {
          basename => self.to_h.merge(opts).merge(
            name: basename,
            variables: opt_variables.reject { |k, _| k.include?('.') }
          )
        }
      end
```

I'm not aware of any way to reference the child stack variables from the parent?

server/app/mutations/stacks/create.rb
SpComb: I'm not sure what you mean.

See the original discussion around this exact same thing: https://github.com/kontena/kontena/pull/2707#discussion_r138290028

server/app/mutations/stacks/create.rb
SpComb: This could probably be fixed by using `string :name` instead of `string :id`. That implies that the existence of the `name` stack does not get validated.

Brief reasoning of examples:
* `{"parent": {"name": "foo"}}` is more consistent with the API response format than `{"parent_name": "foo" }`
* `{"parent": {"name": null}}` is certainly wrong
* `{"parent": null}` makes more sense than `{"parent_name": null}`

SpComb: If we're changing this as part of this PR, then the `kontena stack install` commands also needs to be fixed to POST with the new syntax.

SpComb: Using `name` instead of `id` would also be consistent with the response format:

* `{"parent": null}` is a standalone stack without any dependent child stacks
* `{"parent": {"name": "foo", "id": null}}` is an orphaned child stack that does not have its parent
* `{"parent": {"name": "foo", "id": "grid/foo"}}` is a child stack that has an existing parent stack - that is not a valid scenario for `POST /v1/grids/.../stacks`

jakolehm: @SpComb yeah, let's change `parent.id` to `parent.name`.

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: > I'm not aware of any way to reference the child stack variables from the parent?

It's not yet possible, but it should be in the future.. see #2856

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: It could, was not really sure does that make sense? Output show "changes"...

SpComb: It's a little hard to tell exactly what stacks/services you have present in a complex set of dependent stacks. I'd be more comfortable seeing the complete list of stacks including an explicit list of things not changed, rather than having to wonder if I somehow accidentally mis-indented the YAML and all the dependent stacks got lost somewhere at some point...

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: Fair enough. Should this somehow be passed as a flag to the `ChangeResolver` instead, so that it only returns the top-level stack in `upgraded_stacks`?

jakolehm:
Rebased (waiting for Travis results).

server/app/mutations/stacks/create.rb
jakolehm: I meant reverting this change. Currently it's not possible to validate parent because it's created afterwards.

server/app/mutations/stacks/create.rb
jakolehm: Yes, fixed.

server/spec/mutations/stacks/create_spec.rb
```diff
@@ -23,6 +23,31 @@
 
     it 'creates a new grid stack with parent' do
       grid
+      grid.stacks.create!(name: 'stack-parent')
```
SpComb: I'd still recommend removing this + the other `Stack.create!` in the other spec. The context doesn't actually match up with reality.

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: I had it first in the `ChangeResolver` but somehow it felt weird. I think we can improve this in another pr (easier to review smaller pr's)?

cli/lib/kontena/cli/stacks/upgrade_command.rb
jakolehm: But if user mis-indents (or makes typos) then output should say that those will be removed?

cli/lib/kontena/stacks/change_resolver.rb
jakolehm: Yes, makes sense.

cli/lib/kontena/cli/stacks/upgrade_command.rb
SpComb: Yes, but only during the first `upgrade`, and also not e.g. during the initial `stack install`

server/spec/mutations/stacks/create_spec.rb
jakolehm: Yeah, I can remove those.. although there is also a spec that checks that stack can be created before the parent exists.

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -81,91 +81,108 @@ def process_data(old_data, new_data)
...
+      set_env_variables(stackname, current_grid) # set envs for execution time
+      parsed_stack = reader.execute(
+        values: values,
+        defaults: old_data[stackname].nil? ? nil : old_data[stackname][:stack_data]['variables'],
+        parent_name: data[:parent_name],
+        name: data[:name]
+      )
+      data[:stack_data] = parsed_stack
+    ensure
+      prev_env.each { |k, v| ENV[k] = v }
```
kke: Can be just: `ENV.update(prev_env)`

cli/lib/kontena/cli/stacks/upgrade_command.rb
```diff
@@ -81,91 +81,108 @@ def process_data(old_data, new_data)
...
       puts
     end
 
     # requires heavier confirmation when something very dangerous is going to happen
     def get_confirmation(changes)
       unless force?
-        unless changes.removed_services.empty? && changes.removed_stacks.empty? && changes.replaced_stacks.empty?
+        unless changes.safe?
           puts "#{pastel.red('Warning:')} This can not be undone, data will be lost."
           confirm
```
kke: (these could be `confirm("This can't be undone, data will be lost.") unless safe?`)

SpComb:
> ➖ /➕ Not sure about the --reuse-values but can't think of a better word either

My proposal for what this `--reuse-values` should eventually look/feel like:

* `kontena stack upgrade --prompt` for current behavior (use saved variable values as defaults for prompts)
* `kontena stack upgrade` for #3185 behavior (used saved variable values, prompt for new variables without defaults or values)
* `kontena stack upgrade --no-prompt` for #2897 behavior (use saved variable values, error out for new variables without defaults or chained resolvers)

Hence 👎 for having it in this PR... do we really want to introduce that flag?

PR #3188, opened 2018-01-03
Improve stack install deps handling

jakolehm:
First install all stacks and then deploy them. Better UX.

### DIFF
--- a/cli/lib/kontena/cli/stacks/install_command.rb
+++ b/cli/lib/kontena/cli/stacks/install_command.rb
@@ -36,7 +36,11 @@ module Kontena::Cli::Stacks
       dump_variables if values_to
 
       create_stack
-      deploy_stack if deploy?
+
+      if deploy?
+        deploy_dependencies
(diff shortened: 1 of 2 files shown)


jakolehm:
Yeah, working on the specs ...

cli/lib/kontena/cli/stacks/install_command.rb
```diff
@@ -46,18 +50,25 @@ def install_dependencies
...
-        cmd << '--no-deploy' unless deploy?
-
         cmd << dependency['stack']
         Kontena.run!(cmd)
       end
+    end
+
+    def deploy_dependencies
+      dependencies = loader.dependencies
+      return if dependencies.nil?
```
kke: If loader returned [] instead of nil, these could be just `loader.dependencies.each`

jakolehm: Yes, but I'm not sure what side-effects that would cause.

PR #3193, opened 2018-01-08
Allow to command multiple items via cli

jakolehm:
Allows user to do things like:

```
$ kontena volume rm --force $(kontena volume ls -q)
```

or

```
$ kontena service stop $(kontena service ls -q | grep test)
```

### DIFF
--- a/cli/lib/kontena/cli/grids/remove_command.rb
+++ b/cli/lib/kontena/cli/grids/remove_command.rb
@@ -5,24 +5,26 @@ module Kontena::Cli::Grids
     include Kontena::Cli::Common
     include Common
 
-    parameter "NAME", "Grid name"
+    parameter "NAME ...", "Grid name", attribute_name: :grids
     option "--force", :flag, "Force remove", default: false, attribute_name: :forced
 
     def execute
       require_api_url
       token = require_token
-      confirm_command(name) unless forced?
-      grid = find_grid_by_name(name)
+      grids.each do |name|
+        confirm_command(name) unless forced?
+        grid = find_grid_by_name(name)
 
-      if !grid.nil?
-        spinner "removing #{pastel.cyan(name)} grid " do
-          response = client(token).delete("grids/#{grid['id']}")
-          if response
-            clear_current_grid if grid['id'] == current_grid
+        if !grid.nil?
+          spinner "removing #{pastel.cyan(name)} grid " do
+            response = client(token).delete("grids/#{grid['id']}")
+            if response
+              clear_current_grid if grid['id'] == current_grid
+            end
           end
+        else
+          exit_with_error "Could not resolve grid by name [#{name}]. For a list of existing grids please run: kontena grid list"
         end
-      else
-        exit_with_error "Could not resolve grid by name [#{name}]. For a list of existing grids please run: kontena grid list"
       end
     end
   end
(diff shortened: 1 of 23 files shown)


kke:
I think this could maybe also be done half-automatically, requiring only an include to each of the "multiple params consuming" commands.

jakolehm:
> I think this could maybe also be done half-automatically

Maybe, but I prefer this way... less magic and still very easy to implement.

cli/lib/kontena/cli/grids/remove_command.rb
```diff
@@ -5,24 +5,26 @@ class RemoveCommand < Kontena::Command
...
-            clear_current_grid if grid['id'] == current_grid
+        if !grid.nil?
+          spinner "removing #{pastel.cyan(name)} grid " do
+            response = client(token).delete("grids/#{grid['id']}")
+            if response
+              clear_current_grid if grid['id'] == current_grid
+            end
           end
+        else
+          exit_with_error "Could not resolve grid by name [#{name}]. For a list of existing grids please run: kontena grid list"
```
kke: I think these should just warn and continue instead of exit with error.

Consider something like:
```
$ kontena stack rm --force stack1 stack2 snack3 stack4
Could not resolve grid by name [snack3]  # <-- btw inconsistent formatting there with the []
```

After such an error you kind of have to guess / figure out which ones already went through.

kke: Hmm, but then again, how about: `kontena stack rm --force $(stack ls -q) && ...` maybe it's actually better to exit with error if even one fails.

kke:
How about plugin install?

kke:
```ruby
module Kontena
  module Callbacks
    class ConsumeStdinParams < Kontena::Callback

      matches_commands 'stacks deploy'

      def before_parse
        command.arguments.concat $stdin.read.split(/\s+/).map(&:strip).reject(&:empty?) unless $stdin.tty?
      end
    end
  end
end
```

with this, it would be possible to have `kontena stack ls -q | kontena stack rm --force` working without `xargs`.

kke:
> How about plugin install?

Looks like there are some weird problems with it. Like getting HTTP 503 or having some rubygems lib give `undefined method `include?' for false:FalseClass`

SpComb:
This seems to have broken the e2e specs with the `after do run 'kontena ... rm --force $(kontena ... ls -q)'` hooks:, that kind shell interpolation syntax doesn't work in kommando:

```
|command="kontena" interpolated_args=["service", "rm", "--force", nil, "service", "ls", "-q)"]
```

```
  1) service remove removes a service
     Failure/Error: Kommando.run(cmd, opts)
     
     TypeError:
       no implicit conversion of nil into String
     # ./vendor/bundle/ruby/2.3.0/gems/kommando-0.1.2/lib/kommando.rb:187:in `spawn'
     # ./vendor/bundle/ruby/2.3.0/gems/kommando-0.1.2/lib/kommando.rb:187:in `run'
     # ./vendor/bundle/ruby/2.3.0/gems/kommando-0.1.2/lib/kommando.rb:21:in `run'
     # ./spec/support/shell.rb:19:in `run'
     # ./spec/features/service/remove_spec.rb:4:in `block (2 levels) in <top (required)>'
```

PR #3195, opened 2018-01-08
Remove deprecated app subcommand from CLI completions

### LINKED ISSUE #3181: Deprecated "kontena app" still in completions
The app-subcommand was removed some versions ago. It still appears in the CLI tab completions.

kke:
Replaced by #3197


kke:
Replaced by #3197

PR #3199, opened 2018-01-08
Fix travis e2e allow_failures clause

kke:
The `allow_failures` env needs to be exactly the env of the job that is allowed to fail.


kke:
It was failing without this change if the e2e failed, as seen in https://travis-ci.org/kontena/kontena/builds/326347377

PR #3200, opened 2018-01-08
Optimize e2e vpn specs

SpComb:
The e2e VPN specs run `kontena vpn create` three times.

`kontena vpn create` is very slow.

Let's only run it once instead.


test/spec/features/vpn/remove_spec.rb
```diff
@@ -6,18 +6,6 @@
...
-  after(:each) do
-    run 'kontena stack rm --force vpn'
-  end
-
-  it 'removes the vpn stack' do
-    run 'kontena vpn create'
-    k = run 'kontena stack rm --force vpn'
-    expect(k.code).to eq(0)
-    k = run 'kontena stack show vpn'
-    expect(k.code).not_to eq(0)
```
SpComb: A spec for "`kontena stack show ...` fails after `kontena stack rm --force ...`" doesn't need to use the VPN stack.

PR #3204, opened 2018-01-09
Upgrade CLI excon dependency to 0.60.0

kke:
Upgrade excon http library from 0.59.0 to 0.60.0


kke:
I wouldn't add this without #2366

SpComb:
Bump to master again to get the e2e specs passing.

PR #3205, opened 2018-01-09
Upgrade CLI hash_validator dependency to 0.8.0

kke:
Upgrade hash_validator dependency from `~> 0.7.1` to 0.8.0.


SpComb:
Hmm... I would approve this if it wasn't for the e2e spec failures: https://travis-ci.org/kontena/kontena/jobs/326723050

However, I don't think those are necessarily specific to this branch... it looks like current git master is failing in the same way: https://travis-ci.org/kontena/kontena/jobs/329412872

kke:
None of the specs seem related to hashes and the validation part is quite well covered. Rebased anyway.

SpComb:
Merge conflict on #3208

kke:
Confict be gone

PR #3208, opened 2018-01-09
Drop CLI launchy dependency

### LINKED ISSUE #2288: CLI: addressable gem unresolved
Installed: https://github.com/kontena/kontena/releases/download/v1.2.2.rc1/kontena-cli-osx-v1.2.2.rc1-amd64.pkg

```
$ kontena -v
WARN: Unresolved specs during Gem::Specification.reset:
      addressable (~> 2.3)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
kontena-cli 1.2.2.rc1 [ruby2.1.8+x86_64-darwin15.0]
```

kke:
Fixes #2288

Drops launchy (used for opening a browser for interactive login) from dependencies (which brings in addressable and public_suffix which both seem to cause some problems when resolving gems).

This PR replaces launchy with a simpler `Kontena::Cli::BrowserLauncher` which should work in most common environments.

Tested on:
- [X] Mac
- [X] Desktop linux
- [X] Windows 10

### DIFF
--- a/cli/lib/kontena/cli/cloud/login_command.rb
+++ b/cli/lib/kontena/cli/cloud/login_command.rb
@@ -137,12 +137,11 @@ module Kontena::Cli::Cloud
       puts
 
       server_thread  = Thread.new { Thread.main['response'] = web_server.serve_one }
-      browser_thread = Thread.new { Kontena::Cli::BrowserLauncher.open(uri.to_s) }
+      Kontena::Cli::BrowserLauncher.open(uri.to_s)
 
       spinner "Waiting for browser authorization response" do
         server_thread.join
       end
-      browser_thread.join
 
       update_token(Thread.main['response'])
     end
--- a/cli/lib/kontena/cli/cloud/login_command.rb
+++ b/cli/lib/kontena/cli/cloud/login_command.rb
@@ -105,7 +105,7 @@ module Kontena::Cli::Cloud
       end
 
       require_relative '../localhost_web_server'
-      require 'launchy'
+      require 'kontena/cli/browser_launcher'
 
       uri = URI.parse(kontena_account.authorization_endpoint)
       uri.host ||= kontena_account.url
@@ -137,7 +137,7 @@ module Kontena::Cli::Cloud
       puts
 
       server_thread  = Thread.new { Thread.main['response'] = web_server.serve_one }
-      browser_thread = Thread.new { Launchy.open(uri.to_s) }
+      browser_thread = Thread.new { Kontena::Cli::BrowserLauncher.open(uri.to_s) }
 
       spinner "Waiting for browser authorization response" do
         server_thread.join
--- a/cli/kontena-cli.gemspec
+++ b/cli/kontena-cli.gemspec
@@ -26,7 +26,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "tty-prompt", "0.13.1"
   spec.add_runtime_dependency "clamp", "~> 1.1.0"
   spec.add_runtime_dependency "ruby_dig", "~> 0.0.2"
-  spec.add_runtime_dependency "launchy", "~> 2.4.3"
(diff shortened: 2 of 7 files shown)


SpComb:
Fixes #2288?

kke:
Yes

kke:
I think this doesn't work yet. The launch is happening inside a thread and the raise will be eaten.

SpComb:
Works on my machine (Ubuntu xenial, using `xdg-open`)

cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -137,7 +137,7 @@ def web_flow
       puts
 
       server_thread  = Thread.new { Thread.main['response'] = web_server.serve_one }
-      browser_thread = Thread.new { Launchy.open(uri.to_s) }
+      browser_thread = Thread.new { Kontena::Cli::BrowserLauncher.open(uri.to_s) }
```
SpComb: > The launch is happening inside a thread and the raise will be eaten.

Using `Thread.new { ... }` like this without any explicit rescue isn't necessarily that great, but that hasn't changed in this PR?

SpComb: It's not 100% clear to me how `thread.join` behaves, but at least the example in the docs for [`thread.value`](https://ruby-doc.org/core-2.3.1/Thread.html#method-i-value) shows it being used to handle any top-level exception raised in the thread:

```ruby
b = Thread.new { raise 'something went wrong' }
b.value   #=> RuntimeError: something went wrong
```

kke: I don't think the browser thread is needed for anything. In case the detection fails, the error raised will not be seen and it just hangs forever waiting.

kke:
@SpComb Happen to have a windows vm / machine around?

SpComb:
> @SpComb Happen to have a windows vm / machine around?

No, I'm pretty sure the dreamspark licenses I've personally used disallow any use for commercial software development.

jakolehm:
https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

kke:
Gem files for CLI with this change:

[kontena-cli-1.5.0.dev.gem.gz](https://github.com/kontena/kontena/files/1634024/kontena-cli-1.5.0.dev.gem.gz)
[kontena-cli-1.5.0.dev.gem.zip](https://github.com/kontena/kontena/files/1634025/kontena-cli-1.5.0.dev.gem.zip)

1. Download + extract 
2. `gem install kontena-cli-1.5.0.dev.gem`
3. verify it's the correct version: `kontena --version` should say `kontena-cli 1.5.0-dev`
4. try `kontena cloud login --force`or `kontena master login --force`

cli/lib/kontena/cli/cloud/login_command.rb
```diff
@@ -137,12 +137,11 @@ def web_flow
       puts
 
       server_thread  = Thread.new { Thread.main['response'] = web_server.serve_one }
-      browser_thread = Thread.new { Launchy.open(uri.to_s) }
+      Kontena::Cli::BrowserLauncher.open(uri.to_s)
```
SpComb: This is assuming that the browser launch command just opens up a new tab/window in the background and exits. That's probably the case 99% of the time, but testing with the browser not already running shows that at least `sensible-browser` will actually end up spawning a new browser process and block waiting for it to exit... that might count as a bug in `sensible-browser`, though.

Either way, yes, it's better for this to explode immediately on errors, rather than hanging in the `spinner` -> `server_thread.join` that will never happen if the browser fails to launch.

kke:
Now tested on Win10, looks like you need to escape urls in a [peculiar way](https://stackoverflow.com/questions/1327431/how-do-i-escape-ampersands-in-batch-files)

Issue #3151, opened 2017-12-18
No specs for "kontena master token" subcommand


kke:
Add some coverage

SpComb:
Still missing specs for `kontena master token remove`.

`kontena master token create` and `kontena master token list/show` are basically covered in the e2e specs, but those don't BTW actually even cleanup that token?

PR #3211, opened 2018-01-10
Add description field to master authentication access tokens

### LINKED ISSUE #3210: Access token should have optional description field

kke:
Fixes #3210

- [X] Adds an optional description field to master access tokens
- [X] Adds an optional argument to `kontena master token create` for setting a description to created tokens
- [X] Displays token description in `kontena master token <ls | show>`

Partial improvement for #3151

### DIFF
--- a/cli/lib/kontena/cli/master/token/create_command.rb
+++ b/cli/lib/kontena/cli/master/token/create_command.rb
@@ -15,6 +15,7 @@ module Kontena::Cli::Master::Token
     option ['-u', '--user'], '[EMAIL]', 'Generate a token for another user'
     option ['--id'], :flag, "Only output the token ID"
     option ['--token'], :flag, "Only output the access_token (or authorization code)"
+    option ['-d', '--description'], '[DESCRIPTION]', 'Token description'
 
     option ['--return'], :flag, "Return the response hash", hidden: true
 
@@ -22,8 +23,9 @@ module Kontena::Cli::Master::Token
       params = {
         response_type: self.code? ? 'code' : 'token',
         scope: self.scopes,
-        expires_in: self.expires_in
+        expires_in: self.expires_in,
       }
+      params[:description] = self.description if self.description
       params[:user] = self.user if self.user
       data = token_data_to_hash(client.post("/oauth2/authorize", params))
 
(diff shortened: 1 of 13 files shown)


kke:
It could be handy if the description was not in `--description` option but in a multivalued param:

```ruby
parameter '[DESCRIPTION ...]', "Add token description"
def execute
  data[:description] = description_list.join(' ') unless description_list.empty?
end
```

This would make something like this work:

```
$ kontena master token create -e 0 a token of my appreciation
```

instead of using `--description "game over, insert token"`

SpComb:
> It could be handy if the description was not in --description option but in a multivalued param:

IMO for a CLI interface it's better to not try and have something too magical for how the whitespace tokens behave, it's just more confusing for tools to define their own rules than learning how to use quotes with any generic CLI tools...

kke:
Yes

cli/lib/kontena/cli/master/token/create_command.rb
```diff
@@ -15,15 +15,17 @@ class CreateCommand < Kontena::Command
     option ['-u', '--user'], '[EMAIL]', 'Generate a token for another user'
     option ['--id'], :flag, "Only output the token ID"
     option ['--token'], :flag, "Only output the access_token (or authorization code)"
+    option ['-d', '--description'], '[DESCRIPTION]', 'Token description'
```
jakolehm: Just noticed that here we use `[DESCRIPTION]` (with brackets) and somewhere else without brackets.. maybe we should unify these?

kke: Yes it seems to be  quite half and half. Functionally I don't think there's any difference but semantically it looks like the value would be optional and without it it would work like a flag.

kke: Actually we do have some required options like:

```
$ kontena certificate import
[ERROR] --private-key is required
```

In that case, the help should be (and is):

```
--private-key PATH
```

In cases where the option is optional, maybe it should say:

```
[ -d, --description DESCRIPTION] Token description
```

instead of the current:

```
-d, --description [DESCRIPTION] Token description
```

kke:
Ref: https://github.com/matti/kommando/issues/110
Ref: https://github.com/mdub/clamp/issues/82

PR #3217, opened 2018-01-15
Fix server Celluloid::Proxy::Async<MongoPubsub> leak from RPC /container/health handler

SpComb:
Fixes  #3207

* Fix the `Rpc::ContainerHandler#health` specs: they were only testing `expect(MongoPubsub).not_to receive(:publish)`, which is meaningless: it's actually `MongoPubsub.publish_async(...)`.
* Fix `EventStream#publish_async` to not return `Celluloid::Proxy::Async<MongoPubsub>`
* Improve the server RPC to log all request/notify timings and return values at `DEBUG` level

## Example `DEBUG` output
```
D, [2018-01-15T11:11:14.694776 #8] DEBUG -- WebsocketBackend: node core-01 connected using node token with node_id XI4K:NPOL:EQJ4:S4V7:EN3B:DHC5:KZJD:F3U2:PCAN:46EV:IO4A:63S5
I, [2018-01-15T11:11:14.696643 #8]  INFO -- WebsocketBackend: node core-01 agent version 1.5.0.dev connected at 2018-01-15 11:11:14 UTC, 0.06s ago
I, [2018-01-15T11:11:14.718142 #8]  INFO -- Agent::NodePlugger: Connected node development/core-01 at 2018-01-15 11:11:14 UTC
D, [2018-01-15T11:11:14.756983 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.814566 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.832138 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.844192 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.854505 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.866495 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.881725 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.886687 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
D, [2018-01-15T11:11:14.908923 #8] DEBUG -- Rpc::NodeServicePodHandler: pod_cache miss
[...]

PR #3038, opened 2017-11-13
[WiP] Fix server MongoPubsub supervision

SpComb:
TODO: fix spec flakyness

The way the server `MongoPubsub` was using [`@supervision = Celluloid.supervise(...)`](https://github.com/kontena/kontena/blob/dd68aadceb3f8d0cc6e447b54095ae60ea2f1e37/server/app/services/mongo_pubsub.rb#L142) and [`@supervisor.actors.first`](https://github.com/kontena/kontena/blob/dd68aadceb3f8d0cc6e447b54095ae60ea2f1e37/server/app/services/mongo_pubsub.rb#L116) was wrong: calling `MongoPubsub.supervise` or `Celluloid.supervise(type: MongoPubsub` does not return a supervisor for that specific actor: it returns the default ` Celluloid::Supervision::Service::Public` supervisor actor used for **all** actors. The `MongoPubsub` class methods only worked because `MongoPubsub` was the first (and only?) actor being supervised outside of the non-default `JobSupervisor` and `WorkerSupervisor` supervision containers.

Changing the `boot_jobs.rb` to call `RpcServer.supervise` before `MongoPubsub.start!` to demonstrate this confusion:

```
MongoPubsub#start!: @supervisor=#<Celluloid::Supervision::Service::Public:0x005649a306ec10> actors=["#<RpcServer:0x005649a40b3ee0>", "#<MongoPubsub:0x005649a40a0fe8>"]
E, [2017-11-13T15:31:36.041286 #3246] ERROR -- : Actor crashed!
NoMethodError: undefined method `subscribe' for #<RpcServer:0x2b24d2059f70>
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:42:in `rescue in check'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:39:in `check'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/calls.rb:26:in `dispatch'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/call/sync.rb:16:in `dispatch'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/cell.rb:50:in `block in dispatch'
[...]

PR #3218, opened 2018-01-15
Fix server MongoPubsub to restart subscriptions after crashing

### LINKED ISSUE #3206: Server RPCs stop working if the MongoPubsub actor crashes and restarts
If some exception causes the server's `MongoPubsub` actor to crash, the actor will normally get restarted by the Celluloid supervisor. Because it is the only supervised top-level actor, the `@supervisor.actors.first` will point to the new actor, and any future `publish` etc calls will work.

However, the server `WebsocketBackend` middleware class only ever calls `MongoPubsub.subscribe` once, when the worker starts up. If the `MongoPubsub` actor crashes, then the [`WebsocketBackend#subscribe_to_rpc_channel`](https://github.com/kontena/kontena/blob/364629689a5bc00721f4dde0c3b59b8206672292/server/app/middlewares/websocket_backend.rb#L44) subscription will get lost.

This causes all server -> agent RPCs to stop working.

The `/v1/ping` API will still return HTTP 200, because the `MongoPubsub.actor.alive?` is still true - it's just lost the subscriptions.

## Examples

```
D, [2018-01-09T09:38:41.292809 #13] DEBUG -- MongoPubsub: subscribe rpc_client:1159137003
D, [2018-01-09T09:38:41.293663 #13] DEBUG -- MongoPubsub: publish rpc_client
E, [2018-01-09T09:38:41.293788 #13] ERROR -- : Actor crashed!
RuntimeError: test
	/app/app/services/mongo_pubsub.rb:99:in `publish'
I, [2018-01-09T09:38:41.293877 #13]  INFO -- MongoPubsub: terminated
I, [2018-01-09T09:38:41.294765 #13]  INFO -- MongoPubsub: initialized
I, [2018-01-09T09:38:41.297421 #13]  INFO -- MongoPubsub: starting to tail collection
172.17.0.1 - - [09/Jan/2018:09:38:51 +0000] "GET /v1/nodes/development/core-01/health HTTP/1.1" 422 52 10.1472
```

All RPCs for that server worker will timeout, because while the `RpcClient` -> `MongoPubsub.publish/subscribe` works fine, the `WebsocketBackend` never sees the messages.

SpComb:
Fixes #3206

* Store subscriptions in a persistent class variable that survives across actor crashes, instead of an instance variable that gets reset on actor crashes
* Resume subscription processing after crash
* More debug logging of pubsub actions

## Testing

RPCs now continue working after the pubsub actor crashes.

```patch
diff --git a/server/app/services/mongo_pubsub.rb b/server/app/services/mongo_pubsub.rb
index df734ac..a0d61d3 100644
--- a/server/app/services/mongo_pubsub.rb
+++ b/server/app/services/mongo_pubsub.rb
@@ -130,6 +130,8 @@ class MongoPubsub
   def publish(channel, data)
     debug "publish #{channel}"
 
+    fail "crash!" if channel == 'rpc_client' && data.to_s =~ /crash/
+
     self.collection.insert_one(
       channel: channel,
       data: BSON::Binary.new(MessagePack.pack(data)),
```

```
$ bundle exec bin/kontena node health core-01
⊛ Node is online for 8s
⊛ Node core-01 etcd is healthy
$ bundle exec bin/kontena etcd ls /crash
 [fail] Retrieving keys from etcd     \
 [error] 500 : Internal Server Error (/v1/etcd/development//crash)
$ bundle exec bin/kontena node health core-01
⊛ Node is online for 17s
⊛ Node core-01 etcd is healthy
```

```
D, [2018-01-15T12:22:32.153687 #9] DEBUG -- TokenAuthentication: Access token found
D, [2018-01-15T12:22:32.182212 #9] DEBUG -- MongoPubsub: subscribe rpc_client:41634832
D, [2018-01-15T12:22:32.183001 #9] DEBUG -- MongoPubsub: publish rpc_client
D, [2018-01-15T12:22:32.186282 #13] DEBUG -- MongoPubsub::Subscription: queued rpc_client...
D, [2018-01-15T12:22:32.187050 #9] DEBUG -- MongoPubsub::Subscription: queued rpc_client...
D, [2018-01-15T12:22:32.441491 #13] DEBUG -- MongoPubsub: publish rpc_client:41634832
D, [2018-01-15T12:22:32.443250 #9] DEBUG -- MongoPubsub::Subscription: queued rpc_client:41634832...
172.17.0.1 - - [15/Jan/2018:12:22:32 +0000] "GET /v1/nodes/development/core-01/health HTTP/1.1" 200 248 0.3937

D, [2018-01-15T12:22:34.047500 #9] DEBUG -- TokenAuthentication: Access token found
D, [2018-01-15T12:22:34.054504 #9] DEBUG -- MongoPubsub: subscribe rpc_client:1247718998
D, [2018-01-15T12:22:34.055083 #9] DEBUG -- MongoPubsub: publish rpc_client
E, [2018-01-15T12:22:34.055259 #9] ERROR -- : Actor crashed!
RuntimeError: crash!
	/app/app/services/mongo_pubsub.rb:132:in `publish'
I, [2018-01-15T12:22:34.055344 #9]  INFO -- MongoPubsub: terminated
D, [2018-01-15T12:22:34.055491 #9] DEBUG -- MongoPubsub: freezing master:leader_elect
D, [2018-01-15T12:22:34.055857 #9] DEBUG -- MongoPubsub: freezing rpc_client
D, [2018-01-15T12:22:34.056050 #9] DEBUG -- MongoPubsub: freezing FirehoseApiEvent
I, [2018-01-15T12:22:34.056394 #9]  INFO -- MongoPubsub: initialized
D, [2018-01-15T12:22:34.056474 #9] DEBUG -- MongoPubsub: freezing rpc_client:1247718998
D, [2018-01-15T12:22:34.056540 #9] DEBUG -- MongoPubsub: restoring master:leader_elect...
D, [2018-01-15T12:22:34.056670 #9] DEBUG -- MongoPubsub: restoring rpc_client...
D, [2018-01-15T12:22:34.056707 #9] DEBUG -- MongoPubsub: restoring FirehoseApiEvent...
D, [2018-01-15T12:22:34.056737 #9] DEBUG -- MongoPubsub: restoring rpc_client:1247718998...
I, [2018-01-15T12:22:34.059862 #9]  INFO -- MongoPubsub: starting to tail collection

RpcClient::TimeoutError (Connection timeout (5s)):
  /app/app/services/rpc_client.rb:90:in `wait_for_response'
  /app/app/services/rpc_client.rb:53:in `request'
  /app/app/routes/v1/etcd_api.rb:25:in `block (4 levels) in <class:EtcdApi>'
  /usr/lib/ruby/gems/2.4.0/gems/roda-2.26.0/lib/roda.rb:767:in `always'
  /usr/lib/ruby/gems/2.4.0/gems/roda-2.26.0/lib/roda.rb:460:in `is'
  /app/app/routes/v1/etcd_api.rb:22:in `block (3 levels) in <class:EtcdApi>'
  ...
  /usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/thread_pool.rb:120:in `block in spawn_thread'

172.17.0.1 - - [15/Jan/2018:12:22:39 +0000] "GET /v1/etcd/development//crash HTTP/1.1" 500 4315 5.0804

D, [2018-01-15T12:22:40.693216 #13] DEBUG -- TokenAuthentication: Access token found
D, [2018-01-15T12:22:40.700187 #13] DEBUG -- MongoPubsub: subscribe rpc_client:1805502268
D, [2018-01-15T12:22:40.700795 #13] DEBUG -- MongoPubsub: publish rpc_client
D, [2018-01-15T12:22:40.701959 #9] DEBUG -- MongoPubsub::Subscription: queued rpc_client...
D, [2018-01-15T12:22:40.702210 #13] DEBUG -- MongoPubsub::Subscription: queued rpc_client...
D, [2018-01-15T12:22:40.962856 #13] DEBUG -- MongoPubsub: publish rpc_client:1805502268
D, [2018-01-15T12:22:40.965305 #13] DEBUG -- MongoPubsub::Subscription: queued rpc_client:1805502268...
172.17.0.1 - - [15/Jan/2018:12:22:40 +0000] "GET /v1/nodes/development/core-01/health HTTP/1.1" 200 248 0.3502
```


SpComb:
For review context: the `Subscription#wait` should be more or less exactly equivalent to `defer { ... }`

```ruby
  # Perform a blocking or computationally intensive action inside an
  # asynchronous group of threads, allowing the sender to continue processing other
  # messages in its mailbox in the meantime
  def defer(&block)
    # This implementation relies on the present implementation of
    # Celluloid::Future, which uses a thread from InternalPool to run the block
    Future.new(&block).value
  end
```

server/app/services/mongo_pubsub.rb
```diff
@@ -21,14 +21,29 @@ def terminate
...
+    def start
+      @future ||= Celluloid::Future.new do
+        # process in a separate thread
+        process
+      end
+    end
+
+    # Suspend the calling celluloid task until the subscription is stopped
+    def wait
+      start.value
```
SpComb: This should be more or less exactly equivalent to `defer { ... }`.

```ruby
  # Perform a blocking or computationally intensive action inside an
  # asynchronous group of threads, allowing the sender to continue processing other
  # messages in its mailbox in the meantime
  def defer(&block)
    # This implementation relies on the present implementation of
    # Celluloid::Future, which uses a thread from InternalPool to run the block
    Future.new(&block).value
  end
```

SpComb:
Note: this isn't strictly necessary for 1.5 because #3217 fixes the only known case of `MongoPubsub` crashes, but considering how serious the #3206 effects of those actor crashes are, it would definitely be worth fixing to avoid any future issues.

server/app/services/mongo_pubsub.rb
```diff
@@ -21,14 +21,29 @@ def terminate
       stop
     end
 
+    # @return [Celluloid::Future]
+    def start
```
jakolehm: I think `start` / `wait` is a bit confusing as an api. Method `wait` is called  from `process_subscription` and that actually triggers `process`.  If `process` should be protected always then maybe it should return a future?

SpComb: Refactored to remove the `@future ||= ...` lazy-init... no reason to not spawn the future right from `initialize`.

kke:
Pending approval

server/app/services/mongo_pubsub.rb
```diff
@@ -51,28 +62,54 @@ def send_message(data)
 
   attr_accessor :collection
 
+  finalizer :finalize
+
+  # @return [Array<Subscription>]
+  def subscriptions
+    @@subscriptions ||= []
```
kke: This will be shared if the class is inherited from. A safer way to implement might be:

```ruby
def self.subscriptions
  @subscriptions ||= []
end

def subscriptions
  self.class.subscriptions
end
```

Also as this is used in a multi-threaded setup, I believe this should be some kind of thread-safe primitive, as seen in [Celluloid docs](https://github.com/celluloid/celluloid/wiki/Exclusive#all-methods)

SpComb: Per design only the actor thread should be accessing the `subscriptions`, so it shouldn't need to be threadsafe.

SpComb:
Looks like the `MongoPubsub with existing subscriptions after crashing is able to restart and process publishes` is racy on the same `initialize` => `async.tail!` race as in https://github.com/kontena/kontena/pull/3038/#issuecomment-344624933

```
    after crashing
D, [2018-02-14T16:35:58.917945 #6820] DEBUG -- MongoPubsub: subscribe test1
I, [2018-02-14T16:35:58.918531 #6820]  INFO -- MongoPubsub: terminated
D, [2018-02-14T16:35:58.918660 #6820] DEBUG -- MongoPubsub: preserving test1 on crash
I, [2018-02-14T16:35:58.919366 #6820]  INFO -- MongoPubsub: initialized
D, [2018-02-14T16:35:58.919419 #6820] DEBUG -- MongoPubsub: restoring test1...
I, [2018-02-14T16:35:58.922628 #6820]  INFO -- MongoPubsub: starting to tail collection
D, [2018-02-14T16:35:58.924762 #6820] DEBUG -- MongoPubsub: publish test1
D, [2018-02-14T16:35:58.960383 #6820] DEBUG -- MongoPubsub: stopped test1
      is able to restart and process publishes (FAILED - 1)

Failures:

  1) MongoPubsub with existing subscriptions after crashing is able to restart and process publishes
     Failure/Error: expect(test1_msgs).to eq [{'test' => 1}]
       
       expected: [{"test"=>1}]
            got: []
       
       (compared using ==)
       
       Diff:
       @@ -1,2 +1,2 @@
       -[{"test"=>1}]
       +[]
       
     # ./spec/services/mongo_pubsub_spec.rb:169:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:94:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:88:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'

```

The `publish` gets lost because the `tail! => defer { self.collection.find(...) do ... } ` loop isn't running yet...

server/spec/services/mongo_pubsub_spec.rb
```diff
@@ -131,4 +144,40 @@
...
+      described_class.subscribe('test1') do |msg|
+        test1_msgs << msg
+      end
+    }
+
+    describe 'after crashing' do
+      before do
+        described_class.actor.async.crash!
+        WaitHelper.wait_until!("restarted", timeout: 1.0) { (actor = described_class.actor) && actor.alive? && (actor.ping rescue nil)}
+        sleep 1 # XXX: the async.tail! => defer { self.collection.find(...) do ... } loop won't be running yet...
```
SpComb: This is a real bug in the `MongoPubsub#initialize` => `async.tail!` implementation: it can lose messages on restarts.

Not going to try and fix that here, hence the crude `sleep`...

SpComb: NVM, there's also a race on the `publish` => mongodb => `queue_message`... bypassed that entirely, because it's not really relevant for the restart behavior being tested here.

server/app/services/mongo_pubsub.rb
```diff
@@ -51,28 +63,59 @@ def send_message(data)
 
   attr_accessor :collection
 
+  finalizer :finalize
+
+  # @return [Array<Subscription>]
+  def self.subscriptions
+    @subscriptions ||= []
```
jakolehm: This feels wrong but I get why it's done like this. Storing global state outside of actor is not usually a good idea. Do we have any alternatives?

SpComb: Yeah, it's not entirely in line with the actor philosophy, but I'm not convinced the other approaches would be more viable: 

> EDIT: There are also other alternative approaches to fixing #3206, but I think this is the least disruptive to implement: https://github.com/kontena/kontena/issues/3206#issuecomment-357659898

SpComb: Turning each `MongoPubsub::Subscription` itself into an actor that takes care of re-subscribing might vaugely be more correct, but OTOH I think the actual practical complexity of implementing that would be higher than this class-instance-variable-array + `Celluloid::Future` hack...

SpComb: You could also nominally fix the "global state" by passing in the subscriptions array via the supervisor args, so that the actor gets the same array on each restart, but it would effectively still be the same thing.

PR #3223, opened 2018-01-30
Fix potential XSS vulnerability in master remote login code display

kke:
It's possible to inject javascript into the remote login HTML page.

Exploitability is quite small, the attacker would have to trick the user to go to something like `master_url/code#code=abcd<script>alert("foo")</script>`.

Since the master does not have any cookie authentication and the browser won't send any bearer auth headers, the attacker would seem to have very little to gain by doing so.

One possible attack could be to display a fake cloud login screen.

### DIFF
--- a/server/app/views/static/code.html
+++ b/server/app/views/static/code.html
@@ -7,7 +7,7 @@
       Authorization code:
     </p>
     <div style='background: black; color: white; font-family: "Lucida Console", Monaco, monospace; padding: 10px 10px 10px 10px; font-size: 24pt'>
-      <script>document.write(window.location.hash.split('code=')[1].split('&')[0]);</script>
+      <script>document.write(window.location.hash.split('code=')[1].split('&')[0].replace(/[^a-zA-Z0-9\-\_]/g, ''));</script>
     </div>
   </body>
 </html>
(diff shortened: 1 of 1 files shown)


kke:
ping @nevalla

kke:
image

Not sure if the regex is too harsh, maybe the code can contain other characters? Maybe just replace `<`?

SpComb:
I can't repro this, at least on firefox the `window.location.hash` is URL-encoded, so the resulting `document.write(...)` not actually HTML:

```
>>> window.location.hash
"#code=foo%3Cscript%3Ealert(%22foo%22)%3C/script%3Ebar"
```

SpComb:
Anyways, the right way to fix this XSS is to replace the `document.write` with an insintrically safe method like `innerText = ...`: https://www.owasp.org/index.php/DOM_based_XSS_Prevention_Cheat_Sheet#RULE_.237_-_Fixing_DOM_Cross-site_Scripting_Vulnerabilities

```
    <div id="code" style='...'></div>
    <script>
      var code = window.location.hash.split('code=')[1].split('&')[0];
      document.getElementById("code").innerText = code;
    </script>
```

Trying to regexp your way out of XSS injection usually just ends in tears.

SpComb:
👍  with `innerText`:

2018-02-05-113606_826x589_scrot

PR #3224, opened 2018-01-31
Some proposed refactorings to ChangeResolver refactoring PR

kke:
Add a `ChangeResolver::ResultSet` class.

- Removes the need for `upgraded_services.select { |s| s.split('/')[0] == stack }` type of filtering in report generation.
- Changes things such as `changes.added_stacks` to `changes.stacks.added`


SpComb:
Should this be milestoned for 1.5?

kke:
> Should this be milestoned for 1.5?

Doesn't change functionality or fix any errors, so it's not that important. 

Just a byproduct of trying to make sense of the original #3185 PR.

PR #3222, opened 2018-01-30
Enable gzip response parsing in CLI API client where supported

kke:
Sets the `Accept-Encoding: gzip` header to API requests. The HTTP client library handles the decompression (except when raising errors, in which case the response is decompressed in rescue for displaying)

JSON-API responses compress to about 10% of their original size, almost always more efficient than for example MsgPack (which is fine for streaming requests).

If the server does not support `Accept-Encoding: gzip` nothing is changed.

Enabling server side support in Roda-based applications such as the kontena master or any of the micro service API's is simply an addition of:

```ruby
use Rack::Deflater, include: %w(application/json application/vnd.api+json application/yaml)
```

in the `server.rb` or equivalent.

Extracted from #3219

PR #3241, opened 2018-02-05
Enable server API gzip encoding when KONTENA_SERVER_GZIP=true

SpComb:
Use `Rack::Deflater` on the server for JSON API response compression.

Tested together with  #3222 for CLI `Accept-Encoding: gzip` support. Might need some `pre`-release integration testing re `kontena/haproxy`?

PR #3242, opened 2018-02-05
Fix CLI to not accept gzip responses for streaming get requests

SpComb:
Fixes #3222

Commands like `kontena service logs -f` are broken with `Content-Encoding: gzip`, because the `response_block` gets the raw gzip response data: https://github.com/kontena/kontena/pull/3241#issuecomment-363033113


kke:
Merging to original PR

Issue #3246, opened 2018-02-05
The server VAULT_KEY cannot be rotated


SpComb:
The server only supports a single `VAULT_KEY` env for [configuring the private key used for the symmetric encryption](https://github.com/kontena/kontena/blob/ff5bb476ce3852b61f7a40cbdeb2539e69531d87/server/app/initializers/01_symmetric_encryption.rb#L5-L9). This means that the `VAULT_KEY` cannot be changed without breaking all existing vault secrets / certificates / domain-authorization challenges and (secret) configuration values stored in mongo.

The `SymmetricEncryption` gem supports the use of multiple [`SymmetricEncryption.secondary_ciphers = [...]`](https://github.com/rocketjob/symmetric-encryption/blob/v3.9.1/lib/symmetric_encryption/config.rb#L16) ciphers for key rotation. This requires the encrypted values to stored with a header containing a version number (from `SymmetricEncryption::Cipher.new(version: ...)`, or passed to `SymmetricEncryption.decrypt(version: ..)`. The main `SymmetricEncryption.cipher = ...` is always used for encryption, or as a fallback when there is no ciphertext version (as for all existing encrypted secrets).

The `kontena/server` should support the configuration of multiple vault keys, allowing for key rotation... however, this also requires each key to have an associated version number (0..255): when a new key is configured, the old keys must keep the same version numbers in order to decrypt existing secrets. If old keys are removed, the version numbers of the remaining keys must stay the same.

At the same time, the new `VAULT_KEY` should also be changed to use a Key Derivation Function (KDF) to preserve all of the entropy in the configured key, rather than truncating it to 32 bytes... this currently results in fewer than 256 bits of resulting entropy assuming that the configured key only uses printable chars, or worst-case 128-bits if only using hexadecimal characters.

PR #3248, opened 2018-02-05
Server: Derive stronger SymmetricEncryption key from the configured VAULT_KEY

### LINKED ISSUE #3247: The server VAULT_KEY gets truncated to 32 bytes
The server `VAULT_KEY` is typically configured by generating either a base64 or hexadecimal string containing the desired number of bits of entropy. However, the `ENV['VAULT_KEY']` gets used as-is for the [`SymmetricEncryption::Cipher.new(key: ...)`](https://github.com/kontena/kontena/blob/5064602c62992ae5d701c6d01d83704c486562d4/server/app/initializers/01_symmetric_encryption.rb#L6). As the `OpenSSL::Cipher.new('AES-CBC-256')` requires a key that is exactly 32 bytes (256 bits) in length:

* `VAULT_KEY` values shorter than 32 bytes fail: `ArgumentError: key must be 32 bytes`

* `VAULT_KEY` values longer than 32 bytes get truncated to 32 bytes

    Prior to Kontena 1.4 / Ruby 2.4, this was done by `OpenSSL::Cipher.key=` itself: https://github.com/kontena/kontena/pull/2456#issuecomment-307761832

    In Kontena 1.4, this is done by the `kontena/server` for compatibility: https://github.com/kontena/kontena/blob/5064602c62992ae5d701c6d01d83704c486562d4/server/app/initializers/01_symmetric_encryption.rb#L5-L9

This truncation means that for a `SecureRandom.hex(24)` value containing `48 * 4 = 192` bits of entropy, only the first `32 * 4` bits are used, resulting in 128 effective bits of entropy in the key. A 128-bit AES key is still effective, but not as effective as a 192-bit or 256-bit key.

The `VAULT_KEY` should be changed to use a Key Derivation Function (KDF) to preserve all of the entropy in the configured key, rather than truncating it to 32 bytes.

SpComb:
Fixes #3247 by re-encrypting all encrypted values in MongoDB using a stronger derived key utilizing all of the entropy in the `VAULT_KEY`

Instead of truncating the `VAULT_KEY` to 32 bytes, use [HKDF](https://tools.ietf.org/html/rfc5869) ([gem](https://rubygems.org/gems/hkdf/versions/0.3.0), also included in the new [ruby 2.5 `OpenSSL::KDF`](https://ruby-doc.org/stdlib-2.5.0/libdoc/openssl/rdoc/OpenSSL/KDF.html#method-c-hkdf)) to derive a stronger key utilizing all of the entropy provided in the complete `ENV['VAULT_KEY']`.

Because the `SymmetricEncryption::Cipher.new(key: ...)` is now different, this requires setting up a primary cipher used for encryption, plus a secondary cipher for decrypting legacy stored secrets. This is implemented by configuring the primary key with `version: 1`, such that the legacy cipher with `version: 0` gets used for decrypting secrets that either do not have any header, or have a header with the previously used default cipher `version: 0`.

The new primary cipher also does not require any `VAULT_IV` to be configured, because it is configured to `always_add_header`, assuming that all encryption happens using `random_iv: true`.

Also contains a migration to re-encrypt all vault secrets using the new, stronger cipher.

## TODO
- [x] test what happens for `SymmetricEncryption.encrypt(..., random_iv: false)` without configuring any `SymmetricEncryption::Cipher.new(iv: ...)`
- [x] specs
- [x] also migrate other encrypted values:
  - [x] `GridSecret.encrypted_value`
  - [x] `Configuration.encrpyted_value`
  - [x] `Certificate.encrypted_private_key`
  - [x] `GridDomainAuthorization.encrypted_tls_sni_certificate`


server/app/initializers/01_symmetric_encryption.rb
```diff
@@ -1,10 +1,38 @@
...
+    legacy_cipher = SymmetricEncryption::Cipher.new(
+      version:     0, # default, as used for encrypting existing secrets with headers
+      cipher_name: 'aes-256-cbc',
+      key:         vault_key[0...32],
+      iv:          vault_iv[0...16],
+    )
+
+    # used to lookup cipher for vault values having a header with version=0 using the legacy key
+    SymmetricEncryption.secondary_ciphers = [
+      legacy_cipher,
```
SpComb: This secondary cipher with `version=0` is a little bit silly, because it will never actually get used in any production release, assuming that this PR gets merged before then... it will only ever get used for decrypting values generated using git master at #3184, which still entrpys values using the truncated key and writes them out to mongo with a `version: 0` header.

Existing secrets from previous releases that don't have any header will use the `SymmetricEncryption.select_cipher => legacy_cipher`.

SpComb:
Also bump symmetric-encryption from 3.8.1 => 3.9.1. I don't see any backwards-incompatible changes: https://github.com/rocketjob/symmetric-encryption/compare/v3.8.1...v3.9.1

Saving the symmetric-encryption 4.0 bump for a later PR, because that has backwards-incompatible changes that also require changes to the server code.

server/app/initializers/01_symmetric_encryption.rb
SpComb: Actually, that's not true... the [`Configuration`](https://github.com/kontena/kontena/blob/v1.4.3/server/app/models/configuration.rb#L32) model already stores encrypted values with a header, so we do really need this `version: 0` secondary cipher, also for release versions.

server/app/initializers/01_symmetric_encryption.rb
```diff
@@ -1,10 +1,39 @@
...
+require 'hkdf'
 
-if ENV['VAULT_KEY'] && ENV['VAULT_IV']
+if vault_key = ENV['VAULT_KEY']
+  # generate strong keys using entropy derived from the weak (human-readable) input key
+  hkdf = HKDF.new(vault_key)
+
+  # primary cipher used for encryption
+  # not configured using any fixed IV, assuming that we always encrypt with a random IV
+  # default IV implies an IV of all zeroes, in case something does encrypt without random_iv: true
```
SpComb: > test what happens for SymmetricEncryption.encrypt(..., random_iv: false) without configuring any SymmetricEncryption::Cipher.new(iv: ...)

You end up with `iv: nil`, which ends up using the OpenSSL default of an all-zeroes IV:

```
irb(main):024:0> SymmetricEncryption::Cipher.parse_header!(SymmetricEncryption.cipher.decode(SymmetricEncryption.encrypt('foobar', false)))
=> #<struct SymmetricEncryption::Cipher::HeaderStruct compressed=false, iv=nil, key=nil, cipher_name=nil, version=1, decryption_cipher=#<SymmetricEncryption::Cipher:0x2adcedbbd6b4 @key="[FILTERED]" @iv=nil @cipher_name="aes-256-cbc", @version=1, @encoding=:base64, @always_add_header=true>
```

From: https://www.reddit.com/r/ruby/comments/ndl1p/be_careful_rubys_openssl_rulez/

> Currently the "default IV" being used is an all zeroes IV (see https://gist.github.com/1480700), which is also not ideal, but note that this default is actually OpenSSL's default and it needs to be kept for the same reason of backwards compatibility.

See https://gist.github.com/emboss/1480700

SpComb:
Ready for review.

IMO this should be included in 1.5 together with #3184. Not sure about the #3246 `VAULT_KEY` rotation issue.

server/app/initializers/01_symmetric_encryption.rb
```diff
@@ -1,10 +1,39 @@
...
+    )
+
+    # used to lookup cipher for vault values having a header with version=0 using the legacy key
+    SymmetricEncryption.secondary_ciphers = [
+      legacy_cipher,
+    ]
+
+    # used to lookup cipher for vault values not having any header
+    SymmetricEncryption.select_cipher do
+      legacy_cipher
```
jakolehm: I think this needs a bit more explanation.

SpComb: See:

* https://github.com/rocketjob/symmetric-encryption/blob/v3.9.1/lib/symmetric_encryption/symmetric_encryption.rb#L219-L246
* https://github.com/rocketjob/symmetric-encryption/blob/v3.9.1/lib/symmetric_encryption/symmetric_encryption.rb#L123-L124

SpComb:
Cross-checked the `hkdf` gem against the [python implementation](https://pypi.python.org/pypi/hkdf) to ensure that they both generate the same output key for the `VAULT_KEY` that I use in development.

PR #3257, opened 2018-02-09
Warn if a service or stack name includes invalid characters for a hostname

### LINKED ISSUE #3256: Underscore should not be allowed in service names
Underscore ('_') is not allowed in DNS host names and can cause trouble with certain runtimes (e.g. Java). Kontena should display a warning when validating stack files and perhaps even remove underscores or replace them with a dash ('-') when deploying services.

kke:
Fixes #3256 

image

### DIFF
--- a/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
@@ -69,8 +69,8 @@ module Kontena::Cli::Stacks
         if yaml.key?('services')
           if yaml['services'].kind_of?(Hash)
             yaml['services'].each do |service, options|
-              unless service =~ /\A[a-z0-9\-]+\z/
-                result[:notifications] << { 'services' => { service => { 'name' => 'A service name should only include a-z, 0-9 and - characters' } } }
+              unless service =~ /\A(?!-)[a-z0-9\-]+\z/
+                result[:notifications] << { 'services' => { service => { 'name' => 'A service name should only include a-z, 0-9 and - characters and not start with the - character' } } }
               end
               unless options.kind_of?(Hash)
                 result[:errors] << { 'services' => { service => { 'options' => "must be a mapping not a #{options.class}"}  } }
--- a/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
+++ b/cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
@@ -69,6 +69,9 @@ module Kontena::Cli::Stacks
         if yaml.key?('services')
           if yaml['services'].kind_of?(Hash)
             yaml['services'].each do |service, options|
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
```diff
@@ -69,6 +69,9 @@ def validate(yaml)
         if yaml.key?('services')
           if yaml['services'].kind_of?(Hash)
             yaml['services'].each do |service, options|
+              unless service =~ /\A[a-z0-9\-]+\z/
```
SpComb: Is there any reason for the CLI and the server `GridServices::Create` to use a different regexp for validation?

https://github.com/kontena/kontena/blob/6f9de9d5518b3e3d72516906fbf25d4151b18fcf/server/app/mutations/grid_services/create.rb#L15

kke: `\w includes underscore`. But maybe should add `(?!-)` here

kke:
image

Regexp updated.

cli/lib/kontena/cli/stacks/yaml/validator_v3.rb
```diff
@@ -69,6 +69,9 @@ def validate(yaml)
         if yaml.key?('services')
           if yaml['services'].kind_of?(Hash)
             yaml['services'].each do |service, options|
+              unless service =~ /\A(?!-)[a-z0-9\-]+\z/
```
SpComb: Compared to `\w`, this also rejects `A-Z`?

```
- test.yaml:
  - services:
      Redis:
        name: A service name should only include a-z, 0-9 and - characters and not
          start with the - character
```

SpComb: OTOH DNS names are case-insensitive, so this might actually be a server bug if it lets you create two services called `redis` and `Redis`... they will overlap eachother in DNS?

kke: Doesn't reject, just warns.

SpComb: It might indeed make sense to warn about uppercase characters in the service name, yes...

SpComb:
@jakolehm third opinion on this?

I don't think we should tighten the server validations to reject names that technically work, so this kind of CLI warning for anything that's not a best-practice name seems like a good idea to me.

SpComb:
The exact same regexp checks also apply to the stack name BTW... this PR should probably validate both?

kke:
> The exact same regexp checks also apply to the stack name BTW... this PR should probably validate both?

Maybe. Btw, the dependency thing creates `stack1_stack2` stacks. Should we change it to dash?

PR #3261, opened 2018-02-12
Remove project root .ruby-version, update agent/ and server/ .ruby-version

kke:
The build system does not need them (the ruby version is set in `.travis.yml`)

Mostly it's an annoyance, constantly you run into something like:

image

and have to do something like `rbenv local 2.5.0` and then you bump into `you have unstaged changes that would be overwritten ...` when you try to switch branches.


jakolehm:
`.ruby-version` is mainly (only?) for helping local development. Not sure if it's good idea to remove these?

kke:
Then I think they should either point to an up-to-date ruby or the oldest supported ruby or the ruby version offered by the version of alpine in that component's Dockerfile.

`agent/.ruby-version` currently points to 2.3.3, but travis and alpine 3.6 run on 2.4.

The project root `.ruby-version` also says 2.3.3, can't see any benefit in that.

Maybe it would make sense to have `agent/` and `server/` .ruby-versions point to the one offered by the current version of alpine.

At least in the project root it's mostly an annoyance.

jakolehm:
Yes, project root `.ruby-version` file is not needed. Agent/server should indeed point to the same version that Dockerfile uses. Cli probably should point to the oldest supported version (2.1)?

kke:
2.1 is EOL. Also something that works on 2.1 might not work on 2.5, so it doesn't add any compatibility guarantees.

Changing this to update `.agent` to 2.4 and remove root.

jakolehm:
> 2.1 is EOL. Also something that works on 2.1 might not work on 2.5, so it doesn't add any compatibility guarantees.

I would not say that it does not add any.. it adds some (like forces not to use new style hashes and stuff).

Bigger question (and separate issue) is: should we drop 2.1 support (which means also update to omnibus stuff)?

kke:
I think yes. MacOs high sierra has 2.3. Older ones have 2.0 so you would have had to upgrade to be able to use any recent version of CLI anyway.

PR #3265, opened 2018-02-12
Change agent health check to accept HTTP 3xx as healthy

### LINKED ISSUE #1790: health_check to support other states than 200

SpComb:
Replaces #1760
Fixes #1790

It turns out that the agent `ContainerHealthCheckWorker#check_http_status` probe would previously follow any HTTP 3xx `Location: ...` redirects, and act based on the resulting HTTP response from the redirected location, or consider the service as unhealthy if the redirected request failed.

Change the agent `health_check: protocol: http` probe to match the default/hardcoded behavior of various other HTTP healthcheck implementations, including the `kontena/lb`: the HTTP healthcheck probe no longer follows HTTP 3xx redirects, but immediately accepts them as healthy.

* https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#option%20httpchk

    > ... responses 2xx and 3xx are
considered valid, while all other ones indicate a server failure

* https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes

    > The diagnostic is considered successful if the response has a status code greater than or equal to 200 and less than 400.

This PR also adds e2e tests for HTTP service healthchecks.

## TODO
- [x] fix the flaky e2e LB healthcheck specs


SpComb:
Per #1760 I don't necessarily see this as being such a great idea myself, but it makes more sense to have the defaults here match the behavior of other implementations, and in particular, the `kontena/lb` haproxy behavior.

If this is a detail that really turns out to matter, then we should consider making the healthcheck statuses configurable (as supported by e.g. haproxy), but maybe we're better off just avoiding that complexity?

SpComb:
(milestoning to 1.5 because #1790 is also milestoned)

SpComb:
e2e specs reveal an interesting detail: the agent http healthcheck probe already follows HTTP redirects... 

```
  1) kontena service health_check for a http test service returning HTTP 302 => 500 redirect for healthchecks has a healthy status
     Failure/Error: expect(check_service_health()).to eq 'healthy'
     
       expected: "healthy"
            got: "unhealthy"
     
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:61:in `block (4 levels) in <top (required)>'
```

```
curl -vL 'http://10.81.128.71:8000/health?status=302&location=/health?status=500'
*   Trying 10.81.128.71...
* Connected to 10.81.128.71 (10.81.128.71) port 8000 (#0)
> GET /health?status=302&location=/health?status=500 HTTP/1.1
> Host: 10.81.128.71:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 302 Found
< Server: BaseHTTP/0.6 Python/3.6.4
< Date: Mon, 12 Feb 2018 14:58:02 GMT
< Content-Type: text/plain; charset=utf-8
< Content-Length: 23
< Location: /health?status=500
< 
* Closing connection 0
* Issue another request to this URL: 'http://10.81.128.71:8000/health?status=500'
* Hostname 10.81.128.71 was found in DNS cache
*   Trying 10.81.128.71...
* Connected to 10.81.128.71 (10.81.128.71) port 8000 (#1)
> GET /health?status=500 HTTP/1.0
> Host: 10.81.128.71:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 500 Internal Server Error
< Server: BaseHTTP/0.6 Python/3.6.4
< Date: Mon, 12 Feb 2018 14:58:02 GMT
< Content-Type: text/plain; charset=utf-8
< Content-Length: 23
< 
Response status is 500
* Closing connection 1
```

SpComb:
The agent `Excon.get(...)` following the 3xx redirects means that this PR is useless, the healthcheck probe is not going to normally return any 3xx status codes 😖

SpComb:
So the interesting question is: should the agent accept HTTP 3xx responses as-is as healthy, or should it follow those 3xx redirects and expect the redirected response to return HTTP 2xx?

SpComb:
The e2e spec seems to be racy on the LB, I'm assuming it can hit the LB with a HTTP request before it has reconfigured itself... need to figure out something to deal with that, but unfortunately it's the same HTTP 503 response both for an unconfigured LB and the negative healthcheck case?

```
  1) kontena service health_check for a http test service returning HTTP 200 for healthchecks returns HTTP 200 via the LB
     Failure/Error: expect(check_lb_response_code).to eq 200
       expected: 200
            got: 503
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:46:in `block (4 levels) in <top (required)>'
```

The other e2e spec failure is about the design decision:

```
  2) kontena service health_check for a http test service returning HTTP 302 => 500 redirect for healthchecks has a healthy status
     Failure/Error: expect(check_service_health('healthcheck-test-302-500/server')).to eq 'healthy'
       expected: "healthy"
            got: "unhealthy"
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:86:in `block (4 levels) in <top (required)>'
```

jakolehm:
> So the interesting question is: should the agent accept HTTP 3xx responses as-is as healthy, or should it follow those 3xx redirects and expect the redirected response to return HTTP 2xx?

`3xx` redirect can point to `https` and it's not probably a valid hostname. This probably can be fixed by not validating certs?

SpComb:
>> So the interesting question is: should the agent accept HTTP 3xx responses as-is as healthy, or should it follow those 3xx redirects and expect the redirected response to return HTTP 2xx?
> 
> 3xx redirect can point to https and it's not probably a valid hostname. This probably can be fixed by not validating certs?

Based on the choice of matching what haproxy etc does, the straightforward thing to do would be to not have the agent follow HTTP 3xx redirects for healthcheck probes, and accept the 3xx response as-is as healthy... because that's what haproxy seems to do.

jakolehm:
@jnummelin wdyt?

SpComb:
Amusingly excon itself doesn't default to following HTTP redirects, but [`docker-api` configures it to do so globally by default](https://github.com/swipely/docker-api/blob/v1.33.6/lib/docker.rb#L17):

```
irb(main):001:0> require 'excon'
=> true
irb(main):002:0> Excon.defaults[:middlewares]
=> [Excon::Middleware::ResponseParser, Excon::Middleware::Expects, Excon::Middleware::Idempotent, Excon::Middleware::Instrumentor, Excon::Middleware::Mock]
irb(main):003:0> require 'docker-api'
=> true
irb(main):004:0> Excon.defaults[:middlewares]
=> [Excon::Middleware::Hijack, Excon::Middleware::ResponseParser, Excon::Middleware::Expects, Excon::Middleware::Idempotent, Excon::Middleware::Instrumentor, Excon::Middleware::Mock, Excon::Middleware::RedirectFollower]
```

SpComb:
Overriding the middlewares to leave out the `RedirectFollower` now results in the intended behavior:

```
D, [2018-02-13T09:15:54.083408 #1] DEBUG -- Kontena::Workers::ContainerHealthCheckWorker: checking health for container: /healthcheck.server-1 using url: http://10.81.128.104:8000/health?status=302&location=/health?status=500
D, [2018-02-13T09:15:54.088026 #1] DEBUG -- Kontena::Workers::ContainerHealthCheckWorker: got status: 302
```

SpComb:
Hmm... the lb healthchecks must not fail fast enough...

```
  1) kontena service health_check for a http test service returning HTTP 500 for healthchecks returns HTTP 503 via the LB
     Failure/Error: expect(check_lb_response_code).to eq 503
     
       expected: 503
            got: 200
     
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:117:in `block (4 levels) in <top (required)>'
```

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -78,10 +76,17 @@ def check_http_status(ip, port, uri, timeout)
           connect_timeout: timeout,
           headers: {
             "User-Agent" => "Kontena-Agent/#{Kontena::Agent::VERSION}"
-          }
+          },
+          middlewares: [
```
kke: How about `Excon.defaults[:middlewares] - [Excon::Middleware::RedirectFollower]`

agent/lib/kontena/workers/container_health_check_worker.rb
```diff
@@ -78,10 +76,17 @@ def check_http_status(ip, port, uri, timeout)
...
+            Excon::Middleware::ResponseParser,
+            Excon::Middleware::Expects,
+            Excon::Middleware::Idempotent,
+            Excon::Middleware::Instrumentor,
+            Excon::Middleware::Mock,
+          ]
         )
         debug "got status: #{response.status}"
-        data['status'] = HEALTHY_STATUSES.include?(response.status) ? 'healthy' : 'unhealthy'
+        data['status'] = (response.status >= 200 && response.status < 400) ? 'healthy' : 'unhealthy'
```
kke: How about `(200..399).cover?(response.status)` or `response.status.between?(200, 399)`

SpComb:
Looks like there's still some source of flakyness in the e2e LB specs... not sure what's causing it

```
  1) kontena service health_check for a http test service returning HTTP 302 => 500 redirect for healthchecks returns HTTP 200 via the LB
     Failure/Error: expect(check_lb_response_code).to eq 200
       expected: 200
            got: 503
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:99:in `block (4 levels) in <top (required)>'
```

agent/lib/kontena/workers/container_health_check_worker.rb
SpComb: I considered `HEALTHY_STATUSES = (200...400)` and `HEALTHY_STATUSES.include?(response.status)`, but the `...` vs `..` is way too subtle to actually use. I would vaguely prefer the more explicit use of `>=` and `<` for the open vs closed range... there's many different ways to write this condition.

SpComb: Reverted this to just keep using `HEALTHY_STATUSES.include?`... but got rid of the unnecessary `to_a` which starts to hurt me now that the resulting array actually contains ~200 entries.

jnummelin:
> @jnummelin wdyt?

I think it makes sense NOT to follow redirects.

SpComb:
> I think it makes sense NOT to follow redirects.

That is what this PR should do now.

The e2e specs for the LB healthcheck behavior seem tricky. I suspect it's because the LB config update / haproxy healthcheck can be delayed... I assume the e2e specs will either need some liberal `sleep` or other more complex trickery to work... maybe I should just leave those out?

SpComb:
Changed to just retry on 503 to make sure it's stable... 🤞 

```
kontena service health_check
  for a http test service
    returning HTTP 200 for healthchecks
      has a healthy status
GET http://localhost/ => 503
GET http://localhost/ => 200
      returns HTTP 200 via the LB
    returning HTTP 302 => 200 redirect for healthchecks
      has a healthy status
GET http://localhost/ => 503
GET http://localhost/ => 200
      returns HTTP 200 via the LB
    returning HTTP 302 => 500 redirect for healthchecks
      has a healthy status
GET http://localhost/ => 503
GET http://localhost/ => 200
      returns HTTP 200 via the LB
    returning HTTP 500 for healthchecks
      has a healthy status
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 503
      returns HTTP 503 via the LB
```

agent/lib/kontena/workers/container_health_check_worker.rb
kke: `Range#cover?` does not `.to_a`, `Comparable#between?` neither.

SpComb: The existing code does, but I don't think there's any good reason for it to do so: https://github.com/kontena/kontena/blob/e95dc72104e7207a8dd70e6b31ce4666f7682aa4/agent/lib/kontena/workers/container_health_check_worker.rb#L12

test/spec/fixtures/stack/healthcheck/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:latest
+
+ADD server.py .
+
+CMD ["python3", "server.py"]
```
jakolehm: Why 🐍 ?

SpComb: First thing at hand. I can rewrite this in ruby using `WEBrick`...

SpComb:
Looks like it is still possible for haproxy to temporarily return HTTP 200 before the healtchecks kick in:

```
    returning HTTP 500 for healthchecks
      has a healthy status
GET http://localhost/ => 200
      returns HTTP 503 via the LB (FAILED - 1)

  1) kontena service health_check for a http test service returning HTTP 500 for healthchecks returns HTTP 503 via the LB
     Failure/Error: expect(check_lb_response_code(retry_503: 3)).to eq 503
       expected: 503
            got: 200
       (compared using ==)
     # ./spec/features/stack/healthcheck_spec.rb:121:in `block (4 levels) in <top (required)>'
```

The other e2e spec failure is unrelated flakyness from container name conflicts.

SpComb:
Testing with just the `returns HTTP ... via the LB` examples shows that the LB tends to take up to 3-4s after the `kontena stack install` completes to start returning responses from the backend... the retry logic can handle it, although normally the `has a healthy status` spec in between is slow enough to avoid those:

```
kontena service health_check
  for a http test service
    returning HTTP 200 for healthchecks
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 200
GET http://localhost/ => 200
      returns HTTP 200 via the LB
```

SpComb:
The new retry code should also be able to  handle the random HTTP 200 responses before the healthcheck kicks in... unless the healthcheck interval is also exactly 1s?

```
    returning HTTP 500 for healthchecks
GET http://localhost/ => 503
GET http://localhost/ => 200
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 503
GET http://localhost/ => 503
      returns HTTP 503 via the LB
```

SpComb:
e2e specs are now green and I've been running the LB specs locally in a loop for a while now, and they're all passing.

Seeing some weird transient stuff in the response patterns that might be considered actual bugs in the LB behavior, but eventually it does stabilize.

PR #3276, opened 2018-02-13
Fix server RpcServer to not hang the actor thread

### LINKED ISSUE #3040: Server RpcServer actor does not respond to any actor calls, hangs worker shutdown
See #2348 for an example

The server [`RpcServer`](https://github.com/kontena/kontena/blob/dd68aadceb3f8d0cc6e447b54095ae60ea2f1e37/server/app/services/rpc_server.rb#L41) has a task running the `process!` method, which sits in a loop calling [`SizedQueue.pop`](https://github.com/kontena/kontena/blob/dd68aadceb3f8d0cc6e447b54095ae60ea2f1e37/server/app/services/rpc_server.rb#L41). This is a blocking operation, so the Celluloid task never yields, and the actor thread never gets a chance to process any mailbox messages.

This can easily lead to deadlocks when other actors (inadvertently) attempt sync method calls on the `RpcServer` actor: this can also happen implicitly as as a part of `inspect`, which can even happen implicitly while handling `NameError` exceptions: #2348

This also causes the server shutdown to timeout, because the `RpcServer` actor does not respect the `terminate!` call... as revealed by [some celluloid hooking](https://gist.github.com/SpComb/1700fd30fca8f8f90bd52d5217c5c777):

```
JOIN RpcServer... @ /app/app/initializers/celluloid.rb:21:in `join'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:134:in `block (2 levels) in shutdown'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:132:in `each'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:132:in `block in shutdown'
	/usr/lib/ruby/2.4.0/timeout.rb:93:in `block in timeout'
	/usr/lib/ruby/2.4.0/timeout.rb:33:in `block in catch'
	/usr/lib/ruby/2.4.0/timeout.rb:33:in `catch'
	/usr/lib/ruby/2.4.0/timeout.rb:33:in `catch'
	/usr/lib/ruby/2.4.0/timeout.rb:108:in `timeout'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid/actor/system.rb:121:in `shutdown'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:184:in `shutdown'
	/usr/lib/ruby/gems/2.4.0/gems/celluloid-0.17.3/lib/celluloid.rb:173:in `block in register_shutdown'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cluster.rb:126:in `fork'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cluster.rb:126:in `block in spawn_workers'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cluster.rb:122:in `times'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cluster.rb:122:in `spawn_workers'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cluster.rb:447:in `run'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/launcher.rb:174:in `run'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/lib/puma/cli.rb:77:in `run'
	/usr/lib/ruby/gems/2.4.0/gems/puma-3.9.1/bin/puma:10:in `<top (required)>'
	/usr/bin/puma:22:in `load'
	/usr/bin/puma:22:in `<main>'
E, [2017-11-13T17:44:17.166423 #8] ERROR -- : Couldn't cleanly terminate all actors in 10 seconds!
```

Disabling the `RpcServer#process!` task allows the server to shutdown cleanly.

SpComb:
Fixes #3040 to keep the `RpcServer` actor responsive to actor calls

The server now shuts down in <1s on Ctrl-C... that should lead to at least 10% development productivity improvements :)

* Refactor the `WebsocketBackend` <-> `RpcServer` interface to be less of a mess and add some specs
* Change the  `RpcServer#process` task to run in a `defer { ... }` thread, leaving the actor thread free to respond to actor calls

   However, this now means that all of the `handle_*` functions run in the `defer { ... }` thread, not in the actor thread... shouldn't be a problem, but it's ugly.

### DIFF
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -293,7 +293,7 @@ class WebsocketBackend
     if client = client_for_ws(ws)
       RpcServer.handle_rpc_request(client[:grid_id].to_s, data) do |message|
         EM.next_tick { # important to push sending back to EM reactor thread
-          send_message(message)
+          send_message(ws, message)
         }
       end
     end
--- a/server/app/middlewares/websocket_backend.rb
+++ b/server/app/middlewares/websocket_backend.rb
@@ -15,8 +15,6 @@ class WebsocketBackend
   CLOCK_SKEW = Kernel::Float(ENV['KONTENA_CLOCK_SKEW'] || 1.seconds)
 
   RPC_MSG_TYPES = %w(request notify)
-  QUEUE_WATCH_PERIOD = 60 # once in a minute
-  QUEUE_DROP_NOTIFICATIONS_LIMIT = (RpcServer::QUEUE_SIZE * 0.8)
 
   class CloseError < StandardError
     attr_reader :code
@@ -43,7 +41,6 @@ class WebsocketBackend
     @msg_dropped = 0
     subscribe_to_rpc_channel
     watch_connections
-    watch_queue
     watchdog
   end
 
@@ -293,25 +290,20 @@ class WebsocketBackend
   # @param [Faye::WebSocket::Event] ws
   # @param [Array] data
   def handle_rpc_request(ws, data)
-    client = client_for_ws(ws)
-    if client
-      self.rpc_queue << [ws, client[:grid_id].to_s, data]
+    if client = client_for_ws(ws)
+      RpcServer.handle_rpc_request(client[:grid_id].to_s, data) do |message|
+        EM.next_tick { # important to push sending back to EM reactor thread
+          send_message(message)
+        }
+      end
     end
   end
 
   # @param [Faye::WebSocket::Event] ws
   # @param [Array] data
   def handle_rpc_notification(ws, data)
-    rpc_queue = self.rpc_queue
-
-    if rpc_queue.size > QUEUE_DROP_NOTIFICATIONS_LIMIT # too busy to handle notifications
-      @msg_dropped += 1
-      return
-    end
-
-    client = client_for_ws(ws)
-    if client
-      rpc_queue << [client[:grid_id].to_s, data]
+    if client = client_for_ws(ws)
+      RpcServer.handle_rpc_notification(client[:grid_id].to_s, data)
     end
   end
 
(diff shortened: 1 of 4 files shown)


server/app/services/rpc_server.rb
```diff
@@ -36,68 +68,98 @@ def initialize(code, message)
...
-      size = data.size
-      if size == 2
-        handle_notification(data[0], data[1])
-      elsif size == 3
-        handle_request(data[0], data[1], data[2])
+
+    # separate thread for blocking queue.pop
+    # XXX: all of the handle_* methods run in the defer thread, not the actor thread...
+    defer {
+      while @processing && data = self.queue.pop
```
SpComb: The `@processing` is still needed, because otherwise the actor crashing and restarting would spawn a new `process => defer { ... }` thread while the old one would keep running, I think...

server/app/services/rpc_server.rb
```diff
@@ -128,15 +190,9 @@ def handling_instance(grid_id, name)
     @handlers[grid_id][name]
   end
 
-  # @param [Faye::Websocket] ws
-  # @param [Object] message
-  def send_message(ws, message)
-    EM.next_tick { # important to push sending back to EM reactor thread
-      ws.send(MessagePack.dump(message.as_json).bytes)
```
SpComb: The `RpcServer` class shouldn't need to know anything about `EventMachine`, the `Faye::Websocket` or `MessagePack`...

server/app/services/rpc_server.rb
```diff
@@ -36,68 +68,98 @@ def initialize(code, message)
...
-    while @processing && data = @queue.pop
-      @counter += 1
-      size = data.size
-      if size == 2
-        handle_notification(data[0], data[1])
-      elsif size == 3
-        handle_request(data[0], data[1], data[2])
+
+    # separate thread for blocking queue.pop
+    # XXX: all of the handle_* methods run in the defer thread, not the actor thread...
```
SpComb: very difficult to avoid this without digging into the `Celluloid` internals, I think.

server/app/middlewares/websocket_backend.rb
```diff
@@ -293,25 +290,20 @@ def handle_rpc_response(data)
   # @param [Faye::WebSocket::Event] ws
   # @param [Array] data
   def handle_rpc_request(ws, data)
-    client = client_for_ws(ws)
-    if client
-      self.rpc_queue << [ws, client[:grid_id].to_s, data]
+    if client = client_for_ws(ws)
+      RpcServer.handle_rpc_request(client[:grid_id].to_s, data) do |message|
```
jakolehm: Does this improve performance (because it skips mailbox)?

SpComb: No, the `RpcServer.queue` is unchanged, it's just refactored so that the `WebsocketBackend` class doesn't need to know about it.

PR #3277, opened 2018-02-14
Change service start/stop to execute as deploys

SpComb:
Fixes #2290 to have `GridServices::Scale` set the service `state` to `running`
Maybe fixes #2710 to cancel running service deploys if service is redeployed

Change the `kontena service start/stop` => `POST /v1/services/.../{start,stop}`=> `GridServices::Start/Stop` to just update the `GridService.state = 'running' / 'stopped'` and create a `GridServiceDeploy`. The API now returns a deploy object, and the CLI will spin waiting for the services to actually start/stop. The new API should be backwards-compatible.

## TODO
- [ ] Change the `GridServiceDeployer` to run all of the instance `start/stop` deploys in parallel?
- [ ] Fix the CLI to show the `start/stop` deploy result correctly (not `Deployed ...` but `Stopped ...`)
- [ ] Test the redeploy => cancel mechanism, also for stack deploys

PR #3278, opened 2018-02-14
[WiP] Wait for service containers to terminate when removing

### LINKED ISSUE #3030: stack install immediately after stack rm on a slow node will cause ConflictError
When a node is not fast enough to remove containers:
```
$ kontena stack rm --force mystack && kontena stack install mystack
ServiceError: Docker::Error::ConflictError: Conflict. The name "/mystack.myservice-1" is already in use by container
```

SpComb:
Follows #3277
Fixes #3030 Docker container name conflicts when deploying new services immediately after removing an earlier version

Adds a new `POST /v1/services/.../terminate` => `GridServices::Terminate` that deploys the service in the `terminated` state. The CLI `kontena service remove` waits for the service to terminate before removing it, ensuring that no lingering containers get left behind.

The agent is otherwise already capable of dealing with `service_pod.desired_state = 'terminated'`, but it just needs to be fixed to sync the terminated state back to the master instead of terminating the `ServicePodWorker` actor. Now the `ServicePodWorker` actor only gets terminated when the `ServicePodManager` destroys it.

## TODO
- [ ] specs
- [ ] Update stack remove to use the same semantics

### DIFF
--- a/agent/lib/kontena/workers/service_pod_worker.rb
+++ b/agent/lib/kontena/workers/service_pod_worker.rb
@@ -19,6 +19,7 @@ module Kontena::Workers
 
     attr_reader :node, :prev_state, :service_pod
     attr_accessor :service_pod, :hook_manager
+    finalizer :finalize
 
     # @param node [Node]
     # @param service_pod [ServicePod]
@@ -135,7 +136,11 @@ module Kontena::Workers
 
     def destroy
       @service_pod.mark_as_terminated
-      apply
+      if apply
+        # Only terminate this actor after we have succesfully ensure_terminated the Docker container and sync'd back to master
+        # Otherwise, stick around... the manager will notice we're still there and re-signal to destroy
+        self.terminate
+      end
     end
 
     def apply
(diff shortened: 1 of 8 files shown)


agent/lib/kontena/workers/service_pod_worker.rb
```diff
@@ -435,10 +439,13 @@ def sync_state_to_master(service_pod, service_container, error = nil)
           debug "skip #{service_pod} state sync at #{service_pod.deploy_rev}: unchanged"
         else
           debug "sync #{service_pod} state at #{service_pod.deploy_rev}: #{state}"
-          rpc_client.async.request('/node_service_pods/set_state', [node.id, state])
+          rpc_client.request('/node_service_pods/set_state', [node.id, state])
```
SpComb: This change to drop the `async` isn't entirely necessary for this PR, but I think this is probably a good idea.

As commented below, this means that the `@prev_state` doesn't get updated if the RPC fails, so the agent could re-sync the state later... although that would require something to call `sync_state_to_master` again, which won't happen just because the `sync_state_to_master` failed...

Issue #2928, opened 2017-10-10
Stack name should allow dot (.) symbol


ptsurbeleu:
It would be great if `USERNAME/STACK` naming schema would allow use of dot (.) symbol. This naming convention follows precisely the already familiar convention of FQDNs and usernames, for example:

```
stack: john.doe/blog.about.me
version: 0.1.1
...
```
*versus this*
```
stack: john-doe/blog-about-me
version: 0.1.1
...
```

Allowing stacks to be named after their respective resources is very handy and useful when you troubleshoot/monitor any of the customer issues on the cluster.

Otherwise, I would have to either hash the original string or "sanitize" it and honestly I dislike both of these techniques...

SpComb:
Stack names are not URLs, though... The stack YAML has a free-form `description` field that might be useful for describing the resources, but it's not really visible anywhere in Kontena.

I don't know if allowing domain names in the stack owner (user/organization) part is feasible or not, that would mainly depend on the oauth provider / Kontena cloud side... it doesn't really have any effect on the actual Kontena platform.

For the actual stack name part after the `/`, it would still need to be mangled for the actual installed stack name, because that gets used as part of the grid DNS domain, so the `.` already has a meaning there. We don't want a stack called `test/example.com` to result in services called `*.example.com.gridname.kontena.local`.

ptsurbeleu:
@SpComb, okie, conjuring `stack` field for that matter was an obvious choice from my side and as you explained the field has actually a reference in the networking model of the grid it is deployed to.

I do agree with you that it isn't a good idea to mess with it.

How about introducing an optional `display_name: test/example.com` property that is for humans and not machines?
```
stack: fe34ae10/d0a1bcef89
version: 0.1.1
display_name: john.doe/test.example.com
...
```

Or as a more far-reaching alternative an optional dictionary-alike `attributes` property.
```
stack: fe34ae10/d0a1bcef89
version: 0.1.1
...
attributes:
  display_name: john.doe/test.example.com
  content_revision: xyz
  custom_property: aaa
...
```

In that case, when CLI displays grid of stacks it can use `display_name` attribute as a substitute for `NAME` column. That seems to be less of a radical change and more like a cosmetic one and seems to be adding up to `developer-friendly` portfolio of Kontena.

I volunteer to write code for that if you folks are okay with that proposal.

Appreciate your feedback very much!

kke:
You can use `data:` in YAML top level.

[...]

PR #3280, opened 2018-02-15
WIP: Allow stacks to define labels (#2928)

ptsurbeleu:


test/docker-compose.yml
```diff
@@ -1,6 +1,15 @@
 version: '2'
 
 services:
+  dev:
```
ptsurbeleu: Developing in a container probably is not for everyone and I would remove that change before starting a final PR.

kke:
Looking nice this far, but what are the labels here used for?

Something like filtering by `kontena stack ls -l foo=bar`?

In the shortly upcoming 1.5 the stack files have a top level `meta` keyword that can be used to add tags:

```yaml
stack: foo/bar
version: 0.1.0
services:
  redis:
    image: redis:latest
meta:
  tags:
    - foobar
    - bar=foo
```

Those could perhaps be utilized here also.

It would require that the master also stores the meta separately and returns it in the `GET` response.

ptsurbeleu:
The idea (originates from https://github.com/kontena/kontena/issues/2928) is to provide the same experience as with `kontena node ls`:
```
NAME                  VERSION   STATUS      INITIAL   LABELS
⊛ muddy-night-65      1.4.3     online 1d   1 / 3     region=nyc1,az=nyc1,provider=digitalocean
⊛ wild-rain-24        1.4.3     online 4d   2 / 3     region=lon1,az=lon1,provider=digitalocean
⊛ wandering-hill-55   1.4.3     online 4d   3 / 3     region=sfo2,az=sfo2,provider=digitalocean
⊛ icy-dust-28         1.4.3     online 4d   -         region=ewr1,provider=packet
```
It helps a lot to discern between stacks when you need to do troubleshooting and etc, in a case like this:
```
NAME                                 STACK                                    SERVICES   STATE               EXPOSED PORTS
⊝ 15dce04b995f05842a99a2c22011336b   15dce04b995f05842a99a2c22011336b:0.0.1   1          running
⊝ 1fe3c510e8b6ef2d03a8c104fe2a6eb0   1fe3c510e8b6ef2d03a8c104fe2a6eb0:0.0.1   1          running
⊝ 75c6383b92e3a7591b9e2a50b78cf272   75c6383b92e3a7591b9e2a50b78cf272:0.0.1   1          running
```
I'm totally cool with relying on `meta` if it is rendered the output of `kontena stack ls`, since it is very helpful for troubleshooting and monitoring of customers stacks in the cluster.

Hope that makes sense.

server/app/models/stack_revision.rb
```diff
@@ -12,10 +12,12 @@ class StackRevision
   field :expose, type: String
   field :services, type: Array
   field :volumes, type: Array
+  field :labels, type: Array, default: []
 
   belongs_to :stack
 
   index({ stack_id: 1 })
+  index({ labels: 1 })
```
ptsurbeleu: Not very sure about indexing `labels`... I used `host_node` as a primer to derive from and my assumptions might be incompetent at best. 😃

kke:
Yeah that makes sense, I think meta could be the solution there, the PR for that is #3281

ptsurbeleu:
I looked at #3281 and have a question, can we enhance it with support in Kontena CLI? As it is `meta` is in the stack definition but when you run `kontena stack ls` you won't see any of `meta's` tags... 😢

kke:
> can we enhance it with support in Kontena CLI? As it is meta is in the stack definition but when you run kontena stack ls you won't see any of meta's tags.

Yes, once #3281 is in, the master starts to include the meta in stack listings and CLI can then display it.

But thinking about it now, I'm not sure how practical that is for adding the label functionality. Perhaps this PR is still relevant.

ptsurbeleu:
Okie, I will proceed with the work on `labels` then and we can port/cherry-pick parts as we see them fit together.

ptsurbeleu:
Here is a screenshot to clarify what the end-user experience in CLI looks like when stacks are allowed to defined their own labels:

image

kke:
I'm uncertain if the labels should be in `Stack` instead of `StackRevision`.

Also I'm not sure if they should be present in the Stack YAML.

Adding labels probably should not create a new revision? Labels should probably persist between revisions?

```console
$ kontena stack install -n foo -l fqdn=example.com redis.yml
$ kontena stack labels add foo env=staging
$ kontena stack upgrade foo redis2.yml
$ kontena stack ls -l env=staging
```

kke:
Ref: #2928

ptsurbeleu:
> I'm uncertain if the labels should be in Stack instead of StackRevision.

@kke If you ask me, I also prefer `labels` to be in `Stack` rather than in `StackRevision` for the very same reasons you mentioned. However `StackRevision` is a bigger hub of fields and that drove me to plug the change and yet another field there - I'm totally up to move it to `Stack`.

> Also I'm not sure if they should be present in the Stack YAML.

Having labels in `Stack` YAML, make `labels` very durable and portable which I think makes `Stack` YAML concept even more competitive and compelling. Here are the use cases:

A) YAML checked into a Git repository, helps to find out how `labels` evolved over time;

B) In case of disaster recovery, having `labels` in YAML saves Kontena operators from an infamous mistake - when stack YAML here and `labels` of the stack there in a Shell script that you should not forget to run to get back to the expected state;

C) `labels` in YAML give a way to accomplish 2 things in 1 API request, create a stack and assign `labels`;

> Adding labels probably should not create a new revision? Labels should probably persist between revisions?

Yup, I agree that is a good idea. So, are we good moving `labels` up to the `Stack`?

kke:
> labels in YAML give a way to accomplish 2 things in 1 API request

`kontena stack install -l label=xyz foo.yml`

I think the new `meta[tags][]` could be used for the "static labels" that come from YAML.

I would think of the stack labels as a way to differentiate between multiple installations of the same stack YAML, such as:

```
$ kontena stack install -l client=client1 -n client1app
$ kontena stack install -l client=client2 -n client2app
```

And perhaps in the future do something like:

```yaml
variables:
  api_stack_url:
    type: string
    from:
      installed_stack:
        search:
          stack: user/fooapi
          labels:
            - client=${client_name}
        return:
           label: fqdn

services:
  foofoo:
    environment:
      - API_URL=https://${api_stack_url}/api/v2
```

ptsurbeleu:
> I would think of the stack labels as a way to differentiate between multiple installations of the same stack YAML

Exactly the point of having labels, just to be able visually discern & filter by in `kontena stack ls -l client=client1` output (_BTW, I love filtering idea a lot_)! And it is so convenient to have these labels in YAML file - to me that is kind of a deal-breaker, I suffer from missing it a lot and hence my motivation to contribute.

> kontena stack install -l label=xyz foo.yml

That definitely works for `2 in 1` point made before, however still doesn't work for disaster recovery case - when you need to restore the stack back to the expected state (including `labels`). Also it is neither durable nor portable since `labels` part of `Stack` definition couldn't live with it which seems to be a major drawback - usually parts are supposed to be stored alongside with the definition to avoid being accidentally lost or forgotten (happens even to the best & most experienced of us).

> I think the new meta[tags][] could be used for the "static labels" that come from YAML.

It could be but that might be confusing to use properly. As far as I can tell, it is challenging to adequately document how `meta[…]` can be interpolated into `static labels` and then used for display purposes in Kontena CLI. It would be great if `meta` could be used for one purpose and `labels` for the other - unless there is a desire to merge two together.

> And perhaps in the future do something like

Ouch, my visual scanner blew up with `unexpected syntax exception` :-) Are you okay to postpone that line of thought until we settle and commit on the baseline between `meta` and `labels`? My understanding on the subject is not deep enough to make meaningful contributions to the conversation of the future concept and I need to enlighten myself on variables and Liquid support in `Stack` YAML quite a bit.

Thanks!

kke:
Yeah I think the labels should be in the Stack instead of StackRevision. I think their purpose is mostly to be an extension of the stack instance name.

Having them in the YAML is ok.

Maybe @jakolehm @jnummelin or @SpComb have some thoughts.

cli/lib/kontena/cli/stacks/yaml/reader.rb
```diff
@@ -206,6 +206,7 @@ def execute(service_name = nil, name: loader.stack_name.stack, parent_name: nil,
           result['stack']         = raw_yaml['stack']
           result['version']       = loader.stack_name.version || '0.0.1'
           result['name']          = name
+          result['labels']        = raw_yaml['labels'] || []
```
kke: This needs to be:

`fully_interpolated_yaml['labels'] || []` 

The `raw_yaml` is from before any variable/liquid interpolation has taken place. With `fully_interpolated_yaml` it's possible to insert variable values into labels.

For example:
```yaml
variables:
  customer:
    type: string
    from:
      prompt: Enter customer name
labels:
  - customer=${customer}
```

ptsurbeleu: @kke thanks for the catch - that's a good one!

ptsurbeleu:
@kke thanks for the feedback! I just pushed a change that moves `labels` to `Stack` level with test coverage around it.

Should we also put together `kontena stack label add | remove | ls` and `kontena stack ls -l 'foo=bar'` before merging changes to upstream or it is okay to implement these in 2 separate pull requests?

Let me know what option closely follows Kontena's development workflow and many thanks for your help to keep the branch up-to-date! 😄

kke:
The stack list API endpoint does not have any existing search functionality, so for now I think the best way to implement the label filtering would be to have something like this in the `stacks/list_command.rb`

```ruby

option ['--label', '-l'], '[LABEL]', "Filter by labels", multivalued: true

def execute
...
      print_table(stacks) do |row|
        next if quiet?

        unless label_list.empty?
          next unless (label_list - row['labels']).empty?
        end

        row['name'] = health_icon(stack_health(row)) + " " + tree_icon(row) + " " + row['name']
        row['stack'] = "#{row['stack']}:#{row['version']}"
        row['services_count'] = row['services'].size
        row['ports'] = stack_ports(row).join(',')
        row['state'] = pastel.send(state_color(row['state']), row['state'])
        row['labels'] = stack_labels(row)
      end
...
```

The `add|remove|ls` can be in the same PR and I think should be there when the rest of it is merged.

Also note the `stack inspect` in the master branch.

ptsurbeleu:
@kke, @SpComb, @jakolehm and @jnummelin need help and critical eye to review this fairly innocent snippet that actually doesn't work out of the box. 

Here are details about the screenshot:
- Top section (left-hand block) with `add_command.rb` and `execute` method is the code that tries to update `labels` for an existing stack;
- Top section (right-hand block with `stack-with-labels.yml`) is the stack definition that was originally deployed;
- Bottom part is the dump of `original_data` variable and response from my Master that is not happy with the stack's json I'm sending;

image

ptsurbeleu:
As far as I can tell after reviewing validation code that Master executes against these two troublesome fields `hooks` and `stop_grace_period` there is something sneaky going on... `hooks` validation expects to receive a `Hash`, however `hooks` are moving around as an `Array` of hashes...

I also spent some time in `stacks/update_command.rb` and found interesting nuisance, `hooks` received from Master as an `Array` but travels back as `Hash`...

Also `stop_grace_period` is expected to be an interval string, eq. `10s` or `1h30m` in stack's definition but it is exposed via API as an `integer`...

The least I can infer from my 2-day adventure full of findings and experiments - these parts of API for some reason don't behave as advertised. Sadness...

kke:
Hmm yes, there is no "stack update" endpoint / mutation that would not create a revision.

You could create a new endpoint, such as:

```ruby
        r.put do # this is in app/routes/stacks_api.rb
          # PUT /v1/stacks/:grid/:name
          r.is do
            data = parse_json_body
            update_stack(@stack, data)
          end
          r.on 'labels' do
            data = parse_json_body
            task = Stacks::UpdateLabels(stack: @stack, labels: data['labels'])
          end
        end
```

and create a mutation called `Stacks::UpdateLabels` that would just update labels.

the `stack labels rm` can then just do an update with unwanted labels removed from existing and `add` with new labels added.

But this design is something that @jakolehm may have something to say about.

But the situation is: with the current API endpoints, it is not possible to update the stack without creating a revision.

Tinkering with the differing JSON request/response formats is not worth your time, that's for sure, it's not going to work without an api change unless we want to create new revisions (meaning that you would have to deploy the stack before label changes take effect, which in my opinion is not what the labels would be ideal for)

SpComb:
> Yeah I think the labels should be in the Stack instead of StackRevision. I think their purpose is mostly to be an extension of the stack instance name.

Yes. If the stack labels were to get deployed as service container labels, then the labels would need to go into the `StackRevision` services. But if the labels are only used as server/API metadata without affecting the actual services/containers, then keeping them on the `Stack` is the right choice IMO.

> The least I can infer from my 2-day adventure full of findings and experiments - these parts of API for some reason don't behave as advertised. Sadness...

The problem you're running into is that the JSON data model used for the stack/services `GET` and `POST`/`PUT` APIs are not exactly the same. You can't `GET` a stack/services JSON from the server and then `PUT` the same stack/services JSON back to the server, the fields are subtly different ([`GridServices::Common#common_validations`](https://github.com/kontena/kontena/blob/462a5653623f16a521d970948a0f21cfdc2d3a7b/server/app/mutations/grid_services/common.rb#L224) vs [`v1/grid_services/_grid_service.json.jbuilder`](https://github.com/kontena/kontena/blob/462a5653623f16a521d970948a0f21cfdc2d3a7b/server/app/views/v1/grid_services/_grid_service.json.jbuilder) - the stack YAML is also again slightly different).

> Hmm yes, there is no "stack update" endpoint / mutation that would not create a revision.

I think the significant distinction here is that the `PUT /v1/stacks/...` API actually behaves as a real REST API `PUT` should: it replaces the entire stack, creating a new revision. If you were to leave out something like the `"services": [...]`, that would imply removing all of the existing stack services. That's not the API that you want for modifying the stack labels in-place.

If you look at how the other APIs like `PUT /v1/services/...` or `PUT /v1/nodes/...` behave, they're not actually complete replacement operations... all of the parameters are optional, so semantically, they're actually more like `PATCH` operations.

So the solution would be to either preserve the existing API semantics and fix the `PUT /v1/stacks/...` => `Stacks::Update` to make the `services` etc optional, or start fixing the API semantics by introducing a new `PATCH /v1/stacks/... {"labels": [ ... ] }` API where all of the fields are optional. The later is probably easier assuming that it only needs to supports `labels`, whereas changing the `Stacks::Update` to make the `StackRevision` fields optional would be harder.

ptsurbeleu:
@kke, @SpComb thanks folks for the prompt response & ideas! Have been extremely busy at work for the last 2 days and just got the chance to get back working on PR.

> You could create a new endpoint, such as:...

That sounds like a great plan! 👍 Somehow I was fixated on the constraints of the existing `PUT` endpoint and completely overlooked that option. 🔨 

> ..., or start fixing the API semantics by introducing a new PATCH /v1/stacks/... {"labels": [ ... ] } API where all of the fields are optional. The later is probably easier assuming that it only needs to supports labels, whereas changing the Stacks::Update to make the StackRevision fields optional would be harder.

Love this option better as it offers more granular & isolated change surface and doesn't change existing business logic on stack revisions.

Gone working on updates to the PR. 😃

ptsurbeleu:
@kke and @SpComb, support for `kontena stack label ls/rm/add` is finally here. Let me know folks what you think and if there is anything that needs to be done differently.

Thanks! 😄

cli/lib/kontena/cli/stacks/label_command.rb
```diff
@@ -0,0 +1,10 @@
+module Kontena::Cli::Stacks
+  class LabelCommand < Kontena::Command
+    subcommand "add", "Add label to stack", load_subcommand('stacks/labels/add_command')
+    subcommand ["remove", "rm"], "Remove label from stack", load_subcommand('stacks/labels/remove_command')
+    subcommand ["list", "ls"], "List node labels", load_subcommand('stacks/labels/list_command')
```
kke: `node` => `stack`

cli/lib/kontena/cli/stacks/labels/add_command.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    parameter "NAME", "Stack name"
+    parameter "LABEL ...", "Labels"
+
+    requires_current_master
+    requires_current_master_token
+
+    def execute
+      original_labels = fetch_master_data(name)
+      update_stack(name, { 'labels' => (original_labels + label_list).uniq })
```
kke: Can be `'labels' => original_labels | label_list`

cli/lib/kontena/cli/stacks/labels/list_command.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    parameter "NAME", "Stack name"
+
+    requires_current_master
+    requires_current_master_token
+
+    def execute
+      stack = client.get("stacks/#{current_grid}/#{name}")
+      # safeguard from nil, since 'labels' field is optional
+      puts Array(stack['labels'] || []).join("\n")
```
kke: See the table generator thing on other `list_command`s

cli/lib/kontena/cli/stacks/labels/remove_command.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def update_stack(name, data)
+      client.patch(stack_url(name), data)
+    end
+
+    def stack_url(name)
+      "stacks/#{current_grid}/#{name}"
+    end
+
+    def fetch_master_data(stackname)
```
kke: Maybe create a `stacks/labels/common.rb` with these duplicate methods

cli/lib/kontena/cli/stacks/list_command.rb
```diff
@@ -37,27 +37,30 @@ def get_stacks
...
 
-    print_table(stacks) do |row|
+      print_table(stacks) do |row|
         next if quiet?
         row['name'] = health_icon(stack_health(row)) + " " + tree_icon(row) + " " + row['name']
         row['stack'] = "#{row['stack']}:#{row['version']}"
         row['services_count'] = row['services'].size
         row['ports'] = stack_ports(row).join(',')
         row['state'] = pastel.send(state_color(row['state']), row['state'])
+        row['labels'] = stack_labels(row)
```
kke: I think this could use some kind of max length thing?

Something like:

```ruby
labels = stack_labels(row)
labels = "#{labels[0..17]}..." if labels.length > 20
row['labels'] = labels.length > 20 ? "#{labels[0..17]}..." : labels
```

cli/spec/fixtures/kontena_v3.yml
```diff
@@ -1,5 +1,6 @@
 stack: user/stackname
 version: 0.1.1
+labels: ['fqdn=wordpress.com']
```
kke: I'd prefer the multi-line array representation:

```yaml
labels:
  - fqdn=wordpress.com
```

cli/spec/fixtures/stack-with-labels.yml
```diff
@@ -0,0 +1,11 @@
+stack: user/labels
+version: 0.1.1
+labels: ['fqdn=about.wp.com', 'about.wp.com']
```
kke: Same here

cli/spec/kontena/cli/stacks/labels/add_command_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'kontena/cli/stacks/label_command'
+require 'kontena/cli/stacks/labels/add_command'
+
+describe Kontena::Cli::Stacks::Labels::AddCommand do
+  include ClientHelpers
+  include OutputHelpers
+
+  # stack name
+  let :name do
```
kke: Other specs tend to have parenthesis and the oneliner syntax for these: 

```ruby
let(:name) { 'test-stack' }
```

cli/spec/kontena/cli/stacks/labels/add_command_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  before do
+    # stub Kontena Master call
+    allow(client).to receive(:get).with(endpoint).and_return(stack)
+  end
+
+  context "for a stack without labels" do
+    # original stack definition
+    let(:stack) { scratch }
+    it "adds the labels" do
+      # arrange
```
kke: No need for this level of comments

cli/spec/kontena/cli/stacks/labels/remove_command_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    'test-stack'
+  end
+
+  # shared endpoint - the same for all cases
+  let :endpoint do
+    "stacks/test-grid/#{name}"
+  end
+
+  # scratch stack definition - use it as is, or modify to accomodate specific test case
+  let :scratch do
```
kke: I think it's cleaner if the base form is `:stack` and then the modified versions in different contexts are something like `let(:stack_with_labels) { stack.merge('labels' => %w(foo=bar xyz=zyx)) }`

cli/spec/kontena/cli/stacks/list_command_spec.rb
```diff
@@ -60,4 +61,77 @@
       end
     end
   end
+
+  context "command output layout" do
+    before do
+      allow(client).to receive(:get).with('grids/test-grid/stacks').and_return({
```
kke: No need for  `({})`, can be just `('stacks' => [])`

cli/spec/kontena/cli/stacks/list_command_spec.rb
```diff
@@ -60,4 +61,77 @@
...
+        'stacks' => []
+      })
+    end
+
+    it "renders NAME as 1st column" do
```
kke: This series of specs is maybe a bit overly strict. If there are any changes to the fields, this whole set will blow up.

There's also a helper for testing the table outputs, It works something like:

```ruby
expect{subject.run([])}.to output_table([
  %w(xyz abc dfg)
]).with_header(
  %w(NAME FOO LABELS)
)
```

(see `table_generator_spec.rb`)

cli/spec/kontena/cli/stacks/list_command_spec.rb
```diff
@@ -60,4 +61,77 @@
...
+          'variables' => { 'foo' => 'bar' },
+          'children' => [],
+          'labels' => ['fqdn=about.me']
+        }]
+      })
+    end
+
+    it "outputs stack labels" do
+      expect{subject.run([])}.to output_table [
+        ["\e[2m\u229D\e[0m", 'stack-a', 'foo/stack-a:', '0', 'fqdn=about.me'],
```
kke: Maybe these icon sequences could be replaced with something.

I think the output helper understands rspec matchers such as:

```ruby
[anything, 'stack-a', .. ]
[an_instance_of(String), 'stack-a', .. ]
```

server/app/mutations/stacks/patch.rb
```diff
@@ -0,0 +1,31 @@
...
+module Stacks
+  class Patch < Mutations::Command
+    include Common
+
+    required do
+      model :stack_instance, class: Stack
+    end
+
+    optional do
+      model :grid, class: Grid
```
SpComb: Doesn't need a `grid` parameter.

(not for this PR, but this should also get cleaned up from the `Stacks::Update` as well - it's only used for the `validate_volumes`, and should get passed as a parameter instead).

server/app/mutations/stacks/create.rb
```diff
@@ -52,9 +52,15 @@ def validate_services
 
     def execute
       attributes = self.inputs.clone
-      %w( parent parent_name ).each { |k| attributes.delete(k.to_sym)}
+      # filter out some attributes and ensure StackRevision receives only
+      # ones that are supported to avoid Mongoid::Errors::UnknownAttribute error
+      %w( parent parent_name labels ).each { |k| attributes.delete(k.to_sym)}
```
kke: This by the way could be `%i( parent parent_name labels ).each { |k| attributes.delete(k) }` and avoid the `.to_sym`.

server/app/mutations/stacks/patch.rb
```diff
@@ -0,0 +1,31 @@
...
+      array :labels
```
SpComb: For validation, use something like this (not 100% sure about the exact regexp, should see what docker specs for labels)

```ruby
array :labels do
  string matches: /\A[a-z0-9.-]+=/
end
```

server/app/mutations/stacks/patch.rb
```diff
@@ -0,0 +1,31 @@
...
+        return
+      end
+    end
+
+    def execute
+      # update labels only if assigned, empty array removes existing labels
+      stack_instance.labels = self.labels if self.labels
+      stack_instance.save if stack_instance.changed?
+      # reload changes (if any)
+      self.stack_instance.reload
```
SpComb: Shouldn't be needed, the model object will already be updated if using `stack_instance.save`.

server/app/routes/v1/stacks_api.rb
```diff
@@ -15,6 +15,22 @@ class StacksApi < Roda
...
+      # @param [Stack] stack
+      # @param [Hash] data
+      def patch_stack(stack, data)
+        data[:stack_instance] = stack
+        outcome = Stacks::Patch.run(data)
+
+        if outcome.success?
+          audit_event(request, @grid, @stack, 'patch')
+          response.status = 200
+          {}
```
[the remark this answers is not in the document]
SpComb: First intuition would be to return HTTP 200 + `render('stacks/show')` to match the existing `PUT` behavior in other APIs.

server/spec/api/v1/stacks_spec.rb
```diff
@@ -214,6 +222,22 @@
     end
   end
 
+  describe 'PATCH /:name' do
+    it 'saves non-empty stack labels' do
+      labels = ['foo=bar', 'timezone=PDT']
+      patch "/v1/stacks/#{stack.to_path}", {labels: labels}.to_json, request_headers
+      expect(response.status).to eq(200)
+      expect(stack.reload.labels).to eq(labels)
```
SpComb: I recommend writing these specs as

```ruby
expect{
  patch ...
}.to change{stack.reload.lablels}.from([...]).to([...])
```

server/spec/api/v1/stacks_spec.rb
```diff
@@ -234,6 +259,7 @@
       expect {
         put "/v1/stacks/#{stack.to_path}", data.to_json, request_headers
         expect(response.status).to eq(200)
+        expect(stack.reload.labels).to eq(['foo=1', 'bar=2'])
```
SpComb:
```ruby
expect {

}.to change{...}.by(...).and change{stack.reload.labels}.from(...).to(...)
```

test/.kontena_client.json
```diff
@@ -0,0 +1,24 @@
...
+      "token": "50f480e2c6a62dc54ba4aeb219c447329e5634a5e9fb9582e994d94ae113ac16",
+      "token_expires_at": 0,
+      "refresh_token": null
+    }
+  ],
+  "accounts": [
+    {
+      "name": "kontena",
+      "username": "pabloduo",
+      "token": "0bd4d4155c364de79ef264089a3e8d0f",
```
SpComb: oops?

server/app/mutations/stacks/patch.rb
```diff
@@ -0,0 +1,31 @@
...
+    required do
+      model :stack_instance, class: Stack
+    end
+
+    optional do
+      model :grid, class: Grid
+      array :labels
+    end
+
+    def validate
```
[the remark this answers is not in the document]
kke: the `array` clause already validates that it is an array.

Looks like this can be done:

```ruby
array :labels do
  string
end
```

it works a bit strangely though:

```
task.new(labels: [1,2,3])
=>
:labels => ["1", "2", "3"]
```

but catches this:

```
task.new(labels: [nil])
=> 
{"labels" => ["Labels[0] can't be nil"]}
```

server/app/routes/v1/stacks_api.rb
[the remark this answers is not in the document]
kke: I think there are several variations in the API currently, both are fine by me, but generally I suppose it would be a good practice to return the object.

server/spec/mutations/stacks/create_spec.rb
```diff
@@ -101,6 +101,22 @@
...
+        stack: 'foo/bar',
+        version: '0.1.0',
+        registry: 'file://',
+        labels: ['fqdn=about.me'],
+        source: '...',
+        variables: {foo: 'bar'},
+        services: [{name: 'redis', image: 'redis:2.8', stateful: true }]
+      ).run
+      expect(outcome).to be_success
+      expect(outcome.result.labels).to eq(['fqdn=about.me'])
```
kke: maybe replace these test fqdn:s with something like `example.com` instead of some actual sites.

cli/spec/fixtures/kontena_v3.yml
kke: maybe replace these test fqdn:s with something like `example.com` instead of some actual sites.

test/.kontena_client.json
ptsurbeleu: @SpComb yes... I should exercise extreme caution and not push changes to this file 😃 thanks for the catch - I've already revoked this token and confirmed both `token` and `refresh_token` are no longer authorized to access my Kontena Cloud account.

server/app/mutations/stacks/patch.rb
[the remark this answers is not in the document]
ptsurbeleu: @kke thanks for the feedback - adjusted the code for that + test coverage

cli/spec/fixtures/kontena_v3.yml
ptsurbeleu: yup, will replace with something test-alike...

cli/lib/kontena/cli/stacks/labels/list_command.rb
ptsurbeleu: @kke not sure if that is comparable to things like `kontena stack ls` or `kontena service ls`... I took the inspiration from `kontena node labels ls <node name>`:
```ruby
def execute
  node = client.get("nodes/#{current_grid}/#{self.node}")
  puts Array(node['labels']).join("\n")
end
```
If you feel the output should be table-alike, could you suggest an example?

cli/lib/kontena/cli/stacks/label_command.rb
ptsurbeleu: ✅ DONE

cli/lib/kontena/cli/stacks/labels/add_command.rb
ptsurbeleu: ✅ RESOLVED

cli/lib/kontena/cli/stacks/labels/remove_command.rb
ptsurbeleu: ✅ RESOLVED

server/app/mutations/stacks/patch.rb
ptsurbeleu: ✅ RESOLVED

server/app/mutations/stacks/patch.rb
ptsurbeleu: ✅ RESOLVED

server/spec/api/v1/stacks_spec.rb
ptsurbeleu: ✅ RESOLVED

server/spec/api/v1/stacks_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/lib/kontena/cli/stacks/labels/list_command.rb
kke: Hmm, indeed. I need to look into this at office hours, I don't know why `node labels ls` is not a regular list command. Maybe @jakolehm has an opinion about this?

cli/lib/kontena/cli/stacks/list_command.rb
kke: And maybe only do that if `$stdin.tty?` so that something like `label ls | grep xyz` will not get cut.

cli/spec/fixtures/kontena_v3.yml
kke: example.com is reserved for things like that by whoever runs this internet thing.

cli/lib/kontena/cli/stacks/labels/common.rb
```diff
@@ -0,0 +1,19 @@
+module Kontena::Cli::Stacks::Labels
+  module Common
+    private
```
kke: 👍

server/app/mutations/stacks/patch.rb
ptsurbeleu: @SpComb it seems Docker makes it ridiculously easy to work with labels, here is the snippet from their CLI:
```go
// ValidateLabel validates that the specified string is a valid label, and returns it.
// Labels are in the form on key=value.
func ValidateLabel(val string) (string, error) {
	if strings.Count(val, "=") < 1 {
		return "", fmt.Errorf("bad attribute format: %s", val)
	}
	return val, nil
}
```

cli/lib/kontena/cli/stacks/list_command.rb
ptsurbeleu: @kke sounds like a good idea, will adjust for that case as well.

cli/spec/kontena/cli/stacks/labels/add_command_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/spec/kontena/cli/stacks/labels/add_command_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/spec/fixtures/stack-with-labels.yml
ptsurbeleu: ✅ RESOLVED

cli/spec/fixtures/kontena_v3.yml
ptsurbeleu: ✅ RESOLVED

cli/lib/kontena/cli/stacks/list_command.rb
ptsurbeleu: ✅ RESOLVED

server/spec/mutations/stacks/create_spec.rb
ptsurbeleu: ✅ RESOLVED

server/app/routes/v1/stacks_api.rb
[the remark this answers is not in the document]
ptsurbeleu: ✅ RESOLVED

cli/spec/kontena/cli/stacks/labels/remove_command_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/spec/kontena/cli/stacks/list_command_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/spec/kontena/cli/stacks/list_command_spec.rb
ptsurbeleu: ✅ RESOLVED

cli/spec/kontena/cli/stacks/list_command_spec.rb
ptsurbeleu: ✅ RESOLVED

server/app/mutations/stacks/create.rb
ptsurbeleu: ✅ RESOLVED

server/app/mutations/stacks/patch.rb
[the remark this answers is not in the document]
ptsurbeleu: ✅ RESOLVED

ptsurbeleu:
@kke and @SpComb, folks I'm hitting an unexpected difference in behavior between `dev env` and `ci env` for this CI run: https://travis-ci.org/kontena/kontena/builds/353689272. Any suggestions how to cater it so that the code works seamlessly in `dev` and `ci`?

kke:
It's same problem as in #3331

The screen width on your local dev is larger than what travis uses and the table thing goes into single-column mode because it thinks it can't fit stuff to the terminal width.

Will investigate.

ptsurbeleu:
@kke, seems I've found the cure to the problem - the default setting of viewport in Travis CI worker (`trusty`) doesn't accommodate Kontena's CLI output. Adjusting the settings via `stty columns 157` does return the viewport into reasonable dimensions and makes `cli` tests __PASS__.

PS. `columns=157` is the current viewport setting on my `dev env` therefore the __"magic"__ number.

cli/lib/kontena/cli/stacks/list_command.rb
```diff
@@ -91,6 +94,17 @@ def tree_icon(row)
...
 
+    # Converts an array of stack labels into a comma-separated string
+    # or '-' when labels field is not defined.
+    #
+    # @param [Hash] stack
+    # @return [<String>]
+    def stack_labels(stack)
+      labels = (stack['labels'] || ['-']).join(',')
+      # trim labels to fit viewport when exceed 20 chars
+      labels.length > 20 && $stdin.tty? ? "#{labels[0..17]}..." : labels
```
kke: Wrap this inside `if $stdin.tty?` (so that something like `stack ls | grep foo=bar` won't fail because of the cutting.

build/travis/before_install.sh
```diff
@@ -10,4 +10,6 @@ if [ "$TEST_DIR" = "server" ]; then
 fi
 
 gem update --system
-gem install bundler-audit --no-ri --no-rdoc
\ No newline at end of file
+gem install bundler-audit --no-ri --no-rdoc
+stty cols 157
```
SpComb: #3331 will fake the `TTY::Screen.width` for all specs: https://github.com/kontena/kontena/pull/3331/files#diff-cf82573a2ae83c839d397f095fdc8b33

kke: Yep that's better than messing with the travis shell script.

cli/lib/kontena/cli/stacks/list_command.rb
ptsurbeleu: @kke, it seems we should be testing the output device `$stdout` instead of input device `$stdin`. Also I don't get why it needs to wrapped inside of `if $stdin.tty?`?

As it is, the logic of `stack_labels` helper:
 - truncate + ellipsis when `lengthy && tty`;
 - don't truncate when `short && tty`;
 - don't truncate when `short && non-tty`;
 - don't truncate when `lengthy && non-tty`;

Is anything else missing? 😄 

Here is a screenshot of the behavior captured as of now (just increased size up to `43`, since `20` is way too short even for a single reasonably-sized label):

image

build/travis/before_install.sh
ptsurbeleu: @kke and @SpComb, agree - stubbing `TTY::Screen` is way more portable than `before_install.sh`. Therefore I'm reverting my changes and back-porting yours to unblock my CI runs.

ptsurbeleu: ✅ RESOLVED

server/app/mutations/stacks/patch.rb
ptsurbeleu: ✅ RESOLVED

ptsurbeleu:
@kke  and @SpComb, got a few more changes and validations in - let me know what you think.

cli/lib/kontena/cli/stacks/list_command.rb
kke: Yes, right, stdout. 

Also missed the use of $stdin.tty? in the existing version for some reason.

server/spec/mutations/stacks/create_spec.rb
```diff
@@ -117,6 +117,22 @@
...
     end
 
+    it 'rejects invalid stack labels' do
+      outcome = described_class.new(
+        grid: grid,
+        name: 'stack',
+        stack: 'foo/bar',
+        version: '0.1.0',
+        registry: 'file://',
+        labels: ['noop', 'loop'],
```
kke: I suppose these should fail if the label validator checks for `=`

ptsurbeleu:
@kke, thanks for helping out with the changes and approving the pull request! I'm still wondering why would that part of `e2e` test is failing so often...?

Is the test too fragile or pull request changes are too invasive? 😄

image

kke:
I don't think that spec has anything to do with this PR. The e2e suite and also some agent tests are a bit too timing critical and fail randomly.

ptsurbeleu:
@SpComb and @kke, folks let me know if we need to make any other adjustments before this PR can be merged. 

If we are good to merge it, then I can switch the context and work to contribute updates to the  documentation around `kontena stack label` and `kontena stack ls` commands.

Thanks!

kke:
I think it's quite kosher

ptsurbeleu:
@kke and @SpComb, very excited to see this PR being merged! Do you have any `guesstimates` when that might happen? 😃

server/app/models/stack.rb
```diff
@@ -16,6 +17,7 @@ class Stack
   index({ grid_id: 1 })
   index({ name: 1 })
   index({ parent_name: 1 })
+  index({ labels: 1 })
```
jakolehm: This requires a migration (see `db/migrations/`).

kke: basically just `def up; Stacky.create_indexes; end`

ptsurbeleu: thanks for the hint - will update migrations to reflect the change.

ptsurbeleu: ✅ RESOLVED

kke:
The failing specs are unrelated and are fixed in #3339

ptsurbeleu:
@kke and @jakolehm, folks - when are you planning to start merging and preparing `1.6.0` changes for release?

test/docker-compose.yml
kke: This is still here btw

kke:
Shouldn't be too long

test/.kontena_client.json
```diff
@@ -0,0 +1,16 @@
+{
```
kke: I don't think this belongs here

ptsurbeleu: @kke, yup - I will remove that one, thanks for the catch! 😄

PR #3219, opened 2018-01-16
Switch to Stack registry V2 API and add support the new stack YAML metadata fields

kke:
Makes the CLI communicate to the stack registry via the new V2 API which offers some benefits, such as speed improvement, better searches, JSONAPI compliance, gzip response compression and private stacks for organizations.

Adds optional metadata fields to Stack YAML:

```yaml
stack: user/stack
version: 1.2.3
services:
  redis:
    image: redis:latest
meta: # (optional)
  description: Short description / summary # (optional)
  icon: A URL to an SVG or PNG image to be used as an icon # (optional).
  home: The URL of this project's home page # (optional)
  source: The URL of this project's source code # (optional)
  readme: The README.md for this stack file (URL or markdown) # (optional) 
  issues: The URL of this project's issue tracking # (optional)
  required_kontena_version: ">= 1.5.0" # (optional)
  tags: # (optional)
    - A list of tags 
  app_version: # (optional)
    mongo: "3.2" 
    app: 1.2.3
  maintainers: # (optional)
    - name: name 
      email: 
      url: 
```

If the stack metadata has `required_kontena_version` defined, the `kontena stack install` and `kontena stack upgrade` commands check master version before the operation and prompt for confirmation unless `--force` is given.

PR #3281, opened 2018-02-15
Send and return stack metadata to master

kke:
Add `metadata` to the stack-json on master

Adds to #3219

### DIFF
--- a/server/app/models/stack_revision.rb
+++ b/server/app/models/stack_revision.rb
@@ -6,13 +6,13 @@ class StackRevision
   field :stack_name, type: String
   field :registry, type: String
   field :source, type: String
-  field :variables, type: Hash
+  field :variables, type: Hash, default: {}
   field :version, type: String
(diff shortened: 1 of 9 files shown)


server/app/models/stack_revision.rb
```diff
@@ -12,6 +12,7 @@ class StackRevision
   field :expose, type: String
   field :services, type: Array
   field :volumes, type: Array
+  field :metadata, type: Hash
```
jakolehm: Maybe add `default: {}` .

kke: Thought about it but variables didn't have one either.

jakolehm: I think this should be fixed. Otherwise api will send `null` or `{ .. }` and it's quite ugly to handle on client side.

kke:
Merging to parent PR

PR #3285, opened 2018-02-16
Try speeding up e2e by caching docker

kke:
Trying to speed up e2e builds by caching the docker images/layers:

https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-248915326
https://stackoverflow.com/a/35374332/681520
https://gist.github.com/hc2p/9e284cee3d585eefbc59454e44cc247a

Also fixes this problem (unless this PR makes any progress, that should be fixed separately)

```yaml
cache: bundler
cache:
  directories:
    - ...
```

=> bundler will not be cached because the second occurance of `cache:` overwrites the value.


kke:
First run 19min28sec

kke:
The attempt so far has been unsuccesful, but perhaps worth it to extract the waiting stuff

kke:
And the bundler cache fix.

PR #3297, opened 2018-02-19
[WiP] Use stack deploys to have the CLI kontena stack stop/remove commands wait for the service containers to stop/terminate

### LINKED ISSUE #3030: stack install immediately after stack rm on a slow node will cause ConflictError

SpComb:
Requires #3277 for service deploys in the `stopped` state
Requires #3278 for service deploys in the `terminated` state
Fixes #3030 Docker container name conflicts when installing a new stack immediately after removing a similarly named stack

Changes the `POST /v1/stacks/.../stop` => `Stacks::Stop` API to return a `StackDeploy` tracking all of the `GridServices::Stop` => `GridServiceDeploy`s for the stack services. Instead of creating a new `StackStopWorker`, the `StackDeploy` object is changed to passively determine its state from the `GridServiceDeploys`, removing the need for the stack deploy workers to update the deploy state. The CLI `kontena stack stop` waits for the stack deploy to complete.

Adds a new `POST /v1/stacks/.../terminate` => Stacks::Terminate` API to return a `StackDeploy` tracking all of the `GridServices::Terminate` => `GridServiceDeploy`s for the stack services. The  `DELETE /v1/stacks/...` => `Stacks::Delete` no longer uses an async worker to stop the stack services, it just destroys them from mongo. The CLI `kontena stack rm` first terminates the stack and waits for the deploy to complete before deleting the stack.


SpComb:
The specs and e2e specs are now green for the this entire series of PRs, so this should be feature-complete.

cli/lib/kontena/cli/stacks/remove_command.rb
```diff
@@ -1,16 +1,19 @@
...
     include Kontena::Cli::GridOptions
     include Common
+    include StacksHelper # XXX: must be after Common, because that includes ServiceHelper, which has similarly named methods >_>
 
     banner "Removes a stack in a grid on Kontena Master"
 
     parameter "NAME ...", "Stack name", attribute_name: :names
     option "--force", :flag, "Force remove", default: false, attribute_name: :forced
     option "--keep-dependencies", :flag, "Do not remove dependencies"
+    option '--[no-]wait', :flag, 'Do not wait for service deployment', default: true
```
kke: Misleading description.

`--wait` is true
`--no-wait` is false

cli/lib/kontena/cli/stacks/stop_command.rb
```diff
@@ -1,23 +1,28 @@
...
 module Kontena::Cli::Stacks
   class StopCommand < Kontena::Command
     include Kontena::Cli::Common
     include Kontena::Cli::GridOptions
     include Common
+    include StacksHelper # XXX: must be after Common, because that includes ServiceHelper, which has similarly named methods >_>
 
     banner "Stops all services of a stack"
 
+    option '--[no-]wait', :flag, 'Do not wait for service deployment', default: true
```
kke: As above

server/app/models/stack_deploy.rb
```diff
@@ -1,12 +1,66 @@
 class StackDeploy
   include Mongoid::Document
   include Mongoid::Timestamps
-  include Mongoid::Enum
 
-  index({ stack_id: 1 })
+  field :error, type: String
+  field :error_at, type: DateTime
```
kke: `errored_at` / `failed_at`?

server/app/models/stack_deploy.rb
```diff
@@ -1,12 +1,66 @@
...
-  enum :deploy_state, [:created, :ongoing, :success, :error], default: :created
+  index({ stack_id: 1 })
 
   belongs_to :stack
   has_many :grid_service_deploys, dependent: :destroy
+
+  def error!(message)
+    self.error_at = Time.now
+    self.error = message
+    self.save!
```
kke: Setters need `self.`, `save!` does not.

server/app/models/stack_deploy.rb
```diff
@@ -1,12 +1,66 @@
...
+  end
+
+  # @return [DateTime]
+  def started_at
+    timestamps = self.grid_service_deploys.map{|service_deploy| service_deploy.started_at}
```
kke: Inconsistent whitespacelessness between `{|`

kke: And space before `{`

server/app/models/stack_deploy.rb
SpComb: I wonder if the model method should even do the `save!`

kke: It has the `!`so I guess yes. It could do `update(error: message, error_at: Time.now)` instead of the `=` + `save`.

PR #3299, opened 2018-02-19
Fix service scale to set service running state

### LINKED ISSUE #2290: Service can be deployed in initialized state
It is possible for a service to be deployed with running instances while the service itself is still in the `initialized` state. This seems like a weird inconsistency.

The server has explicit code to deal with services in the `initialized` state while deploying, but those should not be needed if the services were fixed to be `running`:

* [`GridServiceSchedulerWorker`](https://github.com/kontena/kontena/blob/416ac835903b145b7251e6da0dc13418f1c9a123/server/app/jobs/grid_service_scheduler_worker.rb#L33) `elsif service_deploy.grid_service.running? || service_deploy.grid_service.initialized?`
* [`GridServiceDeployer`](https://github.com/kontena/kontena/blob/cadde67cf43b55de032f548529787301c2f15fd5/server/app/services/grid_service_deployer.rb#L61) `unless self.grid_service.running? || self.grid_service.initialized?`

### Create and scale a service without deploying it
```
$ kontena service create redis-test redis
 [done] Creating redis-test service      
$ kontena service ls
⊝ redis-test                                                 0 / 1      no       initialized                                                   
```
```
$ kontena service scale redis-test 1
 [done] Scaling redis-test to 1 instances      
⊛ Deployed instance test/null/redis-test-1 to node terom-test2-1
$ kontena service ls
⊝ redis-test                                                 1 / 1      no       initialized                                                   
```

### Deploying stack with dependent services
Deploying a stack that has services using `volumes_from` to depend on eachother:

```
$ kontena service ls
⊝ vols-from/web                                              1 / 1      no       initialized                                                   
⊝ vols-from/nginx                                            1 / 1      no       initialized                                                   
⊝ vols-from/redis                                            1 / 1      no       running   
```

SpComb:
Split out of #3277
Fixes #2290 `GridServices::Scale` can leave the service in the `initialized` state

Set the `GridService.state` to `running` for `GridServices::Scale` when deploying the service.

Includes `GridServices::Deploy` refactoring for #3277.

Services will no longer get deployed in the `initialized` state, removing some special cases.

### DIFF
--- a/server/app/mutations/grid_services/helpers.rb
+++ b/server/app/mutations/grid_services/helpers.rb
@@ -42,5 +42,17 @@ module GridServices
 
       save_grid_service(grid_service)
     end
+
+    # @param grid_service [GridService]
+    # @param state [String] desired state
+    # @param force [Boolean] force-update revision
+    def deploy_grid_service(grid_service, state = 'running', force: false)
+      grid_service.deploy_requested_at = Time.now.utc
+      grid_service.state = state
+
+      if force ? update_grid_service(grid_service, force: true) : save_grid_service(grid_service)
+        GridServiceDeploy.create(grid_service: grid_service)
+      end
+    end
   end
 end
--- a/server/app/jobs/grid_service_scheduler_worker.rb
+++ b/server/app/jobs/grid_service_scheduler_worker.rb
@@ -30,7 +30,7 @@ class GridServiceSchedulerWorker
         info "delaying #{service_deploy.grid_service.to_path} deploy because there is another deploy in progress"
         return nil
 
-      elsif service_deploy.grid_service.running? || service_deploy.grid_service.initialized?
+      elsif service_deploy.grid_service.running?
         info "starting #{service_deploy.grid_service.to_path} deploy"
         service_deploy.set(started_at: Time.now.utc)
         return service_deploy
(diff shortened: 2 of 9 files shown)


SpComb:
This is the first PR in the associated series of PRs to fix `kontena stack rm --wait`. The remainder are WiP waiting for this to get merged.

server/app/mutations/grid_services/helpers.rb
```diff
@@ -42,5 +42,17 @@ def update_grid_service(grid_service, force: false)
...
     end
+
+    # @param grid_service [GridService]
+    # @param state [String] desired state
+    # @param force [Boolean] force-update revision
+    def deploy_grid_service(grid_service, state = 'running', force: false)
+      grid_service.deploy_requested_at = Time.now.utc
+      grid_service.state = state
+
+      if force ? update_grid_service(grid_service, force: true) : save_grid_service(grid_service)
```
jakolehm: A bit hard to read.

SpComb: This is existing code that was just moved from the `GridServices::Deploy#execute`. The control flow is for dealing with the non-exceptions-based mutations `add_error` case: do not create the deploy if the update/save fails.

Cleaning this up would probably be having the `update_grid_service`/`save_grid_service` return an error instead, but then it would get messy figuring out where to deal with those exceptions and convert them into mutations `add_error` instead.

PR #3301, opened 2018-02-20
Remove debug output from e2e healthcheck test

kke:
image

I assume this was left here accidentally / temporarily

### DIFF
--- a/test/spec/features/stack/healthcheck_spec.rb
+++ b/test/spec/features/stack/healthcheck_spec.rb
@@ -29,8 +29,6 @@ describe 'kontena service health_check' do
         response = Net::HTTP.get_response(uri)
         status = response.code.to_i
 
-        puts "GET #{uri} => #{status}"
-
         if status == 503 && ((count += 1) < retry_503)
           # LB can return 503 temporarily during configuration, retry to make sure it's stable before returning it
           sleep 1


SpComb:
I wouldn't mind having some way to include optional debug output in the e2e specs themselves... debugging any e2e spec failures for those healthcheck specs without those is going to be hard.

kke:
We could introduce some commit-msg flags:

- `[ci_skip]` for example a doc change PR
- `[ci_debug]` run specs with `DEBUG=true`
- `[ci_e2e]` just run e2e

Then in the test script have something that takes the current commit message and does `exit 0` when the test in question should not run.

jakolehm:
Travis already supports `[skip ci]`, see: https://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build

SpComb:
I don't need to be able to debug the e2e specs in travis, it's enough if I can do something like `DEBUG=true rspec spec/features/stacks/healthcheck_spec.rb`... but that specific example is currently unusable because it causes kommando to spew.

kke:
Yeah the spammy kommando `process xxyyzz` output thing should go away.

jakolehm:
>  that specific example is currently unusable because it causes kommando to spew.

@matti any tips for kommando?

matti:
@jakolehm not sure if I understand the problem?

kommando accepts output: false?

jakolehm:
@matti probably `output: false` solves the problem.

SpComb:
> @matti probably output: false solves the problem.

I can't remember exactly what the problem was anymore, but kommando itself also uses the `DEBUG` env: https://github.com/matti/kommando/blob/38a8a0ad2d38fc2715a31df0a05ed70f0b8dfe0b/lib/kommando.rb#L334

There was something in kommando somewhere that caused it to spew something along the lines of "process ..." for each and every stdout character separately...

PR #3303, opened 2018-02-21
Fix kontena wrapper when PATH contains spaces.

colinmollenhour:
For example when I'm on Windows Subsystem for Linux and have installed the deb package the PATH contains some Windows directories so I get:

```
~ $ kontena whoami
/usr/local/bin/kontena: 3: export: (x86)/NVIDIA: bad variable name
```

This fixes it.


kke:
I didn't know the deb package works in WSL :)

colinmollenhour:
It does now! :D  Thanks for the merge.

PR #3305, opened 2018-02-22
Fix flaky agent EventWorker#start spec

### LINKED ISSUE #2321: Flaky agent spec: Kontena::Workers::EventWorker#start starts processing events
```
  1) Kontena::Workers::EventWorker#start starts processing events

     Failure/Error: expect(subject.wrapped_object).to receive(:stream_events)

       (#<Kontena::Workers::EventWorker:0x000000022220e8>).stream_events(any args)

           expected: 1 time with any arguments

           received: 0 times with any arguments

     # ./spec/lib/kontena/workers/event_worker_spec.rb:38:in `block (3 levels) in <top (required)>'
```

SpComb:
Fixes #2321

The specs assumed that the `async.foo` tasks would run by the time the sync call to `actor.start` completed and the spec example finished, which is racy.

The `streams and processes events` spec is also racy as in https://github.com/kontena/kontena/issues/2321#issuecomment-329765755, but fixing that would need changes to the `EventWorker`, which I'd rather avoid right now.

### DIFF
--- a/agent/spec/lib/kontena/workers/event_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/event_worker_spec.rb
@@ -46,7 +46,7 @@ describe Kontena::Workers::EventWorker do
         times.times {
           block.call(Docker::Event.new(event))
         }
-        sleep 0.1
+        sleep 0.5
         subject.stop_processing
       }
       subject.start
--- a/agent/spec/lib/kontena/workers/event_worker_spec.rb
+++ b/agent/spec/lib/kontena/workers/event_worker_spec.rb
@@ -33,7 +33,9 @@ describe Kontena::Workers::EventWorker do
     it 'starts processing events' do
       expect(subject.wrapped_object).to receive(:stream_events)
       expect(subject.wrapped_object).to receive(:process_events)
+
       subject.start
+      subject.nil? # ping it with something inconsequential to allow the async tasks to run
     end
 
     it 'streams and processes events' do


agent/spec/lib/kontena/workers/event_worker_spec.rb
```diff
@@ -44,7 +46,7 @@
         times.times {
           block.call(Docker::Event.new(event))
         }
-        sleep 0.1
+        sleep 0.5
```
SpComb: > The streams and processes events spec is also racy as in #2321 (comment), but fixing that would need changes to the EventWorker, which I'd rather avoid right now.

Alternatively just bump the `sleep` to avoid the race condition. With a 10ms sleep, I'm getting about 40-50/100 events on my laptop... 20ms still fails every time at 80-90, but 50ms mostly passes but still fails occasionally.  It's currently at 100ms, so if we bump that to say 500ms, then it should be enough to make the probably of travis spec failures rare enough to not be an issue that we need to care about anymore. 

The entire `stop_processing` mechanism is flawed, but it's probably not worth fixing because it's only really used in the specs... it's also used in the actor finalizer, but the only thing it really serves to do there is to suppress any error logging... but I wouldn't except any of the `!processing?` error cases to necessarily even get hit during the agent shutdown process, because AFAIK the `Docker::Event.stream` in the `Celluloid::Future` thread won't get any signal leading to a ruby exception during shutdown?

PR #3323, opened 2018-03-08
Change CLI cert authorize --type=http-01 to be the default

SpComb:


kke:
1.5 or 1.6?

SpComb:
> 1.5 or 1.6?

Technically 1.6

Issue #3338, opened 2018-04-03
CLI specs fail because of a broken dependency


kke:
![image](https://user-images.githubusercontent.com/224971/38234827-afcfbb4a-3727-11e8-9873-39041cfcdd25.png)

kke:
I think bundler got updated or something and doesn't now install `--with test` by default.

PR #3344, opened 2018-04-24
Do not allow to remove connected node from grid

### LINKED ISSUE #3343: kontena node rm can be used to remove connected Kontena Cloud nodes without terminating them
Because Kontena Cloud nodes are provisioned using per-node tokens, the `kontena node rm` command will allow removing the nodes from the grid, even though the host machine is still running and the agent is connected. However, disconnecting the agent from the master and removing the node from the grid will still leave the host machine itself running, and the user will continue to be billed for the host machine even though it is no longer visible in the `kontena node list` CLI or the Kontena Cloud dashboard. The node will still be visible in `kontena cloud node list`, and the user should be using `kontena cloud node terminate` instead of `kontena node rm` to terminate the node and stop the billing.

The same thing re `kontena $plugin node terminate` vs `kontena node rm` also applies to other provider plugins, but this issue is mitigated for the case of nodes not using node tokens by `kontena node rm` only allowing offline nodes to be removed, which mostly avoids this issue. If the agent is temporarily disconnected for whatever reason, then the CLI will still allow removing the node, but once the agent reconnects using the grid token, then the node will also reappear in `kontena node list`, making it more obvious what's happening.

The usecase for this `kontena node rm` behavior with node tokens is to allow isolating misbehaving/compromised nodes from the master if you are unable to manage them, but the resulting confusion hurts more than this obscure usecase. Removing a connected node should either require some explicit flag (`kontena node rm --yes-really-just-disconnect-the-agent-but-leave-the-host-machine-running`), or just rely on `kontena node reset-token --reset-connection` instead.

The command should also print out some very clear warning/hint about using `kontena $provider node terminate` instead, which will terminate the node and stop any billing before removing it from the grid. The current behavior also prints a warning in this case, but it's clearly not explicit enough about the consequences of leaving the host machine running:

> Node #{node['name']} is still connected using a node token, but will be force-disconnected

nevalla:
Fixes #3343

### DIFF
--- a/cli/lib/kontena/cli/nodes/remove_command.rb
+++ b/cli/lib/kontena/cli/nodes/remove_command.rb
@@ -13,9 +13,14 @@ module Kontena::Cli::Nodes
 
       nodes.each do |node_name|
         node = client(token).get("nodes/#{current_grid}/#{node_name}")
-
-        if node['connected']
-          exit_with_error "Node #{node['name']} is still connected. You must terminate the node before removing it, for example: kontena cloud node rm"
+        provider = Array(node["labels"]).find{ |l| l.start_with?('provider=')}.to_s.split('=').last
+        if node['connected'] && provider
+          plugin = provider == 'kontena' ? 'cloud' : provider
+          exit_with_error "Node #{node['name']} is still connected. You should terminate the node instead: kontena #{plugin} node terminate #{node['name']}"
+        elsif node['connected']
+          exit_with_error "Node #{node['name']} is still connected. You must terminate the node before removing it"
+        else
+          warning "Removing a node from the grid does not terminate the host machine from any cloud provider, so your cloud provider may continue to bill you for the machine until you terminate it"
         end
 
         confirm_command(node_name) unless forced?
--- a/cli/lib/kontena/cli/nodes/remove_command.rb
+++ b/cli/lib/kontena/cli/nodes/remove_command.rb
@@ -14,10 +14,8 @@ module Kontena::Cli::Nodes
       nodes.each do |node_name|
         node = client(token).get("nodes/#{current_grid}/#{node_name}")
 
-        if node['has_token'] && node['connected']
-          warning "Node #{node['name']} is still connected using a node token, but will be force-disconnected"
-        elsif node['connected']
-          exit_with_error "Node #{node['name']} is still connected using a grid token. You must terminate the node before removing it."
+        if node['connected']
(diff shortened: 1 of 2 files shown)


cli/lib/kontena/cli/nodes/remove_command.rb
```diff
@@ -14,10 +14,8 @@ def execute
       nodes.each do |node_name|
         node = client(token).get("nodes/#{current_grid}/#{node_name}")
 
-        if node['has_token'] && node['connected']
-          warning "Node #{node['name']} is still connected using a node token, but will be force-disconnected"
-        elsif node['connected']
-          exit_with_error "Node #{node['name']} is still connected using a grid token. You must terminate the node before removing it."
+        if node['connected']
+          exit_with_error "Node #{node['name']} is still connected. You must terminate the node before removing it, for example: kontena cloud node rm"
```
SpComb: The `kontena cloud node rm` command will also remove the node from the grid, so the `You must terminate the node before removing it` doesn't make sense in that case.

SpComb: Something like this would be more useful, with explicit `kontena $plugin node terminate` hints and a general warning for offline nodes.

```ruby
provider = Array(node["labels"]).find{ |l| l.start_with?('provider=')}.to_s.split('=').last

if node['connected'] && provider
  exit_with_error "Node #{node['name']} is still connected. You should terminate the node instead: kontena #{provider} node terminate #{node['name']}"
elsif node['connected']
  exit_with_error "Node #{node['name']} is still connected. You must terminate the node before removing it"
else
  warning "Removing a node from the grid does not terminate the host machine from any cloud provider, so your cloud provider may continue to bill you for the machine until you terminate it"
end
```

The warning is to cover the edge cases of nodes that are just temporarily offline...

SpComb: Although maybe the warning should only be `if provider`... I suppose the main usecase for `kontena node rm` is self-hosted custom nodes, which should be provisioned without any `provider=...` labels, and warning about generic cloud providers is unnecessary.

nevalla:
I've changed the format of error messages based on @SpComb comment.

SpComb:
A spec for the case of a node without any `labels`, or missing the `provider=` label might be a good idea.

nevalla:
More specs added

nevalla:
Some weird CI error with ruby 2.5 🤔 @kke any ideas?

SpComb:
> Some weird CI error with ruby 2.5 thinking @kke any ideas?

That would be #3338

PR #3287, opened 2018-02-16
Fix agent to unregister LB service backends earlier during container shutdown

SpComb:
Fixes #3283

This is a one-line change to make the agent unregister service containers from LB service backends earlier: immediately on the `kill` event, rather than waiting for the `die` event (see #3283 for details). Unregistering the container on the `die` event is still needed to handle containers that die spontaneously without getting killed.

More importantly, this adds an e2e spec for graceful shutdowns across LB service deploys. The server service simulates long-running requests, proxied via an LB service. The client service does continuous HTTP requests in parallel with randomized skew, ensuring that there are always active in-flight requests on each backend service. The e2e specs then force-deploy the server service to do a rolling re-create of each service container. The client service will crash with an error if any request fails or returns a non-200 status, and the e2e specs pick up those request errors using the new `service:instance_exit` event from #3286

There are two kinds of shutdown behaviors in the e2e spec:
* The `using ungraceful shutdown to fail requests` spec case has the service exit immediately on `SIGTERM`, causing requests to get interrupted
* The `graceful shutdown` has the service close its listening socket while allowing any remaining requests to complete, so it passes the e2e specs without changes
* The `no-healthcheck` does not configure any kontena `healthcheck: ...` on the service, but this is equivalent to the `graceful shutdown` case because `kontena/lb` [unconditionally configures L4 healthchecks](https://github.com/kontena/kontena-loadbalancer/blob/v1.1.3/lib/kontena/templates/haproxy/http_backends.text.erb#L34) for service backends
* The `delayed shutdown` relies on the fix here to unregister the LB service backend in order to drain the service, so it fails the e2e specs without this change

```
kontena service shutdown
  for a http test service
    using delayed shutdown without closing the listening socket
      after re-deploying the server
        does not result in any client errors (FAILED - 1)

Failures:

  1) kontena service shutdown for a http test service using delayed shutdown without closing the listening socket after re-deploying the server does not result in any client errors
     Failure/Error: expect(events.select{|e| e[:type] == 'instance_exit'}).to eq []
     
       expected: []
            got: [{:time=>"2018-02-16T13:11:21.953Z", :type=>"instance_exit", :data=>"service shutdown-test-delay/clie...tdown-test-delay/client-1 instance exited with code 1, restarting (delay: 0s) (terom-e2e-dev)\r\n"}]
     
       (compared using ==)
     
       Diff:
       @@ -1,2 +1,37 @@
       -[]
       +[{:time=>"2018-02-16T13:11:21.953Z",
[...]

PR #3354, opened 2018-05-08
E2e: Mark the graceful shutdown spec as pending

kke:
The e2e graceful shutdown spec from #3287 always fails.

Marked as pending.


kke:
And of course now it doesn't fail.

kke:
Doesn't fail anymore.

PR #3357, opened 2018-06-21
Update README to Kontena Classic

jakolehm:


jakolehm:
@kke any idea why e2e is failing?

PR #3360, opened 2018-07-05
Fix e2e test specs debug output

SpComb:
Alternative to #3301

Add a new `DebugHelper` to the e2e test specs, allowing specs to output optional debugging information if running with `DEBUG_SPECS=true` (ref `DEBUG_KOMMANDO=true`).

### DIFF
--- a/test/spec/support/debug_helper.rb
+++ b/test/spec/support/debug_helper.rb
@@ -1,0 +1,11 @@
+module DebugHelper
+  def debug?
+    ENV.has_key?('DEBUG_SPECS')
+  end
+
+  def debug(msg = nil, &block)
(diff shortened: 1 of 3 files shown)


test/spec/support/debug_helper.rb
```diff
@@ -0,0 +1,11 @@
+module DebugHelper
+  def debug?
+    ENV.has_key?('DEBUG_SPECS')
+  end
+
+  def debug(msg = nil, &block)
+    msg = msg || yield
```
kke: `puts(block_given? ? yield : msg) if debug?` would keep the intent (not yielding when not debugging) of the logger-alike message block.

PR #3364, opened 2018-08-09
Ignore CVE-2018-1000539

kke:
Whitelists [CVE-2018-1000539](https://nvd.nist.gov/vuln/detail/CVE-2018-1000539)

`Json::JWT` has a vulnerability when validating JWT tokens. The `Acme::Client` is using it only for generating requests, not validation.

Fixing will require changes to mongo schema and some code rewrite.


kke:
E2e keeps timing out in all branches.

